PR #39, opened 2023-07-26
Add basic workflows

EmilijusS:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
+name: CI
+on: pull_request
+permissions: {}
+jobs:
+  lint:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
```
bartoszWojciechO, line 8: Is it possible to use tags instead of commit hashes?

ci/jobs/analyse.yml
```diff
@@ -34,15 +34,6 @@ analyse/binaries:
   parallel:
     matrix:
       - ARCH: [i386, amd64, armel, armhf, aarch64]
-analyse/correctness:
```
bartoszWojciechO, line 37: Do we have to remove gitlab stuff?

bartoszWojciechO (approved):

.github/workflows/ci-gitlab.yml
```diff
@@ -7,11 +7,17 @@ on:
     branches: [dev, main, github-ci-setup]
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
+  workflow_run:
```
keliramu, line 10: AFAIU, now this workflow will be triggered by 3 events: 1) pull request, 2) tag assigned, 3) other workflow named CI completed. Simple jobs workflow CI is triggered only by pull request created, so, if tag is assigned this workflow is not triggered. I would assume, that assigning tag full CI pipeline has to be executed - simple jobs as well.

ci/jobs/analyse.yml
keliramu: These job should be removed from Gitlab - but in `nordvpn-linux-build` project

.github/workflows/ci.yml
EmilijusS: It's possible, but AppSec recommends to use hashes.

ci/jobs/analyse.yml
EmilijusS: Okay will remove in that project.
On the other hand - shouldn't they be removed here as well? We can't perform these jobs without appropriate Docker containers anyways. And I'm not sure if we should keep all of this stuff if it will not be used anyways?

keliramu (dismissed):

.github/workflows/ci-gitlab.yml
EmilijusS: Thanks, changed

keliramu (approved):

ggediminass (approved):

ggediminass (approved):

keliramu (approved):

PR #54, opened 2023-08-28
LAN Discovery fixes

EmilijusS:


daemon/firewall/allowlist/allowlist_routing.go
```diff
@@ -34,8 +36,12 @@ func NewAllowlistRouting(commandFunc runCommandFunc) *IPTables {
 
 // Adds allowlist routing rules for ports
 func (ipt *IPTables) EnablePorts(ports []int, protocol string, mark string) error {
-	for _, port := range ports {
-		err := routePortsToIPTables(ipt.runCommandFunc, fmt.Sprintf("%d", port), protocol, mark)
+	for _, portRange := range iptables.PortsToPortRanges(ports) {
```
EmilijusS, line 39: Would insert 1000 rules into firewall before if port range 1000-2000 was allowlisted :slightly_smiling_face:

daemon/rpc_set_allowlist.go
```diff
@@ -11,12 +11,12 @@ import (
...
 )
 
-// containsPrivateNetwork returns true if subnets contains a private network not conainted in oldWhitelist
+// containsPrivateNetwork returns true if subnets contains a private network
 func containsPrivateNetwork(subnets []string) bool {
 	for _, subnet := range subnets {
 		if net, err := netip.ParsePrefix(subnet); err != nil {
 			log.Println("Failed to parse subnet: ", err)
-		} else if net.Addr().IsPrivate() {
+		} else if net.Addr().IsPrivate() || net.Addr().IsLinkLocalUnicast() {
```
EmilijusS, line 19: There were several places where we incorrectly wouldn't be testing for link local addresses. We consider such addresses as private when forwarding traffic or enabling lan discovery.

magefiles/mage.go
```diff
@@ -367,7 +367,6 @@ func (Test) CgoDocker(ctx context.Context) error {
 	env["ARCH"] = build.Default.GOARCH
 	env["CI_PROJECT_DIR"] = "/opt"
 	env["ENVIRONMENT"] = "dev"
-	env["GOPATH"] = build.Default.GOPATH
```
EmilijusS, line 370: CGO tests wouldn't work locally. This somewhat helped because tests run now, but there is also permission issue and tests that are manipulating iptables for example are failing now... Decided not to bother with that now.

daemon/rpc_set_lan_discovery_test.go
```diff
@@ -224,74 +215,3 @@ func TestSetLANDiscovery_Error(t *testing.T) {
 		})
 	}
 }
-
-func TestSetLANDiscovery_MeshInteraction(t *testing.T) {
```
keliramu, line 228: Why this test is removed?

EmilijusS: Because this was basically testing whether the correct mesh peers are being passed to networker and after my changes mesh peers are not being passed at all anymore.

keliramu (approved):

PR #56, opened 2023-08-28
Add missing manual DNS tests

ggediminass:


test/qa/lib/dns.py
```diff
@@ -21,6 +21,21 @@ def _is_set_to_threat_protection_lite_ipv6():
     return "2400:bb40:4444::103" and "2400:bb40:8888::103" in sh.cat("/etc/resolv.conf")
 
 
+# returns True when one Custom DNS server is being used
+def _is_set_to_one_custom_dns_server():
+    return "1.1.1.1" in sh.cat("/etc/resolv.conf")
```
keliramu, line 26: Why not using dns from array defined in init.py?

test/qa/lib/dns.py
```diff
@@ -21,6 +21,21 @@ def _is_set_to_threat_protection_lite_ipv6():
...
+
+
+# returns True when two Custom DNS servers are being used
+def _is_set_to_two_custom_dns_servers():
+    return "1.1.1.1" and "9.9.9.9" in sh.cat("/etc/resolv.conf")
```
keliramu, line 31: Why not using dns from array defined in init.py?

test/qa/lib/dns.py
```diff
@@ -21,6 +21,21 @@ def _is_set_to_threat_protection_lite_ipv6():
...
+
+
+# returns True when three Custom DNS servers are being used
+def _is_set_to_three_custom_dns_servers():
+    return "1.1.1.1" and "9.9.9.9" and "208.67.222.222" in sh.cat("/etc/resolv.conf")
```
keliramu, line 36: Why not using dns from array defined in init.py?

test/qa/lib/dns.py
```diff
@@ -31,8 +46,16 @@ def is_unset():
     )
 
 
-def is_set_for(threat_protection_lite, ipv6="off"):
+def is_set_for(threat_protection_lite="", ipv6="off", custom_dns=[]):
     logging.log(sh.cat("/etc/resolv.conf"))
+
+    if custom_dns == ["1.1.1.1"]:
```
keliramu, line 52: Why not using dns from array defined in init.py?

test/qa/test_dns.py
```diff
@@ -114,3 +150,114 @@ def test_set_dns_connected(tech, proto, obfuscated):
...
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(40)
+def test_set_custom_dns_errors_disconnected(tech, proto, obfuscated, stdin, expected_error):
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+    lib.set_dns("off")
+
+    assert dns.is_unset()
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        sh.nordvpn.set.dns(stdin)
```
keliramu, line 230: Why not make param name by its actual meaning? eg. `nameserverip`

test/qa/test_dns.py
```diff
@@ -114,3 +150,114 @@ def test_set_dns_connected(tech, proto, obfuscated):
...
+    lib.set_dns("off")
+
+    assert dns.is_unset()
+
+    output = sh.nordvpn.connect()
+    print(output)
+    assert lib.is_connect_successful(output)
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        sh.nordvpn.set.dns(stdin)
```
keliramu, line 251: Why not make param name by its actual meaning? eg. nameserverip

keliramu (approved):

test/qa/test_dns.py
```diff
@@ -114,3 +150,114 @@ def test_set_dns_connected(tech, proto, obfuscated):
...
+    assert lib.is_connect_successful(output)
+
+    sh.nordvpn.set.dns(nameserver)
+
+    assert dns.dig(nameserver)
+    assert dns.dns_visible_in_settings(nameserver)
+
+    with lib.ErrorDefer(sh.nordvpn.disconnect):
+        assert network.is_connected()
+        assert dns.is_unset()
```
EmilijusS, line 176: Why is dns unset if it was set just above?

test/qa/test_dns.py
```diff
@@ -85,10 +87,44 @@ def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
...
+
+    sh.nordvpn.set.dns(nameserver)
+    assert dns.dns_visible_in_settings(nameserver)
+    assert dns.is_unset()
+
+    sh.nordvpn.set.dns("off")
+    assert dns.dns_visible_in_settings("disabled")
+    assert dns.is_unset()
+
+    output = sh.nordvpn.connect()
```
EmilijusS, line 105: Is there a specific reason why printing the output of connect is necessary? Should just do `assert lib.is_connect_successful(sh.nordvpn.connect())` I think. Although I'm not even sure what purpose this `is_connect_successful` serves, I think if connect fails an exception will happen anyway.

test/qa/test_dns.py
```diff
@@ -85,10 +87,44 @@ def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
...
+    print(output)
+    assert lib.is_connect_successful(output)
+
+    assert not dns.dig(nameserver)
+    assert dns.is_unset()
+
+    with lib.ErrorDefer(sh.nordvpn.disconnect):
+        assert network.is_connected()
```
EmilijusS, line 113: Is this check really necessary? I understand checking this in connect tests, but I think this is an overkill in DNS tests. I don't think everything should be checked every time, DNS tests should check DNS stuff. But maybe I'm wrong, this is just my opinion.

test/qa/test_dns.py
```diff
@@ -85,10 +87,44 @@ def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
...
+        assert dns.is_unset()
```
EmilijusS, line 114: Already checked above. Applicable to some other tests as well.

test/qa/test_dns.py
```diff
@@ -85,10 +87,44 @@ def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
...
+    assert dns.is_unset()
+
+    output = sh.nordvpn.connect()
+    print(output)
+    assert lib.is_connect_successful(output)
+
+    assert not dns.dig(nameserver)
+    assert dns.is_unset()
+
+    with lib.ErrorDefer(sh.nordvpn.disconnect):
```
EmilijusS, line 112: I think if you want to ensure robustness of tests (always disconnecting after connecting), then it should look like this:
```
  sh.nordvpn.connect()
  # asserts while connected
  assert bla
  assert bla
# asserts while disconnected
assert bla
assert bla
```

test/qa/lib/__init__.py
```diff
@@ -114,6 +117,12 @@
 ]
 
 
+# Used for DNS test parametrization
+DNS_CASES = [
```
EmilijusS, line 121: Should rename to `DNS_ERROR_CASES` IMO.

test/qa/lib/__init__.py
```diff
@@ -51,7 +51,10 @@
 ]
 
 # Used for test parametrization, when the same test has to be run for different values of custom dns.
-DNS = ["1.1.1.1"]
+DNS = [["1.1.1.1"], ["1.1.1.1", "9.9.9.9"], ["1.1.1.1", "9.9.9.9", "208.67.222.222"]]
```
EmilijusS, line 54: And this one should be renamed to `DNS_CASES` :smile:

test/qa/test_dns.py
```diff
@@ -58,7 +57,7 @@ def test_dns_connect(tech, proto, obfuscated, threat_protection_lite):
 
 
 @pytest.mark.parametrize("nameserver", lib.DNS)
-@pytest.mark.parametrize("tech,proto,obfuscated", lib.TECHNOLOGIES_BASIC1)
+@pytest.mark.parametrize("tech,proto,obfuscated", lib.TECHNOLOGIES)
 @pytest.mark.flaky(reruns=2, reruns_delay=90)
 @timeout_decorator.timeout(40)
 def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
```
EmilijusS, line 63: I just encountered this test failing on reruns. Please recheck that every test here which has a `flaky` decorator could be rerun successfully on connect failure. To be more precise, the problem is that if connect fails, dns state is not reset. Then the test is rerun, tries to set dns, but as it is already set - it fails again, so in essence reruns are useless.

bartoszWojciechO (approved):

test/qa/test_dns.py
```diff
@@ -84,23 +153,183 @@ def test_custom_dns_connect(tech, proto, obfuscated, nameserver):
...
+        assert dns.is_set_for(dns.DNS_NORD)
+
+    assert dns.is_unset()
+
+
+@pytest.mark.parametrize("nameserver", dns.DNS_CASES_CUSTOM)
+@pytest.mark.parametrize("tech,proto,obfuscated", lib.TECHNOLOGIES)
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(40)
+def test_set_custom_dns_already_set_disconnected(tech, proto, obfuscated, nameserver):
```
EmilijusS, line 261: This is a very minor one, but the name of this test confused me, had to read the body to understand what is being tested. So now I know that this test tests what happens when setting custom dns while it is already set and while we are connected to VPN. So my suggestion for the name fix would be "test_set_custom_dns_while_already_set_and_connected". Now I know that this name is long, but as far as I know in test automation industry long names are generally OK and even preferable if it is possible to understand exactly what is being tested just from the name. Comment is applicable to other test names as well.

EmilijusS (approved):

PR #61, opened 2023-08-31
Add masquerade rules for OpenVPN (simpler)

EmilijusS:


meshnet/exitnode/server.go
```diff
@@ -30,7 +31,9 @@ type Server struct {
 }
 
 // NewServer create & initialize new Server
-func NewServer(interfaceNames []string, commandFunc runCommandFunc) Node {
+func NewServer(interfaceNames []string, commandFunc runCommandFunc) *Server {
+	// To allow peer routing when connected to OpenVPN as well
+	interfaceNames = append(interfaceNames, openvpn.InterfaceName)
```
keliramu, line 36: Does it work when OpenVPN is not used? `fw.enableMasquerading` should return error when trying to add rules to non existing interface

EmilijusS: Iptables allows inserting rules with non-existing network devices.

bartoszWojciechO (approved):

keliramu (approved):

PR #62, opened 2023-09-01
Introduce testing policy

EmilijusS:


bartoszWojciechO (approved):

test/mock/networker/networker.go
```diff
@@ -0,0 +1,141 @@
+package networker
```
keliramu, line 1: Our initial approach was to have project code structure as flat as possible, now it's getting 3 levels deep. But probably `test` folder adds to it as special case. Ok, then no problem.

keliramu (approved):

test/mock/networker/networker.go
EmilijusS: Libtelio is still deeper :smile: I only did this because putting it straight into `mock` created a circular dependency - it's impossible to implement Networker interface while not depending on `networker` package since some return values are defined in that package. And `networker` package tests of course uses some mocks. This was the simplest solution, as this PR is already bloated as is.

PR #70, opened 2023-09-11
Add masquerade rules per mesh peer

bartoszWojciechO:


meshnet/exitnode/fw.go
```diff
@@ -211,7 +168,11 @@ type TrafficPeer struct {
 	LocalNetwork bool
 }
 
-func resetPeersTraffic(peers []TrafficPeer, commandFunc runCommandFunc) error {
+func resetPeersTraffic(peers []TrafficPeer, interfaceNames []string, commandFunc runCommandFunc) error {
```
EmilijusS, line 171: There will be merge conflicts with release branch. I think it's better to allow to merge release branch into main first and only then merge this PR.

meshnet/exitnode/fw.go
```diff
@@ -224,6 +185,12 @@ func resetPeersTraffic(peers []TrafficPeer, commandFunc runCommandFunc) error {
 	}
 
 	for _, peer := range peers {
+		if peer.Routing {
```
EmilijusS, line 188: Actually it might be beneficial to move this to the very bottom of this function. As these rules are inserted into a different table, their insertion sequence to other rules doesn't matter. But moving their insertion to the very end would help with a small leak that is currently present in this code. At the moment the rules inserted just below allow full routing to the peer, until the `refreshPrivateSubnetsBlock` finishes execution (so while routing to the local network is forbidded, it is allowed for a very brief moment). By moving the insertion of masquerading rules to the bottom I think we could prevent this leak.

meshnet/exitnode/fw.go
```diff
@@ -211,7 +168,11 @@ type TrafficPeer struct {
...
 }
 
-func resetPeersTraffic(peers []TrafficPeer, commandFunc runCommandFunc) error {
+func resetPeersTraffic(peers []TrafficPeer, interfaceNames []string, commandFunc runCommandFunc) error {
+	if err := clearMasquerading(commandFunc); err != nil {
+		return fmt.Errorf("clearing masquerade rules: %w", err)
+	}
+
 	for _, peer := range peers {
 		err := modifyPeerTraffic(peer.IP, "-D", true, true, commandFunc)
```
EmilijusS, line 177: It's not directly related to this ticket, but as you are working here it might be the best moment to fix one more thing. I think there is an issue that when a peer is removed then their filtering rules will remain undeleted because the removed peer is simply not in the list anymore. On the other hand we cannot simply delete all rules with the `nordvpn` comment from the `FORWARD` chain, because there is the all blocking rule which we do not touch. What do you think? Of course this can be a separate issue as well if the fix is not trivial, there are probably enough changes in this PR already :D

EmilijusS (approved):

meshnet/exitnode/server_test.go
```diff
@@ -100,6 +160,93 @@ func TestResetPeers_LANDiscoveryDisabled(t *testing.T) {
...
+		"Firewall was configured incorrectly after meshnet peer restart. \nEXPECTED:\n%s\nGOT:\n%s",
+		strings.Join(expectedCommands, "\n"), strings.Join(commandExecutor.executedCommands, "\n"))
+}
+
+func TestDisable(t *testing.T) {
+	category.Set(t, category.Firewall)
+
+	commandExecutor := newCommandExecutorMock(t)
+	// The contents of the rules is not so important. Basically, we want to make sure that Disable
+	// will remove every rule from filter FORWARD chain and nat POSROUTING returned by querying commands.
```
keliramu, line 172: Typo: POSROUTING

keliramu (approved):

PR #71, opened 2023-09-12
More accurate error message when failing to fetch VPN subscription status

EmilijusS:


auth/auth_test.go
```diff
@@ -38,3 +41,95 @@ func TestIsTokenExpired(t *testing.T) {
 		assert.Equal(t, tt.expected, got)
 	}
 }
+
+type authConfigManager struct {
```
bartoszWojciechO, line 45: Shouldn't this go to `test/mocks`, according to new guidelines?

EmilijusS: It should, but the guideline isn't merged yet :smile:

daemon/rpc_account.go
```diff
@@ -18,6 +18,18 @@ func (r *RPC) AccountInfo(ctx context.Context, _ *pb.Empty) (*pb.AccountResponse
 		return nil, internal.ErrNotLoggedIn
 	}
 
+	accountInfo := &pb.AccountResponse{}
+
+	vpnExpired, err := r.ac.IsVPNExpired()
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "checking VPN expiration: ", err)
+		return &pb.AccountResponse{Type: internal.CodeTokenRenewError}, nil
```
bartoszWojciechO, line 26: These internal error codes don't fit our current approach I think. Currently we are relying on dedicated protobuff responses defined for each RPC. I sometimes refactor it while I'm working on other stuff, but sometimes it results in a lot of extra work and bloated PR.

bartoszWojciechO (approved):

auth/auth_test.go
keliramu: It's not strict requirement to have every mock in there, I guess. That would be reasonable for bigger mocks, imo.

keliramu (approved):

daemon/rpc_account.go
EmilijusS: The code changes aren't very extensive so I decided not to refactor this time.

PR #73, opened 2023-09-13
Add git-secrets hook when building application via mage

bartoszWojciechO:


magefiles/mage.go
```diff
@@ -44,6 +44,41 @@ type Test mg.Namespace
...
+	if err != nil {
+		if werr, ok := err.(*exec.ExitError); ok {
+			if werr.ExitCode() != 1 {
+				return fmt.Errorf("getting user.email from git config: %w", err)
+			}
+		} else {
+			return fmt.Errorf("unknown error when getting user.email from git config: %w", err)
+		}
+
+		fmt.Println("Warnirng: git user.email is not configured.")
```
EmilijusS, line 59: Typo

magefiles/mage.go
```diff
@@ -44,6 +44,41 @@ type Test mg.Namespace
 // View is used for viewing information about the project.
 type View mg.Namespace
 
+// installHookIfNordsec installs git-secrets hook if git user belongs to Nord Security.
+func installHookIfNordsec() error {
+	output, err := exec.Command("git", "config", "--get", "user.email").CombinedOutput()
+	if err != nil {
+		if werr, ok := err.(*exec.ExitError); ok {
+			if werr.ExitCode() != 1 {
```
EmilijusS, line 52: Should add a comment why we fail on every error except when error code is 1

magefiles/mage.go
```diff
@@ -44,6 +44,41 @@ type Test mg.Namespace
...
+	output, err = exec.Command("git", "secrets", "--list").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("fetching git-secrets providers: %w", err)
+	}
+
+	if strings.Index(string(output), "llt-secrets") == -1 {
+		return fmt.Errorf("Secret provider was not configured.")
+	}
+
+	if _, err := exec.Command("git", "secrets", "--install", "--force").CombinedOutput(); err != nil {
```
EmilijusS, line 75: This is installed every time. Is it basically like a noop if it is already installed?

bartoszWojciechO: It's not, it basically is overriding the existing hooks. So not a noop but should not change the files either. Checking would be somewhat complicated. We'd need to read the git.config file to get the hooks location, then read the hook scripts and grep them for git-secrets commands, I didn't find any better way. So I'm not sure if it's worth extra complexity.

EmilijusS (approved):

PR #78, opened 2023-09-20
Improve filter table exitnode related rules cleanup

bartoszWojciechO:


meshnet/exitnode/fw.go
```diff
@@ -168,20 +172,37 @@ type TrafficPeer struct {
 	LocalNetwork bool
 }
 
+func clearExitnodeForwardRules(commandFunc runCommandFunc) error {
+	output, err := commandFunc("iptables", "-S", "FORWARD")
+
```
EmilijusS, line 177: This is very minor, but there's an empty line between almost every statement in this function. At the very least I'd suggest grouping error returning statements with the error checks together. It's also possible to do a one liner `if err := func(); err != nil {`, but personally I'm not a fan of those, they just seem too dense for me :smile: 
Though personally this function could have no empty lines at all, it's short enough and indentation makes it clear anyway :smile:

EmilijusS (approved):

keliramu (approved):

PR #80, opened 2023-09-21
Fix initial http failures

Savolro:


events/moose/moose.go
```diff
@@ -300,7 +300,13 @@ func (s *Subscriber) NotifyTechnology(data config.Technology) error {
 }
 
 func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
+	// Even in meshnet peer case keep calculating connected at so we don't have extremely long sessions in case meshnet peer connections are used
 	s.connectedAt = time.Now()
+
+	if data.IsMeshnetPeer {
```
EmilijusS, line 306: This doesn't seem necessary to me. The only new thing I see happening that when we connect to exit node the `connectedAt` time will be reset. But if we do a VPN connection after the exit node connection, it will be reset again. So am I missing something? Seems like a noop to me as it is.

keliramu: It's distinction of 2 cases: connect to vpn and connect to/via exit node. Do not register vpn connect to analytics in case it is exit not connect.

keliramu (approved):

keliramu (approved):

EmilijusS (approved):

PR #92, opened 2023-09-28
Fix gitlab coverage job

EmilijusS:
Since all jobs jobs in Gitlab run on `main` branch of build repo, the app would be built without `-race` option and would use different coverage calculation method, hence breaking combined coverage calculations (unit tests are always run with `-race`)


ci/compile.sh
```diff
@@ -8,9 +8,8 @@ source "${WORKDIR}"/ci/archs.sh
 
 # Since race detector has huge performance price and it works only on amd64 and does not
 # work with pie executables, its enabled only for development builds.
-branch="${CI_COMMIT_REF_NAME:=$(git describe --contains --all HEAD)}"
```
EmilijusS, line 11: Firstly I tried fixing this by adding `--git-dir=${WORKDIR}` to the git command. Somehow it didn't work and then I thought whether we need this branch condition altogether. We have clearly separated dev and candidate builds now, so building differently based on source branch is just confusing (as proved by the issue we are having). So I decided to simply remove that.

bartoszWojciechO (approved):

bartoszWojciechO (approved):

keliramu (approved):

PR #95, opened 2023-10-02
Incoming connections to local interface control

bartoszWojciechO:


meshnet/server.go
```diff
@@ -2704,3 +2743,21 @@ func (s *Server) GetPrivateKey(ctx context.Context, _ *pb.Empty) (*pb.PrivateKey
 		},
 	}, nil
 }
+
+// refreshIncoming will block and unblock firewall rules for incoming traffic, applying routing, icoming and local network
+// access rules
+func (s *Server) refreshIncoming(peer mesh.MachinePeer) error {
```
Savolro, line 2749: This does not need any information from the server and it is always used in combination with `netw.ResetRouting`. Would it make sense to move this to networker and merge those methods?

meshnet/server_test.go
```diff
@@ -58,8 +59,13 @@ func (n *workingNetworker) AllowFileshare(address UniqueAddress) error {
 	return nil
 }
 
-func (n *workingNetworker) AllowIncoming(address UniqueAddress) error {
+func (n *workingNetworker) AllowIncoming(address UniqueAddress, lanAllowed bool) error {
 	n.allowedIncoming = append(n.allowedIncoming, address)
+
+	if !lanAllowed {
```
Savolro, line 65: This mock implementation is becoming too big. Generally, you shouldn't care about the internal state of mocked object in unit tests. Actual internal state of actual networker can be tested in networker's unit tests. But for using it in e. g. meshnet server you only care that this method was called correctly.

Savolro (approved):

meshnet/networker.go
```diff
@@ -21,7 +21,7 @@ type Networker interface {
 	UnSetMesh() error
 	// AllowIncoming creates an allowing fw rule for the given
 	// address
-	AllowIncoming(UniqueAddress) error
+	AllowIncoming(UniqueAddress, bool) error
```
keliramu, line 24: I would recommend to add description of what are expected parameters - added new parameter but it's meaning is unknown (because even name is not present)

keliramu (approved):

meshnet/networker.go
bartoszWojciechO: Added parameter names, should be more meaningful now.

PR #96, opened 2023-10-02
Implement route to remote LAN

keliramu:


networker/networker.go
```diff
@@ -261,16 +261,19 @@ func (netw *Combined) start(
...
-			netw.vpnet.Tun().Interface(),
-			serverData.IP.Is6(),
-		)
+	netw.publisher.Publish("Setting the routing rules up")
 
-		if err != nil {
-			return err
-		}
+	// need to re-set routing setup
+	if err = netw.policyRouter.CleanupRouting(); err != nil {
```
keliramu, line 267: There is still a problem with LAN Discovery setup - its decision and setup is done early i.e. on cli command (set on/off, add allowlist) because LAN discovery is about local (private) addresses be enabled to access when exit routing is enforced. But also that is related very much with Allowlist - need to merge addresses. But when we have to make decision about Route LAN access to remote peer or not - here we need to disable LAN discovery in case remote peer allows its LAN access, but LAN Discovery setup is already blended/mixed with Allowlist setup and we have no clear way to manage LAN Discovery setting without risk screwing Allowlist setup. IMO, routing setup decisions has to be done as late as possible or need possibility to re-play LAN Discovery and Allowlist setup.

keliramu: short term solution: disable lan-discovery when mesh peer connected;
long term solution: design and implement networker state machine;

keliramu:
> What are the conditions for blocking local LANs? i.e when is `routeToRemoteLan` set to true? I cannot find it in the code somehow.

See meshnet/server.go https://github.com/NordSecurity/nordvpn-linux/pull/96/files#diff-96ca84766083d590bc0bed6793d9440b684721fee325861802b344baaa687dfb

Condition is target peer allows LAN access or not.

bartoszWojciechO (approved):

daemon/routes/iprule/iprule.go
```diff
@@ -80,16 +81,72 @@ func (r *Router) SetupRoutingRules(
 			return err
 		}
 
-		otherRuleID, err := calculateRulePriority(ipv6)
-		if err != nil {
-			return err
+		// PeerA (LAN-a 192.168.1.x) connects to PeerB (LAN-b 192.168.1.x)
+		// if PeerB allows its LAN access when used as Exit node
+		// then PeerB LAN access is the priority over PeerA LAN
+		if !routeToRemoteLan {
```
EmilijusS, line 87: The code is duplicated with PreventLocalTraffic. I think it would be possible to refactor in a way that this would just call preventLocalTraffic.

daemon/routes/iprule/iprule.go
```diff
@@ -80,16 +81,72 @@ func (r *Router) SetupRoutingRules(
...
-		if err = addSuppressprefixLengthRule(
-			otherRuleID,
-			ipv6,
-		); err != nil {
-			return err
+	return nil
+}
+
+// PreventLocalTraffic adjust policy based routing rules
+func (r *Router) PreventLocalTraffic(
```
EmilijusS, line 106: Subjective, but I would rather call this `EnableLocalTraffic`, that way it's more clear that it's dependent on a bool argument whether the traffic is enabled and disabled. And even then, since it has two bool arguments it's impossible to tell whether `EnableLocalTraffic(interface, false, true)` would enable it or disable it. 

Another way to deal with bad readability of bool arguments is to simply use two public functions: EnableLocalTraffic and DisableLocalTraffic (both these functions would then just call the same private function setLocalTraffic) - this way it would be totally clear what is happening. Of course calling code would get more verbose because the `if routeToRemoteLan` condition would become the responsibility of the caller, but I think that is still better for readability.

daemon/routes/iprule/iprule.go
```diff
@@ -80,16 +81,72 @@ func (r *Router) SetupRoutingRules(
...
+	vpnInterface net.Interface,
+	ipv6Enabled bool,
+	routeToRemoteLan bool,
```
EmilijusS, line 109: Continuing with my above suggestion to rename the function, this argument should simply be called `enable`

daemon/routes/iprule/iprule.go
```diff
@@ -80,16 +81,72 @@ func (r *Router) SetupRoutingRules(
...
 
-		if err = addSuppressprefixLengthRule(
-			otherRuleID,
-			ipv6,
-		); err != nil {
-			return err
+	return nil
+}
+
+// PreventLocalTraffic adjust policy based routing rules
```
EmilijusS, line 105: Comment could be more specific

networker/networker.go
```diff
@@ -149,6 +150,11 @@ type Combined struct {
 	fwmark             uint32
 	mu                 sync.Mutex
 	lanDiscovery       bool
+	// need to memorize route to remote LAN state set on mesh peer connect
+	// according how remote peer has set its permission, for later when
+	// doing mesh refresh which may happen in background e.g. when new peer
+	// was added and NC update message is received
+	routeToRemoteLan bool
```
EmilijusS, line 157: Is this state really necessary? The PreventLocalTraffic function already doesn't return errors on cases when trying to insert the rule while it already exists or trying to delete it when it doesn't exist. So I'm not sure why we need to remember the last state, when we can always just use the current one which is passed by an argument.

networker/networker.go
```diff
@@ -1138,13 +1167,25 @@ func (netw *Combined) setMesh(
 		if err = netw.policyRouter.SetupRoutingRules(
 			netw.mesh.Tun().Interface(),
 			false,
+			netw.routeToRemoteLan,
```
EmilijusS, line 1170: This should always be false and the else is not necessary, because we are still connected to VPN and the suppress prefix rule will never change.

networker/networker.go
```diff
@@ -1257,6 +1298,10 @@ func (netw *Combined) defaultMeshUnBlock() error {
 func (netw *Combined) UnSetMesh() error {
 	netw.mu.Lock()
 	defer netw.mu.Unlock()
+	// clear this flag only when user turns mesh off, cannot do that in internal func
+	// because it is used during refresh, and during refresh we need to know what
```
EmilijusS, line 1302: Maybe I missed something but I don't see refresh case being handled at the moment.

daemon/routes/iprule/iprule.go
EmilijusS, line 1: Missing tests. Networker as well.

networker/networker.go
keliramu: Sorry, comment is incorrect about the source of refreshVPN trigger - it's not NC event, it's network change event, `netw.routeToRemoteLan` we have state which was set during `mesh peer connect`, we keep this state for `refreshVPN` case (https://github.com/NordSecurity/nordvpn-linux/blob/3bd6de0bbccd1648fcdbed6eab0b6b98817631a4/networker/vpn.go#L35) which is invoked on network state change (https://github.com/NordSecurity/nordvpn-linux/blob/3bd6de0bbccd1648fcdbed6eab0b6b98817631a4/networker/listener.go#L10)

networker/networker.go
keliramu: right, this `else` part is not logical, but current logic would create duplicate fwMark rules during network change event

networker/networker.go
keliramu: Ok, I think I got your idea when analyzing network change event handling with refreshVPN and setMesh - all can be handled in `SetupRoutingRules` only need to add extra check to not create duplicate rules

daemon/routes/iprule/iprule.go
keliramu: Working on those after main approach has been confirmed (not rejected)

daemon/routes/iprule/iprule.go
keliramu: Removed this function, added extra logic to `SetupRoutingRules`

networker/networker.go
keliramu: Still need to keep state for `refreshVPN` case

daemon/routes/iprule/iprule.go
```diff
@@ -45,57 +46,66 @@ func (r *Router) SetupRoutingRules(
...
-		if err = addSuppressprefixLengthRule(
-			otherRuleID,
-			ipv6,
-		); err != nil {
-			return err
+		// PeerA (LAN-a 192.168.1.x) connects to PeerB (LAN-b 192.168.1.x)
+		// if PeerB allows its LAN access when used as Exit node
+		// then PeerB LAN access is the priority over PeerA LAN
+		if enableLocal {
+			enableLocalTraffic(ipv6)
```
EmilijusS, line 79: Some inconsistency - code above returns errors, this private function doesn't, and the private function below returns them but we are only logging it. I think instead of logging we should return all errors, especially because we are checking for existence of rules before acting, so any error would be unexpected.

networker/networker.go
EmilijusS: Got it. So some other points:

1. Does it really need to be reset here? I mean it will be set anyways when a new connect will be done.
2. I think the case where the peer changes `lan-access` permission while we are routing through them is still unhandled.

daemon/routes/iprule/iprule.go
```diff
@@ -45,57 +46,66 @@ func (r *Router) SetupRoutingRules(
 		ipv6EnabledList = append(ipv6EnabledList, true)
 	}
 
-	defer func() { // recover if error
```
Savolro, line 48: Why was this removed?

networker/networker.go
```diff
@@ -1134,18 +1151,17 @@ func (netw *Combined) setMesh(
...
-			return fmt.Errorf(
-				"setting routing rules: %w",
-				err,
-			)
-		}
-		routingRulesSet = true
+	if err = netw.policyRouter.SetupRoutingRules(
+		netw.mesh.Tun().Interface(),
+		false,
+		netw.enableLocalTraffic,
```
EmilijusS, line 1157: I think my comment about this always being `true` is still valid. It's subjective, so change only if you agree, but I think this parameter being `netw.enableLocalTraffic` instead of just `true` slightly gives wrong sense that this is dynamic when it isn't :D

daemon/routes/iprule/iprule.go
keliramu: Ok, wrong assumption of mine, getting it back.

networker/networker.go
keliramu: When mesh connect peer (peer allows lan access, enableLocalTraffic=false) -> with your laptop go outside building i.e. switch to mobile network -> Combined.Reconnect(...)->Combined.refreshVPN(...)->Networker.setMesh(...)->policyRouter.SetupRoutingRules(...., netw.enableLocalTraffic)

If you set it to `true` always then you will loose what was set during mesh peer connect, as it does not replay `mesh peer connect` during network change event.
The same should be done for `ipv6enabled` parameters as well, I guess.

daemon/routes/iprule/iprule.go
keliramu: Agree

networker/networker.go
keliramu: 1. enableLocalTraffic state to keep is needed for refreshVPN during network change when we are mesh peer connected;
2. because policy router SetupRoutingRules is used during `mesh on` and `mesh peer connect` and also during those bothactions this is state is used, wee need to reset it to default value `true` when unseting mesh (by user specifically, not by internal func which is used for refreshVPN case).
3. well, we do not handle such cases when peer changes permissions (e.g. do we brake file transfer if peer changes that permission?), probably need new task for this;

bartoszWojciechO (approved):

PR #102, opened 2023-10-05
Refactor DNS modules in Python tests

ggediminass:


test/qa/lib/network.py
```diff
@@ -143,3 +143,17 @@ def unblock():
...
+def block_traffic():
+    for chain in chains:
+        sh.sudo("iptables", "-P", chain, "DROP")
+        sh.sudo("ip6tables", "-P", chain, "DROP")
+
+
+def allow_traffic():
+    for chain in chains:
+        sh.sudo("iptables", "-P", chain, "ACCEPT")
+        sh.sudo("ip6tables", "-P", chain, "ACCEPT")
```
edvyno1, line 159: no longer used

test/qa/lib/dns.py
```diff
@@ -1,45 +1,47 @@
...
 
+    # DNS Adddresses set in Resolver:
+    dns_set_in_os_addresses = dns.resolver.Resolver().nameservers
 
-def is_set_for(threat_protection_lite, ipv6="off"):
-    logging.log(sh.cat("/etc/resolv.conf"))
-    if threat_protection_lite == "on" and ipv6 == "on":
-        return _is_set_to_threat_protection_lite_ipv6()
+    # Check DNS Addresses that were successfully set in NordVPN app and Resolver, and put them in list:
+    dns_used_from_app = [address for address in dns_set_in_app if address in dns_set_in_os_addresses]
```
mariusSincovici, line 42: Wouldn't work to create 2 sets from the lists and compare them?
e.g.: `return set(dns_set_in_os_addresses) == set(dns_set_in_app)`

ggediminass: It seems like this could be a beautiful alternative method to check for same traits, e.g. if list lengths and data is the same.

However, I tried a scenario, where NordVPN application sets `103.86.96.96, 103.86.99.99` DNS on the system, and we slip one additional DNS Address `103.86.99.99` to resolver, making `/etc/resolv.conf` look like the following:
```
nameserver 103.86.96.96
nameserver 103.86.99.99
nameserver 103.86.99.99
```

It seems like current implemented solution would return `False` in scenario described above, but solution with two sets returns `True`.

Since current implemented solution covers more cases, I'd like to stay with it :smile:

test/qa/lib/__init__.py
```diff
@@ -204,6 +201,13 @@ def set_threat_protection_lite(dns):
         print("WARNING:", ex)
 
 
+def set_dns(dns):
```
EmilijusS, line 204: What is the purpose of this? Seems like it just allows ignoring the error, which doesn't seem like a good thing.

test/qa/lib/dns.py
```diff
@@ -1,45 +1,47 @@
...
+# Used for test parametrization, when the same test has to be run for different values of custom dns.
+DNS_CASE_CUSTOM_SINGLE = "2.0.0.0"
+DNS_CASE_CUSTOM_DOUBLE = "2.0.0.1 2.0.0.2"
+DNS_CASE_CUSTOM_TRIPLE = "2.0.0.3 2.0.0.4 2.0.0.5"
+DNS_CASES_CUSTOM = [DNS_CASE_CUSTOM_SINGLE, DNS_CASE_CUSTOM_DOUBLE, DNS_CASE_CUSTOM_TRIPLE]
 
-# returns True when NordVPN DNS servers are used
-def _is_set_to_nord_ipv6():
-    return "2400:bb40:8888::100" and "2400:bb40:4444::100" in sh.cat("/etc/resolv.conf")
+ALL_DNS_CASES = \
```
EmilijusS, line 18: If I understand correctly how this is used, I think it should be called `ALL_TEST_DNS_ADDRESSES` instead. It's just that originally I thought this is also used in parametrization, but looking at code I see that this is used to check whether dns is unset.

test/qa/lib/dns.py
```diff
@@ -1,45 +1,47 @@
...
+    DNS_CASE_CUSTOM_SINGLE.split(" ") + \
+    DNS_CASE_CUSTOM_DOUBLE.split(" ") + \
+    DNS_CASE_CUSTOM_TRIPLE.split(" ")
 
 
-# returns True when Threat Protection Lite DNS servers are used
-def _is_set_to_threat_protection_lite_ipv6():
-    return "2400:bb40:4444::103" and "2400:bb40:8888::103" in sh.cat("/etc/resolv.conf")
+def is_unset() -> bool:
+    """ returns True when NordVPN DNS servers are used """
```
EmilijusS, line 29: Comment doesn't seem to be correct. From the code I understand that True will be returned if none of the resolver DNS addresses are the same as the DNS addresses in `ALL_DNS_CASES`. So I think that the comment that was previously here (`returns True when NordVPN app has not modified the DNS`) would be correct.

test/qa/lib/__init__.py
EmilijusS: Ah, I see there's already a comment above.

test/qa/lib/dns.py
ggediminass: It is also used for parametrization in `test_dns.py` module, in test `test_custom_dns_connect`

test/qa/test_dns.py
```diff
@@ -25,92 +24,73 @@ def teardown_module(module):
...
+    with lib.Defer(sh.nordvpn.disconnect):
+        sh.nordvpn.connect()
 
-    output = sh.nordvpn.disconnect()
-    print(output)
-    assert lib.is_disconnect_successful(output)
-    assert network.is_disconnected()
+        assert dns.is_set_for(nameserver)
 
     sh.nordvpn.set.dns.off()
```
EmilijusS, line 77: What should the DNS be set to after disconnect but before setting DNS off? Should add an additional assert. Unless it is also unset, in which case `sh.nordvpn.set.dns.off()` wouldn't be necessary then.

test/qa/lib/dns.py
EmilijusS: This looks like a bug to me personally. The duplicated addresses I mean.

ggediminass: We currently do not have such case, but in case we would have one like this, my current implementation would detect this issue, and assert would fail, that is why I'd prefer to keep my solution

test/qa/test_dns.py
ggediminass: Since we unset DNS before executing each test, in `setup_function`, I think we can remove the `sh.nordvpn.set.dns.off()`

test/qa/lib/dns.py
mariusSincovici: yes, I was thinking that the records are unique. :smile:

if you want to check also for duplicates, then you can keep your implementation or maybe `sorted` would work too(e.g. `return sorted(dns_set_in_os_addresses) == sorted(dns_set_in_app)`)

EmilijusS: Sorry, I misunderstood your example. I thought that this is an actual bug, but it's just a hypothetical situation :D Yes, I agree then that it's better not to use a set, but Marius `sorted` suggestion should work then.

ggediminass: Solution using `sorted` seems to be working :smile: Will implement this :smile:

EmilijusS (approved):

bartoszWojciechO (approved):

PR #105, opened 2023-10-10
Fix daemon crash when groups list is empty

mariusSincovici:
When groups list is nil, return an error message to let the use know that the list is being fetched.

### DIFF
--- a/daemon/rpc_groups.go
+++ b/daemon/rpc_groups.go
@@ -13,12 +13,30 @@ import (
 // Groups provides endpoint and autocompletion.
 func (r *RPC) Groups(ctx context.Context, in *pb.GroupsRequest) (*pb.Payload, error) {
 	var cfg config.Config
+	if r.cm == nil || r.dm == nil {
+		log.Println(internal.ErrorPrefix, "configuration Manager or DataManager is nil")
+		return &pb.Payload{
+			Type: internal.CodeEmptyPayloadError,
+		}, nil
+	}
+
 	if err := r.cm.Load(&cfg); err != nil {
 		log.Println(internal.ErrorPrefix, err)
+		return &pb.Payload{
+			Type: internal.CodeConfigError,
+		}, nil
+	}
+
+	groups := r.dm.GetAppData().GroupNames[in.GetObfuscate()][cfg.AutoConnectData.Protocol]
+	if groups == nil {
+		log.Println(internal.ErrorPrefix, "groups list are nil")
+		return &pb.Payload{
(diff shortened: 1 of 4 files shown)


daemon/rpc_groups.go
```diff
@@ -13,12 +13,30 @@ import (
 // Groups provides endpoint and autocompletion.
 func (r *RPC) Groups(ctx context.Context, in *pb.GroupsRequest) (*pb.Payload, error) {
 	var cfg config.Config
+	if r.cm == nil || r.dm == nil {
+		log.Println(internal.ErrorPrefix, "configuration Manager or DataManager is nil")
```
Savolro, line 17: Typo: Choose one:
```
ConfigurationManager or DataManager
Configuration Manager or Data Manager
configuration manager or data manager
```

daemon/rpc_groups.go
```diff
@@ -13,12 +13,30 @@ import (
...
 	if err := r.cm.Load(&cfg); err != nil {
 		log.Println(internal.ErrorPrefix, err)
+		return &pb.Payload{
+			Type: internal.CodeConfigError,
+		}, nil
+	}
+
+	groups := r.dm.GetAppData().GroupNames[in.GetObfuscate()][cfg.AutoConnectData.Protocol]
+	if groups == nil {
+		log.Println(internal.ErrorPrefix, "groups list are nil")
```
Savolro, line 32: Typo. Change to one of:
```
group list is nil
group list is empty
```

Savolro (approved):

daemon/rpc_groups.go
```diff
@@ -13,12 +13,30 @@ import (
 // Groups provides endpoint and autocompletion.
 func (r *RPC) Groups(ctx context.Context, in *pb.GroupsRequest) (*pb.Payload, error) {
 	var cfg config.Config
+	if r.cm == nil || r.dm == nil {
```
EmilijusS, line 16: I don't think this check is necessary actually. At least we are not doing such checks elsewhere. Of course there's a theoretical possibility of a null pointer panic, but in practice this struct has a "constructor" so the actual possibility of them being nil is close to zero unless you are doing something unwise deliberately.
My opinion in general is that checking every time if every used interface is nil is just cumbersome, and in practice at least I never got burned by not checking them :smile:

mariusSincovici: yes, this only happened while writing the tests. I'll remove the check in this case.

EmilijusS (approved):

PR #108, opened 2023-10-12
Fix crash for countries command

mariusSincovici:
* Prevent crashing when countries data is missing.
* Fix test for bash autocomplete for connect

### DIFF
--- a/cli/cli_connect_test.go
+++ b/cli/cli_connect_test.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"strings"
-	"sync"
 	"testing"
 
 	"github.com/NordSecurity/nordvpn-linux/client/config"
@@ -31,21 +30,16 @@ func captureOutput(f func()) (string, error) {
 		os.Stdout = stdout
 		os.Stderr = stderr
 	}()
+
 	os.Stdout = writer
 	os.Stderr = writer
-	out := make(chan string)
-	wg := new(sync.WaitGroup)
-	wg.Add(1)
-	go func() {
-		var buf bytes.Buffer
-		wg.Done()
-		io.Copy(&buf, reader)
-		out <- buf.String()
-	}()
-	wg.Wait()
+
 	f()
-	writer.Close()
-	return strings.TrimSuffix(<-out, "\n"), nil
+
+	writer.Close() // close to unblock io.Copy(&buf, reader)
+	var buf bytes.Buffer
+	io.Copy(&buf, reader)
+	return strings.TrimSuffix(buf.String(), "\n"), nil
 }
 
 type mockDaemonClient struct {
--- a/cli/cli_connect_test.go
+++ b/cli/cli_connect_test.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"flag"
-	"log"
+	"io"
 	"os"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/NordSecurity/nordvpn-linux/client/config"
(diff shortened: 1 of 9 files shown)


bartoszWojciechO (approved):

cli/cli_connect_test.go
```diff
@@ -18,39 +20,92 @@ import (
...
+		os.Stderr = stderr
+	}()
+	os.Stdout = writer
+	os.Stderr = writer
+	out := make(chan string)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		var buf bytes.Buffer
+		wg.Done()
```
EmilijusS, line 41: `wg` seems useless here and also it is unnecessary because syncing is done via channel

cli/cli_connect_test.go
```diff
@@ -18,39 +20,92 @@ import (
...
+		io.Copy(&buf, reader)
+		out <- buf.String()
+	}()
+	wg.Wait()
+	f()
+	writer.Close()
```
EmilijusS, line 47: Why `reader` isn't closed? And should probably just do at the top:
```
	reader, writer, err := os.Pipe()
	if err != nil {
		return "", err
	}
    defer reader.Close()
    defer writer.Close()
```

cli/cli_connect_test.go
```diff
@@ -18,39 +20,92 @@ import (
...
+	}()
+	os.Stdout = writer
+	os.Stderr = writer
+	out := make(chan string)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		var buf bytes.Buffer
+		wg.Done()
+		io.Copy(&buf, reader)
```
EmilijusS, line 42: At first I wanted to ask why is pipe used if stdout can be written straight to buffer. Then I realized that `Copy` is used specifically because it only exits on EOF, otherwise we might not capture the full output, and Pipe is needed because Copy requires a Reader argument. So would be nice to leave a comment explaining this I think :D

cli/cli_countries_test.go
```diff
@@ -0,0 +1,53 @@
...
+		expected      string
+		input         string
+		expectedError error
+	}{
+		{
+			name:          "error response",
+			expectedError: formatError(fmt.Errorf(MsgListIsEmpty, "countries")),
+		},
+		{
+			name:      "counties list",
```
EmilijusS, line 32: typo

daemon/rpc_countries.go
```diff
@@ -76,16 +76,25 @@ func (r *RPC) Countries(ctx context.Context, in *pb.CountriesRequest) (*pb.Paylo
...
 		log.Println(internal.ErrorPrefix, err)
+		return &pb.Payload{
+			Type: internal.CodeConfigError,
+		}, nil
 	}
 
-	var countryNames []string
-	for country := range r.dm.GetAppData().CountryNames[in.GetObfuscate()][cfg.AutoConnectData.Protocol].Iter() {
-		countryNames = append(countryNames, country.(string))
+	if countries, ok := r.dm.GetAppData().CountryNames[in.GetObfuscate()][cfg.AutoConnectData.Protocol]; ok {
```
EmilijusS, line 84: Minor, but I'd rather write 
``` 
countries, ok := blabla
if !ok {
    return error
}

do things with countries
```
Basically leaving main unindented code path for the success case.

EmilijusS (approved):

cli/cli_connect_test.go
mariusSincovici: true. I'll refactor this and remove completely the coroutine, because it just ads unneeded complexity.

cli/cli_connect_test.go
mariusSincovici: the `writer` needs to be closed before to unlock `io.Copy` from `reader`.

by default when a `pipe` is created it will add a cleanup function when garbage collector releases the memory, so reader will be closed by GC: `runtime.SetFinalizer(f.file, (*file).close)` from https://github.com/golang/go/blob/release-branch.go1.20/src/os/file_unix.go#L215

PR #121, opened 2023-10-19
Automate some of the status tests

naujalyte:


test/qa/test_combinations.py
```diff
@@ -156,3 +156,70 @@ def test_connect_country_and_city(country, city):
...
+    target_proto,
+    source_obfuscated,
+    target_obfuscated,
+):
+    lib.set_technology_and_protocol(source_tech, source_proto, source_obfuscated)
+
+    sh.nordvpn.connect()
+
+    lib.set_technology_and_protocol(target_tech, target_proto, target_obfuscated)
+    sh.nordvpn.connect()
```
EmilijusS, line 214: I'm not 100% sure, but if test fails on this connect, then on a rerun it might fail again on the first connect because we are already connected (I'm not sure whether after second failed connect we would be left disconnected or still connected to the first server). See this test for example how to avoid this using `lib.defer` https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/test_dns.py#L43
In your test above it's not strictly necessary since there's only one connect and it's at the start of the test. On the other hand I would still personally recommend to do this `lib.defer(disconnect)` before every connect in every test just in case, but it's your choice.

test/qa/test_connect.py
```diff
@@ -468,4 +468,52 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+
+    if tech == "openvpn":
+        assert proto.upper() in sh.nordvpn.status()
+    else:
+        assert "UDP" in sh.nordvpn.status()
+
+    transfer = str(sh.nordvpn.status()).split("\n")[7]
+    transfer_received = float(transfer.split(" ")[1])
+    transfer_sent = float(transfer.split(" ")[4])
+    assert transfer_received > 0
```
EmilijusS, line 512: Right now this only tests whether something is happening with these number, no matter if the numbers make sense or not.
An idea to make these checks better - upload and download 10MB of data after connecting and then check whether `transfer_received` and `transfer_sent` is more than 10MB but lower than 11MB for example (might need to adjust upper interval to avoid flakiness). Not sure where to download and upload an exact amount of data, but that question can be raised to team-level if needed I think.

test/qa/lib/server.py
```diff
@@ -50,3 +52,17 @@ def get_hostname_by(technology="", protocol="", obfuscated="", group_id=""):
     available_server_count = len(requests.get(url).json()) - 1
     server = requests.get(url).json()[random.randint(0, available_server_count)]
     return server["name"], server["hostname"]
+
+
+def get_city_name(server_name):
```
mariusSincovici, line 57: Wouldn't it be better to combine the 2 functions and return some kind of class, e.g. `get_server_info(server)`? In this way the information will be obtained from one server call + if more data is needed in the future, then more members are added to the class?

test/qa/test_connect.py
```diff
@@ -468,4 +468,52 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+    assert transfer_sent > 0
+
+    uptime = str(sh.nordvpn.status()).split("\n")[8]
+    time_connected = int(uptime.split(" ")[1])
+    assert time_connected > 0
```
EmilijusS, line 517: Same with this test, except it's easier to fix - can just count time since connect and check it with some leeway, like `assert time_connected > time_passed-1s && time_connected < time_passed+1s`

EmilijusS (approved):

test/qa/test_connect.py
```diff
@@ -468,4 +468,52 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+    assert server.get_city_name(name) in sh.nordvpn.status()
+
+    assert tech.upper() in sh.nordvpn.status()
+
+    if tech == "openvpn":
+        assert proto.upper() in sh.nordvpn.status()
+    else:
+        assert "UDP" in sh.nordvpn.status()
+
+    transfer = str(sh.nordvpn.status()).split("\n")[7]
```
mariusSincovici, line 509: [minor] played [here](https://www.online-python.com/kEBmGhHZcK) how to convert the status string to a dictionary. In this way if something is added to the status list, it will not break the tests, but it can break if the strings(keys) are changed :smile:.

bartoszWojciechO (approved):

test/qa/test_connect.py
```diff
@@ -483,52 +483,57 @@ def test_status_disconnected(tech, proto, obfuscated):
...
     with lib.Defer(sh.nordvpn.disconnect):
         name, hostname = server.get_hostname_by(technology=tech, protocol=proto, obfuscated=obfuscated)
         sh.nordvpn.connect(hostname.split(".")[0])
+
         connect_time = int(time.time())
 
-        assert hostname in sh.nordvpn.status()
+        data_10MB = requests.get("http://speedtest.tele2.net/10MB.zip")
+        destination_file = "file_10MB"
+        if data_10MB.status_code == 200:
```
keliramu, line 498: need to handle failure of downloading above file, maybe `assert` instead of `if` would be appropriate here

test/qa/test_connect.py
```diff
@@ -483,52 +483,57 @@ def test_status_disconnected(tech, proto, obfuscated):
...
 
         #The app shows transfer size in MiB, 10MB = ~10.49MiB
         assert transfer_received > 10 and transfer_received < 12
         assert transfer_sent > 10 and transfer_sent < 12
 
-        uptime = str(sh.nordvpn.status()).split("\n")[8]
-        time_connected = int(uptime.split(" ")[1])
-        current_time = int(time.time())
+        time_connected = int(status_info['uptime'].split(" ")[0])
         time_passed = current_time - connect_time
```
mariusSincovici, line 537: This might not be problematic because:
* if `uptime` is bigger then 1 minute, the first component will be minutes instead of seconds, and then it needs to be `* 60`
* if it is still in seconds, the diff will not return the uptime since it was connected, but some time before the connection time. Adding the `uptime` might give the correct value.

Previous check might have been better, if instead of `time.time()` `time.monotonic`(https://docs.python.org/3/library/time.html#time.monotonic) was used.

test/qa/test_combinations.py
```diff
@@ -156,3 +156,74 @@ def test_connect_country_and_city(country, city):
...
+
+        lib.set_technology_and_protocol(target_tech, target_proto, target_obfuscated)
+        assert source_tech.upper() in sh.nordvpn.status()
+
+        if source_tech == "openvpn":
+            assert source_tech.upper() in sh.nordvpn.status()
+        else:
+            assert "UDP" in sh.nordvpn.status()
+
+        output = sh.nordvpn.disconnect()
```
EmilijusS, line 192: Minor thing, but the `lib.Defer(sh.nordvpn.disconnect)` already does the disconnect when that `with` block ends, so this and the next line is unnecessary, and the lines below should be unindented. Example:
```
    sh.nordvpn.connect()
    # test some connect stuff

assert lib.is_disconnect_successful(output)
assert network.is_disconnected()
```
Of course the print is omitted then, but I'm personally not sure why it's even there :D As far as I know if the command would fail the output would be printed anyway. Or maybe I'm wrong and that's not the case with these `lib.Defer`. Maybe would be useful to experiment and see what happens on failure there.

test/qa/test_connect.py
```diff
@@ -468,4 +468,73 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+    with lib.Defer(sh.nordvpn.disconnect):
+        name, hostname = server.get_hostname_by(technology=tech, protocol=proto, obfuscated=obfuscated)
+        sh.nordvpn.connect(hostname.split(".")[0])
+
+        connect_time = int(time.time())
+
+        data_10MB = requests.get("http://speedtest.tele2.net/10MB.zip")
+        destination_file = "file_10MB"
+        if data_10MB.status_code == 200:
+            with open(destination_file, "wb") as file:
```
EmilijusS, line 499: 1. Instead of opening the same file two times you could just continue the uploading in this same `with` block. Actually I think you also forgot to close the file after opening it the second time below. By the way that's why this `with` construct is used - can't forget to close the file because it is automatically closed when the `with` block ends.
2. It's likely the case that this file could pollute our codespace: after running tests this file would appear as newly added in git, maybe someone would accidentally commit it, etc. Alternative would be to use an in-memory file, so a physical one would not appear https://stackoverflow.com/a/44672691/10169568

test/qa/test_connect.py
```diff
@@ -468,4 +468,73 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+                file.write(data_10MB.content)
+
+        files = {'file': open(destination_file, 'rb')}
+        requests.post("http://speedtest.tele2.net/upload.php", files=files)
+
+        current_time = int(time.time())
```
EmilijusS, line 505: Maybe should be called `status_time, as it's used much further below and is not really "current" anymore then.

test/qa/test_connect.py
```diff
@@ -468,4 +468,73 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
...
+
+        if tech == "openvpn":
+            assert proto.upper() in status_info['current protocol']
+        else:
+            assert "UDP" in status_info['current protocol']
+
+        transfer_received = float(status_info['transfer'].split(" ")[0])
+        transfer_sent = float(status_info['transfer'].split(" ")[3])
+
+        #The app shows transfer size in MiB, 10MB = ~10.49MiB
```
EmilijusS, line 532: Should also test if the line contains "MiB"

test/qa/test_connect.py
naujalyte: Changed to `time.monotonic` and for the issue that would come up if connection time was more than a minute came up with such workaround:

```
time_connected = int(status_info['uptime'].split(" ")[0])
time_connected_seconds = int(status_info['uptime'].split(" ")[2]) if "minute" in status_info["uptime"] else 0
time_passed = status_time - connect_time
     
if "minute" in status_info["uptime"]:
            assert time_connected * 60 + time_connected_seconds >= time_passed - 1 and time_connected * 60 + time_connected_seconds <= time_passed + 1
else:    
            assert time_connected >= time_passed - 1 and time_connected <= time_passed + 1 
```
            
To test this, temporarily added `time.sleep(50)` before executing `sh.nordvpn.status()` and tests pass (also to double-check, printed values in logs to see if everything is correct and matching) 

Do you think this look okay or is there a better way to approach this?

naujalyte: An example from logs then:

DBG: 1 // `time_connected` from status (1 minute)
DBG: 19 // `time_connected_seconds`
DBG: 79.82627920698724 // `time_passed`
DBG: {'status': 'Connected', 'hostname': 'us5063.nordvpn.com', 'ip': '185.245.87.59', 'country': 'United States', 'city': 'Los Angeles', 'current technology': 'NORDLYNX', 'current protocol': 'UDP', 'transfer': '11.50 MiB received, 11.66 MiB sent', 'uptime': '1 minute 19 seconds'} //The output of status

mariusSincovici: Looks good. :+1: 
if you want, you can move `time_connected_seconds` inside the `if "minute" in status_info["uptime"]:` and you can write just `time_connected_seconds = int(status_info['uptime'].split(" ")[2])`.

bartoszWojciechO (approved):

test/qa/test_connect.py
```diff
@@ -491,20 +492,21 @@ def test_status_connected(tech, proto, obfuscated):
...
-        connect_time = int(time.time())
+        connect_time = time.monotonic()
 
         data_10MB = requests.get("http://speedtest.tele2.net/10MB.zip")
-        destination_file = "file_10MB"
-        if data_10MB.status_code == 200:
-            with open(destination_file, "wb") as file:
-                file.write(data_10MB.content)
+        data_10MB.raise_for_status()
+        assert data_10MB.status_code == 200
```
EmilijusS, line 499: Don't need to assert if using `raise_for_status` already, the exception from that function will fail the test anyway

EmilijusS (approved):

test/qa/test_connect.py
EmilijusS: Just my opinion, but I don't think it was necessary to bother with the "minute" case, this test is never going to take a full minute :D

naujalyte: True, but oh well, just in case :D

test/qa/test_connect.py
naujalyte: Right, removed assert 👍

PR #124, opened 2023-10-19
Automated missing Settings tests for Technology

naujalyte:


test/qa/test_settings.py
```diff
@@ -94,3 +94,34 @@ def test_set_obfuscate_server_obfuscation_mismatch(obfuscateInitialState, server
     assert errorMessage in str(ex.value)
 
     sh.nordvpn.set.autoconnect.off()
+
+
+@pytest.mark.parametrize("tech,proto,obfuscated", lib.TECHNOLOGIES_BASIC2 + lib.TECHNOLOGIES_BASIC1)
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
```
mariusSincovici, line 100: Why is this needed, isn't it always working?

mariusSincovici (commented):
Not sure if it was intended, but it I think it can happen that the first run in each test is not doing anything, depending on the app existing settings and the test value.
e.g. when changing the technology, if `NORDLYNX` is already set, setting it again will have no impact.

test/qa/test_settings.py
EmilijusS: We use these `flaky` decorators on tests that do connects because sometimes they just fail (well not "just", but it's complicated). But here you are right, I also think this is unnecessary. Timeout decorator below as well. FYI explicit timeout decorator is used with flaky decorator because the default timeout method in our tests just kills the python process while with a softer timeout the tests would be left hanging sometimes. But in this case because we need to restart a test we want a softer timeout, so an explicit timeout decorator is used with a lower timeout value than the global timeout.

EmilijusS (approved):

mariusSincovici (approved):

test/qa/test_settings.py
naujalyte: "e.g. when changing the technology, if NORDLYNX is already set, setting it again will have no impact." I don't think that this should be happening since all of these tests start from changing the technology from the default Nordlynx to OpenVPN

naujalyte:
> "e.g. when changing the technology, if NORDLYNX is already set, setting it again will have no impact." 

I don't think that this should be happening since all of these tests start from changing the technology from the default Nordlynx to OpenVPN

test/qa/test_settings.py
naujalyte: Thanks! Will remove those from tests that do not do connects

mariusSincovici (approved):

PR #125, opened 2023-10-20
Route to other lan fix

keliramu:


Savolro (approved):

mariusSincovici (approved):
Also verified and it works

PR #132, opened 2023-10-25
Update message that the device needs to be rebooted after install

mariusSincovici:
The device needs to be rebooted after installing nordvpn, otherwise there are problems with `systemd --user`.
Update error message to let the user know that the machine needs to be rebooted instead of just logging out

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -935,7 +935,7 @@ func (c *cmd) action(err error, f func(*cli.Context) error, daemonURL string, la
 				os.Exit(1)
 			case internal.ErrSocketAccessDenied:
 				color.Red(formatError(internal.ErrSocketAccessDenied).Error())
-				color.Red("Run 'usermod -aG nordvpn $USER' to fix this issue and log out of OS afterwards for this to take an effect.")
+				color.Red("Run 'usermod -aG nordvpn $USER' to fix this issue and reboot your device afterwards for this to take an effect.")
 				os.Exit(1)
 			case internal.ErrDaemonConnectionRefused:
 				color.Red(formatError(internal.ErrDaemonConnectionRefused).Error())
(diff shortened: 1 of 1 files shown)


EmilijusS (approved):

cli/cli.go
```diff
@@ -935,7 +935,7 @@ func (c *cmd) action(err error, f func(*cli.Context) error, daemonURL string, la
 				os.Exit(1)
 			case internal.ErrSocketAccessDenied:
 				color.Red(formatError(internal.ErrSocketAccessDenied).Error())
-				color.Red("Run 'usermod -aG nordvpn $USER' to fix this issue and log out of OS afterwards for this to take an effect.")
+				color.Red("Run 'usermod -aG nordvpn $USER' to fix this issue and reboot your device afterwards for this to take an effect.")
```
EmilijusS, line 938: Rytis (the Meshnet writer who encountered the original issue) also suggested adding "Run '**sudo** usermod -aG nordvpn $USER'" to the message, as casual users often get confused why the copy-pasted command doesn't work.

Arguments for:
1. Would make it simpler and more clear for casual Ubuntu users (the biggest share of our users I believe)
Arguments against:
1. `sudo` is not always necessary: in Docker or in plain Debian for example. On the other hand users using these are most likely more advanced and can figure that out by themselves.
2. Copy pasting commands with `sudo` that you don't understand is probably not a good practice which we shouldn't encourage. 

I would still vote for including `sudo` in the example command because that just seems more pragmatic to me, but more opinions are welcome.

keliramu (approved):

bartoszWojciechO (approved):

cli/cli.go
mariusSincovici: Actually I had it with `sudo` but accidentally reverted that. Also on the [website](https://support.nordvpn.com/Connectivity/Linux/1325531132/Installing-and-using-NordVPN-on-Debian-Ubuntu-Raspberry-Pi-Elementary-OS-and-Linux-Mint.htm) it contains `sudo`.

So I'm also for `sudo` :)

EmilijusS (approved):

PR #135, opened 2023-10-27
Fix meshnet not enabling after account switch

EmilijusS:
Target - release branch

### DIFF
--- a/meshnet/register.go
+++ b/meshnet/register.go
@@ -1,11 +1,13 @@
 package meshnet
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"sync"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/core"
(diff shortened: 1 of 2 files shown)


meshnet/register.go
```diff
@@ -117,6 +119,18 @@ func (r *RegisteringChecker) register(cfg *config.Config) error {
 		OS:              cmesh.OperatingSystem{Name: "linux", Distro: distroName},
 		SupportsRouting: true,
 	})
+	if errors.Is(err, core.ErrConflict) {
+		// We try to keep the same keys as long as possible, but if relogin with different account happens
+		// then they have to be regenerated. There's no way to check if the current mesh device data
```
mariusSincovici, line 124: it is possible to fetch the meshnet map and compare the public key for user devices with the current public key. If none matches then the data is invalid.
But that might be too much logic for this case, if API already reported the error :smile:

mariusSincovici (approved):

Savolro (approved):

PR #137, opened 2023-10-27
Fix VPN disconnected after set meshnet off

piotrjurkiewicz:
When connecting to a Meshnet peer as an VPN exit, the Meshnet server `s.isPeerConnected` variable was set to `true`. Then this variable was being used during Meshnet disable (`nordvpn set mesh off`) to determine whether we are connected to a Meshnet peer to stop that connection along with Meshnet disable.

However, user can in the meantime connect to a regular VPN server (and thus disconnect from a Meshnet peer). The `s.isPeerConnected` variable was not updated, so Meshnet disabling was resulting in regular VPN disconnect in such case.

The fix is to not use `s.isPeerConnected` variable to remember if we had connected to Meshnet peer, but to check during Meshnet disable the address of currently connected VPN and stop connection only if address indicates that we are connected to a Meshnet peer.

### DIFF
--- a/meshnet/server.go
+++ b/meshnet/server.go
@@ -48,7 +48,7 @@ type Server struct {
 	subjectMeshSetting events.Publisher[bool]
 	subjectConnect     events.Publisher[events.DataConnect]
 	lastPeers          string
-	isPeerConnected    bool
+	lastConnectedPeer  string
 	fileshare          service.Fileshare
 	scheduler          *gocron.Scheduler
 	pb.UnimplementedMeshnetServer
@@ -318,12 +318,11 @@ func (s *Server) DisableMeshnet(context.Context, *pb.Empty) (*pb.MeshnetResponse
 		s.pub.Publish(fmt.Errorf("disabling fileshare: %w", err))
 	}
 
-	// stop networker only if mesh peer connected before
-	if s.isPeerConnected {
+	// try to stop networker only if mesh peer connected before
+	if s.netw.LastServerName() == s.lastConnectedPeer {
 		if err := s.netw.Stop(); err != nil {
 			s.pub.Publish(fmt.Errorf("disconnecting: %w", err))
 		}
-		s.isPeerConnected = false
 	}
 
 	if err := s.netw.UnSetMesh(); err != nil {
@@ -2671,7 +2670,7 @@ func (s *Server) Connect(
 			},
 		}, nil
 	}
-	s.isPeerConnected = true
+	s.lastConnectedPeer = peer.Hostname
 	s.subjectConnect.Publish(events.DataConnect{
 		IsMeshnetPeer: true,
 	})
--- a/test/qa/test_meshnet.py
+++ b/test/qa/test_meshnet.py
@@ -1,4 +1,4 @@
-from lib import daemon, info, logging, login, meshnet, ssh
+from lib import daemon, info, logging, login, meshnet, network, ssh
 import lib
 import sh
 import requests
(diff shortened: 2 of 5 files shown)


keliramu (approved):

EmilijusS (changes requested):
So what was the problem and how did the fix help? It's useful to describe such things in PR description.

meshnet/server.go
```diff
@@ -319,11 +318,10 @@ func (s *Server) DisableMeshnet(context.Context, *pb.Empty) (*pb.MeshnetResponse
 	}
 
 	// stop networker only if mesh peer connected before
-	if s.isPeerConnected {
+	if strings.HasSuffix(s.netw.LastServerName(), ".nord") {
```
EmilijusS, line 321: I'm not a fan of this because Meshnet package shouldn't know about the VPN server naming conventions (which also might change, who knows). 
Suggestion: we should still keep information about being connected to a peer, but instead of it being just a boolean which doesn't tell us much we could keep the peer's hostname. Then in this condition we could check whether we are still connected to that hostname. It's still a bit bad because it relies on "LastServerName" returning the hostname and nothing else, but at least we would be doing checks only with meshnet related information. 

Also, the comment above needs to be updated.

test/qa/test_meshnet.py
```diff
@@ -277,6 +277,19 @@ def check_rules_routing() -> (bool, str):
...
 
+def test_connect_set_mesh_off():
+    output = f"{sh.nordvpn.mesh.peer.list(_tty_out=False)}"
+    peer = meshnet.get_peers(output)[0]
+    sh.nordvpn.mesh.peer.connect(peer)
+    sh.nordvpn.disconnect()
+    sh.nordvpn.connect()
+    sh.nordvpn.set.mesh.off()
+
+    with lib.Defer(sh.nordvpn.set.mesh.on):
```
mariusSincovici, line 288: why is needed to have meshnet `on` at the end of the test?

piotrjurkiewicz: To bring daemon back to the same state as after `setup_module`.

meshnet/server.go
```diff
@@ -318,12 +319,11 @@ func (s *Server) DisableMeshnet(context.Context, *pb.Empty) (*pb.MeshnetResponse
 		s.pub.Publish(fmt.Errorf("disabling fileshare: %w", err))
 	}
 
-	// stop networker only if mesh peer connected before
-	if s.isPeerConnected {
+	// stop networker only if we are connected to mesh peer
+	if s.wasPeerConnected && s.netw.LastServerName() == s.lastConnectedPeer {
```
EmilijusS, line 323: 2 more things:
1. I don't think `wasPeerConnected` is necessary (sorry if I hadn't made that clear before). Well maybe in the case when no connections were made at all, maybe then both would be equal to empty string, which brings me to the second point...
2. What happens if we connect to peer, disconnect and then disable mesh? It seems to me by this logic that it will still try to stop the networker even though it is already stopped. Nothing bad will happen it seems, only an error will be logged, but it would still be nice to avoid that.

piotrjurkiewicz: 1. Yes, when no connections were made, both equal to empty strings, so I kept the boolean to avoid calling `Stop` in such case.
2. You are right, `Stop` in being called in such case. However, no error is emitted, because `Stop` checks with appropriate locks whether the VPN is connected before calling its actual implementation: https://github.com/NordSecurity/nordvpn-linux/blob/main/networker/networker.go#L412-L422

So maybe the `wasPeerConnected` should be removed and it should rely `if netw.isVpnSet` check inside `Stop` in both cases?

EmilijusS: At this point it's a very minor thing then, but I guess yeah, `wasPeerConnected` can be removed.

EmilijusS (approved):

PR #139, opened 2023-10-27
Client config removal

Savolro:


bartoszWojciechO (approved):

cli/cli_allowlist_add_port.go
```diff
@@ -36,54 +33,49 @@ func (c *cmd) AllowlistAddPort(ctx *cli.Context) error {
...
 
-	if !success {
-		return formatError(fmt.Errorf(AllowlistAddPortExistsError, data...))
+	settings, err := c.getSettings()
+	if err != nil {
+		return formatError(err)
+	}
+	allowlist := settings.GetAllowlist()
+	if isTCP {
+		allowlist.Ports.Tcp = append(allowlist.Ports.Tcp, int64(port))
```
mariusSincovici, line 72: Isn't needed to check for duplicates?

cli/cli_set_technology.go
```diff
@@ -68,13 +68,11 @@ func (c *cmd) SetTechnology(ctx *cli.Context) error {
 	case internal.CodeSuccess:
 		flag, _ := strconv.ParseBool(resp.Data[0])
 		color.Green(fmt.Sprintf(MsgSetSuccess, "Technology", strings.Join(resp.Data[1:], " ")))
-		c.config.Technology = tech
-		c.config.Obfuscate = false
```
mariusSincovici, line 72: The `obfuscate` value is not set to `false` on the daemon side when switching to Nordlynx and because of this, when it is `true` for OpenVPN, switching to Norlynx will break all the commands that use `obfuscated`, e.g. `nordvpn cities`, `countries`

PR #140, opened 2023-10-30
Use cached data for remote config when fetching fails

mariusSincovici:

* use cached data as fallback for remote config,when fetching new data fails
* define a default config for telio that also includes direct connections, if fetching fails and there is no cached data

### DIFF
--- a/config/remote/firebase.go
+++ b/config/remote/firebase.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/coreos/go-semver/semver"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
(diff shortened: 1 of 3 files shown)


config/remote/firebase.go
```diff
@@ -63,56 +64,84 @@ func NewRConfig(updatePeriod time.Duration, rs RemoteConfigService, cm config.Ma
 func (rc *RConfig) fetchRemoteConfigIfTime() error {
 	var cfg config.Config
 	if err := rc.configManager.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, "failed to load the cached config:", err)
```
EmilijusS, line 67: Should either log the error or return it, never both. If it needs to be logged, it should be logged by the caller then.

config/remote/firebase.go
```diff
@@ -63,56 +64,84 @@ func NewRConfig(updatePeriod time.Duration, rs RemoteConfigService, cm config.Ma
...
-		err = json.Unmarshal(remoteConfigValue, rc.config)
-		if err != nil {
-			return err
-		}
-		if err := rc.configManager.SaveWith(func(c config.Config) config.Config {
-			s, err := json.Marshal(rc.config)
-			if err == nil {
-				c.RemoteConfig = string(s)
-				c.RCLastUpdate = time.Now()
+	if cfg.RemoteConfig == "" || time.Now().After(cfg.RCLastUpdate.Add(rc.updatePeriod)) {
```
EmilijusS, line 71: This is not a good change, because RC fetching will become more frequent in case of errors. Before it was a hard limit that one fetch is done per day whether it is successful or not, but now fetching will be done constantly if they are unsuccessful.

keliramu:
This fix is not for `release/3.16.7`

bartoszWojciechO (approved):

piotrjurkiewicz (approved):

PR #142, opened 2023-10-31
Add firewall manager

bartoszWojciechO:


daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,599 @@
...
+			return fmt.Errorf("removing fileshare allow rule: %w", err)
+		}
+	}
+
+	delete(f.fileshareRules, peer.UID)
+	return nil
+}
+
+func (f *FirewallManager) addBlockTrafficRules() error {
+	interfaces, err := f.devices()
```
mariusSincovici, line 208: if a new interface is connected, e.g. usb adapter, will this be updated?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,599 @@
...
+		}
+	}
+
+	f.trafficBlocked = true
+
+	return nil
+}
+
+func (f *FirewallManager) removeBlockTrafficRules() error {
+	interfaces, err := f.devices()
```
mariusSincovici, line 247: Same here, if the list of devices is different from the one in enable it will result in an error, because the rule doesn't exist and it will not remove the rules from the remaining ones. 

Wouldn't be better to have a map, like for peers, with the interfaces and the rules?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,599 @@
...
+func (f *FirewallManager) AllowIncoming(peer meshnet.UniqueAddress, allowLocal bool) error {
+	if _, ok := f.allowIncomingRules[peer.UID]; ok {
+		return fmt.Errorf("allow incoming rule for the given peer already exists")
+	}
+
+	rule := fmt.Sprintf("INPUT -s %s/32 -m comment --comment nordvpn -j ACCEPT", peer.Address)
+
+	blockLANRules := []string{}
+	if !allowLocal {
+		lans := []string{
```
mariusSincovici, line 308: What would happen if the user doesn't have one of the RFC IP ranges on the LAN?

mariusSincovici (commented):
Not sure how common is, but in case of errors is it possible to have duplicate rules added to the iptables? or this is the responsibility of the caller to ensure this is not happening? 
e.g. if in the middle of the list from `addBlockTrafficRules` there is an error, calling second time the function, without calling before `remove...` will duplicate the already added rules.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,599 @@
...
+		devices:            devices,
+		allowIncomingRules: make(map[string]allowIncomingRule),
+		fileshareRules:     make(map[string]string),
+		connmark:           connmark,
+		enabled:            enabled,
+	}
+}
+
+func (f *FirewallManager) Disable() error {
+	if !f.enabled {
```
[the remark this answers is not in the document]
bartoszWojciechO: No need, as the caller(`networker`) is protected.

daemon/firewall/firewall_manager.go
bartoszWojciechO: LAN block will not work for them, i.e it will be routable. I guess we could(or even should) make this more sophisticated with arp, but this it out of scope for this PR.

daemon/firewall/firewall_manager.go
bartoszWojciechO: New devices will be returned, but the rules themselves won't be updated. I think we should have a task for that, since this issue also exists on main.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,599 @@
...
+
+type PortRange struct {
+	min int
+	max int
+}
+
+type FirewallManager struct {
+	commandExecutor      CommandExecutor
+	devices              device.ListFunc              // list network interfaces
+	allowIncomingRules   map[string]allowIncomingRule // peer public key to allow incoming rule
```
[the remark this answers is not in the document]
bartoszWojciechO: Well, I checked it and `meshnet.UniqueAddress` defines peers `UID` as string. go will not cast it implicitly to `PeerPubKey`, so the tradeoff to this approach would be having a bunch of casts from string to string alias.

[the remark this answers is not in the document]
mariusSincovici: yeah, that was my concern, not to add too many casts.
then maybe we can do a refactor, if needed, in the future to use same type alias in all places.

mariusSincovici (approved):

PR #145, opened 2023-11-02
Do not change local obfuscation setting on technology change

piotrjurkiewicz:
CLI client sets obfuscate to false on `set technology` in its own config copy, without sending a RPC to daemon to really disable it. This pull request removes changing local obfuscation settings on technology change. This means that obfuscation will remain enabled for OpenVPN when changing a technology to NordLynx. When user will change back to OpenVPN, obfuscation will be enabled.

### DIFF
--- a/cli/cli_set_technology.go
+++ b/cli/cli_set_technology.go
@@ -69,7 +69,6 @@ func (c *cmd) SetTechnology(ctx *cli.Context) error {
 		flag, _ := strconv.ParseBool(resp.Data[0])
 		color.Green(fmt.Sprintf(MsgSetSuccess, "Technology", strings.Join(resp.Data[1:], " ")))
 		c.config.Technology = tech
-		c.config.Obfuscate = false
 		if flag {
 			color.Yellow(SetReconnect)
 		}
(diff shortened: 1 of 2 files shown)


cli/cli_set_technology.go
```diff
@@ -69,7 +69,6 @@ func (c *cmd) SetTechnology(ctx *cli.Context) error {
 		flag, _ := strconv.ParseBool(resp.Data[0])
 		color.Green(fmt.Sprintf(MsgSetSuccess, "Technology", strings.Join(resp.Data[1:], " ")))
 		c.config.Technology = tech
-		c.config.Obfuscate = false
```
mariusSincovici, line 72: The application uses this bool also for `Nordlynx`. Not setting it to false will break the commands that are using it.
 e.g.: `nordvpn countries`, `nordvpn cities Germany`

mariusSincovici:
PR https://github.com/NordSecurity/nordvpn-linux/pull/139 will remove the client settings, so maybe it will fix also this issue?

piotrjurkiewicz:
Yes, it should fix this issue also.

PR #149, opened 2023-11-07
Update RCLastUpdate before fetching remote config

mariusSincovici:

Before fetching update RCLastUpdate to be sure that the remote configuration is fetched only at the specified interval.


config/remote/firebase.go
```diff
@@ -92,6 +92,15 @@ func (rc *RConfig) fetchRemoteConfigIfTime() error {
 }
 
 func (rc *RConfig) fetchAndSaveRemoteConfig(cfg config.Config) error {
```
EmilijusS, line 94: What is the point of `cfg` being passed here? I see it is being reloaded, but as the argument is not a pointer it is a copy so in essence nothing happens. If the config has to be returned, the function signature should be `func (rc *RConfig) fetchAndSaveRemoteConfig() (config.Config, error)`

bartoszWojciechO (approved):

config/remote/firebase.go
EmilijusS: Also, this whole function just looked confusing to me. I took a shot at rewriting it (used named return variables, I'm not a big fan of using them but seemed to fit well in this case), use it if you like it.
```
func (rc *RConfig) fetchAndSaveRemoteConfig() (cfg config.Config, err error) {
	var remoteConfigValue string

	// Always save the last update time and also save the remote config if it was retrieved successfully
	defer func(){
		cfgSaveErr := rc.configManager.SaveWith(func(c config.Config) config.Config {
			c.RCLastUpdate = time.Now()
			if err != nil {
				c.RemoteConfig = remoteConfigValue
			}
			cfg = c
			return c
		})
		err = errors.Join(err, cfgSaveErr)
	}()

	remoteConfigValue, err = rc.remoteService.FetchRemoteConfig()
	if err != nil {
		return fmt.Errorf("fetching the remote config failed: %w", err)
	}
	err = json.Unmarshal(remoteConfigValue, rc.config)
	if err != nil {
		return fmt.Errorf("parsing the fetched remote config failed: %w", err)
	}
}
```

mariusSincovici: yes, you're right it's confusing. thanks. :smiley: 
I've change using your example, and the function is just fetching and saving the new config returning it. Change not to modify `rc.config`, because maybe `SaveWith` fails and then function returns error but it also modifies things. The caller function is already doing that.

EmilijusS (approved):

PR #155, opened 2023-11-09
Automate login callback test using selenium

ggediminass:


test/qa/lib/login.py
```diff
@@ -51,3 +85,61 @@ def login_as(username, ssh_client: ssh.Ssh = None):
...
+
+class SeleniumBrowser():
+    def __init__(self, preferences:list=None):
+        self.options = webdriver.FirefoxOptions()
+        self.options.binary_location = str(sh.which("firefox"))
+        self.options.add_argument('--headless')
+        self.options.add_argument('--no-sandbox')
+
+        if preferences != None:
+            for preference, value in preferences:
```
mariusSincovici, line 98: [minor] can't you use `self.browser_set_preferences()` instead?

test/qa/test_login.py
```diff
@@ -137,3 +138,81 @@ def test_logout_disconnects():
...
+            selenium.browser_element_interact(login.NA_PASSWORD_PAGE_BUTTON_XPATH)
+
+            # Continue to app page
+            # preferences not set in constructor, so when we click link it does not redirect us to app.
+            callback_link = selenium.browser_element_interact(login.NA_CONTINUE_PAGE_LINK_BUTTON, return_attribute='href')
+        except:
+            browser.save_screenshot(login.BROWSER_LOGS_PATH + "Screenshot.png")
+            pytest.fail()
+
+        assert login.LOGIN_MSG_SUCCESS in sh.nordvpn.login("--callback", callback_link)
```
mariusSincovici, line 217: If the message changes in the app then `logout` will not be called, and it will remain logged in. wouldn't be better to execute the functions in a `defer` block?

test/qa/lib/login.py
ggediminass: Whoops :smile: seems like I left dead code behind, as `browser_set_preferences()` is not really used anywhere. If I am going to use it one single time, perhaps I should get rid of the method itself? :smile:

test/qa/test_login.py
ggediminass: Good idea!

test/qa/lib/login.py
mariusSincovici: if you're not using it, you can just remove the method :smile:

mariusSincovici (approved):

bartoszWojciechO (approved):

PR #161, opened 2023-11-14
Add FirewallManager

bartoszWojciechO:


daemon/firewall/firewall_manager_test.go
```diff
@@ -0,0 +1,995 @@
...
+
+var ErrIptablesFailure = errors.New("iptables failure")
+
+type IptablesMock struct {
+	isErr       bool
+	Commands    []string
+	errCommands map[string]bool
+}
+
+func NewIptablesMock(isErr bool) IptablesMock {
```
bartoszWojciechO, line 25: We have to add IptablesMock here, temporarily, against our uts conventions. Adding it in firewall mock directory will cause cyclical dependency, because `nordvpn-linux/test/mock/firewall/firewall.go` depends on the firewall package:
```
// Add and apply firewall rules
func (mf *FirewallMock) Add(rules []firewall.Rule) error {
	mf.Rules = append(mf.Rules, rules...)

	return nil
}
```

This will be fixed once new firewall layer is integrated into networker, because we will be able to remove the old mock.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+}
+
+type FirewallManager struct {
+	commandExecutor      IptablesExecutor
+	devices              device.ListFunc              // list network interfaces
+	allowIncomingRules   map[string]allowIncomingRule // peer public key to allow incoming rule
+	fileshareRules       map[string]string            // peers public key to allow fileshare rule
+	allowlistRules       []string
+	trafficBlockRules    []string
+	connmark             uint32
```
Savolro, line 87: No way to update this during a runtime. Either add a Set function or make this as an argument where needed

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+	meshnetDeviceAddress string // used for unblocking meshnet after if has been blocked and for tracking meshnet block state
+	enabled              bool
```
Savolro, line 89: What is the reason this is kept here? Why not just make it always add rules and create a different implementation for enabling/disabling by calling the right exported functions?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+}
+
+type PortRange struct {
+	min int
+	max int
+}
+
+type FirewallManager struct {
+	commandExecutor      IptablesExecutor
+	devices              device.ListFunc              // list network interfaces
```
Savolro, line 82: Any reason why is this a func pointer and not interface? Also, why not just make it an argument?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+}
+
+func (f *FirewallManager) Disable() error {
+	if !f.enabled {
+		return fmt.Errorf("firewall is already disabled")
+	}
+
+	// remove traffic block
+	if err := f.removeBlockTrafficRules(); err != nil {
+		log.Printf("unblocking traffic: %s", err.Error())
```
Savolro, line 110: I don't think it should use Println dirrectly. If we still want to execute everything, maybe it would make sense to use https://pkg.go.dev/errors#example-Join and return a joined error?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+		}
+	}
+
+	f.enabled = true
+
+	return nil
+}
+
+// AllowFileshare adds ACCEPT rule for all incoming connections to tcp port 49111 from the peer with given UniqueAddress.
+func (f *FirewallManager) AllowFileshare(peer meshnet.UniqueAddress) error {
```
Savolro, line 202: Full UniqueAddress is not used anywhere. UID should be enough

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+
+	delete(f.allowIncomingRules, peerUID)
+
+	return nil
+}
+
+func (f *FirewallManager) removeMeshnetBlockRules(deviceAddress string) error {
+	// -D INPUT -s 100.64.0.0/10 -m conntrack --ctstate RELATED,ESTABLISHED --ctorigsrc <device address> -m comment --comment nordvpn -j ACCEPT
+	// -D INPUT -s 100.64.0.0/10 -m comment --comment nordvpn -j DROP
+	command := fmt.Sprintf("-D INPUT -s 100.64.0.0/10 -m conntrack --ctstate RELATED,ESTABLISHED --ctorigsrc %s -m comment --comment nordvpn -j ACCEPT", deviceAddress)
```
Savolro, line 381: Please move 10.64.0.0/10 to a constant

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+	return nil
+}
+
+// AllowFileshare adds ACCEPT rule for all incoming connections to tcp port 49111 from the peer with given UniqueAddress.
+func (f *FirewallManager) AllowFileshare(peer meshnet.UniqueAddress) error {
+	if _, ok := f.fileshareRules[peer.UID]; ok {
+		return ErrRuleAlreadyActive
+	}
+
+	rule := fmt.Sprintf("INPUT -s %s/32 -p tcp -m tcp --dport 49111 -m comment --comment nordvpn -j ACCEPT", peer.Address.String())
```
Savolro, line 207: Please move 49111 to a constant

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
+package firewall
```
Savolro, line 1: Please rename `api` and `Api` to `API` where it does not unnecessarily export code. Also, please update where it is mentioned in comments

daemon/firewall/firewall_manager_test.go
```diff
@@ -0,0 +1,995 @@
...
+
+	err := firewallManager.BlockMeshnet(deviceAddress)
+	assert.NoError(t, err, "Unexpected error when blocking meshnet.")
+	iptablesMock.PopCommands()
+
+	err = firewallManager.BlockMeshnet(deviceAddress)
+	assert.ErrorIs(t, err, ErrRuleAlreadyActive)
+	iptablesMock.PopCommands()
+
+	// using different device address shoudl yield the same result
```
Savolro, line 627: `s/shoudl/should`

daemon/firewall/firewall_manager_test.go
```diff
@@ -0,0 +1,995 @@
+package firewall
```
Savolro, line 1: Please add `category.Set` to each of the tests

Savolro (changes requested):

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+)
+
+var ErrRuleAlreadyActive = errors.New("this rule is already active")
+
+const (
+	iptables  = "iptables"
+	ip6tables = "ip6tables"
+)
+
+type IptablesExecutor interface {
```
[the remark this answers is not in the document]
bartoszWojciechO: Moved here temporarily, will be moved into the external `test` module after integration with the `networker`.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,621 @@
...
+	allowIncomingRule string
+	blockLANRules     []string
+}
+
+type PortRange struct {
+	min int
+	max int
+}
+
+type FirewallManager struct {
```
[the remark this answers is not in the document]
bartoszWojciechO: Is it needed? `networker` is already thread safe, adding a mutex here will result in interlocking.

daemon/firewall/firewall_manager.go
bartoszWojciechO: Old implementation, was always provided as func pointer. Adding it as a member will allow us to remove any references to device listing from the `networker`.

daemon/firewall/firewall_manager.go
bartoszWojciechO: IP address is used for actually adding the rule, see line 207

daemon/firewall/firewall_manager.go
bartoszWojciechO: As it is now, `connmark` is never updated during runtime. App tells the user to restart the daemon after configuring it. This implementation mostly tries to replicate functionality of the old implementation, just in a simpler way. Should it allow for dynamic `connmark` configuration?

Savolro: Well, we have a configuration for it so it should be handled during runtime without full daemon restart. Anything in between is on the table. Since this is likely never re-created, connmark will never be updated during runtime

bartoszWojciechO: I guess there's also more than goes into connmark update than just firewall. By skimming the code, routing and domain rotator use it, neither can be dynamically updated. So the scope here is wider than just firewall.

PR #162, opened 2023-11-14
Add FirewallManager

bartoszWojciechO:


daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// -I INPUT -i <iface> -m comment --comment nordvpn -j DROP
+	// -I OUTPUT -o <iface> -m comment --comment nordvpn -j DROP
+	for _, iface := range interfaces {
+		inputCommand := fmt.Sprintf("INPUT -i %s -m comment --comment nordvpn -j DROP", iface.Name)
+		outputCommand := fmt.Sprintf("OUTPUT -o %s -m comment --comment nordvpn -j DROP", iface.Name)
+		f.trafficBlockRules = append(f.trafficBlockRules, inputCommand)
+		f.trafficBlockRules = append(f.trafficBlockRules, outputCommand)
+
+		if f.enabled {
```
mariusSincovici, line 109: when enabled, if one of the commands fails, then `removeBlockTrafficRules` will also fail because  `f.trafficBlockRules` will have rules that are not inserted into the firewall. 
Since the rules are appended at the end of `f.trafficBlockRules` probably the firewall will not remain in a bad state, but `removeBlockTrafficRules` will return error for rule not found.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,316 @@
...
+	// -A INPUT -i <interface> -p <protocol> -m <protocol> --dport <port> -m comment --comment nordvpn -j ACCEPT
+	// -A INPUT -i <interface> -p <protocol> -m <protocol> --sport <port> -m comment --comment nordvpn -j ACCEPT
+	// -A OUTPUT -o <interface> -p <protocol> -m <protocol> --sport <port> -m comment --comment nordvpn -j ACCEPT
+	// -A OUTPUT -o <interface> -p <protocol> -m <protocol> --dport <port> -m comment --comment nordvpn -j ACCEPT
+	inputDportRule := fmt.Sprintf("INPUT -i %s -p %s -m %s --dport %d:%d -m comment --comment nordvpn -j ACCEPT", iface, protocol, protocol, portRange.min, portRange.max)
+	inputSportRule := fmt.Sprintf("INPUT -i %s -p %s -m %s --sport %d:%d -m comment --comment nordvpn -j ACCEPT", iface, protocol, protocol, portRange.min, portRange.max)
+	outputDportRule := fmt.Sprintf("OUTPUT -o %s -p %s -m %s --dport %d:%d -m comment --comment nordvpn -j ACCEPT", iface, protocol, protocol, portRange.min, portRange.max)
+	outputSportRule := fmt.Sprintf("OUTPUT -o %s -p %s -m %s --sport %d:%d -m comment --comment nordvpn -j ACCEPT", iface, protocol, protocol, portRange.min, portRange.max)
+
+	if f.enabled {
```
mariusSincovici, line 183: here it might be a problem, if one of the commands fails, the ones that passed are not added to the list => `UnsetAllowlist` will not be able to remove them.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,316 @@
...
+			if f.enabled {
+				if err := f.commandExecutor.ExecuteCommand("-I " + inputRule); err != nil {
+					return fmt.Errorf("adding input accept rule for subnet: %w", err)
+				}
+				if err := f.commandExecutor.ExecuteCommand("-I " + outputRule); err != nil {
+					return fmt.Errorf("adding output accept rule for subnet: %w", err)
+				}
+			}
+
+			f.allowlistRules = append(f.allowlistRules, inputRule)
```
mariusSincovici, line 229: same here.
Since there are more places, it might be easier to have some helper function that inserts a rule into the FW when enabled and appends into a list on success?

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,316 @@
...
+	return nil
+}
+
+func (f *FirewallManager) manageApiAllowlist(allow bool) error {
+	iptablesMode := "-I "
+	if !allow {
+		iptablesMode = "-D "
+	}
+
+	ifaces, err := f.devices()
```
mariusSincovici, line 275: Shouldn't this also be stored into a list? If the devices list changes between add and remove then the remove will fail, leaving rules into the FW.

daemon/firewall/firewall_manager.go
bartoszWojciechO: Fixed, but not detected by GH as outdated.

daemon/firewall/firewall_manager.go
bartoszWojciechO: Fixed, not detected by GH as outdated.

mariusSincovici (approved):

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,333 @@
...
+var ErrRuleAlreadyActive = errors.New("this rule is already active")
+
+const (
+	iptablesCommand  = "iptables"
+	ip6tablesCommand = "ip6tables"
+)
+
+// IptablesExecutor is an abstraction over iptables command. It should accept any input that is also viable for iptables
+// command.
+type IptablesExecutor interface {
```
Savolro, line 23: I know it would require some rework. But could this have a priority functionality. E. g. instead of always inserting to the top, we would need to calculate index based on priority and insert to the right place with `-I index`.  I see 2 options for this:
1. We can use `comment` module for storing priority in `iptables`. If we do this, we could have fixed priorities for each of the rule types and not care about correct function call order anymore. So it would look like this:
```go
InsertRule(command string, priority int) error
```
2. We can use `comment` module for specific key of a rule. E. g. `nordvpn_block_traffic` and then have a priority list somewhere in the executor. Then it would look like this:
```go
func NewIPTablesExecutor(order []string) *IPTablesExecutor { ... }
// ...
InsertRule(command string, key string) error
```
Then make sure that rule with a given key is always inserted in between the already existing rules with other keys in the order.

Another benefit of this would be that we don't need to provide a full rule to `DeleteRule` call and could have it as
```go
DeleteRule(priority int) error
// or
DeleteRule(key string) error
```

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,333 @@
...
+	DeleteRuleIPv6(command string) error
+	// Enable changes the Executor state so that it performs the commands provided to ExecuteCommand and
+	// ExecuteCommandIPv6.
+	Enable()
+	// Disable changes the Executor state so that it does not perform commands provided to ExecuteCommands and
+	// ExecuteCommandIPv6.
+	Disable()
+}
+
+type Iptables struct {
```
Savolro, line 36:
```suggestion
type IPTables struct {
```
And same for the all functions

Savolro (changes requested):

daemon/firewall/firewall_manager.go
bartoszWojciechO: Added a commit with priority tracking, to visualize how it would look like of such approach(didn't adjust uts yet though).

It increases complexity quite a bit. State management/book keeping in `FirewallManager` got more involved and `IptablesExecutor` got more complicated, so we'd probably have create a layer of abstraction over `Exec` calls so that it can be unit tested.

What is the benefit of using priority system over ensuring ordering in the implementation?

Rule deletion needs to stay the same, for some types of rules(meshnet peer permissions), we do not want to delete all the rules with the same key/priority. So we still need to provide specific rules.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,333 @@
+package firewall
```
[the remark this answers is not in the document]
bartoszWojciechO: `iptables` subpackage could be removed eventually.`firewall` will be greatly simplified, we'll have `firewall_manager` and maybe `allowlist_routing`.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+func (f FwRule) ToDeleteCommand() string {
+	return fmt.Sprintf("-D %s %s %s", f.chain, f.params, f.priority.toCommentArgs())
+}
+
+type FirewallManager struct {
+	iptablesManager   iptablesManager
+	devices           device.ListFunc // list network interfaces
+	allowlistRules    []FwRule
```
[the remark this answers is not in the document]
bartoszWojciechO: We need to keep track of the actual rules for the `enable`/`disable` functionality. When firewall is disabled we still want to store the rules so that they can be re-applied once it is enabled.

daemon/firewall/firewall_manager.go
```diff
@@ -0,0 +1,502 @@
...
+//     rules.
+//  2. Last nordvpn rule in the chain is our boundary. New rule should be inserted either above it, or at most at one
+//     line bellow it. This enforces priority of nordvpn rules over non-nordvpn rules.
+//  3. Non-nordvpn rules located between last nordvpn rule in the chain and the first rule in the chain are ignored.
+func (i iptablesManager) getRuleLine(command string, chain iptablesChain, priority rulePriority) (int, error) {
+	// Run command with --numeric to avoid reverse DNS lookup. This takes a long time and is unecessary for the purpose
+	// of line number calculation(we ignore everything but the 'nordvpn-<priority>' comment or the lack of thereof).
+	args := "-L " + chain.String() + " --numeric"
+
+	output, err := i.cmdRunner.runCommand(command, args)
```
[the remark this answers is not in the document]
bartoszWojciechO: What is the desired simplification? Offload conversion of the output from `[]byte` to `string` to the caller?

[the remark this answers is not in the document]
Savolro: We would need to have implementations for this cmdRunner. Less implementations and simplier testing

daemon/firewall/firewall_manager.go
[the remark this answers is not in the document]
bartoszWojciechO: Created a sub package to avoid circular dependency shenanigans. Will move it to `iptables`(or just rename the new package) once we can get rid of the old stuff and restructure.

Savolro (approved):

PR #165, opened 2023-11-15
Use default MTU when there is no default gateway

mariusSincovici:
If the app is not able to use default gateway's MTU use as fallback `1500`


internal/command.go
```diff
@@ -0,0 +1,32 @@
+package internal
+
+import "os/exec"
+
+type Command interface {
```
mariusSincovici, line 5: I didn't saw to have something similar, if there is let me know and I'll remove this.

bartoszWojciechO: Exitnode firewall has something similar: https://github.com/NordSecurity/nordvpn-linux/blob/main/meshnet/exitnode/fw.go#L22
It's not 1:1 but maybe it would be worth it to have single abstraction over `exec` across the code. I like yours better to be honest, it's more robust.

mariusSincovici: thanks for info.
I'm changing the code and I won't need this anymore. So I'll delete it for now.

Savolro (approved):

keliramu (approved):

PR #167, opened 2023-11-16
Automate missing manual login tests

naujalyte:


bartoszWojciechO (approved):

test/qa/test_login.py
```diff
@@ -137,3 +137,53 @@ def test_logout_disconnects():
...
+
+
+def test_invalid_url_callback_login():
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        sh.nordvpn.login("--callback", "https://www.google.com/")
+    
+    assert "Expected a url with nordvpn scheme." in str(ex.value)
+
+
+def test_repeated_login_callback():
```
mariusSincovici, line 162: If for some reason one of the asserts fails, the next 3 tests will not execute the `logout` part, and maybe another test will do the logout without `--persist-token`. wouldn't be better to wrap them in a defer logout block?

naujalyte: Good point, thanks! Fixed.

mariusSincovici (approved):

PR #168, opened 2023-11-17
Clean NordVPN rules from iptables after daemon crash

piotrjurkiewicz:


contrib/systemd/system/nordvpnd.service
```diff
@@ -6,6 +6,8 @@ Wants=network-online.target
 
 [Service]
 ExecStart=/usr/sbin/nordvpnd
+ExecStopPost=/bin/sh -c "iptables-save -c | grep -v '\\-\\-comment nordvpn' | iptables-restore -c"
```
mariusSincovici, line 9: If the daemon crashes, wouldn't this result in leaks for ~5 seconds, until the OS restarts it?

PR #172, opened 2023-11-20
Handle SIGHUP gracefully

piotrjurkiewicz:


internal/wait.go
```diff
@@ -10,6 +10,6 @@ import (
 // WaitSignal for app to shutdown
 func WaitSignal() {
 	signals := make(chan os.Signal, 1)
-	signal.Notify(signals, os.Interrupt, linux.SIGTERM)
+	signal.Notify(signals, os.Interrupt, linux.SIGTERM, linux.SIGHUP)
```
mariusSincovici, line 13: Shouldn't it handle all the standard signals?

mariusSincovici (approved):

keliramu (approved):

internal/wait.go
piotrjurkiewicz: These are common user-generated signals which can be expected. Maybe `SIGUSR1` and `SIGUSR2` should be added to the list. But I don't think that all standard signals should be handled, as the rest of them are generated by the system and have some special meaning. And `SIGKILL` cannot be caught anyway.

PR #174, opened 2023-11-21
Store r.tableID also when fwmark rule is already present

piotrjurkiewicz:
When fwmark rule already exists in the system, `r.tableID` is not set to the actual table ID. Instead it stays initialized to `0`, which means that new routes are then added to the `main` table instead of table `205` (or any custom table).

This pull requests changes `checkFwmarkRule` function to `findFwmarkRule`. It will check if fwmark rule already exist and find its table ID. This table ID will be than stored in `r.tableID`.

### DIFF
--- a/daemon/routes/iprule/iprule.go
+++ b/daemon/routes/iprule/iprule.go
@@ -63,21 +63,20 @@ func (r *Router) SetupRoutingRules(
 	}()
 
 	for _, ipv6 := range ipv6EnabledList {
-		fwmarkRulePresent, err := checkFwmarkRule(r.fwmark, ipv6)
+		routingTableID, err := findFwmarkRule(r.fwmark, ipv6)
 		if err != nil {
 			return err
 		}
 
-		if !fwmarkRulePresent {
+		if routingTableID == 0 {
 			fwMarkRuleID, err := calculateRulePriority(ipv6)
 			if err != nil {
 				return err
 			}
-			routingTableID, err := calculateCustomTableID(ipv6)
+			routingTableID, err = calculateCustomTableID(ipv6)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


daemon/routes/iprule/iprule.go
```diff
@@ -345,8 +345,8 @@ func addFwmarkRule(
 	return nil
 }
 
-// checkFwmarkRule check if fwmark rule is set
-func checkFwmarkRule(fwMarkVal uint32, ipv6 bool) (bool, error) {
+// findFwmarkRule check if fwmark rule is set and find its table ID
+func findFwmarkRule(fwMarkVal uint32, ipv6 bool) (bool, uint, error) {
```
mariusSincovici, line 349: Is the boolean needed, because when is false it also returns error?

daemon/routes/iprule/iprule.go
```diff
@@ -368,22 +368,30 @@ func checkFwmarkRule(fwMarkVal uint32, ipv6 bool) (bool, error) {
...
+			tbldID, err := strconv.ParseUint(words[len(words)-1], 10, 64)
+			if err != nil {
+				return false, 0, fmt.Errorf("converting '%s' to uint: %w", words[len(words)-1], err)
+			}
+			return true, uint(tbldID), nil
 		}
 	}
 
-	return false, nil
+	return false, 0, nil
```
mariusSincovici, line 394: It might be better to return some custom created error in this case, e.g. `not found`, and the caller to check the error type and decide what to do?

Savolro (changes requested):

keliramu (approved):

Savolro (approved):

PR #175, opened 2023-11-21
Move into Networker the 30s timer when starting the VPN

mariusSincovici:
* Move the 30 seconds timer from Nordlynx and OpenVPN into the Networker. In this way it can be controlled when the timer should be used and when not.
* Activate the timer only when the VPN is started after a user command. But don't use the timer at network changes, because in some situations, the daemon can disconnect by itself from VPN without notifying the user.


daemon/vpn/nordlynx/kernel_space.go
```diff
@@ -145,10 +147,19 @@ func (k *KernelSpace) stop() error {
...
+	k.setState(vpn.ExitedState)
 	return nil
 }
 
+func (k *KernelSpace) StateChanged() <-chan vpn.State {
+	return k.stateChanged
+}
+
+func (k *KernelSpace) setState(newState vpn.State) {
+	k.state = newState
```
keliramu, line 159: What if same state is being set, so, no change happened why bother informing about that someone

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -259,11 +259,21 @@ func (l *Libtelio) Start(
...
-	defer cancel()
 	// Start monitoring connection events before connecting to not miss any
-	isConnectedC := isConnected(ctx, l.events, serverPublicKey)
+	l.monitorNodesEvents(serverPublicKey)
+
+	if err := l.connectToExitNode(serverIP, serverPublicKey); err != nil {
+		return err
+	}
+
+	// Check if the connection actually happened. Disconnect if
```
keliramu, line 269: Looks like this comment is not relevant anymore

internal/helpers.go
```diff
@@ -0,0 +1,12 @@
+package internal
+
+import "time"
+
+// Will try to send the value non-blocking. Possible loosing value.
+// if there is no receiver and the channel buffer is full or unbuffered, then the value is lost
+func SendNonBlocking[T any](c chan<- T, value T) {
+	select {
+	case c <- value:
+	case <-time.After(500 * time.Millisecond):
```
keliramu, line 10: Why timeout is needed?
Non blocking send was like this:
```
select {
	case c <- value:
	default: // drop if nobody is listening
}
```

mariusSincovici:
drop for now

PR #182, opened 2023-11-23
Automate Meshnet invites tests

naujalyte:


test/qa/test_meshnet.py
```diff
@@ -317,4 +319,167 @@ def test_account_switch():
...
     teardown_module(None)
     setup_module(None)
-    
\ No newline at end of file
+
+
+def test_invite_send_repeated():
+
+    with lib.Defer(lambda: sh.nordvpn.meshnet.invite.revoke("test@test.com")):
+        process = subprocess.Popen(sh.nordvpn.meshnet.invite.send("test@test.com"), shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
```
mariusSincovici, line 327: it might be good to move the entire `send` flow into a helper function.

mariusSincovici (approved):

test/qa/test_meshnet.py
naujalyte: Good idea, made changes accordingly, thanks!

mariusSincovici (approved):

test/qa/test_meshnet.py
```diff
@@ -317,4 +319,147 @@ def test_account_switch():
...
+@pytest.mark.skip(reason="A different error message is expected - LVPN-262")
+def test_invite_send_long_email():
+    
+    # A long email address containing more than 256 characters is created
+    email = "test" * 65 + "@test.com"
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        meshnet.send_meshnet_invite(email)
+
+    assert not "It's not you, it's us. We're having trouble with our servers. If the issue persists, please contact our customer support." in str(ex.value)
```
keliramu, line 360: Does not look like correct logic in the app - I would expect clear message saying that email is invalid. Now this test would legitimize incomplete error handling in app.

test/qa/test_meshnet.py
```diff
@@ -317,4 +319,147 @@ def test_account_switch():
...
+
+
+@pytest.mark.skip(reason="A different error message is expected - LVPN-262")
+def test_invite_send_email_special_character():
+    
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        meshnet.send_meshnet_invite("\u2222@test.com")
+
+    assert not "It's not you, it's us. We're having trouble with our servers. If the issue persists, please contact our customer support." in str(ex.value)
```
keliramu, line 369: Same as with too long email - app should respond with clear message about invalid email.

test/qa/test_meshnet.py
naujalyte: Yes, this is why this test (and the test right after this one) is skipped, we do expect a clear error message but we do not know what exactly it should say, a bug is registered. Benas suggested to write the tests but skip them until the bug is fixed.

keliramu (approved):

PR #189, opened 2023-12-01
Don't recreate tun at network changes

mariusSincovici:
For Nordlynx try to reconfigure internally the VPN at network changes, instead of recreating the tunnel.
For OpenVPN is not possible at the moment, because the server config can change between pulls and that internally will trigger the recreation of the tunnel.

### DIFF
--- a/daemon/vpn/openvpn/openvpn.go
+++ b/daemon/vpn/openvpn/openvpn.go
@@ -35,7 +35,7 @@ var (
 	errServerTimeout  = errors.New("server timeout")
 	ErrServerVersion  = errors.New("invalid openvpn server version")
 	errExited         = errors.New("exited")
-	errNotSupported   = errors.New("not supported")
+	errNotImplemented = errors.New("not implemented")
 )
 
 type OpenVPN struct {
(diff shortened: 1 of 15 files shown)


bartoszWojciechO (approved):

networker/networker_test.go
```diff
@@ -207,7 +207,8 @@ func (e *workingExitNode) SetAllowlist(_ config.Allowlist, lan bool) error {
 func (e *workingExitNode) ResetFirewall(lan bool) error { e.LanAvailable = lan; return nil }
 
 type workingMesh struct {
-	enableErr error
+	enableErr         error
+	errNetworkChanged error
```
keliramu, line 211: Plz unify naming - `err` prefix or postifx

meshnet/mesh.go
```diff
@@ -25,6 +25,8 @@ type Mesh interface {
 	// StatusMap retrieves the current status map for the related
 	// meshnet peers
 	StatusMap() (map[string]string, error)
+	// handle network changes
```
keliramu, line 28: Plz make comments style same as all above

daemon/vpn/openvpn/openvpn.go
```diff
@@ -210,6 +211,32 @@ func (ovpn *OpenVPN) stop() error {
 	return nil
 }
 
+func (ovpn *OpenVPN) NetworkChanged() error {
+	// application uses the flag --persist-tun and it should not close the TUN interface at restarts
+	// but because the servers sends different configuration at pull, then the TUN is closed + recreated
+	// https://github.com/OpenVPN/openvpn/blob/a1cb1b47b138b9f654cd0bca5de6d08dbca61888/src/openvpn/init.c#L2421
+
+	return errNotSupported
```
keliramu, line 219: IMO, we should say into log that OpenVPN should handle network change by itself, rather that saying it is not supported?

keliramu (approved):

PR #190, opened 2023-12-04
Make firewall.py more readable

ggediminass:


bartoszWojciechO (approved):

ggediminass:
I may have lied about the `more readable` part :smile: 

However, now we can check if rules are added to iptables, when MULTIPLE subnets and ports are whitelisted :partying_face:

test/qa/lib/firewall.py
```diff
@@ -93,98 +96,195 @@
...
+    print("Expected rules:\n")
+    for rule in expected_rules:
+        print(rule)
+    print("\n")
 
     current_rules = _get_iptables_rules()
-    print("Current rules:\n", current_rules)
+    print("Current rules:\n")
+    for rule in current_rules:
+        print(rule)
```
mariusSincovici, line 274: shouldn't each rule be printed on a new line?

mariusSincovici (approved):

test/qa/lib/firewall.py
ggediminass: Checked this once again, it seems like we are indeed printing elements to new lines :smile:

test/qa/lib/firewall.py
```diff
@@ -92,99 +95,231 @@
...
+    [
+        f"-A INPUT -i {interface} -m connmark --mark 0xe1f1 -m comment --comment nordvpn -j ACCEPT",
+        f"-A INPUT -i {interface} -m comment --comment nordvpn -j DROP",
+        f"-A OUTPUT -o {interface} -m mark --mark 0xe1f1 -m comment --comment nordvpn -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff",
+        f"-A OUTPUT -o {interface} -m connmark --mark 0xe1f1 -m comment --comment nordvpn -j ACCEPT",
+        f"-A OUTPUT -o {interface} -m comment --comment nordvpn -j DROP"
+    ]
+
+
+def _get_rules_connected_to_vpn_server(interface:str):
```
keliramu, line 110: Minor point, but what is the point of having function which adds nothing? Maybe just use original function?

test/qa/lib/firewall.py
```diff
@@ -95,157 +95,220 @@
...
+def is_active(ports: list[str]="", protocol="", subnets: list[str]="") -> bool:
+    """ returns True when all expected rules are found in iptables, in matching order """
+    print(sh.ip.route())
 
-    expected_rules = _get_firewall_rules(
-        daemon.is_killswitch_on(), server_ip, default_route_interface, ports, protocol, subnets
-    )
-    print("Expected rules:\n")
+    expected_rules = _get_firewall_rules(ports, protocol, subnets)
+    print("\nExpected rules:")
```
bartoszWojciechO, line 302: Are prints even visible when running the tests? I thought that pytest captures the output.

ggediminass: Same as in previous code, prints are only visible, if test fails :smile:

test/qa/lib/firewall.py
```diff
@@ -92,99 +95,231 @@
...
-        _, server_ip = status.split(None, 2)
-    except sh.ErrorReturnCode:
-        server_ip = ""
-    print("Default gateway:", iface, "Server's IP:  ", server_ip)
-
-    rules = _get_firewall_rules(
-        daemon.is_killswitch_on(), server_ip, iface, port, protocol, subnet
-    )
-    print("Expected rules:\n", rules)
+def _get_firewall_rules(ports: list[str]="", protocol: str="", subnets: list[str]="") -> list[str]:
```
piotrjurkiewicz, line 258: I think it should rather get `None` as a default value, not an empty string. The same for `is_active()`.

test/qa/lib/daemon.py
```diff
@@ -150,4 +150,4 @@ def is_peer_running(ssh_client: ssh.Ssh) -> bool:
     except:
         return False
     else:
-        return True
\ No newline at end of file
+        return True
```
mariusSincovici, line 153: [minor] probably you can remove the else and always return true?

test/qa/lib/daemon.py
```diff
@@ -150,4 +150,4 @@ def is_peer_running(ssh_client: ssh.Ssh) -> bool:
     except:
```
mariusSincovici, line 150: [minor] Cannot add comment on the function name, but it is a bit confusing. maybe in future can be changed to something like is daemon/service running for other peer?

mariusSincovici (approved):

test/qa/lib/daemon.py
ggediminass: Hey, probably you are right :smile: However, changes like this may not be limited to only that function name, it may be, that other related functions will also need renaming, so I think this may need a separate task. Maybe even module itself needs refactoring. However, this is sort of getting out of the scope of my current task

test/qa/lib/firewall.py
```diff
@@ -92,99 +97,219 @@
...
-    except sh.ErrorReturnCode:
-        server_ip = ""
-    print("Default gateway:", iface, "Server's IP:  ", server_ip)
-
-    rules = _get_firewall_rules(
-        daemon.is_killswitch_on(), server_ip, iface, port, protocol, subnet
-    )
-    print("Expected rules:\n", rules)
+def _get_firewall_rules(ports: list[lib.Port]=None, subnets: list[str]=None) -> list[str]:
+    if subnets:
```
Savolro, line 236: Would be nice if this was commented why do we need this

test/qa/lib/firewall.py
```diff
@@ -92,99 +97,219 @@
...
+def _get_rules_allowlist_port_on(interface:str, ports:list[lib.Port]):
+    # Port(range) whitelist rules for SPECIFIC protocols
+
+    ports_udp: list[lib.Port]
+    ports_tcp: list[lib.Port]
+    ports_udp, ports_tcp = _sort_ports_by_protocol(ports)
+
+    result = []
+
+    for port in ports_udp:
```
Savolro, line 150: This is the same as in function `_get_rules_allowlist_subnet_and_port_on`. Please move this to a separate function and re-use that. Also, check if the same can be done with other sets of FW rules.

test/qa/test_firewall6.py
```diff
@@ -111,7 +111,7 @@ def test_firewall_enable_connect(tech, proto, obfuscated):
 
 
 @pytest.mark.parametrize("tech,proto,obfuscated", lib.OVPN_STANDARD_TECHNOLOGIES)
-@pytest.mark.parametrize("port", lib.PORTS)
+@pytest.mark.parametrize("port", lib.PORTS, ids=[lib.Protocol.UDP, lib.Protocol.TCP, lib.Protocol.ALL])
```
Savolro, line 114: there is no relation between `ids` and `PORTS`. If we add more ports, this will show incorrect logs

Savolro (changes requested):

PR #197, opened 2023-12-12
Implement fileshare history clear feature

piotrjurkiewicz:


bartoszWojciechO:
Can you add some tests? Mostly looks like just calling the `libdrop` API, so maybe too simple to get in there with uts. Maybe just one integration test to have this covered.

cli/cli_fileshare.go
```diff
@@ -203,6 +208,28 @@ func (c *cmd) FileshareAutoCompletePeers(ctx *cli.Context) {
 	}
 }
 
+// FileshareAutoCompleteClear implements bash autocompletion for history clearing
+func (c *cmd) FileshareAutoCompleteClear(ctx *cli.Context) {
+	if ctx.NArg() == 0 {
+		fmt.Println("all\n<time_period>")
+	} else {
```
mariusSincovici, line 215: how many arguments can it have?
wouldn't be better to just allow max 2 params `<num>` `<s|m|h|d>` , to keep things simpler? :smile:

cli/cli_fileshare.go
```diff
@@ -203,6 +208,28 @@ func (c *cmd) FileshareAutoCompletePeers(ctx *cli.Context) {
...
+		last := ctx.Args().Get(ctx.Args().Len() - 1)
+		i, err := strconv.Atoi(last)
+		if err == nil {
+			if i == 1 {
+				fmt.Println("second\nminute\nhour\nday\nweek\nmonth\nyear")
+			} else {
+				fmt.Println("seconds\nminutes\nhours\ndays\nweeks\nmonths\nyears")
+			}
+		} else {
+			fmt.Println("<time_period>")
```
mariusSincovici, line 228: if the first param is not an int, wouldn't be better not to print anything otherwise I think the autocomplete will display `<time_period`?

Savolro (changes requested):

cli/cli_fileshare.go
```diff
@@ -299,6 +326,82 @@ func (c *cmd) FileshareCancel(ctx *cli.Context) error {
...
+
+func parseTimespan(ts string) (int64, error) {
+	var timespan int64 = 0
+	matches := timespanRegexp.FindAllString(ts, -1)
+
+	if matches == nil {
+		return 0, fmt.Errorf("Time span parsing error: '%s'", ts)
+	}
+
+	for _, m := range matches {
```
mariusSincovici, line 369: does it make sense to support multiple values for time?

cli/cli_fileshare.go
```diff
@@ -299,6 +326,82 @@ func (c *cmd) FileshareCancel(ctx *cli.Context) error {
...
+
+		if err == io.EOF && n == 1 && unit == "" {
+			// pass
+		} else if err != nil {
+			return 0, fmt.Errorf("Time span argument '%s' parsing error: %s", m, err)
+		}
+
+		// Using time span syntax as defined by systemd
+		// https://www.freedesktop.org/software/systemd/man/latest/systemd.time.html
+		switch unit {
```
mariusSincovici, line 382: maybe `unit` should be converted to lowercase first

cli/cli_fileshare.go
piotrjurkiewicz: I tried to keep it compatible with the systemd specification, which allows for example 2h 30min as an equivalent to 150min: https://www.freedesktop.org/software/systemd/man/latest/systemd.time.html

cli/cli_fileshare.go
piotrjurkiewicz: It prints <time_period> if the previous (last) argument is a complete time period (e.g. ... 5d) or a unit (e.g. ... 5 days). This gives a hint to user that he can write next part of time span.

cli/cli_fileshare.go
piotrjurkiewicz: See the first answer.

cli/cli_fileshare.go
piotrjurkiewicz: Unit can be `m` for minutes or `M` for months, so the case should not be modified.

cli/messages.go
```diff
@@ -284,6 +285,10 @@ Provide a [transfer_id] argument to list files in the specified transfer.`
 	MsgFileshareAcceptArgsUsage   = "<transfer_id> [file_id1] [file_id2...]"
 	MsgFileshareAcceptDescription = MsgFileshareAcceptUsage + "\n\nTo cancel a transfer in progress, press Ctrl+C"
 	MsgFileshareAcceptPathUsage   = "Specify download path (default: $XDG_DOWNLOAD_DIR or $HOME/Downloads)"
+	MsgFileshareClearUsage        = "Clear entries older than time period from the file transfer history."
```
Savolro, line 288: Please send user facing texts for a review

cli/cli_fileshare.go
```diff
@@ -299,6 +328,82 @@ func (c *cmd) FileshareCancel(ctx *cli.Context) error {
...
+	var until = time.Now()
+
+	args := ctx.Args()
+	if args.Get(0) != "all" {
+		argsJoined := strings.Join(args.Slice(), " ")
+		ts, err := parseTimespan(argsJoined)
+		if err != nil {
+			return formatError(err)
+		}
+		until = until.Add(time.Duration(-1000000000 * ts))
```
Savolro, line 348:
```suggestion
		until = until.Add(time.Duration(-time.Second * ts))
```

Savolro: I see you are using average values for a month and year which will end up in not correct timestamps. Considering you still need a reference timestamp, which is `var until = time.Now()`, I recommend reworking `parseTimespan`  function so it would return years and months separately. E. g. 
```go
parseTimestamp(input string) (years int, months int, seconds int) { ... }
```
Then you could use a combination of https://pkg.go.dev/time#Time.AddDate and https://pkg.go.dev/time#Time.Add to have correct timestamps

Savolro (changes requested):

cli/cli_fileshare.go
```diff
@@ -203,6 +207,28 @@ func (c *cmd) FileshareAutoCompletePeers(ctx *cli.Context) {
 	}
 }
 
+// FileshareAutoCompleteClear implements bash autocompletion for history clearing
+func (c *cmd) FileshareAutoCompleteClear(ctx *cli.Context) {
+	if ctx.NArg() == 0 {
+		fmt.Println("all\n<time_period>")
```
mariusSincovici, line 213: I think `<time_period>` is not ok, specially for `zsh`. Since there the value appears into the autocompletion list and it can be selected.

cli/cli_fileshare.go
mariusSincovici: same here, `\<time_period\>` is automatically added to the command when pressing tab.

fileshare/server.go
```diff
@@ -465,3 +465,18 @@ func (s *Server) SetNotifications(ctx context.Context, in *pb.SetNotificationsRe
 		}
 	}
 }
+
+func (s *Server) PurgeTransfersUntil(ctx context.Context, req *pb.PurgeTransfersUntilRequest) (*pb.Error, error) {
```
mariusSincovici, line 469: This doesn't delete the transfers from previous version of fileshare, which are stored in another place. it only deletes the ones from libdrop db. Maybe it should also go over the other list and delete them too.

bartoszWojciechO (approved):

keliramu (approved):

PR #198, opened 2023-12-12
Add nickname for peers

mariusSincovici:
* Add nickname for a peer
* update DNS to include the nickname 
* fix tests


bartoszWojciechO (approved):

cli/cli_mesh_peers.go
```diff
@@ -188,8 +188,12 @@ func peerToOutputString(peer *pb.Peer) string {
 		{Key: "Allows Sending Files", Value: nstrings.GetBoolLabel(peer.IsFileshareAllowed)},
 		{Key: "Accept Fileshare Automatically", Value: nstrings.GetBoolLabel(peer.AlwaysAcceptFiles)},
 	}
+	hostname := peer.Hostname
+	if peer.Nickname != "" {
+		hostname = peer.Nickname
+	}
 	return titledKeyvalListToColoredString(keyval{
-		Key: "Hostname", Value: peer.Hostname,
+		Key: "Hostname", Value: hostname,
```
keliramu, line 196: Nickname to be shown instead of full original peer name. How to know original peer name? Will it be shown somewhere?

cli/cli.go
```diff
@@ -736,6 +736,26 @@ func meshnetCommand(c *cmd) *cli.Command {
...
+						Subcommands: []*cli.Command{
+							{
+								Name:         "set",
+								Usage:        MsgMeshnetPeerSetNicknameUsage,
+								ArgsUsage:    MsgMeshnetPeerSetNicknameArgsUsage,
+								Action:       c.MeshPeerSetNickname,
+								BashComplete: c.MeshPeerNicknameAutoComplete,
+							},
+							{
+								Name:         "reset",
```
keliramu, line 751: Should it be changed to `remove` ?

keliramu: Or is it bound by API: nickname operations `set` and `reset` (ie. rename)?

meshnet/server.go
```diff
@@ -1254,16 +1255,126 @@ func (s *Server) RemovePeer(
...
+
+	if peer == nil {
+		return &pb.RenamePeerResponse{
+			Response: &pb.RenamePeerResponse_UpdatePeerErrorCode{
+				UpdatePeerErrorCode: pb.UpdatePeerErrorCode_PEER_NOT_FOUND,
+			},
+		}, nil
+	}
+
+	if req.Nickname == "" && peer.Nickname == "" {
```
keliramu, line 1338: What is this situation? When can it happen? I why we give empty response to user?

cli/cli.go
keliramu: But I see API has single method to update machine peer. Why have to CLI commands to do the same? `set` and `reset` essentially do the same. Better to have `set` and `remove`, IMO. Otherwise how to remove nickname? Set it to empty?

keliramu: Or, does API operation mean that we `rename` peer and original peer name is gone? But I see that API gives back original peer name and nickname as well. So, it is not `peer rename` it is `set nickname`

protobuf/meshnet/service.proto
```diff
@@ -38,6 +38,8 @@ service Meshnet {
 	rpc GetPeers(Empty) returns (GetPeersResponse);
 	// RemovePeer removes a peer from the meshnet
 	rpc RemovePeer(UpdatePeerRequest) returns (RemovePeerResponse);
+	// RenamePeer set or clear a nickname for a meshnet peer
+	rpc RenamePeer(RenamePeerRequest) returns (RenamePeerResponse);
```
keliramu, line 42: Semantically it would be more clear to have 2 methods: `Set` and `Remove` as by API it is `peer add nickname` and `peer remove nickname` but not `peer rename`

cli/cli_mesh_peers.go
keliramu: IMO, there should be somewhere shown peer original name to be able to clear nickname in case of something messy gets. Also, IMO, peer should be accessible by nickname and by original name.

mariusSincovici: The peer original name is not displayed anymore, but is still accessible with the autocomplete. I'm thinking to change the meshnet map and display both, nickname and original hostname.

The idea is that a peer can be accessed using 3 names in the app:
* original Nord name
* peer ID
* nickname

The DNS is configured:
* original name, with and without `.nord`
* nickname

cli/cli.go
mariusSincovici: I've changed the commands names in the next PR https://github.com/NordSecurity/nordvpn-linux/pull/200/files#diff-ef44b92f1ba3916c39c98fc4a25b67da847cb2a07a5d1253e1978365338888a6R740 to `set`/`remove`
```
nordvpn mesh peer nickname set <ID|hostname|nickname> new-nickname
nordvpn mesh peer nickname remove <ID|hostname|nickname>
```
API exposes only one call, when calling with empty nickname means remove it.

meshnet/server.go
mariusSincovici: The user tries to remove the nickname from a peer that already doesn't have nickname.

protobuf/meshnet/service.proto
mariusSincovici: I've tried to have 2 separate function when setting nickname for own machine, but in the end one of them calls the other one. Because of this I'm not sure if it brings anything to have 2 calls.
Instead I was thinking to change it to `ChangePeerNickname` or something similar?

meshnet/server.go
mariusSincovici: updated and in this case an error will be returned if peer doesn't have a nickname.

cli/cli_mesh_peers.go
mariusSincovici: for the moment changed to display `Nickname [original_host_name]`. I'll update when I'll get more info how to be displayed.

protobuf/meshnet/service.proto
mariusSincovici: changed to `ChangePeerNickname`

keliramu (approved):

PR #200, opened 2023-12-13
Add nickname for current machine(PR 2)

mariusSincovici:
* Add commands to set/remove the nickname for the current machine.
* When there is a NC notification for current machine, update meshnet info because maybe the nickname was modified by another peer


cli/cli_mesh_peers.go
```diff
@@ -624,6 +628,72 @@ func (c *cmd) renameMeshnetPeer(peer *pb.Peer, nickname string) error {
...
+		return formatError(err)
+	}
+
+	if resp == nil {
+		return errors.New(AccountInternalError)
+	}
+	switch resp := resp.Response.(type) {
+	case *pb.RenamePeerResponse_ServiceErrorCode:
+		return serviceErrorCodeToError(resp.ServiceErrorCode)
+
```
bartoszWojciechO, line 651: Why the extra newline here(and in line 656)?

cli/cli_mesh_peers.go
```diff
@@ -624,6 +628,72 @@ func (c *cmd) renameMeshnetPeer(peer *pb.Peer, nickname string) error {
...
+			return fmt.Errorf(MsgMeshnetPeerSetNicknameTheSame, nickname)
+		}
+	}
+
+	color.Green(MsgMeshnetSetNicknameSuccessful, nickname)
+	return nil
+}
+
+func (c *cmd) MeshRemoveMachineNickname(ctx *cli.Context) error {
+	resp, err := c.meshClient.RemoveMachineNickname(context.Background(), &pb.ChangeMachineNicknameRequest{
```
bartoszWojciechO, line 669: Why not call `ChangeMachineNickname` RPC directly?

cli/cli_mesh_peers.go
```diff
@@ -166,8 +166,12 @@ func selfToOutputString(peer *pb.Peer) string {
 		{Key: "OS", Value: peer.Os},
 		{Key: "Distribution", Value: peer.Distro},
 	}
+	hostname := peer.Hostname
+	if peer.Nickname != "" {
+		hostname = peer.Nickname
```
keliramu, line 171: My peer's original name is not shown? I guess it is preserved by API and it can be used to access my peer, so, it should be shown somewhere. Or is it not so?

cli/messages.go
```diff
@@ -160,6 +160,19 @@ Example: nordvpn set meshnet on`
 	MsgMeshnetAllowLocalNetworkUsage          = "Allow the peer to access local network when routing traffic through this device."
 	MsgMeshnetAllowFileshare                  = "Allow the peer to send you files."
 
+	// Meshnet set commands group
+	MsgMeshnetSetUsage = "COPY: Sets meshnet configuration option"
+
+	MsgMeshnetSetMachineNicknameUsage = "COPY: set current machine nickname"
```
keliramu, line 166: Is it marked with `COPY:` for copywriters to review?

core/mesh/peer.go
```diff
@@ -59,9 +60,19 @@ func (s Machine) ToProtobuf() *pb.Peer {
...
 	}
 }
 
+func (s *Machine) IsEqual(b Machine) bool {
+	return s.ID == b.ID &&
+		s.HardwareID == b.HardwareID &&
+		s.Hostname == b.Hostname &&
+		s.Address == b.Address &&
+		s.SupportsRouting == b.SupportsRouting &&
+		s.Nickname == b.Nickname
```
keliramu, line 73: Looks like too many conditions: `Nickname` is not needed because of `Hostname`, and why you need to check `SupportsRouting`? it's not part of ID. Or what situation you are trying to handle?

keliramu: Ok, I got it - initially I thought it is to find Machine, but below I see it is to check if there were any changes made to machine data.

meshnet/server.go
```diff
@@ -1369,6 +1374,125 @@ func (s *Server) RenamePeer(
...
+	if !s.mc.IsRegistrationInfoCorrect() {
+		return &pb.RenamePeerResponse{
+			Response: &pb.RenamePeerResponse_ServiceErrorCode{
+				ServiceErrorCode: pb.ServiceErrorCode_CONFIG_FAILURE,
+			},
+		}, nil
+	}
+
+	if req.Nickname == "" {
+		if cfg.MeshDevice.Nickname == "" {
```
keliramu, line 1420: Why need this check? Plz add comment in the code. Probably it is when trying to remove nickname when it was not set. But in that case maybe we should say to user exactly that?

cli/cli_mesh_peers.go
mariusSincovici: I'll change this to display in this case both. 
But maybe put the nickname as `titledKeyvalListToColoredString` and hostname lower? Because it seams that other apps are treating the nickname as the main "hostname" and the Nord secondary.

cli/messages.go
mariusSincovici: yes, I've marked all the new strings with this to find them easier so I don't miss something.

core/mesh/peer.go
mariusSincovici: we store locally the info about the current machine. So I needed to check if 2 objects of this type are equal. Not sure if the other 3 `OS`, `PublicKey` and `Endpoints` makes sense to check too...

meshnet/server.go
mariusSincovici: Ok, I'll return an error also for this case.

keliramu (approved):

cli/cli_mesh_peers.go
mariusSincovici: probably leftovers :). I'll remove it.

PR #202, opened 2023-12-14
Add meshnet rules to FirewallManager

bartoszWojciechO:


daemon/firewall/firewall_manager.go
```diff
@@ -39,10 +56,111 @@ func NewFirewallManager(devices device.ListFunc,
...
+		iptablesmanager.Input,
+		iptablesmanager.IPv4,
+		fmt.Sprintf("-s %s/32 -j ACCEPT", peer.Address),
+		MeshnetIncoming)
+
+	if err := f.iptablesManager.InsertRule(rule); err != nil {
+		return fmt.Errorf("allowing incoming traffic for peer: %w", err)
+	}
+
+	f.allowIncomingRules[peer.UID] = meshIncomingRule{
```
mariusSincovici, line 105: What would happen if one of the `InsertRule` will fail until reaching here, so some of the rules will be set but not saved in the list? Will it be possible to remove the already added rules?

bartoszWojciechO: For now the rules will remain. Depending on how the new layer is integrated, it might be possible to remove them. I want to add some sort of cleanup on failure after all the kinds of rules are implemented(to generalize it if possible).

piotrjurkiewicz (approved):

mariusSincovici (approved):

PR #203, opened 2023-12-14
Add tests for nicknames(PR 3)

mariusSincovici:
* remove duplicate classes and move them into mock package
* when a nickname is set, check with DNS resolver no to already be assigned
* add tests to setting the nickname


meshnet/server.go
```diff
@@ -1340,22 +1339,37 @@ func (s *Server) RenamePeer(
 		}, nil
 	}
 
-	if req.Nickname == "" && peer.Nickname == "" {
-		log.Println(internal.InfoPrefix, "peer doesn't have a nickname")
+	if req.Nickname == "" {
+		if peer.Nickname == "" {
+			log.Println(internal.InfoPrefix, "peer doesn't have already a nickname")
```
keliramu, line 1344: Is this remove nickname when it is not set situation? Should be say to the user about that instead pretending that operation succeeded?

bartoszWojciechO (approved):

meshnet/server.go
mariusSincovici: changed and will return an error in this case

piotrjurkiewicz (approved):

PR #204, opened 2023-12-15
Standalone daemon

Savolro:


bartoszWojciechO (approved):

internal/filesystem.go
```diff
@@ -72,9 +72,40 @@ func SystemDListener() (net.Listener, error) {
...
+}
+
 // ManualListener returns manually created listener with provided permissions
 func ManualListener(socket string, perm fs.FileMode) func() (net.Listener, error) {
 	return func() (net.Listener, error) {
+		dirCreated := true
+		if err := os.MkdirAll(
+			RunDir, PermUserRWGroupRW,
+		); err != nil && !errors.Is(err, os.ErrExist) {
+			if errors.Is(err, os.ErrExist) {
```
mariusSincovici, line 103: This will never be true, because it can enter into the if, only if error is not `ErrExitst`. Because of this if the error is that the folder already exists, then the function will return error.

internal/filesystem.go
```diff
@@ -72,9 +72,40 @@ func SystemDListener() (net.Listener, error) {
...
+				dirCreated = false
+			}
+			return nil, fmt.Errorf("ceating run dir: %w\n", err)
```
mariusSincovici, line 106: [miss-spelling] `ceating`

internal/filesystem.go
```diff
@@ -72,9 +72,40 @@ func SystemDListener() (net.Listener, error) {
...
+	listener   net.Listener
+	dirCreated bool
+}
+
+func (rdl *runDirListener) Accept() (net.Conn, error) {
+	return rdl.listener.Accept()
+}
+
+func (rdl *runDirListener) Close() error {
+	var removeErr error
```
mariusSincovici, line 85: shouldn't also close also the `listener` here?

Savolro: It is closed in the return line

keliramu (approved):

PR #205, opened 2023-12-18
Handle api errors when changing the meshnet nickname(PR4)

mariusSincovici:
* handle API error
* handle DNS conflicts when setting the nickname 
* add tests


cli/cli_mesh_peers.go
```diff
@@ -764,7 +775,7 @@ func (c *cmd) MeshPeerNicknameAutoComplete(ctx *cli.Context) {
 
 func peerByIdentifier(id string) func(*pb.Peer) bool {
 	return func(peer *pb.Peer) bool {
-		return peer.GetIp() == id || peer.GetHostname() == id || peer.GetPubkey() == id || peer.GetNickname() == id
+		return peer.GetIp() == id || peer.GetHostname() == id || peer.GetPubkey() == id || strings.EqualFold(peer.GetNickname(), id)
```
piotrjurkiewicz, line 778: We should probably also compare hostnames using `strings.EqualFold`, as they are by definition case-insensitive.

keliramu (approved):

bartoszWojciechO (approved):

PR #207, opened 2023-12-22
Implement meshnet peers nicknames

piotrjurkiewicz:


core/mesh/peer.go
```diff
@@ -59,9 +60,19 @@ func (s Machine) ToProtobuf() *pb.Peer {
 		Os:         s.OS.Name,
 		Distro:     s.OS.Distro,
 		Hostname:   s.Hostname,
+		Nickname:   s.Nickname,
 	}
 }
 
+func (s *Machine) IsEqual(b Machine) bool {
+	return s.ID == b.ID &&
```
keliramu, line 68: Use `strings.EqualFold` here as well - for `Nickname` and for `Hostname` as in other places

keliramu (approved):

bartoszWojciechO (approved):

core/mesh/peer.go
mariusSincovici: In this case nicknames must not be compared with `strings.EqualFold`. Otherwise if another peer changes the current machine nickname(but only changes the caps), then the current machine will not update its information and will display the older nickname.

PR #211, opened 2023-12-27
Fix random subnet rule order in firewall.py

ggediminass:


bartoszWojciechO (approved):

test/qa/lib/firewall.py
```diff
@@ -120,7 +120,16 @@ def __rules_allowlist_subnet_chain_input(interface: str, subnets: list[str]):
...
+    try:
+        current_subnet_rules_input_chain = sh.grep(r"\-s ", _in=sh.grep("INPUT", _in=sh.sudo.iptables("-S"))).split("\n")
+
+        # We need to sort rules, since they appear in iptables in random fashion, unlike port rules
+        # otherwise rule lists in `is_active` wont be the same, and assert will fail
+        return sort_list_by_other_list(result, current_subnet_rules_input_chain)
+    except:
+        # If "-s" is nowhere to be found in iptables, we get exception, and we do not sort
+        # `result`, since we do not have other list to sort `result` list by
+        return result
```
piotrjurkiewicz, line 133: I think something like below would be more Pythonic and more readable:
``` Python
current_subnet_rules_input_chain = []

for line in sh.sudo.iptables("-S").splitlines():
    if "INPUT" in line and "-s" in line:
        current_subnet_rules_input_chain.append(line)

if current_subnet_rules_input_chain:
    return current_subnet_rules_input_chain
else:
    return result
```

ggediminass: Your solution indeed looks more readable :thinking: I will try to use it instead

piotrjurkiewicz (approved):

PR #213, opened 2023-12-28
Fetch new NC token after 24 hours

bartoszWojciechO:


nc/creds_fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	credentials, err := core.GetNCCredentials(cf.api, tokenData.Token, ncData.UserID)
+	if err != nil {
+		return config.NCData{}, fmt.Errorf("getting NC credentials: %w", err)
+	}
+
+	return credentials, cf.cm.SaveWith(func(c config.Config) config.Config {
+		user := c.TokensData[userID]
+		defer func() { c.TokensData[userID] = user }()
```
keliramu, line 68: why cannot do simple assignment just before return statement?

nc/creds_fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+type RealTime struct{}
+
+func (r RealTime) GetSecondsSinceTimestamp(timestamp int64) float64 {
+	return time.Since(time.Unix(timestamp, 0)).Seconds()
+}
+
+type CredentialsGetter struct {
+	api  core.CredentialsAPI
+	cm   config.Manager
+	time Time
```
keliramu, line 24: Is this `lastFetchTime` ? if keeping in stile with short names, then comment about field meaning would be nice

keliramu: or is this `validityPeriod` ?

keliramu (approved):

nc/creds_fetcher.go
bartoszWojciechO: This is an interface for getting the time, introduced so that it can be mocked in uts. I have changed the name so that it is more meaningful.

nc/creds_fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/core"
+)
+
+type TimeSource interface {
+	GetSecondsSinceTimestamp(int64) float64
```
Savolro, line 12:
```suggestion
	GetDurationSinceTimestamp(time.Time) time.Duration
```

nc/creds_fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+func (cf *CredentialsGetter) GetCredentials() (config.NCData, error) {
+	var cfg config.Config
+	if err := cf.cm.Load(&cfg); err != nil {
+		return config.NCData{}, fmt.Errorf("reading cfg: %w", err)
+	}
+	userID := cfg.AutoConnectData.ID
+	tokenData := cfg.TokensData[userID]
+	ncData := tokenData.NCData
+
+	const tokenValidityPeriodSec = 86400
```
Savolro, line 54: Where does this come from? Also, use `time.Duration` for this

nc/nc.go
```diff
@@ -159,13 +191,32 @@ func (c *Client) newClient(opts *mqtt.ClientOptions) error {
...
+	}
+}
+
 func (c *Client) connectWithBackoff(ctx context.Context, backoff func(int) time.Duration) {
 	for tries := 0; ; tries++ {
-		err := c.connect()
+		ncData, err := c.credsFetcher.GetCredentials()
+		if err != nil {
+			log.Println("failed to fetch new credentials, retrying with old client: ", err)
+			continue
```
Savolro, line 213: This will result in an uncontrolled infinite loop

nc/nc.go
```diff
@@ -97,13 +98,24 @@ type AcknowledgementPayload struct {
...
+type MqttClientBuilder struct{}
+
+func (MqttClientBuilder) Build(opts *mqtt.ClientOptions) mqtt.Client {
+	return mqtt.NewClient(opts)
+}
+
 // Client is a client for Notification center
 type Client struct {
-	client mqtt.Client // nil if not started or stopped, always check for nil before using
+	clientBuilder ClientBuilder
```
Savolro, line 117: Why is this added? Any reasons creating it with `mqtt.NewClient(opts)` when needed is not enough?

Savolro (changes requested):

nc/creds_fetcher.go
bartoszWojciechO: Got the value from the core team.

nc/nc.go
bartoszWojciechO: We have to reinitialize the client when reconnecting https://github.com/NordSecurity/nordvpn-linux/pull/213/files#diff-3ef8346631bf7e381d05cd8c7af5211833424bae0c8a08f8b79831acf2ff0755R217. This function is unit tested, so we have to use a mock so that we do not build a real mqtt client.

test/mock/nc/creds_fetcher.go
```diff
@@ -0,0 +1,11 @@
+package nc
+
+import "time"
+
+type MockTime struct {
+	SecondsSinceTimestamp int
+}
+
+func (m *MockTime) GetDurationSinceTimestamp(time.Time) time.Duration {
+	return time.Duration(m.SecondsSinceTimestamp) * time.Second
```
Savolro, line 10: Unnecessary conversion to `time.Duration` here.

Savolro (approved):

PR #217, opened 2024-01-03
Do not overwrite resolv.conf when its permissions are read-only

piotrjurkiewicz:
Currently there is no way for a user to prevent from overwriting `resolv.conf` file on `nordvpn connect`. Setting the file read-only does not work, because daemon runs as root and thus does not obey file permissions. Setting the file immutable also does not work, because we use the immutable attribute to "lock" the file after we write NordVPN DNS servers to it and we clear that attribute before connect and after disconnect.

This pull request adds an explicit check whether the `resolv.conf` file is read-only before trying to overwrite it. It will allow a user to control whether to allow overwriting it or not. It also moves file "locking" to functions which perform actual operations on `resolv.conf`, as it does not make sense to lock the file when `systemd-resolved` or `resolveconf` are used to change DNS. It also removes warning printing on unsuccessful file unlock (`chattr`) to not spam logs with warnings when `resolv.conf` resides on a file system which does not support attributes (`/etc/resolv.conf` is often a symlink to a file located on `/run` tmpfs, which does not support attributes).

Note that this fix applies only to the case when we use direct `resolv.conf` manipulation to set DNS servers. That is, when there is no `systemd-resolved` running on the system AND the command `resolveconf` is not available.

Allowing a user to turn off DNS changing on modern distributions with `systemd-resolved` running will require addition of an appropriate option to the `nordvpn set dns` command.


daemon/dns/dns.go
```diff
@@ -340,6 +334,16 @@ func unsetDNSWithResolvconf(iface string) error {
 }
 
 func setDNSDefault(addresses []string) error {
+	if internal.FileExists(resolvconfFilePath) {
+		info, err := os.Stat(resolvconfFilePath)
```
bartoszWojciechO, line 338: We have an abstraction over filesystem stuff in fileshare. Could you extract it to some top level package(maybe `nordvpn-linux/internal` and use it here? Would be god for consistency.

bartoszWojciechO (approved):

keliramu (approved):

daemon/dns/dns.go
piotrjurkiewicz: Do you mean creating a function which combines `internal.FileExists` and `os.Stat`? Because `os.Stat` is used in various places in the code and I don't see what would be a point in making its one-to-one copy in `internal`.

mariusSincovici: maybe move the entire logic in a function e.g.`IsWritable` so it is easier to reuse in future?

mariusSincovici (approved):

mariusSincovici (approved):

PR #233, opened 2024-01-17
Socket permissions for Snap

keliramu:
Socket permissions for Snap environment and requesting user authentication

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -484,6 +484,7 @@ func main() {
 			if os.Getenv(internal.InSnapOperation) != "" {
 				internal.UpdateSocketFilePermissions(ConnURL)
 			}
+			listener = internal.NewLimitListener(listener)
 		case sockTCP:
 			listener, err = net.Listen("tcp", ConnURL)
 			if err != nil {
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -481,7 +481,9 @@ func main() {
 			if err != nil {
 				log.Fatalf("Error on listening to UNIX domain socket: %s\n", err)
 			}
-			internal.UpdateSocketFilePermissions(ConnURL)
+			if os.Getenv(internal.InSnapOperation) != "" {
+				internal.UpdateSocketFilePermissions(ConnURL)
+			}
 		case sockTCP:
 			listener, err = net.Listen("tcp", ConnURL)
 			if err != nil {
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -481,6 +481,7 @@ func main() {
 			if err != nil {
 				log.Fatalf("Error on listening to UNIX domain socket: %s\n", err)
 			}
+			internal.UpdateSocketFilePermissions(ConnURL)
 		case sockTCP:
 			listener, err = net.Listen("tcp", ConnURL)
 			if err != nil {
@@ -489,6 +490,7 @@ func main() {
 		default:
 			log.Fatalf("Invalid predefined connection type: %s", ConnType)
 		}
+
 		if err := s.Serve(listener); err != nil {
 			log.Fatalln(err)
 		}
--- a/internal/chmod_sock.go
+++ b/internal/chmod_sock.go
@@ -11,7 +11,7 @@ func UpdateSocketFilePermissions(sockFileName string) {
 		return
 	}
 	// #nosec G302 -- need world writable permissions
-	if err := os.Chmod(sockFileName, PermUserRWXGroupRWXOthersRWX); err != nil {
+	if err := os.Chmod(sockFileName, PermUserRWGroupRWOthersRW); err != nil {
 		log.Println(err)
 	}
 }
--- a/internal/chmod_sock.go
+++ b/internal/chmod_sock.go
@@ -12,6 +12,6 @@ func UpdateSocketFilePermissions(sockFileName string) {
 	}
 	// #nosec G302 -- need world writable permissions
 	if err := os.Chmod(sockFileName, PermUserRWGroupRWOthersRW); err != nil {
-		log.Println(err)
+		log.Println(ErrorPrefix, err)
 	}
 }
--- a/internal/chmod_sock.go
+++ b/internal/chmod_sock.go
@@ -1,0 +1,17 @@
+package internal
+
+import (
+	"log"
+	"os"
+)
+
+// UpdateSocketFilePermissions set socket file permissions
+func UpdateSocketFilePermissions(sockFileName string) {
+	if !FileExists(sockFileName) {
+		return
+	}
+	// #nosec G302 -- need world writable permissions
+	if err := os.Chmod(sockFileName, 0777); err != nil {
+		log.Println(err)
+	}
+}
--- a/internal/chmod_sock.go
+++ b/internal/chmod_sock.go
@@ -11,7 +11,7 @@ func UpdateSocketFilePermissions(sockFileName string) {
 		return
 	}
 	// #nosec G302 -- need world writable permissions
-	if err := os.Chmod(sockFileName, 0777); err != nil {
+	if err := os.Chmod(sockFileName, PermUserRWXGroupRWXOthersRWX); err != nil {
 		log.Println(err)
 	}
 }
--- a/internal/socket.go
+++ b/internal/socket.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os/user"
 	"strconv"
 	"strings"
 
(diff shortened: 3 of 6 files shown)


internal/socket.go
```diff
@@ -70,9 +71,39 @@ func getUnixCreds(conn net.Conn) (*unix.Ucred, error) {
...
 	}
 
+	if err := authenticateUser(ucred); err != nil {
+		return nil, err
+	}
+
 	return ucred, nil
 }
 
+func authenticateUser(ucred *unix.Ucred) error {
```
Savolro, line 81: Could this be done by using syscalls instead of calling executables?

internal/chmod_sock.go
```diff
@@ -0,0 +1,19 @@
...
+	"log"
+	"os"
+	"os/exec"
+)
+
+func CheckSockFile(sockDirName, sockFileName string) {
+	if _, err := os.Stat(sockFileName); err != nil {
+		return
+	}
+	if _, err := exec.Command("chmod", "777", sockDirName).CombinedOutput(); err != nil {
```
Savolro, line 13:
```suggestion
	if _, err := exec.Command("chmod", "775", sockDirName).CombinedOutput(); err != nil {
```
In this case watch mechanism is not needed anymore

internal/chmod_sock.go
```diff
@@ -0,0 +1,19 @@
...
+		log.Println(err)
+	}
+	if _, err := exec.Command("chmod", "777", sockFileName).CombinedOutput(); err != nil {
```
Savolro, line 16: Use https://pkg.go.dev/os#Chmod instead

Savolro (changes requested):

internal/socket.go
keliramu: Converted to syscalls what was possible, sudo check is cmd.Exec

internal/chmod_sock.go
keliramu: Done

internal/chmod_sock.go
keliramu: Done

internal/socket.go
```diff
@@ -70,9 +72,48 @@ func getUnixCreds(conn net.Conn) (*unix.Ucred, error) {
...
+	if err != nil {
+		return fmt.Errorf("authenticate user, lookup user info: %s", err)
+	}
+	// user has sudo?
+	// #nosec G204 -- input is properly validated
+	out, err := exec.Command("sudo", "-l", "-U", userInfo.Username).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("authenticate user, check user has sudo: %s", err)
+	}
+	if !strings.Contains(string(out), "not allowed to run sudo") {
```
mariusSincovici, line 97: what would happen if the OS is in another language than English?

internal/socket.go
```diff
@@ -70,9 +72,48 @@ func getUnixCreds(conn net.Conn) (*unix.Ucred, error) {
...
+		return nil
+	}
+	// user belongs to nordvpn group?
+	groups, err := userInfo.GroupIds()
+	if err != nil {
+		return fmt.Errorf("authenticate user, check user groups: %s", err)
+	}
+	for groupId := range groups {
```
mariusSincovici, line 105: wouldn't it be safer to check if the user is part of `sudo` group here, or is it better with the above check?

mariusSincovici (approved):

internal/chmod_sock.go
```diff
@@ -0,0 +1,21 @@
+package internal
+
+import (
+	"log"
+	"os"
+)
+
+// CheckSockFile set socket file permissions
+func CheckSockFile(sockDirName, sockFileName string) {
```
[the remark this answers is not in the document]
keliramu: Right, will change name.

internal/socket.go
keliramu: There are 2 ways to have sudo permissions: 1) via sudo group 2) via sudoers file.
Cannot cover both just with checking group.

internal/socket.go
keliramu: Good question. But there is no way to do simple `sudo` check. Maybe need to implement some combined workaround check.

internal/chmod_sock.go
```diff
@@ -0,0 +1,21 @@
...
+	// #nosec G302 -- need owner and group permissions
+	if err := os.Chmod(sockDirName, 0775); err != nil {
+		log.Println(err)
+	}
+	if _, err := os.Stat(sockFileName); err != nil {
```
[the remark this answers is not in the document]
keliramu: Right, it is more convenient

internal/socket.go
keliramu: Left only check by `sudo` group.

internal/socket.go
keliramu: Converted all to syscalls

internal/chmod_sock.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+import (
+	"log"
+	"os"
+)
+
+// UpdateSocketFilePermissions set socket file permissions
+func UpdateSocketFilePermissions(sockFileName string) {
```
Savolro, line 9: Already done in `ManualListener`, we only need differentvalue for permissions

internal/socket.go
Savolro, line 81: This is quite a sensitive part. Could it have unit/integration tests?

Savolro (changes requested):

internal/chmod_sock.go
keliramu: What is the use case (purpose) of `ManualListener`?

cmd/daemon/main.go
```diff
@@ -480,6 +480,7 @@ func main() {
 			if err != nil {
 				log.Fatalf("Error on listening to UNIX domain socket: %s\n", err)
 			}
+			internal.UpdateSocketFilePermissions(ConnURL)
```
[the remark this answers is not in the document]
keliramu: Added condition by `SNAP_NAME` env var which is available only under Snap.

internal/chmod_sock.go
keliramu: `ManualListener` is for running non-systemd env e.g. Docker. But in this case it is not relevant except that it also does `Chmod`

internal/socket_test.go
```diff
@@ -24,3 +25,32 @@ func FuzzUcredToStringToUcred(f *testing.F) {
...
+
+func Test_authenticateUser(t *testing.T) {
+	tests := []struct {
+		name    string
+		uid     int
+		grps    []string
+		wantErr bool
+	}{
+		{
+			name:    "test1: existing user id 1000",
```
keliramu, line 37: Expecting in test environment to be available user with id 1000 and to belong to group `nordvpn` - now we have such in our tester docker container. Otherwise would need to setup groups and users which requires root access - dont have it in unit tests.

Savolro: We can assume that user running the tests can read the system users and find ID of the user matching the criteria

internal/socket.go
```diff
@@ -70,9 +71,41 @@ func getUnixCreds(conn net.Conn) (*unix.Ucred, error) {
 		return nil, fmt.Errorf("doing rawConn Control: %w", err)
 	}
 
+	if err := authenticateUser(ucred, []string{"nordvpn", "sudo", "wheel"}); err != nil {
```
Savolro, line 74: `sudo` and `wheel` can be removed as belonging to these groups does not necessarily mean user has full admin rights on the machine

internal/chmod_sock.go
```diff
@@ -0,0 +1,17 @@
...
+	if !FileExists(sockFileName) {
+		return
+	}
+	// #nosec G302 -- need world writable permissions
+	if err := os.Chmod(sockFileName, PermUserRWXGroupRWXOthersRWX); err != nil {
```
Savolro, line 14:
```suggestion
	if err := os.Chmod(sockFileName, PermUserRWGroupRWOthersRW); err != nil {
```

keliramu: Right! it's just a file ;)

internal/socket.go
keliramu: Removed

internal/socket_test.go
```diff
@@ -24,3 +26,39 @@ func FuzzUcredToStringToUcred(f *testing.F) {
...
+		{
+			name:    "test2: non existing user",
+			uid:     9000,
+			grps:    []string{"nordvpn"},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := authenticateUser(&unix.Ucred{Uid: uint32(tt.uid)}, tt.grps); (err != nil) != tt.wantErr {
```
mariusSincovici, line 59: [minor] 
```
err := authenticateUser(&unix.Ucred{Uid: uint32(tt.uid)}, tt.grps)
if tt.wantErr {
  assert.Error(t, err)
} else {
   assert.NoError(t, error)
}
```

mariusSincovici (approved):

internal/chmod_sock.go
```diff
@@ -0,0 +1,17 @@
...
+	"os"
+)
+
+// UpdateSocketFilePermissions set socket file permissions
+func UpdateSocketFilePermissions(sockFileName string) {
+	if !FileExists(sockFileName) {
+		return
+	}
+	// #nosec G302 -- need world writable permissions
+	if err := os.Chmod(sockFileName, PermUserRWGroupRWOthersRW); err != nil {
```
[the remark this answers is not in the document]
keliramu: Under snap environment, in case of error changing permissions, the client will not be able to communicate with daemon - we will log error into log file and that's it, nothing more we can do about it.

internal/chmod_sock.go
```diff
@@ -0,0 +1,17 @@
...
+		log.Println(err)
```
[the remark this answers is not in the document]
keliramu: Thx, added.

internal/socket_test.go
keliramu: Thx, modified.

internal/socket.go
```diff
@@ -70,9 +34,41 @@ func getUnixCreds(conn net.Conn) (*unix.Ucred, error) {
...
+func authenticateUser(ucred *unix.Ucred, allowGroups []string) error {
+	// root?
+	if ucred.Uid == 0 {
+		return nil
+	}
+	userInfo, err := user.LookupId(fmt.Sprintf("%d", ucred.Uid))
+	if err != nil {
+		return fmt.Errorf("authenticate user, lookup user info: %s", err)
+	}
+	// user belongs to 'nordvpn' or 'sudo' group?
```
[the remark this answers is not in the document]
keliramu: Modified comment.

Savolro (approved):

PR #240, opened 2024-01-22
Direct user to log in when attempting to use fileshare

bartoszWojciechO:


cli/messages.go
```diff
@@ -278,6 +278,7 @@ Learn more:
 	MsgFileshareDisconnectedPeer          = "Peer is disconnected."
 	MsgFileshareFileNotFound              = "File not found."
 	MsgFileshareSocketNotFound            = "Enable Meshnet to share files. If Meshnet is already enabled, try disabling and enabling it again. Use \"nordvpn set meshnet on\" to enable it."
+	MsgFileshareUserNotLoggedIn           = "You are not logged in. Please log in and enable meshnet to share files."
```
bartoszWojciechO, line 281: Message will be updated with input from copywriting team.

cli/cli_fileshare.go
```diff
@@ -109,7 +110,12 @@ func statusLoop(fileshareClient pb.FileshareClient, client transferStatusClient,
 // IsFileshareDaemonReachable returns error if fileshare daemon is not reachable, daemon not running
 // being the most likely cause
 func (c *cmd) IsFileshareDaemonReachable(ctx *cli.Context) error {
-	_, err := c.fileshareClient.Ping(context.Background(), &pb.Empty{})
+	resp, err := c.client.IsLoggedIn(context.Background(), &dpb.Empty{})
+	if err != nil || !resp.GetValue() {
+		return formatError(fmt.Errorf(MsgFileshareUserNotLoggedIn))
```
mariusSincovici, line 115: [minor] isn't good the error `internal.ErrNotLoggedIn`? If not, maybe change also in `fileshareServiceErrorCodeToError` to return the same error message?

mariusSincovici (approved):

cli/cli_fileshare.go
bartoszWojciechO: `ServiceErrorCode_NOT_LOGGED_IN` is not returned anywhere anyway, so we might just delete this branch from `fileshareServiceErrorCodeToError`. It was probably used back when fileshare was just a part of main daemon.

cli/cli_fileshare.go
```diff
@@ -109,7 +110,12 @@ func statusLoop(fileshareClient pb.FileshareClient, client transferStatusClient,
 // IsFileshareDaemonReachable returns error if fileshare daemon is not reachable, daemon not running
 // being the most likely cause
 func (c *cmd) IsFileshareDaemonReachable(ctx *cli.Context) error {
-	_, err := c.fileshareClient.Ping(context.Background(), &pb.Empty{})
+	resp, err := c.client.IsLoggedIn(context.Background(), &dpb.Empty{})
+	if err != nil || !resp.GetValue() {
```
keliramu, line 114: Why `OR !resp.GetValue()` part condition is used, in what situation it can happen?

keliramu (approved):

cli/cli_fileshare.go
bartoszWojciechO: Actually the response is a `*pb.Bool`, so `GetValue` returns a bool where true = user is logged in, false = user is not logged in.

PR #245, opened 2024-01-29
Fix notification send timeout

bartoszWojciechO:


mariusSincovici:
Why are the changes fixing the issue? :smile: 
Because `WithTimeout`calls `WithDeadline` and `CombinedOutput` calls `Run`

bartoszWojciechO:
> Why are the changes fixing the issue? 😄 Because `WithTimeout`calls `WithDeadline` and `CombinedOutput` calls `Run`
This is just to make the code nicer, the real fix is in line 73. https://jarv.org/posts/command-with-timeout/

mariusSincovici:
> > Why are the changes fixing the issue? 😄 Because `WithTimeout`calls `WithDeadline` and `CombinedOutput` calls `Run`
> 
> This is just to make the code nicer, the real fix is in line 73. https://jarv.org/posts/command-with-timeout/

:exploding_head:

daemon/notify.go
```diff
@@ -72,9 +70,9 @@ func notify(id int64, body string) error {
 	cmd.Env = append(cmd.Env, "DISPLAY=:0.0")
 	cmd.Env = append(cmd.Env, dbusAddr)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Credential: &syscall.Credential{Uid: uint32(id)}}
-	out, err := cmd.CombinedOutput()
+	err := cmd.Run()
```
mariusSincovici, line 73: Because this is a bit tricky `CombinedOutput` vs `Run` with timeout, would it make sense to have a function that runs the command to have the explanation? if not maybe add comment here why is `Run` used?

mariusSincovici (approved):

PR #250, opened 2024-02-01
Update Go version to 1.20.13

piotrjurkiewicz:


ci/docker/builder/Dockerfile
```diff
@@ -1,15 +1,15 @@
-FROM golang:1.20.1-buster
+FROM ghcr.io/nordsecurity/nordvpn-linux/golang:debian-buster-20240201
 
 LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
 
-ENV GOBIN=/usr/bin
-
 RUN dpkg --add-architecture i386; \
```
keliramu, line 5: `&&` should be used instead of `;` - because using `;` would not stop on error and try to execute following command which could depend on previous command and fail also, so, it's wasting time.

ci/docker/generator/Dockerfile
```diff
@@ -1,22 +1,20 @@
...
+    apt-get -y --no-install-recommends install unzip; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*
 
-RUN apt-get update \
-    && apt-get install -y zip \
-    # cleanup
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+RUN wget --progress=dot:giga https://github.com/protocolbuffers/protobuf/releases/download/v21.6/protoc-21.6-linux-x86_64.zip -O /tmp/protoc.zip && \
```
keliramu, line 10: `RUN` directives can be minimized, because each `RUN` creates another Docker layer and increases image size.

ci/docker/generator/Dockerfile
```diff
@@ -1,22 +1,20 @@
-FROM golang:1.20.1-buster
+FROM ghcr.io/nordsecurity/nordvpn-linux/golang:debian-buster-20240201
 
 LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
 
-ENV GOBIN=/usr/bin
+RUN apt-get update; \
```
keliramu, line 5: `;` --> `&&`

ci/docker/golang/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM debian:buster-slim
+# TODO: Change base image to ubuntu:20.04 on 2024-06-30
+# Then to debian:bullseye-slim on 2025-05-29
+# Then to ubuntu:22.04 on 2026-06-30
+# Then to debian:bookworm-slim on 2027-06-01
+# Then to ubuntu:24:04 on 2028-06-30
+
+LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
+
+RUN apt-get update; \
```
keliramu, line 10: `;` --> `&&`

ci/docker/golang/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        ca-certificates \
+        git; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*
+
+# Based on: https://github.com/docker-library/golang/blob/a6fd6eceb0cb26da2fceefb4353768c472f84420/1.21/bookworm/Dockerfile
+ENV GOLANG_VERSION 1.21.6
+ENV GOLANG_SHA256 3f934f40ac360b9c01f616a9aa1796d227d8b0328bf64cb045c7b8c4ee9caea4
+
+RUN wget --progress=dot:giga "https://go.dev/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz" -O /tmp/go.tar.gz && \
```
keliramu, line 26: `RUN` directives can be minimized and image size can be improved

ci/docker/packager/Dockerfile
```diff
@@ -1,25 +1,21 @@
...
-RUN go install github.com/goreleaser/nfpm/v2/cmd/nfpm@v2.5.1 && \
-    go install github.com/goreleaser/chglog/cmd/chglog@v0.1.2
-
 RUN apt-get update; \
-    apt-get -y install elfutils gettext-base sudo; \
+    apt-get -y --no-install-recommends install elfutils gettext-base sudo; \
     apt-get clean; \
     rm -rf /var/lib/apt/lists/*
 
+RUN go install github.com/goreleaser/nfpm/v2/cmd/nfpm@v2.5.1 && \
```
keliramu, line 10: `RUN` optimize

ci/docker/scanner/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
 
 LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
 
-RUN apt-get update -y \
-    && apt-get -y install devscripts \
-    && apt-get -y install curl sudo \
-    && curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s v2.13.0 \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+RUN apt-get update; \
```
keliramu, line 5: `;` --> `&&`

ci/docker/scanner/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
+    apt-get -y --no-install-recommends install \
+        devscripts \
+        sudo; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*
+
+RUN wget --progress=dot:giga https://raw.githubusercontent.com/securego/gosec/master/install.sh -O /tmp/install.sh && \
```
keliramu, line 12: `RUN` optimize

keliramu (dismissed):
`;` --> `&&`
`RUN` optimizations

mariusSincovici (approved):

ci/docker/golang/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM debian:buster-slim
+# TODO: Change base image to ubuntu:20.04 on 2024-06-30
```
Savolro, line 2: Specifying that this is the current `oldstable` should be enough

ci/docker/golang/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# Then to debian:bullseye-slim on 2025-05-29
+# Then to ubuntu:22.04 on 2026-06-30
+# Then to debian:bookworm-slim on 2027-06-01
+# Then to ubuntu:24:04 on 2028-06-30
+
+LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
+
+# Based on: https://github.com/docker-library/golang/blob/a6fd6eceb0cb26da2fceefb4353768c472f84420/1.21/bookworm/Dockerfile
```
Savolro, line 10: https://github.com/docker-library/golang/blob/a6fd6eceb0cb26da2fceefb4353768c472f84420/1.20/bullseye/Dockerfile
is probably the correct link.

I also think it would make sense to simply copy everything below the line 1 and specify it instead of modifying the generated code

keliramu, line 10: Comment shows go ver 1.21 - should be adjusted to 1.20, imo

Savolro (changes requested):

ci/docker/golang/Dockerfile
piotrjurkiewicz: The `github.com/docker-library/golang` Dockerfile installs some dependencies (for example `g++`) which add hundreds of MB to image size, but are not needed for us. The builder image, which is downloaded several times in the pipeline, built from the unmodified `github.com/docker-library/golang` Dockerfile is 500MB larger then built from our `ci/docker/golang/Dockerfile`.

ci/docker/golang/Dockerfile
piotrjurkiewicz: It won't. The current `oldstable` is `bullseye`, but we need also to support `buster` which is now `oldoldstable`. Also, for example after 2026-06-30 the oldest maintained Debian will be `bookworm`, which contains libc 2.36, but we will still need to support Ubuntu 22.04 until 2027-06-01, which contains libc 2.35. So switching every year between the oldest supported Debian and Ubuntu versions is the only solution.

Savolro (approved):

bartoszWojciechO (approved):

PR #255, opened 2024-02-05
Automate some of the Meshnet tests

naujalyte:


test/qa/test_meshnet.py
```diff
@@ -27,26 +34,15 @@ def setup_module(module):
...
     meshnet.revoke_all_invites()
     meshnet.remove_all_peers()
-    ssh_client.exec_command("nordvpn set mesh off")
-    sh.nordvpn.set.meshnet.off()
-    ssh_client.exec_command("nordvpn logout --persist-token")
-    sh.nordvpn.logout("--persist-token")
+    ssh_client.exec_command("nordvpn set defaults")
+    sh.nordvpn.set.defaults()
     daemon.stop_peer(ssh_client)
     daemon.stop()
```
piotrjurkiewicz, line 45: In `setup_function` we are logging in before each test. Shouldn't we also log out on teardown after each test?

piotrjurkiewicz: So it is also not needed here? :point_down:
https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/test_routing.py#L34
https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/test_allowlist_port.py#L31
https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/test_allowlist_subnet.py#L36

naujalyte: Yes, not needed there as well, just double-checked :D

test/qa/test_meshnet.py
```diff
@@ -464,3 +455,64 @@ def test_invite_accept_non_existant_special_character():
...
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+            sh.nordvpn.set(meshnet_allias, "off")
+
+    assert "Meshnet is already disabled." in str(ex.value)
+
+    sh.nordvpn.set.meshnet("on")
+
+
+@pytest.mark.parametrize("meshnet_allias", meshnet.MESHNET_ALIAS)
+def test_set_meshnet_on_logged_out(meshnet_allias):
```
mariusSincovici, line 493: [minor] wouldn't it be better to add a `when`?  test_set_meshnet_on_`when`_logged_out ? same for the next one :smile:

test/qa/test_meshnet.py
```diff
@@ -464,3 +455,64 @@ def test_invite_accept_non_existant_special_character():
...
+def test_set_meshnet_off_logged_out(meshnet_allias):
+    
+    sh.nordvpn.logout("--persist-token")
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+            sh.nordvpn.set(meshnet_allias, "off")
+
+    assert "You are not logged in." in str(ex.value)
+
+    login.login_as("default")
```
mariusSincovici, line 517: is `login` and `meshnet on` needed or just to check that it is successful?

test/qa/test_meshnet.py
```diff
@@ -464,3 +455,64 @@ def test_invite_accept_non_existant_special_character():
...
+
+    sh.nordvpn.set.meshnet("on")
+
+
+@pytest.mark.parametrize("meshnet_allias", meshnet.MESHNET_ALIAS)
+def test_set_meshnet_on_logged_out(meshnet_allias):
+    
+    sh.nordvpn.logout("--persist-token")
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
```
mariusSincovici, line 497: at logout shouldn't be also checked that into the app settings meshnet is off?

test/qa/test_meshnet.py
naujalyte: This is needed because otherwise tests would error at `teardown` because if we are logged out  and meshnet is off, we could not execute the revoking of peers and so on. Could not think of a better way to avoid this :D

test/qa/test_meshnet.py
naujalyte: Good point, will add this, thanks!

test/qa/test_meshnet.py
naujalyte: Yeah, probably would make a bit more sense :D

test/qa/test_meshnet.py
naujalyte: Changed to only remove peers and revoke invites at setup function instead of both setup and teardown and now those additional logins and enabling of Meshnet are not needed

piotrjurkiewicz (approved):

mariusSincovici (approved):

PR #256, opened 2024-02-05
Fix connect duration for events

mariusSincovici:
Send the correct event duration for connecting to VPN server.

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -313,7 +313,7 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
 		threatProtection = moose.Enum_SS_NordvpnappOptBool(moose.OptBoolFalse)
 	}
 
-	eventDuration := -1
+	eventDurationMs := -1
 	var result moose.Enum_SS_NordvpnappEventStatus
 	switch data.Type {
 	case events.ConnectAttempt:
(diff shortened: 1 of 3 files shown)


events/moose/moose.go
```diff
@@ -365,7 +364,7 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
 	}
 	return s.response(moose.Send_serviceQuality_servers_connect(
 		-1,
-		dnsResolutionTime,
+		int(time.Since(s.connectedAt).Seconds()),
```
Savolro, line 367: This will still not be correct as this function is called with `EventStatusAttempt`  after potential API request is already done

Savolro (changes requested):
This approach will still not be working

events/moose/moose.go
mariusSincovici: The `connecting` event is sent before picking up the VPN server.

events/moose/moose.go
```diff
@@ -306,19 +306,20 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
...
-	s.connectedAt = time.Now()
-	dnsResolutionTime := int(data.DNSResolutionTime.Milliseconds())
 	var threatProtection moose.Enum_SS_NordvpnappOptBool
 	if data.ThreatProtectionLite {
 		threatProtection = moose.Enum_SS_NordvpnappOptBool(moose.OptBoolTrue)
 	} else {
 		threatProtection = moose.Enum_SS_NordvpnappOptBool(moose.OptBoolFalse)
 	}
 
+	eventDuration := int(time.Since(s.connectedAt).Seconds())
```
piotrjurkiewicz, line 316: Maybe initialize here to -1, and then overwrite with actual calculated values in `case events.ConnectSuccess:` and `case events.ConnectFailure:`? The outcome will be the same, but it would be more logical and will also send -1 instead some incorrect value in `default:` case.

events/moose/moose.go
```diff
@@ -306,19 +306,20 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
...
 	var result moose.Enum_SS_NordvpnappEventStatus
 	switch data.Type {
 	case events.ConnectAttempt:
 		result = moose.Enum_SS_NordvpnappEventStatus(moose.EventStatusAttempt)
+		s.connectedAt = time.Now()
```
piotrjurkiewicz, line 321: Maybe rename the variable to `s.connectionStart` or something like that? Because in fact this is the timestamp of connection attempt start, not the moment of connection success.

Savolro (approved):

piotrjurkiewicz (approved):

PR #257, opened 2024-02-07
Arrange DNS handling methods

keliramu:


daemon/dns/dns_resolvectl.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (m *Resolvectl) IsAvailable() bool {
+	return internal.IsCommandAvailable(execResolvectl)
+}
+
+func (m *Resolvectl) Name() string {
+	return "resolvectl"
+}
+
+func setDNSWithResolvectl(iface string, addresses []string) error {
```
mariusSincovici, line 36: the [documentation](https://github.com/poettering/systemd/commit/8cedb0aef94da880e61b4c8cfeb7f450f8760ec6) suggests that following should be used to configure:
```
resolvectl domain <interface> '~.'
resolvectl default-route <interface> true
resolvectl dns <interface> 8.8.8.8
```

same here with the default-route, might be optional.

daemon/dns/dns_resolvd.go
```diff
@@ -0,0 +1,177 @@
...
+	return internal.IsServiceActive(serviceSystemdResolved)
+}
+
+func (m *Resolved) Name() string {
+	return "resolved"
+}
+
+// setDNSWithSystemdResolve uses systemd-resolve dbus API to manage DNS
+// https://www.freedesktop.org/wiki/Software/systemd/resolved/
+func setDNSWithSystemdResolve(ifname string, addresses []string) error {
```
mariusSincovici, line 45: The [docs](https://github.com/poettering/systemd/commit/8cedb0aef94da880e61b4c8cfeb7f450f8760ec6) suggest to use:
```
SetLinkDomains()
SetLinkDefaultRoute()
```
for the default route, might be optional but:
```
If managing a *privacy* VPN, you include `~.` in the routing domains,
 the value for `default-route` is actually irrelevant, but I'd set it to true. 
No need to configure any reverse lookup routing domains for it.
```

internal/filesystem.go
```diff
@@ -284,6 +284,9 @@ func FileSha256(filepath string) (sum []byte, err error) {
 // checks if command is in PATH
 func IsCommandAvailable(command string) bool {
 	_, err := exec.LookPath(command)
+	if err != nil {
+		log.Println(WarningPrefix, "command:", command, " is-available check result:", err)
```
mariusSincovici, line 288: shouldn't log message say that the command is not available?

internal/filesystem.go
```diff
@@ -343,6 +346,7 @@ func CliDimensions() ([]string, error) {
 func IsServiceActive(service string) bool {
 	out, err := exec.Command(SystemctlExec, "is-active", service).Output()
 	if err != nil {
+		log.Println(WarningPrefix, "service:", service, " is-active check result:", err)
```
mariusSincovici, line 349: same here?

daemon/dns/dns_test.go
```diff
@@ -0,0 +1,125 @@
+package dns
```
mariusSincovici, line 1: it might be good to have a test that checks that all the available commands are all iterated?
and that it stops at the first that doesn't return error?

daemon/dns/dns_resolvectl.go
```diff
@@ -0,0 +1,68 @@
...
+
+	cmdStr := []string{"dns", prefix + iface}
+	cmdStr = append(cmdStr, addresses...)
+	// #nosec G204 -- input is properly validated
+	cmd := exec.Command(execResolvectl, cmdStr...)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("setting dns with resolvconf: %s: %w", strings.TrimSpace(string(out)), err)
+	}
+	return nil
```
mariusSincovici, line 50: dns flush is missing: `resolvectl flush-caches`

daemon/dns/dns_resolvectl.go
```diff
@@ -0,0 +1,68 @@
...
+func unsetDNSWithResolvectl(iface string) error {
+	prefix, err := resolvconfIfacePrefix()
+	if err != nil {
+		return fmt.Errorf("determining interface prefix: %w", err)
+	}
+
+	// #nosec G204 -- the code would have failed already if iface did not belong
+	// to an actual network interface on the system.
+	// Just set empty/no DNS server for interface
+	cmd := exec.Command(execResolvectl, "dns", prefix+iface, "")
```
mariusSincovici, line 62: `resolvectl flush-caches`

daemon/dns/dns.go
```diff
@@ -52,32 +17,51 @@ type Setter interface {
...
+
 /*
 DefaultSetter handleds DNS in this order:
 
 1. If systemd-resolve command is available and systemd-resolved.service is
 running, systemd-resolve is used. It replaces /etc/resolv.conf file with
 a generated one.
 
-2. In absence of systemd-resolve, resolvconf command line utility is used, which
+2. In case of systemd-resolve is not accessible, resolvectl (which is part of
```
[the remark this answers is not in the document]
piotrjurkiewicz: This is just an alternative to dbus API. The symlink from `/etc/resolv.conf` to `/run/systemd/resolve/stub-resolv.conf` is created during `systemd-resolved` installation.

internal/filesystem.go
```diff
@@ -284,6 +284,9 @@ func FileSha256(filepath string) (sum []byte, err error) {
 // checks if command is in PATH
 func IsCommandAvailable(command string) bool {
 	_, err := exec.LookPath(command)
+	if err != nil {
```
piotrjurkiewicz, line 287: I don't think it should print anything to log when a command is unavailable. For example on Debian the last method is used, so it would always print two warnings.

daemon/dns/dns.go
```diff
@@ -52,32 +17,51 @@ type Setter interface {
 	Unset(iface string) error
 }
 
+// Method is abstraction of DNS handling method
+type Method interface {
```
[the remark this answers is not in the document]
keliramu: Well, it's in the same style as `Setter` and it's in the `dns` scope then that reads as `dns` `Setter` with `Method`.

daemon/dns/dns.go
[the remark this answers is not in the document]
keliramu: Right, comment is a bit missleading.

daemon/dns/dns_default.go
```diff
@@ -0,0 +1,100 @@
+package dns
```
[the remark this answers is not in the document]
keliramu: Right, I had this feeling as well, need to adjust.

daemon/dns/dns_test.go
keliramu: This kind of test i.e. iterate over methods, check availability, handle errors - it is implemented right now. Just it is abstracted from specific dns handling implementations. See `dns_test.go` `Test_Method`

internal/filesystem.go
keliramu: Probably you are right, at first I had this logging for my own testing/validation, later I left it thinking maybe it could be useful in real life when investigating end user situations to know which dns handling methods were tried (with what outcome) and which used in fact. To be removed.

internal/filesystem.go
keliramu: To be removed. See last comment to @piotrjurkiewicz feedback.

internal/filesystem.go
keliramu: To be removed. See last comment to @piotrjurkiewicz feedback.

daemon/dns/dns_resolvectl.go
keliramu: Thx, added.

daemon/dns/dns_resolvd.go
keliramu: Done

mariusSincovici (approved):

bartoszWojciechO (approved):

PR #259, opened 2024-02-07
Adjustments for clean builds

Savolro:


piotrjurkiewicz (approved):

build/openvpn/check_dependencies.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
mariusSincovici, line 1: [minor] it might be better to to use `#!/usr/bin/env bash`.

mariusSincovici (approved):

PR #260, opened 2024-02-07
Directory config

Savolro:


cmd/daemon/main.go
```diff
@@ -259,7 +260,7 @@ func main() {
 	dnsSetter := dns.NewSetter(infoSubject)
 	dnsHostSetter := dns.NewHostsFileSetter(dns.HostsFilePath)
 
-	eventsDbPath := fmt.Sprintf("%smoose.db", internal.DatFilesPath)
+	eventsDbPath := path.Clean(internal.DatFilesPath + "/moose.db")
```
mariusSincovici, line 263: using `filepath.Join` might be cleaner, and don't need to worry about forgetting `/`.

internal/filesystem.go
```diff
@@ -485,3 +485,28 @@ func IsNetworkLinkUnmanaged(link string) bool {
 	}
 	return false
 }
+
+// PrefixCommonPath is supposed to be used for files which are version specific and not persistent
+func PrefixCommonPath(p string) string {
+	return prefixPath(p, "PREFIX_COMMON")
+}
+
+// PrefixDataPath is supposed to be used for files which are non version specific and persistent
```
mariusSincovici, line 494: will there be any difference between this and the next function, because the description is the same?

mariusSincovici (approved):

bartoszWojciechO (approved):

daemon/constants.go
```diff
@@ -12,24 +16,26 @@ const (
 	// RepoURL is the url for NordVPN repository
 	RepoURL = "https://repo.nordvpn.com"
 
-	// IconPath defines icon file path
-	IconPath = "/usr/share/icons/hicolor/scalable/apps/nordvpn.svg"
+	// RandomComponentMin defines minimal value of random component
+	RandomComponentMin = 0
 
+	// RandomComponentMin defines maximum value of random component
```
keliramu, line 22: In comment variable name should be fixed

keliramu (approved):

PR #261, opened 2024-02-08
Start fileshare daemon as an orphan process

bartoszWojciechO:
### New startup flow for `nordfileshared`
`fileshared` is now independent of meshnet. Meshnet still needs to be enabled for fileshare, but enabling it will not start the daemon automatically. Instead, it will be started on first fileshare command, for user who ran that command. 

**New startup flow**

1. Check if fileshare is already running.
We do that by attempting a connection to the unix socket. With that attempts, we can infer the status, `nordfileshared` command will have different return codes for those statuses, so that nordvpn cli app can perform appropriate actions.

- **socket doesn't exist** - fileshare is not running, this doesn't return an error code, instead, `nordfileshared` will proceed with the startup.
- **socket exists but we do not have permissions to write to it** - fileshare is running for different user, this returns an error code, cli app can then show appropriate message to the user.
- **socket exists and we have permissions to write to it** - fileshare is running for the current user, this returns an error code that is silently discarded, instead, the cli app proceeds with executing the user command.

2. Check if meshnet is enabled.
If meshnet is not running, `nordfileshared` returns an error code and doesn't proceed with init.

3. Startup the gRPC server.

Disabling meshnet shuts down the fileshare process. This has to be done from the daemon, because user disabling the meshnet might be different than the user running the fileshare process. To that end, a `Stop` RPC has been added to the fileshare service.
https://github.com/NordSecurity/nordvpn-linux/pull/261/files#diff-fb8db7d837ca7ab60725e080b3128b3c1d5110cd2181cff78259ad4d9fffa738R13
https://github.com/NordSecurity/nordvpn-linux/pull/261/files#diff-51cde53e4d697b94837aff542c2b53ba75ff41a20d2a5c227249b66b55da95e6R166

The fileshare startup is somewhat complicated. User started fileshare with a command, but we cannot ensure that the process is ready to handle RPCs at that time. In order to make sure that server is actually running, on first(startup) command we do a ping with WaitReady option set to true.
https://github.com/NordSecurity/nordvpn-linux/pull/261/files#diff-2ccaa0db90a48ebd307a6e3f2cf4ed412abefe3fb79e02122b0f18d76840ab5bR155


keliramu (approved):
For snap adoption there will be different task/PR

cmd/fileshare/main.go
```diff
@@ -145,25 +177,36 @@ func main() {
...
 	grpcServer := grpc.NewServer()
 	pb.RegisterFileshareServer(grpcServer, fileshareServer)
 
-	var listenerFunction = internal.SystemDListener
-	if os.Getenv(internal.ListenPID) != strconv.Itoa(os.Getpid()) {
-		listenerFunction = internal.ManualListener(ConnURL, internal.PermUserRWX)
+	listener, err := net.Listen("unix", fileshare_process.FileshareSocket)
+
+	if err != nil {
+		log.Printf("Failed to open unix socket: %s", err)
```
mariusSincovici, line 202: If the fileshared process crashes it will never restart because the socket is already open:
`Failed to open unix socket: listen unix /tmp/fileshare.sock: bind: address already in use`

cli/cli_fileshare.go
```diff
@@ -119,14 +142,31 @@ func (c *cmd) IsFileshareDaemonReachable(ctx *cli.Context) error {
 		return formatError(fmt.Errorf(MsgFileshareUserNotLoggedIn))
 	}
 
-	_, err = c.fileshareClient.Ping(context.Background(), &pb.Empty{})
+	errChan := make(chan error)
+	go func() {
+		err := exec.Command("nordfileshared").Run()
```
mariusSincovici, line 147: wouldn't work to use `internal.Fileshared`?

PR #264, opened 2024-02-09
Patch Fix Resource exhaustion vulnerability in h2 may lead to Denial of Service

lamrecognitions:
An attacker with an HTTP/2 connection to an affected endpoint can send a steady stream of invalid frames to force the generation of reset frames on the victim endpoint. By closing their recv window, the attacker could then force these resets to be queued in an unbounded fashion, resulting in Out Of Memory (OOM) and high CPU usage.

```
        if let Err(Error::Reset(stream_id, reason, initiator)) = res {
            debug_assert_eq!(stream_id, stream.id);
            // Reset the stream.
            self.send
                .send_reset(reason, initiator, buffer, stream, counts, &mut self.task);
            Ok(())
```
```rb
   /// The default value is currently 1024, but could change.
    pub fn max_local_error_reset_streams(&mut self, max: Option<usize>) -> &mut Self {
```

### DIFF
--- a/build/foss/Cargo.lock
+++ b/build/foss/Cargo.lock
@@ -360,7 +360,7 @@ dependencies = [
  "bitflags 1.3.2",
  "clap_derive",
  "clap_lex",
- "indexmap",
+ "indexmap 1.9.1",
  "once_cell",
  "strsim 0.10.0",
  "termcolor",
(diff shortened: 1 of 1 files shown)


Savolro:
Hi! Thank you for contribution and sorry for delayed response. Please sign [CLA](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=264) as described in [CONTRIBUTE.md](https://github.com/NordSecurity/nordvpn-linux/blob/main/CONTRIBUTE.md#contributor-license-agreement) so we can accept your contribution.

piotrjurkiewicz:
@lamrecognitions We have not received the signed CLA from you within a week. In the meantime, the vulnerability has been solved by #287. Therefore we are closing this pull request.

PR #267, opened 2024-02-13
Setup initial project and add navigation bar

mariusSincovici:


ui/pubspec.lock
```diff
@@ -0,0 +1,218 @@
+# Generated by pub
```
mariusSincovici, line 1: not sure if this is needed.

PR #269, opened 2024-02-13
Before disabling IPv6 check if the kernel module is loaded

mariusSincovici:
If IPv6 is disabled from GRUB the application will fail to connect to VPN because it cannot disable IPv6.

### DIFF
--- a/kernel/sysctl_setter.go
+++ b/kernel/sysctl_setter.go
@@ -80,7 +80,7 @@ func (s *SysctlSetterImpl) Unset() error {
 }
 
 func (s *SysctlSetterImpl) Exists() bool {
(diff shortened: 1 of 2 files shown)


kernel/sysctl_setter.go
```diff
@@ -73,3 +78,15 @@ func (s *SysctlSetterImpl) Unset() error {
 	}
 	return nil
 }
+
+func (s *SysctlSetterImpl) Exists() bool {
+	if !internal.FileExists(strings.ReplaceAll(s.paramName, ".", ".")) {
```
piotrjurkiewicz, line 83: Shouldn't this be `!internal.FileExists("/proc/sys/" + strings.ReplaceAll(s.paramName, ".", "/"))`?

piotrjurkiewicz (approved):

PR #270, opened 2024-02-13
Fix help for commands with multiple parameters

mariusSincovici:


keliramu (approved):

cli/cli.go
```diff
@@ -1085,14 +1085,37 @@ func argsCountError(ctx *cli.Context) error {
...
 	return fmt.Errorf(
 		ArgumentParsingError,
 		ctx.App.Name,
-		ctx.Command.Name,
+		commandFullName(ctx, os.Args),
 	)
 }
+
+// because ctx.Command.FullName() doesn't work: https://github.com/urfave/cli/issues/1859
+func commandFullName(ctx *cli.Context, args []string) string {
```
Savolro, line 1101: Since this needs whole context, maybe it would make sense to modify template to not use `ctx.App.Name` separately and add it here?

cli/cli_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"context"
+	"flag"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli/v2"
+)
+
+func TestCLICommands(t *testing.T) {
```
Savolro, line 12: Please add test category

cli/cli.go
mariusSincovici: Now the function returns `appName + args`, also modified the message to accept one `%s`.

cli/cli.go
```diff
@@ -1084,15 +1084,36 @@ func meshnetErrorToError(code meshpb.MeshnetErrorCode) error {
...
-		ctx.App.Name,
-		ctx.Command.Name,
+		commandFullName(ctx, os.Args),
 	)
 }
+
+// because ctx.Command.FullName() doesn't work: https://github.com/urfave/cli/issues/1859
+func commandFullName(ctx *cli.Context, args []string) string {
+	if len(args) < 2 {
+		return ctx.Command.Name
```
mariusSincovici, line 1101: need to fix this to return app name when no command is returned.

piotrjurkiewicz (approved):

PR #272, opened 2024-02-13
Removed using interface name prefix for resolvectl method

keliramu:
Removed using interface name prefix for resolvectl method - problem appears when target system has `resolvconf` package installed but `resolvectl` dns configuration handling method is used.


mariusSincovici (approved):

daemon/dns/test/interface-order
```diff
@@ -0,0 +1,34 @@
+# interface-order(5)
```
Savolro, line 1: It would make sense to split the part `resolvconfIfacePrefix` which parses the content of the file and returns the appropriate prefix and then make higher level function which opens a file and does that. By doing this, the function can be tested with unit tests.

keliramu: ok, done

daemon/dns/dns_test.go
```diff
@@ -123,3 +124,42 @@ func Test_Method(t *testing.T) {
 		})
 	}
 }
+
+func Test_InterfacePreifx(t *testing.T) {
+	category.Set(t, category.Unit)
```
Savolro, line 129: This is an integration test

Savolro (approved):

PR #274, opened 2024-02-14
Automate set Defaults tests

naujalyte:


test/qa/lib/settings.py
```diff
@@ -51,3 +51,52 @@ def get_is_notify_enabled():
 def get_is_routing_enabled():
     """Returns True, if Routing is enabled in application settings."""
     return "Routing: enabled" in sh.nordvpn.settings()
+
+
+def get_is_autoconnect_enabled():
```
mariusSincovici, line 56: I saw that also other functions are named like this before, but if the function returns true/false it might be better if it starts with `is`/`has`? e.g.: `is_autoconnect_enabled`?
also for the others.

test/qa/test_settings.py
```diff
@@ -3,32 +3,26 @@
...
 
 
 def teardown_module(module):  # noqa: ARG001
-    sh.nordvpn.logout("--persist-token")
     daemon.stop()
 
 
 def setup_function(function):  # noqa: ARG001
     logging.log()
+    login.login_as("default")
```
mariusSincovici, line 19: wouldn't be good to set to defaults before logging in, in case the tests are already logged in and mesh is on?

piotrjurkiewicz (approved):

test/qa/test_settings.py
```diff
@@ -119,3 +120,251 @@ def test_technology_set_options(tech, proto, obfuscated):
...
+    assert settings.is_meshnet_on()
+    assert settings.get_is_ipv6_enabled()
+    assert not settings.get_is_dns_disabled()
+    
+    if obfuscated == "on":
+        assert settings.get_is_obfuscated()
+    else:
+        assert not settings.get_is_obfuscated()
+
+    sh.nordvpn.logout("--persist-token")
```
mariusSincovici, line 208: this might be an app bug, but if the user logs out, shouldn't the app switch to the default settings automatically?

test/qa/lib/settings.py
naujalyte: Yes, I agree that this would be better, will change for previous settings as well

test/qa/lib/settings.py
```diff
@@ -51,3 +51,52 @@ def get_is_notify_enabled():
...
+    """Returns True, if Analytics are enabled in application settings."""
+    return "Analytics: enabled" in sh.nordvpn.settings()
+
+
+def get_is_ipv6_enabled():
+    """Returns True, if IPv6 is enabled in application settings."""
+    return "IPv6: enabled" in sh.nordvpn.settings()
+
+
+def get_default_application_settings():
```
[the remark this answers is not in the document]
keliramu: `has_default_settings` = { true | false }

test/qa/test_settings.py
```diff
@@ -119,3 +120,251 @@ def test_technology_set_options(tech, proto, obfuscated):
...
+    else:
+        assert not settings.get_is_obfuscated()
+
+    assert MSG_SET_DEFAULTS in sh.nordvpn.set.defaults()
+
+    assert settings.get_default_application_settings()
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES) # Only using standard technologies here because of "LVPN-4601 - Enabling Auto-connect disables Obfuscation"
+def test_set_defaults_when_logged_in_2nd_set(tech, proto, obfuscated):
```
keliramu, line 156: is `2nd_set` related to `1st_set`? are they expected to run in sequence?

naujalyte: I made 2 sets since some settings cannot be disabled/enabled together e.g. routing cannot be disabled when meshnet is enabled, etc., as well as because of that bug we have where enabling auto connect disables obfuscation. So 2 sets were made in order to check if all settings, after they are changed, can be reset to default value.

test/qa/test_settings.py
naujalyte: Good question, settings are not reset to defaults upon logout, we do not have such a bug registered though, so not sure if this is the expected behavior or not. Do you think we should treat this as a bug?

test/qa/lib/settings.py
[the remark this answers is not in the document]
naujalyte: Changed to `app_has_default_settings` , thanks!

test/qa/test_settings.py
naujalyte: Discussed with PO, this will not be treated as a bug

keliramu (approved):

PR #275, opened 2024-02-14
Correctly check error message from sysctl

mariusSincovici:
If error message contains "cannot stat" then the value doesn't exist into the system.

### DIFF
--- a/kernel/sysctl_setter.go
+++ b/kernel/sysctl_setter.go
@@ -85,7 +85,7 @@ func (s *SysctlSetterImpl) Exists() bool {
 	}
 
 	if _, err := Parameter(s.paramName); err != nil {
-		return strings.Contains(err.Error(), "cannot stat")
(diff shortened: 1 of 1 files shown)


piotrjurkiewicz (approved):

kernel/sysctl_setter.go
```diff
@@ -85,7 +85,7 @@ func (s *SysctlSetterImpl) Exists() bool {
 	}
 
 	if _, err := Parameter(s.paramName); err != nil {
-		return strings.Contains(err.Error(), "cannot stat")
+		return !strings.Contains(err.Error(), "cannot stat")
```
keliramu, line 88: Why any other than "cannot stat" error is OK ? Maybe all/any error means parameter does not exist in a form we can use it?

keliramu (approved):

PR #277, opened 2024-02-20
Move OpenVPN build variables to OpenVPN directory

Savolro:
The only values needed from `ci/env.sh` by OpenVPN build scripts are versions of OpenVPN and its dependencies. Therefore, it is safe to move them. This improves build speeds performance on build systems which check for any file changes in source directory.


build/openvpn/check_dependencies.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
-source "${WORKDIR}/ci/env.sh"
+src_dir=$(dirname -- ${BASH_SOURCE[0]})
+source "${src_dir}/env.sh"
+
+cores=$(nproc)
```
keliramu, line 7: `cores` is not used here/below?

keliramu (approved):

mariusSincovici (approved):

PR #278, opened 2024-02-21
Serialize user information handling

keliramu:


internal/socket.go
```diff
@@ -5,79 +5,33 @@ import (
...
-type UnixSocketCredentials struct{}
+type UnixSocketCredentials struct {
+	mu sync.Mutex
+}
 
 // ServerHandshake is called when client connects to daemon.
 // We retrieve user ID which opened the client here.
-func (UnixSocketCredentials) ServerHandshake(c net.Conn) (net.Conn, credentials.AuthInfo, error) {
+func (cr *UnixSocketCredentials) ServerHandshake(c net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	// cgo calls to access user/group info are not thread safe, need to serialize
```
mariusSincovici, line 31: The comment might not be correct, since the code works when not running in snap. 
The problem might be related to snap env instead of the c functions.

internal/socket.go
```diff
@@ -140,3 +94,84 @@ func StringToUcred(ucredStr string) (unix.Ucred, error) {
...
+		}
+	}
+	return fmt.Errorf("requesting user does not have permissions")
+}
+
+// using `netutil.LimitListener“ but it has internal unexported type `limitListenerConn`
+// which wraps original `net.UnixConn` value inside by embeding abstract interface `net.Conn`
+// this way we cannot access `net.UnixConn` value, because of that we use `go reflection`
+// to extract original wrapped value.
+func extractConnection(c interface{}) net.Conn {
```
mariusSincovici, line 166: would it make sense to have a test for this, to catch if `LimitListener` changes?

mariusSincovici (approved):

internal/socket.go
keliramu: Good point, thx

Savolro (approved):

PR #283, opened 2024-02-27
Don't treat OpenVPN non-zero exit as stop failure

Savolro:


daemon/vpn/openvpn/openvpn.go
```diff
@@ -196,7 +196,9 @@ func (ovpn *OpenVPN) stop() error {
 
 	if ovpn.process.Process != nil {
 		if err := ovpn.process.Wait(); err != nil {
-			return err
+			// Don't return here as this could've failed due to non zero return code.
```
keliramu, line 199: What happens if OpenVPN really fails?

Savolro: It is not expected for OpenVPN to be running after stop was called. Therefore attempt to clean up the state (lines 211-214) is a better approach

keliramu (approved):

piotrjurkiewicz (approved):

PR #292, opened 2024-03-05
Add snap package support

Savolro:
This PR adds initial version of `snapcraft.yml` and snap permission checker which recommends to connect required snap interfaces.


cmd/daemon/main.go
```diff
@@ -473,7 +483,7 @@ func main() {
 			if err != nil {
 				log.Fatalf("Error on listening to UNIX domain socket: %s\n", err)
 			}
-			if os.Getenv(internal.InSnapOperation) != "" {
+			if snapconf.IsUnderSnap() {
 				internal.UpdateSocketFilePermissions(ConnURL)
```
piotrjurkiewicz, line 487: Maybe change the signature of `internal.UpdateSocketFilePermissions` to also get permissions parameter to achieve symmetry with `internal.ManualListener` which is called a few lines above:

``` go
listenerFunction = internal.ManualListener(ConnURL, internal.PermUserRWGroupRW)
if snapconf.IsUnderSnap() {
	internal.UpdateSocketFilePermissions(ConnURL, internal.PermUserRWGroupRWOthersRW)
}
```
That way all the changes of permissions will be clearly visible in a single file when following the code.

keliramu (approved):

piotrjurkiewicz (approved):

PR #295, opened 2024-03-07
Add interface and conditional build for fileshare process

bartoszWojciechO:


cmd/cli/fileshare_process.go
```diff
@@ -0,0 +1,11 @@
+//go:build fileshare_process
```
bartoszWojciechO, line 1: GRPC fileshare process is implemented in the next commit. For now this file is not used anywhere thanks to conditional compilation.

Savolro (approved):

keliramu (approved):

cmd/cli/fileshare_process.go
mariusSincovici: will `fileshare_process` be part of the `.env` file, added to `FEATURES`?

cmd/fileshare_process/main.go
```diff
@@ -0,0 +1,108 @@
...
+	// #nosec path is constant
+	logFile, err := os.OpenFile(path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0600)
+	if err != nil {
+		return nil, err
+	}
+	return logFile, nil
+}
+
+func main() {
+	if logFile, err := openLogFile(filepath.Join(fileshare_process.FileshareLogPath, "/nordfileshared.log")); err == nil {
```
mariusSincovici, line 31: is there any file rotation in place or something similar to restrict log file at a certain size?

cmd/fileshare_process/main.go
```diff
@@ -0,0 +1,108 @@
...
+		log.SetOutput(logFile)
+		log.SetFlags(log.LstdFlags | log.Lshortfile | log.Lmicroseconds)
+	}
+
+	processStatus := fileshare_process.GRPCFileshareProcess{}.ProcessStatus()
```
mariusSincovici, line 36: [minor] maybe use the helper function `NewGRPCFileshareProcess`?

cmd/fileshare_process/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if processStatus == fileshare_process.Running {
+		os.Exit(int(fileshare_process.CodeAlreadyRunning))
+	} else if processStatus == fileshare_process.RunningForOtherUser {
+		os.Exit(int(fileshare_process.CodeAlreadyRunningForOtherUser))
+		log.Println("Cannot start fileshare daemon, it is already running for another user.")
```
mariusSincovici, line 41: Is message getting logged?

fileshare_process/grpc_fileshare_process.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
+)
+
+type StartupErrorCode int
+
+const (
```
mariusSincovici, line 20: Will this errors need to be handled also by the GUI app?

cmd/cli/fileshare_process.go
bartoszWojciechO: As is, it's configured in `snapcraft/snap.yml. No reason to add it to `FEATURES` I think.

cmd/fileshare_process/main.go
bartoszWojciechO: I think so, logger is initialized before.

fileshare_process/grpc_fileshare_process.go
bartoszWojciechO: Some will probably need to be, at least to display correct error messages to the user. Depends on the GUI implementation.

cmd/fileshare_process/main.go
bartoszWojciechO: We don't do rotation for local user logs afaik, probably should implement that.

cmd/fileshare_process/main.go
mariusSincovici: but there is an `os.Exit` above it. isn't that stopping immediately the process?

mariusSincovici (approved):

PR #304, opened 2024-03-12
Implement norduserd startup

bartoszWojciechO:


cmd/norduser/main.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+func main() {
+	<-make(chan struct{})
```
keliramu, line 4: Is this the same as `time.Sleep(infinite)` ? What is the purpose of such non-functional main?

bartoszWojciechO: Just for testing purposes, will be removed once main is implemented.

norduser/service/fork.go
```diff
@@ -0,0 +1,105 @@
...
+	"golang.org/x/sys/unix"
+
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+// ErrNotStarted when disabling norduser
+var ErrNotStarted = errors.New("norduserd wasn't started")
+
+// ForkNorduser manages norduser service through exec.Command
+type ForkNorduser struct {
```
Savolro, line 23: This implementation does no fork but rather spawns a child process

norduser/service/combined.go
```diff
@@ -0,0 +1,86 @@
...
+const (
+	systemd processType = iota
+	fork
+)
+
+type Combined struct {
+	mu               sync.Mutex
+	systemd          SystemdNorduser
+	fork             ForkNorduser
+	uidToProcessType map[uint32]processType
```
Savolro, line 20: I assume this is designed to support both systemd and non systemd at the same time. Any valid reason for such environment?

Savolro (changes requested):

norduser/group_monitor.go
```diff
@@ -0,0 +1,165 @@
...
+	if err := watcher.Add(etcPath); err != nil {
+		return fmt.Errorf("adding file to watcher: %w", err)
+	}
+
+	currentGrupMembers, err := n.getNordvpnGroupMembers()
+	if err != nil {
+		return fmt.Errorf("getting initial group members: %w", err)
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
bartoszWojciechO: Would it? You cannot fetch errors from goroutines at the caller, so either channels or waitgroups would be needed.

internal/filesystem.go
```diff
@@ -420,14 +420,19 @@ func DBUSSessionBusAddress(id int64) string {
 	// #nosec G204 -- input is properly sanitized
 	out, err := exec.Command("ps", "-u", fmt.Sprintf("%d", id), "-o", "pid=").CombinedOutput()
 	if err != nil {
+		log.Println("Listing processes for uid: ", err.Error())
```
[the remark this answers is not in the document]
bartoszWojciechO: Callers ignores errors anyways, I have just added logs for some utility.

norduser/group_monitor.go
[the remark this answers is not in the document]
Savolro: Well, it depends on how caller uses the function. IMO very similar example to this use case is https://pkg.go.dev/net/http#ListenAndServe. has some setting up to do and returns errors e. g. if it cannot bind to the address. After setting up it runs indefinitely until some error happens and it returns an error.

internal/filesystem.go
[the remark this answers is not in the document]
Savolro: Generally it is good idea to give more leverage to the caller. Also, this may produce N logs every time daemon tries to send a notification for the user in case of permission denied accesss to `/proc/<pid>/environ` where N is pids per user.

cmd/norduser/main.go
keliramu: if it's WIP then why PR ?

bartoszWojciechO: Make PR fit in <1k changes.

Savolro (approved):

keliramu (approved):

PR #305, opened 2024-03-12
Fix verbose HTTP logging

Savolro:


events/logger/logger.go
```diff
@@ -55,9 +56,18 @@ func (Subscriber) NotifyRequestAPIVerbose(data events.DataRequestAPI) error {
 	// Additional read of response body. Do not use in production builds
 	var respBodyBytes []byte
 	if data.Response != nil {
-		respBodyBytes, _ := io.ReadAll(data.Response.Body)
+		rawRespBodyBytes, _ := io.ReadAll(data.Response.Body)
 		_ = data.Response.Body.Close()
-		data.Response.Body = io.NopCloser(bytes.NewBuffer(respBodyBytes))
+		var reader io.Reader = bytes.NewBuffer(bytes.Clone(rawRespBodyBytes))
+		if data.Response.Header.Get("Content-Encoding") == "gzip" {
```
keliramu, line 62: There are other compression types, we ignore them?

bartoszWojciechO (approved):

events/logger/logger.go
Savolro: This is the only one we we are requesting and the only one we actually handle in the main logic https://github.com/NordSecurity/nordvpn-linux/blob/main/core/core.go#L106

keliramu (approved):

PR #309, opened 2024-03-15
Add snap build to mage; add mage clean

keliramu:


magefiles/mage.go
```diff
@@ -95,6 +96,31 @@ func (View) Docs() error {
...
+		return err
+	}
+	fmt.Println("Cleanup ./dist...")
+	if err := sh.Run("rm", "-rf", "./dist"); err != nil {
+		return err
+	}
+	// cleanup folders left after building snap in docker
+	// if folders do not exist - no problem, no error
+	fmt.Println("Cleanup snapcraft...")
+	if err := sh.Run("sudo", "rm", "-rf", "./parts", "./stage", "./prime"); err != nil {
```
mariusSincovici, line 112: when `-f` is used I think no errors are reported.
Would it work to remove the `-f`, not to delete too much by accident?

mariusSincovici (approved):

magefiles/mage.go
keliramu: ok, agree, removed `-f`

piotrjurkiewicz (approved):

PR #311, opened 2024-03-15
Fix invalid IP in status under snap

keliramu:


networker/networker.go
```diff
@@ -532,7 +532,7 @@ func (netw *Combined) ConnectionStatus() (ConnectionStatus, error) {
 
 	stats, err := netw.vpnet.Tun().TransferRates()
 	if err != nil {
-		return ConnectionStatus{}, err
+		netw.publisher.Publish(fmt.Sprintf("tun transfer rates err: %s", err))
```
mariusSincovici, line 535: what could cause the error, is it permissions related?

Since `stats` is later used, wouldn't it be good to initialize it with empty class? At the moment `TransferRates()` does that, but if implementation changes in the future and that returns a `nil` pointer it might crash?

keliramu: Right, implemented addition snap interface `hardware-observe` connect and have left proper error handling.

mariusSincovici (approved):

bartoszWojciechO (approved):

piotrjurkiewicz (approved):

PR #314, opened 2024-03-18
Improve local builds - do not build openvpn if it exists

keliramu:


magefiles/mage.go
```diff
@@ -208,9 +208,13 @@ func (Build) Notices() error {
 func buildPackage(packageType string, buildFlags string) error {
 	mg.Deps(Build.Data)
 	mg.Deps(mg.F(buildBinaries, buildFlags))
-	mg.Deps(Build.Openvpn)
 	mg.Deps(Build.Notices)
 
+	// do not build openvpn dependency if it already exists
+	if !internal.FileExists(fmt.Sprintf("./bin/deps/openvpn/%s/latest/openvpn", build.Default.GOARCH)) {
```
mariusSincovici, line 214: do you know if it is possible to re-use some older build, or it is cleared at the end of CI pipeline?
For example the CI pipeline reuses same folder for nordvpn and that has openvpn already build from a previous CI pipeline run.

keliramu: This change is about local builds only, not CI. And it is about reusing what was built previously. Specifically about `openvpn` it is not much you can screw - it can be built either on host or in docker, but both binaries should run ok. Unless `openvpn` version is being upgraded - then you can use `mage clean` and start over. 
To start reusing `openvpn` artifacts in CI from previous builds - we need to go the same way as we do with `libnord`, but this is separate task/topic.

mariusSincovici (approved):

magefiles/mage.go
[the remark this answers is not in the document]
keliramu: In `mage` there is no information about which openvpn version is being used.

Savolro (approved):

PR #315, opened 2024-03-18
Fix dns handling on headless/docker systems

keliramu:
There could be a situation when `systemd` and `resolvectl` get installed in the docker image as a dependency e.g. `openssh-server` but is not active. Out DNS handling methods (one of them) was checking if `resolvectl` command is available in the system and then try to handle DNS settings using that command but it was non-functional. Implemented extra validation.

### DIFF
--- a/daemon/dns/dns_resolvectl.go
+++ b/daemon/dns/dns_resolvectl.go
@@ -5,8 +5,6 @@ import (
 	"log"
 	"os/exec"
 	"strings"
-
-	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
 // Executables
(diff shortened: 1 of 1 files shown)


daemon/dns/dns_resolvectl.go
```diff
@@ -27,7 +27,16 @@ func (m *Resolvectl) Unset(iface string) error {
 }
 
 func (m *Resolvectl) IsAvailable() bool {
-	return internal.IsCommandAvailable(execResolvectl)
+	if internal.IsCommandAvailable(execResolvectl) {
```
mariusSincovici, line 30: [minor] wouldn't be enough to have only the second if, without checking if the command exists before? because in the end the command is always executed, and if it doesn't exist will just return error.

mariusSincovici (approved):

daemon/dns/dns_resolvectl.go
keliramu: yes, good point, thx ;)

bartoszWojciechO (approved):

PR #316, opened 2024-03-19
Add socket and pid file handling

keliramu:


internal/filesystem.go
```diff
@@ -107,6 +108,43 @@ func (rdl *runDirListener) Addr() net.Addr {
...
 func ManualListener(socket string, perm fs.FileMode) func() (net.Listener, error) {
 	return func() (net.Listener, error) {
+		pidfile := DaemonPid
+		// check and read pid file
+		if FileExists(pidfile) {
+			out, err := FileRead(pidfile)
+			if err != nil {
+				// pid file exists, but is not readable,
+				// some garbage from previous run/failure? maybe we can cleanup and proceed?
+				return nil, fmt.Errorf("daemon pid file is not readable: %w", err)
```
mariusSincovici, line 118: If the PID file exists, but fails to read, does it mean that the daemon will not start?

internal/filesystem.go
```diff
@@ -107,6 +108,43 @@ func (rdl *runDirListener) Addr() net.Addr {
...
+			}
+			pidFromFile, err := strconv.Atoi(strings.TrimSpace(string(out)))
+			if err != nil {
+				// pid value is not valid integer, some garbage from previous run/failure?
+				log.Println(WarningPrefix, fmt.Errorf("daemon pid file does not contain valid integer value: %w", err))
+			} else {
+				// try to find existing process in process list by pid
+				processes, _ := process.Processes()
+				for _, process := range processes {
+					if process.Pid == int32(pidFromFile) {
```
mariusSincovici, line 128: maybe not a big issue, but since PIDs are reused by the OS, couldn't happen that a process with the same PID exists, but it is something else and then the daemon will not start?

internal/filesystem.go
keliramu: yes, but as we under `root` it would be some kind of very extreme situation.

internal/filesystem.go
keliramu: good point, I also was thinking about that, even simulated such situation, and probably need to add 2nd check by process name to be 100% sure we found what we needed.

cmd/daemon/main.go
```diff
@@ -589,6 +589,11 @@ func main() {
 	if err := rpc.StopKillSwitch(); err != nil {
 		log.Println(internal.ErrorPrefix, "stopping KillSwitch:", err)
 	}
+	if internal.FileExists(internal.DaemonPid) {
```
Savolro, line 592: Please move this logic to `internal.ManualListener`. `pidFilePath` can be an argument to that function.

internal/filesystem.go
```diff
@@ -107,6 +108,43 @@ func (rdl *runDirListener) Addr() net.Addr {
...
+				// some garbage from previous run/failure? maybe we can cleanup and proceed?
+				return nil, fmt.Errorf("daemon pid file is not readable: %w", err)
+			}
+			pidFromFile, err := strconv.Atoi(strings.TrimSpace(string(out)))
+			if err != nil {
+				// pid value is not valid integer, some garbage from previous run/failure?
+				log.Println(WarningPrefix, fmt.Errorf("daemon pid file does not contain valid integer value: %w", err))
+			} else {
+				// try to find existing process in process list by pid
+				processes, _ := process.Processes()
```
Savolro, line 126: There is no need to list all processes. Simply check for `/proc/<pid>` directory

internal/filesystem.go
Savolro: Check if `/proc/<pid>/comm` has the expected name. Make expected name configurable as this may be re-used by file sharing service

internal/filesystem.go
```diff
@@ -107,6 +108,43 @@ func (rdl *runDirListener) Addr() net.Addr {
 // ManualListener returns manually created listener with provided permissions
 func ManualListener(socket string, perm fs.FileMode) func() (net.Listener, error) {
 	return func() (net.Listener, error) {
+		pidfile := DaemonPid
```
Savolro, line 111: Please move pid logic to a separate function. and use it here

Savolro (changes requested):

go.mod
```diff
@@ -23,8 +23,11 @@ require (
 	github.com/jbowtie/ratago v0.0.0-20200401224626-3140c0a9b186
 	github.com/kofalt/go-memoize v0.0.0-20220914132407-0b5d6a304579
 	github.com/magefile/mage v1.14.0
+	github.com/mattn/go-isatty v0.0.17
 	github.com/milosgajdos/tenus v0.0.3
+	github.com/pkg/errors v0.9.1
 	github.com/quic-go/quic-go v0.40.1
+	github.com/shirou/gopsutil v3.21.11+incompatible
```
Savolro, line 30: I don't think we need a big dependency with some transitive dependencies for simply checking 1 file

cmd/daemon/main.go
keliramu: Done

internal/filesystem.go
keliramu: Yes, right, done

internal/filesystem.go
keliramu: Ok, done

Savolro (approved):

PR #318, opened 2024-03-19
Automate missing meshnet peer tests

ggediminass:


test/qa/test_meshnet.py
```diff
@@ -25,10 +31,30 @@ def teardown_function(function):  # noqa: ARG001
...
 
 def test_meshnet_connect():
-    # Ideally peer update should happen through Notification Center, but that doesn't work often
-    sh.nordvpn.meshnet.peer.refresh()
     peer = meshnet.PeerList.from_str(sh.nordvpn.mesh.peer.list()).get_external_peer()
-    assert meshnet.is_peer_reachable(ssh_client, peer)
+    this_device = meshnet.PeerList.from_str(ssh_client.exec_command("nordvpn mesh peer list")).get_external_peer()
+
+    # Local machine -> Remote machine
+    nickname = "A-a-A"
```
mariusSincovici, line 38: [minor] maybe set the nickname as `local-peer/machine` and `remote-peer/machine` instead, so then the comment is not needed anymore?

test/qa/test_meshnet.py
```diff
@@ -170,3 +196,120 @@ def test_set_meshnet_off_repeated(meshnet_allias):
...
+        return len(list) != len(set(list))
+
+    ssh_client.exec_command("sudo iptables -I OUTPUT 1 -p tcp -m tcp --sport 8765 -j DROP")
+    ssh_client.exec_command("sudo iptables -I OUTPUT 1 -p tcp -m tcp --dport 8765 -j DROP")
+
+    daemon.stop_peer(ssh_client)
+    ssh_client.exec_command("echo '' > /var/log/nordvpn/daemon.log")
+    daemon.start_peer(ssh_client)
+
+    time.sleep(30)
```
mariusSincovici, line 244: why 30 seconds? :)

test/qa/test_meshnet.py
```diff
@@ -170,3 +196,120 @@ def test_set_meshnet_off_repeated(meshnet_allias):
...
+
+        if json_match:
+            json_data = json.loads(json_match.group())
+            hostname = json_data.get("body", {}).get("hostname", "")
+            server_list.append(hostname)
+
+    # Same server should not be contacted twice in a row
+    assert not has_duplicates(server_list)
+
+    ssh_client.exec_command("sudo iptables -D OUTPUT -p tcp -m tcp --sport 8765 -j DROP")
```
mariusSincovici, line 262: if there is an error above this, wouldn't the ssh peer remain with this into the iptables?

test/qa/test_meshnet_routing.py
```diff
@@ -86,3 +91,173 @@ def test_killswitch_exitnode(lan_discovery: bool, local: bool):
     sh.nordvpn.disconnect()
     assert not daemon.is_connected()
     assert network.is_available()
+
+
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(60)
+def test_route_to_peer_when_already_routing():
```
mariusSincovici, line 98: is it while `connecting` or is already connected?

test/qa/test_meshnet.py
```diff
@@ -170,3 +196,120 @@ def test_set_meshnet_off_repeated(meshnet_allias):
...
+
+    peer_list = meshnet.PeerList.from_str(sh.nordvpn.mesh.peer.list())
+
+    with open(logging.FILE) as tester_log:
+        log_content = tester_log.read()
+        qapeer_hostname = peer_list.get_external_peer().hostname
+        base_test(log_content, qapeer_hostname)
+
+
+@timeout_decorator.timeout(40)
```
mariusSincovici, line 305: why is needed to fail after 40 seconds?

mariusSincovici (commented):
Do you know why are the majority of added tests flaky and need reruns?

test/qa/test_meshnet.py
ggediminass: Pretty much same test just for VPN connection is done @ test_connect.py::test_status_connected, so I thought we could keep the same logic :smile:

test/qa/test_meshnet_routing.py
ggediminass: It is when already connected :smile: 
I cannot think of better test name for this one. Maybe test_reroute_traffic, but I am just not sure how clear enough it sounds, since when we start routing, we are re-routing traffic, aren't we? :smile:

test/qa/test_meshnet.py
ggediminass: Good point. For me, it has became pretty normal to add this decorator to almost each test :smile:

test/qa/lib/ssh.py
```diff
@@ -37,6 +43,60 @@ def download_file(self, remote_path: str, path: str):
...
         self.client.close()
 
+    class Network:
+        def __init__(self, ssh_class_instance):
+            self.ssh_class_instance: Ssh = ssh_class_instance
+
+        def _is_internet_reachable(self, retry=5) -> bool:
+            """ returns True when remote host is reachable by it's public IP. """
+            i = 0
+            while i < retry:
```
bartoszWojciechO, line 53: Maybe `poll` function(`nordvpn-linux/test/qa/lib/__init__.py`) could be used here? Example of usage: https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/test_fileshare.py#L183

bartoszWojciechO (approved):

PR #319, opened 2024-03-19
Ensure that only regular files are used for logging

mariusSincovici:


internal/filesystem.go
```diff
@@ -508,3 +517,37 @@ func prefixPath(p string, envKey string) string {
 	}
 	return filepath.Clean(dir + p)
 }
+
+// Will open or create the given file.
+// If a file already exists with the given name but is not a regular file,
+// e.g. a symlink, it will be deleted and a regular file re-created instead
+func OpenOrCreateRegularFile(fileName string, flags int, permission fs.FileMode) (*os.File, error) {
```
mariusSincovici, line 524: open for suggestions, if you have a better name in mind :smile:

internal/filesystem.go
```diff
@@ -508,3 +517,38 @@ func prefixPath(p string, envKey string) string {
...
+			return nil, fmt.Errorf("not a regular file %s", fileName)
+		}
+
+		return file, nil
+	}
+
+	file, err := openFile()
+	if err != nil {
+		// try one more time to open the file if it failed for the first time
+		return openFile()
```
keliramu, line 550: Why need to try 2 times? Any public recommendations or experimental findings? Why not 3 times?

keliramu (approved):

bartoszWojciechO (approved):

internal/filesystem.go
mariusSincovici: I've changed and now it is executed only once.
Initial idea was to try second time, in case file was deleted after it was opened.

PR #321, opened 2024-03-20
norduser snap startup

bartoszWojciechO:


child_process/grpc_child_process_manager.go
```diff
@@ -0,0 +1,83 @@
...
+	Stop() error
+}
+
+type GRPCChildProcessManager struct {
+	processClient     ProcessClient
+	processBinaryPath string
+}
+
+func NewGRPCChildProcessManager(processClient ProcessClient, processBinaryPath string) GRPCChildProcessManager {
+	return GRPCChildProcessManager{
```
keliramu, line 23: Why not pointer?

child_process/grpc_child_process_manager.go
```diff
@@ -0,0 +1,83 @@
...
+		processClient:     processClient,
+		processBinaryPath: processBinaryPath,
+	}
+}
+
+func (g GRPCChildProcessManager) StartProcess() (StartupErrorCode, error) {
```
keliramu, line 29: Why not pointer?

child_process/grpc_child_process_manager.go
```diff
@@ -0,0 +1,83 @@
...
+		pingChan <- err
+	}()
+
+	select {
+	case err := <-errChan:
+		var exiterr *exec.ExitError
+		if errors.As(err, &exiterr) {
+			exitCode := StartupErrorCode(exiterr.ExitCode())
+			return exitCode, nil
+		} else {
```
keliramu, line 51: No need for `else`

child_process/grpc_child_process_manager.go
```diff
@@ -0,0 +1,83 @@
...
+	pingChan := make(chan error)
+	// Start another goroutine where we ping the WaitForReady option, so that server has time to start up before we run
+	// the acctual command.
+	go func() {
+		err := g.processClient.Ping(false)
+		pingChan <- err
+	}()
+
+	select {
+	case err := <-errChan:
```
keliramu, line 46: Here you might receive `error` interface pointer with `nil` value. Is this part working ok?

internal/constants.go
```diff
@@ -125,6 +130,10 @@ var (
 
 	// DaemonSocket defines system daemon socket file location
 	DaemonSocket = filepath.Join(RunDir, "/nordvpnd.sock")
+
+	FileshareBinaryPath = AppDataPathStatic + "/" + Fileshare
```
keliramu, line 134: filepath.Join ?

internal/constants.go
```diff
@@ -125,6 +130,10 @@ var (
...
+
+	NorduserBinaryPath = AppDataPathStatic + "/" + Norduserd
```
keliramu, line 136: filepath.Join ?

internal/constants.go
```diff
@@ -210,7 +198,7 @@ func GetNorduserdLogPath(uid string) string {
 
 	if err != nil {
 		log.Printf("users norduserd logs will be stored in %s: %s", LogPath, err.Error())
-		return filepath.Join(LogPath, Norduserd+"-"+uid+LogFileExtension)
+		return filepath.Join(LogPath, Norduser+"-"+uid+LogFileExtension)
```
keliramu, line 201: fmt.Sprintf ?

magefiles/mage.go
```diff
@@ -631,22 +631,19 @@ func (Install) Binaries() error {
...
+	filesharedSrc := fmt.Sprintf("bin/%s/%s", build.Default.GOARCH, internal.Fileshare)
 	if err := cp(filesharedSrc, filesharedDst); err != nil {
 		return err
 	}
 
-	norduserDst, err := sh.Output("which", internal.Norduserd)
+	norduserDst, err := sh.Output("which", internal.Norduser)
 	if err != nil {
-		norduserDst = "/usr/bin/" + internal.Norduserd
+		norduserDst = "/usr/bin/" + internal.Norduser
```
keliramu: fmt.Sprintf ?

norduser/grpc_interceptor.go
```diff
@@ -13,10 +13,10 @@ import (
 
 // StartNorduserdMiddleware provides a way to start/stop norduserd when handling nordvpnd gRPCs.
 type StartNorduserdMiddleware struct {
```
keliramu, line 15: if it is used in daemon, maybe we should implement thread safety measures?

internal/constants.go
bartoszWojciechO: What's the difference?

norduser/grpc_interceptor.go
bartoszWojciechO: Combined norduser service is thread safe.

child_process/grpc_child_process_manager.go
bartoszWojciechO: Can you explain? `errChan` is an `error` channel.

fileshare/fileshare_process/fileshare_process_manager.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return err
+}
+
+func (f *FileshareProcessClient) Stop() error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	client, clientConn, err := getFileshareClient()
+	defer clientConn.Close()
```
mariusSincovici, line 63: same here for `defer`

child_process/grpc_child_process_manager.go
keliramu: Even in case 1st goroutine executes command and no error happens there, here `nil` value will be received and will be reported outside as failure to start binary

fileshare/fileshare_process/fileshare_process_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	client := pb.NewFileshareClient(fileshareConn)
+	return client, fileshareConn, nil
+}
+
+func (f *FileshareProcessClient) Ping(nowait bool) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	client, clientConn, err := getFileshareClient()
+	defer clientConn.Close()
```
mariusSincovici, line 43: shouldn't defer be after error checking? otherwise it might crash if clientConn is nil.

internal/constants.go
keliramu: ok, minor, just code consistency

norduser/process/norduser_process_manager.go
```diff
@@ -0,0 +1,93 @@
...
+	uid uint32
+}
+
+func NewNorduserProcessClient(uid uint32) *NorduserProcessClient {
+	return &NorduserProcessClient{
+		uid: uid,
+	}
+}
+
+func getNorduserClient(uid uint32) (pb.NorduserClient, *grpc.ClientConn, error) {
```
mariusSincovici, line 31: lot of code from here is the same as is [fileshare_process_manager.go](https://github.com/NordSecurity/nordvpn-linux/pull/321/files#diff-8d942a082510951ed70f6e44609977dcd7bfbe7661628b049366afc8e52a7bda)
would it be possible in the future to mode them in one place?

norduser/process/norduser_process_manager.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}()
+
+	if err != nil {
+		return fmt.Errorf("failed to initialize the connection: %w", err)
+	}
+
+	callOptions := []grpc.CallOption{}
+	if !nowait {
+		callOptions = append(callOptions, grpc.WaitForReady(true))
```
mariusSincovici, line 62: same here, maybe the if is not needed?

norduser/server.go
```diff
@@ -4,32 +4,61 @@ import (
...
-	fileshareProcessManager fileshare_process.GRPCFileshareProcess
+	fileshareProcessManager childprocess.ChildProcessManager
 	fileshareRunning        bool
+	stopChan                chan<- interface{}
 }
 
-func NewServer(fileshareProcessManager fileshare_process.GRPCFileshareProcess) *Server {
+func NewServer(fileshareProcessManager childprocess.ChildProcessManager, stopChan chan<- interface{}) *Server {
 	return &Server{
 		fileshareProcessManager: fileshareProcessManager,
```
mariusSincovici, line 30: the `stopChan` is not initialized with the constructor param, shouldn't it be?

norduser/server.go
```diff
@@ -4,32 +4,61 @@ import (
...
+		returnStatus = pb.StartFileshareStatus_SUCCESS
+	} else {
+		log.Println("Failed to start fileshare, return code: ", returnCode)
+		if s, ok := errorCodeToProtobuffError[returnCode]; ok {
+			returnStatus = s
+		}
 	}
 
-	return &pb.StartFileshareResponse{StartFileshareStatus: status}, nil
+	log.Println("Fileshare started")
```
mariusSincovici, line 59: if there is an error, is this log valid?

norduser/service/snap.go
```diff
@@ -0,0 +1,64 @@
...
+	output, err := exec.Command("ps", "-C", internal.Norduserd, "-o", "uid=").CombinedOutput()
+	if err != nil {
+		log.Println("Failed to list running norduser instances: ", err)
+	}
+
+	uids := string(output)
+	for _, uid := range strings.Split(uids, "\n") {
+		uidInt, err := strconv.Atoi(uid)
+		if err != nil {
+			log.Printf("Invalid unix user id, failed to convert from string: %s", uid)
```
mariusSincovici, line 49: shouldn't there be a `continue`, otherwise `uidInt` might contain wrong values?

child_process/grpc_child_process_manager.go
bartoszWojciechO: Ah. In this case we always expect the started binary to outlive this function. If exec returns before the ping(second case in select), it means that something wrong has happened.

cmd/norduser/main.go
```diff
@@ -2,22 +2,113 @@ package main
...
+
+	socketPath := internal.GetNorduserSocketSnap(uint32(uid))
+
+	if err := os.Remove(socketPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		log.Println("Failed to remove old socket file: ", err)
+	}
+
+	listener, err := internal.ManualListener(socketPath, internal.PermUserRWGroupRWOthersRW)()
+	defer func() {
+		if err != nil {
```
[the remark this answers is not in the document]
bartoszWojciechO: `grpcServer.GracefulStop` closes the listener

cmd/norduser/main.go
```diff
@@ -29,7 +120,7 @@ func main() {
 	}
 	listener = netutil.LimitListener(listener, 100)
 
-	server := norduser.NewServer(fileshare_process.GRPCFileshareProcess{})
+	server := norduser.NewServer(fileshare_process.NewFileshareGRPCProcessManager(), make(chan<- interface{}))
```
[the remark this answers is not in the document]
bartoszWojciechO: We don't need it in case of non-snap environments, main daemon or systemd will do the shutdown for us.

cmd/norduser/main.go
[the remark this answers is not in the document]
mariusSincovici: ok, but if `err != nil` then `listener == nil` and wouldn't crash?

norduser/process/norduser_process_manager.go
bartoszWojciechO: Generally most of the code in fileshare/norduser process managers is trivial and making it generic would require some generics or factory convolution(it all comes down to the fact that grpc clients are of a different type). So I just chose to duplicate it, because it was relatively low cost. But there is no problem to change it if preference is to have it commonized.

fileshare/fileshare_process/fileshare_process_manager.go
```diff
@@ -0,0 +1,76 @@
...
+
+type FileshareProcessClient struct {
+	mu sync.Mutex
+}
+
+func NewFileshareProcessClient() *FileshareProcessClient {
+	return &FileshareProcessClient{}
+}
+
+func getFileshareClient() (pb.FileshareClient, *grpc.ClientConn, error) {
```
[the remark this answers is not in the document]
bartoszWojciechO: This struct might be used in multiple different places, at different points in lifetime of fileshare process. For example, an use case where fileshare process is still not running but the client is still making some request(pinging) is valid. So we'd need some 'smart' connection reinitialization. But managers are used only for shutdown/startup and not any kind of regular operations, so imo current implementation is good enough.

internal/constants.go
bartoszWojciechO: No problem, I'll update it. Just thought that there is some functional difference I was not aware of.

child_process/grpc_child_process_manager.go
bartoszWojciechO: Added separate error message for this case.

cmd/norduser/main.go
```diff
@@ -2,22 +2,113 @@ package main
...
+	}
+
+	socketPath := internal.GetNorduserSocketSnap(uint32(uid))
+
+	if err := os.Remove(socketPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		log.Println("Failed to remove old socket file: ", err)
+	}
+
+	listener, err := internal.ManualListener(socketPath, internal.PermUserRWGroupRWOthersRW)()
+	defer func() {
```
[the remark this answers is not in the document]
bartoszWojciechO: Moved, GH didn't mark as outdated for some reason.

cmd/norduser/main.go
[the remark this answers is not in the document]
bartoszWojciechO: This should be resolved by moving defer after post initialization err check, right? Already did that as suggested in the comment bellow.

mariusSincovici (approved):

keliramu (approved):

keliramu:
ToDo (not in this PR):
1. log file prevent not to open existing link (not merged in main);
2. implement `pid` to prevent `address already in use` situation (not yet merged in main for main daemon);

PR #324, opened 2024-03-21
norduser snap autostart

bartoszWojciechO:


cli/cli_user.go
```diff
@@ -0,0 +1,9 @@
+package cli
+
+import "github.com/urfave/cli/v2"
+
+// User is a dummy command, called by an autostart dekstopfile to start up norduser process in snap environment.
```
mariusSincovici, line 5: `dekstopfile`

cmd/norduser/main.go
```diff
@@ -32,6 +36,66 @@ func openLogFile(path string) (*os.File, error) {
...
+		"\nComment=This is an autostart for NordVPN user daemon" +
+		"\nCategories=Utility;"
+
+	path := path.Join(userHomeDir, ".config", "autostart", "nordvpn.desktop")
+	if err := internal.EnsureDir(path); err != nil {
+		return "", fmt.Errorf("ensuring path: %w", err)
+	}
+
+	// #nosec G304 -- no input comes from the user
+	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
mariusSincovici, line 58: `PermUserRW` instead of 0600?

mariusSincovici, line 58: isn't this already implemented into `internal.FileWrite`?

cmd/norduser/main.go
```diff
@@ -32,6 +36,66 @@ func openLogFile(path string) (*os.File, error) {
...
+		daemonURL,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return false, fmt.Errorf("can't connect to daemon: %w", err)
+	}
+
+	defer func() {
+		if err := grpcConn.Close(); err != nil {
+			log.Println("failed to close grpc connection on failure")
```
mariusSincovici, line 83: `failed to close grpc connection`, without `on failure` because this is always executed?

norduser/server.go
```diff
@@ -39,6 +42,13 @@ func (s *Server) Ping(context.Context, *pb.Empty) (*pb.Empty, error) {
 
 func (s *Server) StartFileshare(context.Context, *pb.Empty) (*pb.StartFileshareResponse, error) {
 	log.Println("Starting nordfileshare process")
+
+	fileshareStatus := s.fileshareProcessManager.ProcessStatus()
+	if fileshareStatus == childprocess.Running || fileshareStatus == childprocess.RunningForOtherUser {
+		log.Println("Received start fileshare request but fileshare is already running, status is: ", fileshareStatus)
+		return &pb.StartFileshareResponse{StartFileshareStatus: pb.StartFileshareStatus_SUCCESS}, nil
```
mariusSincovici, line 49: If the process is running for another user, is it ok to report success because current user might not be able to use it?

mariusSincovici (approved):

norduser/server.go
bartoszWojciechO: Should be enough imo, we start it silently, we do not report anything to the user. At best we could log something on the other side, but we already have norduser logs. Later we can make this more sophisticated if needed.

Savolro (approved):

PR #339, opened 2024-03-26
Fix missing downloads directory in snap package

Savolro:
Snap sets the value of `$HOME` to `$SNAP_USER_DATA`, therefore explicit
    usage of `$SHAN_USER_DATA` is not needed. This results in `$HOME` value
    respected in most places with the exception of `addAutostart` as it will
    not work if added to `$SNAP_USER_DATA`.

### DIFF
--- a/cmd/norduser/main.go
+++ b/cmd/norduser/main.go
@@ -125,10 +125,18 @@ func shouldEnableFileshare(uid uint32) (bool, error) {
 func startSnap() {
 	usr, err := user.Current()
 	if err != nil {
+		log.Println("Unable to retrieve current user:", err)
 		os.Exit(int(childprocess.CodeFailedToEnable))
 	}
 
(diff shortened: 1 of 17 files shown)


cmd/norduser/main.go
```diff
@@ -90,14 +90,21 @@ func startSnap() {
 		os.Exit(int(childprocess.CodeFailedToEnable))
 	}
 
-	configDirPath, err := internal.GetConfigDirPath(usr.HomeDir)
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
```
mariusSincovici, line 94: if not message can be logged, shouldn't be different error codes for this errors? otherwise might be hard for customer support and user to identify the error.

mariusSincovici (approved):

bartoszWojciechO (approved):

PR #346, opened 2024-03-31
Refactor connect tests

ggediminass:


test/qa/lib/__init__.py
```diff
@@ -274,10 +274,10 @@ def flush_allowlist():
 def is_connect_successful(output, name="", hostname=""):
     if name and hostname:
```
mariusSincovici, line 275: If `name` and `hostname` are empty, wouldn't it be good to get them from the `output`? 
Then always check that they are the same for connecting and connected message, and always use
```
f"Connecting to {name} ({hostname})" in output
            and f"You are connected to {name} ({hostname})!" in output
```

test/qa/lib/daemon.py
```diff
@@ -156,3 +156,28 @@ def is_peer_running(ssh_client: ssh.Ssh) -> bool:
...
+                  'P2P',
+                  'Standard_VPN_Servers',
+                  'The_Americas']
+
+    current_groups = str(sh.nordvpn.groups(_tty_out=False)).strip().split(", ")
+
+    return set(all_groups) - set(current_groups)
+
+
+def get_status_data() -> dict:
```
mariusSincovici, line 179: In the future, it might be good to have a class that stores the status. It might be easier to work with and also have the code in one place.

test/qa/lib/network.py
```diff
@@ -1,18 +1,73 @@
...
+    packets = tshark_result.strip().split("\n")
+
+    return len(packets)
+
+
+def capture_traffic(connection_settings) -> int:
+    """ Returns count of captured packets. """
+
+    # We try to capture packets using other thread
+    t_connect = PacketCaptureThread(connection_settings)
```
mariusSincovici, line 65: maybe you should also generate some packets, ping or curl. because I've saw that the pipeline failed because no packets were generated in those 3 seconds.
Otherwise the test relies on someone else to generate traffic.

bartoszWojciechO (approved):

mariusSincovici (approved):

PR #352, opened 2024-04-10
Do not enable norduser when its already enabled

bartoszWojciechO:


Savolro (approved):

norduser/service/combined.go
```diff
@@ -40,15 +40,19 @@ func (c *Combined) Enable(uid uint32, gid uint32, home string) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	if _, ok := c.uidToProcessType[uid]; ok {
```
mariusSincovici, line 43: if the norduser process crashes/is killed it will never be restarted.

PR #353, opened 2024-04-10
prevent norduser zombies

bartoszWojciechO:


norduser/service/fork.go
```diff
@@ -78,7 +78,13 @@ func (f *ChildProcessNorduser) Enable(uid uint32, gid uint32) (err error) {
 	cmd.SysProcAttr = &syscall.SysProcAttr{Credential: credential}
 	f.commandHandles[uid] = cmd
 
-	return cmd.Start()
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("starting the process: %w", err)
+	}
+
+	go cmd.Wait()
```
mariusSincovici, line 85: `Wait` waits: `waits for any copying to stdin or copying from stdout or stderr to complete`.
norduser process uses a file for logs, but if if fails to open the file for some reasons, won't there be an issue in that case?

mariusSincovici (approved):

norduser/service/fork.go
bartoszWojciechO: Actually there is a separate case where norduser fails to open the log file, and in such cases it just terminates.

keliramu (approved):

PR #354, opened 2024-04-10
App state

bartoszWojciechO:


cli/cli.go
Savolro, line 1: Don't forget to remove these changes. CLI does not need to listen to the states

state/state.go
```diff
@@ -0,0 +1,77 @@
+package state
+
+import (
+	"log"
+	"sync"
+
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+type Event int
```
Savolro, line 10: Check `events.DataConnect` and `events.DataDisconnect`

state/state.go
```diff
@@ -0,0 +1,77 @@
...
+		case <-sub.stopChan:
+			delete(s.subscribers, sub)
+		case sub.stateChan <- e:
+		}
+	}
+
+	return nil
+}
+
+func (s *StatePublisher) AddSubscriber() (<-chan Event, chan<- struct{}) {
```
Savolro, line 59: This can be a regular `subject.Subject` which has a subscriber that sends events via gRPC on notify

Savolro (changes requested):

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -46,14 +47,27 @@ func maskPublicKey(event string) string {
 	return expr.ReplaceAllString(event, `"public_key":"***"`)
 }
 
-func eventCallback(states chan<- state) eventFn {
+func publish(state string, publisher *vpn.Events) {
+	switch state {
+	case "connected":
+		publisher.Connected.Publish(events.DataConnect{})
+	case "connecting":
+		publisher.Disconnected.Publish(events.DataDisconnect{})
```
mariusSincovici, line 55: How will the app differentiate between `connecting` and `disconnected` state?

protobuf/state/state.proto
```diff
@@ -0,0 +1,13 @@
...
+
+package statepb;
+
+option go_package = "github.com/NordSecurity/nordvpn-linux/state/pb";
+
+message Empty {}
+
+message AppState {
+    oneof state {
+        bool vpn_enabled = 1;
```
mariusSincovici, line 11: I don't think this will be sufficient. This should be an enum with at least 3 states, optionally 4.
* connecting
* connected
* disconnecting (optional)
* disconnected
In the `connecting` and `connected` state it should also include the server(or the peer) info, maybe more info if needed.

state/state.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func newSubscriber(stateChan chan<- Event, stopChan <-chan struct{}) subscriber {
+	return subscriber{
+		stateChan: stateChan,
+		stopChan:  stopChan,
+	}
+}
+
+type subscriberSet map[subscriber]bool
```
mariusSincovici, line 30: If the `bool` value is not important there, `mapset.Set[]` could be used instead?

state/state.go
```diff
@@ -0,0 +1,91 @@
...
+
+type StatePublisher struct {
+	mu          sync.Mutex
+	subscribers subscriberSet
```
mariusSincovici, line 34: why does it need to be a set/map, because there is no lookup or any other functionality specific to maps used?

state/state.go
```diff
@@ -0,0 +1,91 @@
...
+func NewState() StatePublisher {
+	return StatePublisher{
+		subscribers: make(subscriberSet),
+	}
+}
+
+func (s *StatePublisher) notify(e Event) {
+	log.Println(internal.InfoPrefix+" received notification: ", e)
+	for sub := range s.subscribers {
+		select {
```
mariusSincovici, line 46: shouldn't there be a `default` also in this case? if for some reason the event cannot be sent then it will remain blocked in this place?

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -619,30 +627,88 @@ func (l *Libtelio) Public(private string) string {
...
+
+	for {
+		select {
+		case state := <-states:
+			switch state.State {
+			case "connecting":
+				publishConnecting(eventsPublisher, connParameters.server)
+			case "connected":
+				if state.PublicKey == connParameters.pubKey &&
+					state.State == "connected" {
```
mariusSincovici, line 695: `state.State == "connected"` is not needed because it is checked into the `switch`

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -619,30 +627,88 @@ func (l *Libtelio) Public(private string) string {
...
+					if !connectedPublished {
+						isConnected <- true
+						connectedPublished = true
+						close(isConnected)
+					}
+				}
+				publishConnected(eventsPublisher, connParameters.server)
```
mariusSincovici, line 702: shouldn't this be inside the `if state.PublicKey == connParameters.pubKey`? to only notify if the event was for the server

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -619,30 +627,88 @@ func (l *Libtelio) Public(private string) string {
...
+			case "disconnected":
+				publishDisconnected(eventsPublisher)
+			}
+		case <-ctx.Done():
+			publishDisconnected(eventsPublisher)
```
mariusSincovici, line 707: after 30 seconds this will always generate a `disconnected` event and then exit the for loop, even if it is still connected to the VPN

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -315,6 +319,10 @@ func (l *Libtelio) Stop() error {
 
 // disconnect from all the exit nodes, including VPN server
 func (l *Libtelio) disconnect() error {
+	if l.cancelConnectionMonitor != nil {
```
mariusSincovici, line 322: after the 30 seconds timer, this will not generate any disconnected events

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -247,24 +250,24 @@ func (l *Libtelio) Start(
...
-	l.currentServerIP = serverData.IP
-	l.currentServerPublicKey = serverData.NordLynxPublicKey
 	return nil
 }
 
 // connect to the VPN server
 func (l *Libtelio) connect(serverIP netip.Addr, serverPublicKey string) error {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
-	defer cancel()
+	l.cancelConnectionMonitor = cancel
```
mariusSincovici, line 267: shouldn't this be moved after the if from 272 `if err := toError(l.lib.ConnectToExitNode(`? otherwise in case of an error it might generate a `disconnected` at `cancel()`

bartoszWojciechO: I think disconnect could be valid in some cases(for example when connecting event has already been sent). I have added state tracking in the connection monitor function, so now disconnect will be sent only when previous published state was connected or connecting.

mariusSincovici (approved):

daemon/vpn/nordlynx/kernel_space.go
```diff
@@ -10,35 +10,55 @@ import (
...
 type KernelSpace struct {
-	state  vpn.State
-	active bool
-	fwmark uint32
-	tun    *tunnel.Tunnel
+	state           vpn.State
+	active          bool
+	fwmark          uint32
+	tun             *tunnel.Tunnel
+	eventsPublisher *vpn.Events
```
Savolro, line 23: Any reason to use *vpn.Events instead of `vpn.InternalVPNPublisher`?

bartoszWojciechO: `vpn.Events` is the publishing part, vpn implementations use it to publish events, `vpn.InternalVPNPublisher` is the interface that subscribers(`state.StatePublisher`) implement to subscribe to those events.

Savolro (approved):

PR #356, opened 2024-04-10
Add go race detector config

keliramu:


ci/compile.sh
```diff
@@ -7,10 +7,8 @@ source "${WORKDIR}"/ci/archs.sh
 # Since race detector has huge performance price and it works only on amd64 and does not
 # work with pie executables, its enabled only for development builds.
 # shellcheck disable=SC2153
-[ "${ENVIRONMENT}" = "dev" ] && [ "${ARCH}" = "amd64" ] && BUILDMODE="-race" || BUILDMODE="-buildmode=pie"
-
-#TODO/FIXME: just to verify, to be removed
-BUILDMODE="-buildmode=pie"
+[ "${ENVIRONMENT}" = "dev" ] && [ "${ARCH}" = "amd64" ] && [ "${RACE_DETECTOR_ENABLED:-""}" != "" ] \
+		&& BUILDMODE="-race" || BUILDMODE="-buildmode=pie"
```
Savolro, line 11: Position independent executable will make debuggers not work by default, which is not something we want

ci/compile.sh
```diff
@@ -7,10 +7,8 @@ source "${WORKDIR}"/ci/archs.sh
 # Since race detector has huge performance price and it works only on amd64 and does not
 # work with pie executables, its enabled only for development builds.
 # shellcheck disable=SC2153
-[ "${ENVIRONMENT}" = "dev" ] && [ "${ARCH}" = "amd64" ] && BUILDMODE="-race" || BUILDMODE="-buildmode=pie"
-
-#TODO/FIXME: just to verify, to be removed
-BUILDMODE="-buildmode=pie"
+[ "${ENVIRONMENT}" = "dev" ] && [ "${ARCH}" = "amd64" ] && [ "${RACE_DETECTOR_ENABLED:-""}" != "" ] \
```
Savolro, line 10: Position independent executable will make debuggers not work by default, which is not something we want for dev builds.

ci/compile.sh
Savolro, line 11: Position independent executable will make debuggers not work by default, which is not something we want for dev builds.

Savolro (changes requested):
`

ci/compile.sh
keliramu: For local build developer should add new variable to `.env` file. Need sample?

keliramu: Fixed

bartoszWojciechO (approved):

PR #360, opened 2024-04-12
Fix nordtray loggedin notify loop in case of nonet

keliramu:


tray/monitor.go
```diff
@@ -177,9 +177,9 @@ func (ti *Instance) updateSettings() bool {
 
 func (ti *Instance) updateAccountInfo() bool {
 	changed := false
-	loggedIn := false
-	vpnActive := false
-	accountName := ""
+	loggedIn := ti.state.loggedIn
+	vpnActive := ti.state.vpnActive
+	accountName := ti.state.accountName
```
keliramu, line 182: Keep previous values - so, that subsequent request failure would not affect state

bartoszWojciechO (approved):

tray/monitor.go
```diff
@@ -177,9 +177,9 @@ func (ti *Instance) updateSettings() bool {
...
 
 	payload, err := ti.client.AccountInfo(context.Background(), &pb.Empty{})
 	if err != nil {
```
mariusSincovici, line 185: when there is an error, shouldn't it return immediately instead? Then there are no more `lock`s and unneeded comparisons.

mariusSincovici (approved):

tray/monitor.go
keliramu: Yes, right, fixed.

PR #363, opened 2024-04-15
Retry tray startup

bartoszWojciechO:


keliramu (approved):

cmd/norduser/main.go
```diff
@@ -61,12 +61,23 @@ func addAutostart() (string, error) {
...
+	try := 0
+	// Retry checking systray availability, as it might not be availalble on startup.
+	for {
+		if systray.IsAvailable() {
+			break
+		}
+
+		if try == 5 {
+			log.Println("Session tray not available, exiting")
+			break
```
mariusSincovici, line 73: shouldn't be here a `return` instead of `break`?

mariusSincovici (approved):

PR #365, opened 2024-04-17
Remove norduser systemd implementation

bartoszWojciechO:


Savolro (approved):

norduser/service/fork.go
```diff
@@ -3,57 +3,135 @@ package service
...
+		return []int{}, fmt.Errorf("listing norduser pids: %w", err)
+	}
+
+	return parseNorduserPIDs(string(output)), nil
+}
+
+func findPIDOfUID(uids string, uid uint32) int {
+	desiredUID := fmt.Sprint(uid)
+	for _, pidUid := range strings.Split(uids, "\n") {
+		pidUidSplit := strings.Split(strings.TrimSpace(pidUid), " ")
```
mariusSincovici, line 74: [minor] a possible alternative would be to use `Sscanf` to read the 2 ints from the string? and then you wouldn't need to worry about spaces.

norduser/service/fork.go
```diff
@@ -3,57 +3,135 @@ package service
...
+			return true
 		}
 	}
 
-	return false, nil
+	return false
+}
+
+func isRunning(uid uint32) (bool, error) {
+	// list all norduserd processes, restrict output to uid of the owner
```
mariusSincovici, line 120: why not call `getPIDForNorduserUID` in this case and return true if `PID != -1`? 
might be less code to maintain :)

norduser/service/fork.go
```diff
@@ -3,57 +3,135 @@ package service
...
+		}
+
+		return pidInt
 	}
 
+	return -1
+}
+
+func getPIDForNorduserUID(uid uint32) (int, error) {
+	// list all norduserd processes, restrict output to uid of the owner
```
mariusSincovici, line 98: comment not quire correct, the UID and the PID is fetched

norduser/service/fork.go
```diff
@@ -3,63 +3,109 @@ package service
...
+// case.
+func handlePsError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	var exiterr *exec.ExitError
+	if errors.As(err, &exiterr) {
+		// ps returns 1 when no processes are shown
+		if exiterr.ExitCode() == 1 {
```
mariusSincovici, line 38: from what I've saw `ps` is always returning `1` for error, maybe this needs to be combined with the output if it is empty, otherwise all the errors might be ignored?

mariusSincovici (approved):

PR #367, opened 2024-04-17
Change config dir for snap

bartoszWojciechO:


internal/filesystem.go
```diff
@@ -217,16 +217,38 @@ func cleanPidFile(pidFile string) {
 	}
 }
 
+// EnsureDir creates all directories along the path excluding the last element.
 func EnsureDir(path string) error {
 	dir, err := filepath.Abs(filepath.Dir(path))
```
mariusSincovici, line 222: Wouldn't be easier to just call `EnsureDirFull` and then `ensureDir` can be combined with it, e.g.
```
func EnsureDir(path string) error {
   return EnsureDirFull(filepath.Dir(path))
}
```

internal/filesystem.go
```diff
@@ -217,16 +217,38 @@ func cleanPidFile(pidFile string) {
...
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
-		err = os.MkdirAll(dir, PermUserRWX)
+	if err := ensureDir(dir); err != nil {
+		return fmt.Errorf("ensuring dir: %w", err)
+	}
+	return nil
+}
+
+func ensureDir(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
mariusSincovici, line 245: do you know why is `os.Stat` needed?
`os.MkdirAll` does internally a `Stat` and if the path exists and it is a folder will return no error.
But in our case if at the path something exists, including a file, then the folders are not created anymore.

internal/filesystem.go
Savolro: Is there any case where we would not want to use EnsureDirFull?

bartoszWojciechO: There are a few cases where a path with a file at the end is given.

Savolro (approved):

keliramu (approved):

mariusSincovici (approved):

PR #381, opened 2024-04-24
Display server name in notifications instead of hostname

Savolro:


tray/monitor.go
```diff
@@ -112,10 +117,10 @@ func (ti *Instance) updateVpnStatus() bool {
 	if ti.state.vpnStatus != vpnStatus {
 		if vpnStatus == "Connected" {
 			systray.SetIconName(ti.iconConnected)
-			defer ti.notify("Connected to VPN server: %s", vpnHostname)
+			defer ti.notify("Connected to %s", vpnName)
 		} else {
 			systray.SetIconName(ti.iconDisconnected)
-			defer ti.notify("Disconnected from VPN server")
+			defer ti.notify(fmt.Sprintf("Disconnected from %s", ti.state.vpnName))
```
mariusSincovici, line 123: if there is an error when the status if fetched then `vpnStatus==""` and it will display `Disconnected from` because `ti.state.vpnName`.

mariusSincovici (approved):

bartoszWojciechO (approved):

PR #386, opened 2024-04-25
QA tests on Snap

keliramu:


magefiles/mage.go
```diff
@@ -539,10 +539,22 @@ func (Test) Hardening(ctx context.Context) error {
...
 }
 
+// Run QA tests (arguments: {testGroup} {testPattern})
+func (Test) QA(ctx context.Context, testGroup, testPattern string) error {
+	mg.Deps(mg.F(buildPackageDocker, "deb", "-cover"))
+	return qa(ctx, testGroup, testPattern)
+}
+
+// Run QA tests (arguments: {testGroup} {testPattern})
+func (Test) QASnap(ctx context.Context, testGroup, testPattern string) error {
```
keliramu, line 549: Added `mage` target to execute QA tests on SNAP package. 
It should be executed under `systemd` environment - Snap package can be installed only in that environment. One choice is to use Ubuntu VM and execute tests there to not interfere work on host.
Did some adjustments in Python code - but not all tests are adopted.

test/qa/lib/daemon.py
```diff
@@ -71,14 +76,18 @@ def uninstall_peer(ssh_client: ssh.Ssh):
...
+        #sh.sudo.snap("start", "nordvpn")
+        os.popen("sudo snap start nordvpn").read()
+    elif is_init_systemd():
+        #sh.sudo.systemctl.start.nordvpnd()
+        os.popen("sudo systemctl start nordvpn").read()
+    else:
+        # call to init.d returns before the daemon is actually started
+        _rewrite_log_path()
+        sh.sudo("/etc/init.d/nordvpn", "start")
+        while not is_running():
```
mariusSincovici, line 89: shouldn't while check for all types that the daemon started?

test/qa/lib/daemon.py
```diff
@@ -91,13 +100,17 @@ def start_peer(ssh_client: ssh.Ssh):
...
+    if is_under_snap():
+        #sh.sudo.snap("stop", "nordvpn")
+        os.popen("sudo snap stop nordvpn").read()
+    elif is_init_systemd():
+        #sh.sudo.systemctl.stop.nordvpnd()
+        os.popen("sudo systemctl stop nordvpn").read()
+    else:
+        # call to init.d returns before the daemon is actually stopped
+        sh.sudo("/etc/init.d/nordvpn", "stop")
+        while is_running():
```
mariusSincovici, line 112: same here

test/qa/lib/daemon.py
keliramu: Before my changes, `waiting` for daemon start was only for `init.d` - maybe there was a reason for that, so, I left the same way.

test/qa/lib/daemon.py
keliramu: same as above

bartoszWojciechO (approved):

test/qa/lib/daemon.py
mariusSincovici: Probably because it never ran on systemd env :smile:

mariusSincovici (approved):

PR #392, opened 2024-04-29
Fix Firebase token

mariusSincovici:
Pass `main.FirebaseToken` to the fetcher and remove the local variable defined into libtelio module.

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -64,14 +64,15 @@ import (
 
 // Values set when building the application
 var (
-	Salt        = ""
-	Version     = "0.0.0"
-	Environment = ""
-	PackageType = ""
-	Arch        = ""
-	Port        = 6960
-	ConnType    = "unix"
-	ConnURL     = internal.DaemonSocket
+	Salt          = ""
+	Version       = "0.0.0"
+	Environment   = ""
+	PackageType   = ""
+	Arch          = ""
+	Port          = 6960
+	ConnType      = "unix"
+	ConnURL       = internal.DaemonSocket
+	FirebaseToken = "" // If this is moved to another the scripts need to be updated
 )
 
 // Environment constants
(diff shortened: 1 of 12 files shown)


keliramu (approved):

daemon/vpn/libconfig.go
```diff
@@ -2,5 +2,5 @@ package vpn
 
 // LibConfigGetter is interface to acquire config for vpn implementation library
 type LibConfigGetter interface {
-	GetConfig(version string) (string, error)
+	GetConfig(firebaseToken string, version string) (string, error)
```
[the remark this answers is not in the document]
Savolro: I think https://github.com/NordSecurity/nordvpn-linux/blob/main/daemon/vpn/nordlynx/libtelio/config.go#L28 should be reworked instead so it accepts the list of config fetchers and these are configured in https://github.com/NordSecurity/nordvpn-linux/blob/main/cmd/daemon/vpnconfig_telio.go

Savolro (changes requested):

cmd/daemon/main.go
```diff
@@ -64,14 +64,15 @@ import (
...
-	ConnURL     = internal.DaemonSocket
+	Salt          = ""
+	Version       = "0.0.0"
+	Environment   = ""
+	PackageType   = ""
+	Arch          = ""
+	Port          = 6960
+	ConnType      = "unix"
+	ConnURL       = internal.DaemonSocket
+	FirebaseToken = "" // If this is moved to another package the scripts need to be updated
```
Savolro, line 75: Can this be moved to `cmd/daemon/vpnconfig_telio.go`?

mariusSincovici: yes it can be, was only thinking if it would be easier to keep with the other variables that are set at build time
`// Values set when building the application`

Savolro (approved):

PR #396, opened 2024-05-04
Rework test credentials usage

Savolro:
* All test credentials are moved to `NA_TESTS_CREDENTIALS` variable;
* Introduced possibility to override any of the default username keys (default, qa-peer, invalid, expired)


.env.sample
```diff
@@ -22,3 +11,6 @@ FEATURES=telio drop
 # Used for mage targets, when set to 1, fetching docker images will be skipped if it is already fetched.
 # When set to any other value, images will be always re-fetched.
 IDEMPOTENT_DOCKER=1
+
+# Test credentials to be used in tests
+NA_TESTS_CREDENTIALS={"default": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"}
```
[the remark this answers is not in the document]
Savolro: It is configurable. Users can enter whatever

keliramu (approved):

test/qa/lib/login.py
```diff
@@ -88,51 +88,28 @@ def browser_element_interact(self, xpath: str, write:str=None, return_attribute:
...
-        devs = json.loads(ci_credentials)
-        dev_email = os.environ.get("GITLAB_USER_EMAIL")
-        if dev_email in devs:
-            default_username = devs[dev_email]["username"]
-            default_token = devs[dev_email]["token"]
-    return default_username, default_token
+def get_token(key):
+    """Returns token by a given key."""
+    na_credentials = os.environ.get("NA_TESTS_CREDENTIALS")
+    na_credentials_key = os.environ.get("NA_CREDENTIALS_KEY")
```
mariusSincovici, line 94: when will `NA_CREDENTIALS_KEY` be used?

test/qa/lib/login.py
```diff
@@ -88,51 +88,28 @@ def browser_element_interact(self, xpath: str, write:str=None, return_attribute:
...
+    full_key = key if na_credentials_key is None else "{key}_{na_credentials_key}"
+
+    if na_credentials is None:
+        raise Exception("environment variable 'NA_TESTS_CREDENTIALS' is not set")
+    creds = json.loads(na_credentials)
+
+    key = key if creds.get(full_key) is None else full_key
```
mariusSincovici, line 101: wouldn't be good to log the used `key`, because `key_NA_CREDENTIALS_KEY` will be ignored if the value doesn't exist in `NA_TESTS_CREDENTIALS`?

test/qa/lib/login.py
Savolro: If you want to use different credentials for some reason. e. g. `default_test_some_category`. In our case it would be meshnet tests

mariusSincovici (approved):

mariusSincovici (approved):

test/qa/lib/login.py
Savolro: In case of an error token will be logged making it easy to track which key was actually used

PR #397, opened 2024-05-07
Add a delay before starting mesh peer list tests

bartoszWojciechO:


test/qa/test_meshnet_peer_list.py
```diff
@@ -81,8 +83,10 @@ def test_meshnet_peer_list_permission_filters(allows_incoming_traffic, allows_ro
     base_test_peer_list(filter_list)
 
 
-@pytest.mark.skip("LVPN-4860")
 def test_meshnet_peer_list_peer_connected():
+    # Sleep is needed because there is a delay between when peer comes online and when it affect its status when listing peers.
+    time.sleep(1)
```
mariusSincovici, line 88: wouldn't it be good instead of sleep to have a function that waits for some `param` seconds for a peer to be online instead?
then the function can be used in other tests and might give less failures if the servers update slower the peers status.

mariusSincovici (approved):

PR #399, opened 2024-05-08
Use LAN discovery at the daemon startup

mariusSincovici:
When daemon starts check the LAN discovery setting and when is true add the LAN IPs to the allowed list.
Log LAN discovery at startup.


daemon/events_test.go
```diff
@@ -94,6 +96,7 @@ func (mockDaemonSubscriber) NotifyDefaults(any) error                       { re
 func (mockDaemonSubscriber) NotifyMeshnet(bool) error                       { return nil }
 func (mockDaemonSubscriber) NotifyRate(events.ServerRating) error           { return nil }
 func (mockDaemonSubscriber) NotifyHeartBeat(int) error                      { return nil }
+func (m *mockDaemonSubscriber) NotifyLANDiscovery(bool) error               { return nil }
```
keliramu, line 99: Small cleanup is needed here

keliramu (approved):

bartoszWojciechO (approved):

PR #401, opened 2024-05-09
Update registration link, add callback handling

bartoszWojciechO:


cli/cli_click.go
```diff
@@ -2,14 +2,31 @@ package cli
...
+
+		if url.Scheme == "nordvpn-sl" && url.Host == "claim-online-purchase" {
+			_, err := c.client.ClaimOnlinePurchase(context.Background(), &pb.Empty{})
+			if err != nil {
+				return formatError(err)
+			}
+		}
+	}
+
 	if ctx.NArg() >= 1 {
```
mariusSincovici, line 30: this code will always execute also for `if url.Scheme == "nordvpn-sl" && url.Host == "claim-online-purchase" {` and will return the error `Expected a url with nordvpn scheme.`

daemon/rpc_claim_online_purchase.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func (r *RPC) ClaimOnlinePurchase(ctx context.Context, in *pb.Empty) (*pb.Empty, error) {
+	isExpired, err := r.ac.IsVPNExpired()
+	if err != nil {
+		log.Println(internal.ErrorPrefix+" failed to determine if user is registered: ", err)
+	}
+
+	if isExpired {
+		log.Println(internal.DebugPrefix + " user is expired when caliming online purchase.")
```
mariusSincovici, line 18: [spelling] `caliming`

daemon/rpc_claim_online_purchase.go
```diff
@@ -0,0 +1,26 @@
...
+		return &pb.Empty{}, nil
+	}
+
+	log.Println(internal.DebugPrefix + " send user subscribed notification.")
```
mariusSincovici, line 22: this might not be a very helpful message. could you add more info?

daemon/rpc_claim_online_purchase.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+	"log"
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+func (r *RPC) ClaimOnlinePurchase(ctx context.Context, in *pb.Empty) (*pb.Empty, error) {
```
mariusSincovici, line 11: this will never return an error. Shouldn't there be one when `isExpired == true` or for `err`?

daemon/rpc_claim_online_purchase.go
```diff
@@ -0,0 +1,26 @@
...
+	isExpired, err := r.ac.IsVPNExpired()
+	if err != nil {
+		log.Println(internal.ErrorPrefix+" failed to determine if user is registered: ", err)
```
mariusSincovici, line 14: this should be a return otherwise the `isExpired` will have incorrect value

cli/cli_click.go
```diff
@@ -2,14 +2,31 @@ package cli
...
+	"net/url"
 	"os"
 
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 	"github.com/urfave/cli/v2"
 )
 
 // Click is hidden cmd used to open some info when clicking on desktop icon
 func (c *cmd) Click(ctx *cli.Context) error {
+	if ctx.NArg() == 1 {
```
mariusSincovici, line 16: when there is an error, shouldn't the window remain open and wait for key pressed? 
Otherwise the user would not know why it failed.

cli/cli_click.go
```diff
@@ -2,14 +2,31 @@ package cli
...
+		url, err := url.Parse(ctx.Args().First())
+		if err != nil {
+			return formatError(err)
+		}
+
+		if url.Scheme == "nordvpn-sl" && url.Host == "claim-online-purchase" {
+			_, err := c.client.ClaimOnlinePurchase(context.Background(), &pb.Empty{})
+			if err != nil {
```
mariusSincovici, line 24: On successful login there is a message to let the user know `color.Green(LoginSuccess, ctx.App.Name)`, isn't needed also in this case something similar?

cli/cli_click.go
bartoszWojciechO: Imo it should, but it mirrors the behavior for login, which doesn't display wait for user input. It would be best to just wait for key press in every case(success or failure), but I'm not sure why it wasn't implemented that way from the beginning.

daemon/rpc_claim_online_purchase.go
bartoszWojciechO: What kind of info?

mariusSincovici: what is not clear for me is if this is a message that we need to send a notification into the app(kind of a TODO), or if we log that the user subscription was updated.

if it is the second(we log that this function was executed), wouldn't it be better something like `updated user [subscription] plan`?

cli/cli_click.go
```diff
@@ -2,28 +2,81 @@ package cli
...
+	if !timeout {
+		<-inputChan
+		return
 	}
 
-	if err := cli.ShowAppHelp(ctx); err != nil {
-		return err
+	select {
+	case <-inputChan:
+	case <-time.After(10 * time.Second):
```
keliramu, line 32: This timeout should be discussed: 10 seconds might be not enough for user to take action, and after 10 seconds - terminal window is closed? or at least no user input possibility.

cli/messages.go
```diff
@@ -4,6 +4,12 @@ const (
 	ArgumentCountError   = "The command you entered has incorrect number of arguments. Enter '%s --help' to see the options."
 	ArgumentParsingError = "The command you entered is not valid. Enter '%s --help' to see the options."
 
+	ClaimOnlinePurchaseSuccess = `You’re all set!
+You’ve successfully purchased the NordVPN subscription.
+You can use NordVPN on 10 devices at the same time.`
+	ClaimOnlinePurchaseFailure = `Payment failed
```
mariusSincovici, line 10: shouldn't there be some point or exclamation at the end `Payment failed`?

client/messages.go
```diff
@@ -17,6 +17,6 @@ const (
 	MsgTryAgain            = "We're having trouble reaching our servers. Please try again later. If the issue persists, please contact our customer support."
 	UFWDisabledMessage     = "The active UFW firewall on your system prevents us from setting up our firewall properly. We have disabled UFW for the duration of your VPN connection and enabled our firewall to ensure your online security. Your custom UFW rules are imported to our firewall ruleset."
 
-	SubscriptionURL       = "https://join.nordvpn.com/order/?utm_campaign=%s&utm_medium=app&utm_source=linux"
-	SubscriptionNoPlanURL = "https://join.nordvpn.com/order/?utm_medium=app&utm_source=linux"
+	SubscriptionURL      = "https://my.nordaccount.com/plans/?utm_source=linux&utm_medium=app&utm_campaign=desktop-app&redirect_uri=nordvpn-sln://claim-online-purchase"
```
mariusSincovici, line 20: `nordvpn-sln://claim-online-purchase` is this correct? should't be without `n` in nordvpn-sl`n`?

mariusSincovici (approved):

keliramu (approved):

PR #403, opened 2024-05-10
Revoke NC token on 'logout' and 'set defaults'

piotrjurkiewicz:


nc/nc.go
```diff
@@ -474,3 +476,36 @@ func (c *Client) Stop() error {
 
 	return nil
 }
+
+// Revoke revokes the NC communication token
+func (c *Client) Revoke(purgeSession bool) bool {
+	c.startMu.Lock()
+	defer c.startMu.Unlock()
+
+	if c.started {
```
bartoszWojciechO, line 485: Technically management goroutine could still be running at this point, because `Stop` just calls `cancelConnecting` and doesn't wait for any kind of confirmation. So at this point management loop could be attempting to fetch new creds, connecting to NC, etc. Previously it was not an issue because in any context, there would always be a big delay between Starting/Stopping the NC. I'm not sure if now it's completely safe, since `Revoke` is called right after `Stop`. Might not be a big deal, since it's just the dev

nc/nc.go
```diff
@@ -474,3 +476,36 @@ func (c *Client) Stop() error {
...
+		log.Println(logPrefix, "attempt to revoke token for running client")
+		return false
+	}
+
+	var cfg config.Config
+	if err := c.credsFetcher.cm.Load(&cfg); err != nil {
```
bartoszWojciechO, line 491: This entire block(line 490-510) should be encapsulated in `CredsFetcher`.

daemon/rpc_set_defaults.go
```diff
@@ -26,6 +26,16 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.Empty) (*pb.Payload, error
 		log.Println(internal.WarningPrefix, err)
 	}
 
+	if err := r.ncClient.Stop(); err != nil {
+		log.Println(internal.WarningPrefix, err)
+	}
+
+	if internal.IsDevEnv(string(r.environment)) {
```
mariusSincovici, line 33: same here, isn't needed `!IsProdEnv`?

nc/nc.go
```diff
@@ -474,3 +476,36 @@ func (c *Client) Stop() error {
...
+		log.Println(logPrefix, "error reading cfg:", err)
+		return false
+	}
+	userID := cfg.AutoConnectData.ID
+	tokenData := cfg.TokensData[userID]
+
+	log.Println(logPrefix, "revoking token")
```
mariusSincovici, line 498: could you add also the value for `purgeSession` in the log?

nc/nc.go
```diff
@@ -474,3 +476,36 @@ func (c *Client) Stop() error {
...
+	resp, err := c.credsFetcher.api.NotificationCredentialsRevoke(tokenData.Token, tokenData.NCData.UserID.String(), purgeSession)
+	if err != nil {
+		log.Println(logPrefix, "error revoking token:", err)
+		return false
+	}
+	if strings.ToLower(resp.Status) == "ok" {
+		log.Println(logPrefix, "revoked token")
+		return true
+	} else {
+		log.Println(logPrefix, "not revoked token")
```
mariusSincovici, line 508: for error wouldn't make sense to log the `Status`?

nc/nc.go
piotrjurkiewicz: Done.

core/core.go
```diff
@@ -436,6 +446,46 @@ func (api *DefaultAPI) NotificationCredentials(token, appUserID string) (Notific
...
+	// For example, if user has VPN app for android/iOS/mac, whatever, all of his/her devices will be disconnected.
+	// If you provide additionally app_user_id, then only credential for specific app/device will be revoked.
+	// Connection on other devices will stay unaffected.
+	// The purge_session param make sense only in cases when you definitely know, that app_user_id was generated
+	// just for one time usage and after this usage it is not needed anymore. The good example is exactly tests,
+	// where on each run you generate different app_user_id. In usual scenarios in ideal case, app_user_id on the
+	// same device for the same user and same app should stay constant, even if app were reinstalled. So there is no
+	// need to use purge_session at all, and it even can have a bad consequences.
+
+	data, err := json.Marshal(NotificationCredentialsRevokeRequest{
```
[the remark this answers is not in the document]
piotrjurkiewicz: Good idea. Request will not execute when appUserID is empty.

nc/nc.go
piotrjurkiewicz: Done.

nc/nc.go
piotrjurkiewicz: Done.

daemon/rpc_set_defaults.go
piotrjurkiewicz: `IsDevEnv` is literaly defined as `!IsProdEnv`:
```go
func IsDevEnv(env string) bool {
	return !IsProdEnv(env)
}
```

daemon/rpc_logout.go
```diff
@@ -35,12 +35,20 @@ func (r *RPC) Logout(ctx context.Context, in *pb.LogoutRequest) (*pb.Payload, er
...
+		log.Println(internal.WarningPrefix, err)
+	}
+
 	tokenData, ok := cfg.TokensData[cfg.AutoConnectData.ID]
 	if !ok {
 		return &pb.Payload{Type: internal.CodeFailure}, nil
 	}
 
 	if !in.GetPersistToken() {
+		if !r.ncClient.Revoke(internal.IsDevEnv(string(r.environment))) {
```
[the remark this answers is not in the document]
piotrjurkiewicz: It looks like that not. `QA` constant is not used:
```go
const (
	// Development defines development environment
	Development Environment = "dev"
	// QA defines qa environment
	QA = "qa"
	// Production defines production environment
	Production = "prod"
	// Downloader modifies configs and servers jobs
	Downloader = "downloader"
)
```

core/core.go
```diff
@@ -436,6 +446,46 @@ func (api *DefaultAPI) NotificationCredentials(token, appUserID string) (Notific
...
+// NotificationCredentialsRevoke revokes the credentials for notification center appUserID
+func (api *DefaultAPI) NotificationCredentialsRevoke(token, appUserID string, purgeSession bool) (NotificationCredentialsRevokeResponse, error) {
+	// Calling tokens/revoke endpoint with just bearer token will revoke user credentials for every user device.
+	// For example, if user has VPN app for android/iOS/mac, whatever, all of his/her devices will be disconnected.
+	// If you provide additionally app_user_id, then only credential for specific app/device will be revoked.
+	// Connection on other devices will stay unaffected.
+	// The purge_session param make sense only in cases when you definitely know, that app_user_id was generated
+	// just for one time usage and after this usage it is not needed anymore. The good example is exactly tests,
+	// where on each run you generate different app_user_id. In usual scenarios in ideal case, app_user_id on the
+	// same device for the same user and same app should stay constant, even if app were reinstalled. So there is no
```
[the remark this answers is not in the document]
piotrjurkiewicz: I don't know. That's the description an implementation suggestion I received from the API maintainer.

daemon/rpc_set_defaults.go
mariusSincovici: yes, ignore :smile:

mariusSincovici (approved):

nc/nc.go
piotrjurkiewicz: I checked what will happen when credentials will be revoked without stopping the loop. The manager will just try to fetch new credentials. So it does not look like can break anything. In our code stop is esured to be called before revoke. And even if the management loop will not be able to finish after stop call before calling revoke, it will finish when trying to get new creds.

bartoszWojciechO (approved):

PR #410, opened 2024-05-13
Fix norduser pid detection

bartoszWojciechO:


norduser/service/fork.go
```diff
@@ -71,15 +72,18 @@ func getRunningNorduserPIDs() ([]int, error) {
...
+	re := regexp.MustCompile(`\s*(\d+)\s+(\d+)\s*`)
+
+	for _, uidPid := range strings.Split(uids, "\n") {
+		match := re.FindStringSubmatch(uidPid)
+		if len(match) != 3 {
+			log.Println(internal.ErrorPrefix+" invalid input line: ", uidPid)
 			continue
 		}
 
+		uid, _ := strconv.Atoi(match[1])
```
mariusSincovici, line 84: it might be good to check the errors, otherwise `uid` and `pid` might contain wrong values.

mariusSincovici (approved):

norduser/service/fork.go
```diff
@@ -71,14 +71,19 @@ func getRunningNorduserPIDs() ([]int, error) {
 }
 
 func findPIDOfUID(uids string, desiredUID uint32) int {
-	for _, pidUid := range strings.Split(uids, "\n") {
-		var pid, uid int
-		_, err := fmt.Sscanf(pidUid, " %d %d", &uid, &pid)
+	for _, uidPid := range strings.Split(uids, "\n") {
+		var pid int
+		var uid int
+		n, err := fmt.Sscanf(uidPid, "%d%d", &uid, &pid)
```
piotrjurkiewicz, line 77: Is the format string `"%d%d"` correct?

bartoszWojciechO: Yes, the spaces are discarded, so it should work with combinations of variable number of spaces and pids/uids. Confirmed with uts.

piotrjurkiewicz (approved):

PR #412, opened 2024-05-15
Improve killswitch 

keliramu:


contrib/systemd/system/nordvpnd.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=NordVPN Daemon
 Requires=nordvpnd.socket
-After=network-online.target
+Before=network.target
```
keliramu, line 4: Here `nordvpnd` service is started before network - at this time physical network interfaces are already created but not yet activated.

PR #418, opened 2024-05-21
Validate dedicated IP when user tries to connect

bartoszWojciechO:


cli/cli_connect.go
```diff
@@ -102,7 +102,11 @@ func (c *cmd) Connect(ctx *cli.Context) error {
 		case internal.CodeTokenRenewError:
 			rpcErr = errors.New(client.AccountTokenRenewError)
 		case internal.CodeAccountExpired:
-			c.browseToSubscriptionPage()
+			c.browseToSubscriptionPage(client.SubscriptionURL, client.SubscriptionDedicatedIPURLLogin)
```
mariusSincovici, line 105: shouldn't instead of `SubscriptionDedicatedIPURLLogin` be `SubscriptionURLLogin`?

cli/cli_connect.go
```diff
@@ -33,14 +33,14 @@ Provide a <group> argument to connect to a specific servers group. For example:
...
+func (c *cmd) browseToSubscriptionPage(url string, loginURL string) {
 	resp, err := c.client.TokenInfo(context.Background(), &pb.Empty{})
 	if err != nil {
-		browse(client.SubscriptionURL)
+		browse(url, url)
 		return
 	}
 
-	browse(fmt.Sprintf(client.SubscriptionURLLogin, resp.Token, resp.Id))
+	browse(fmt.Sprintf(loginURL, resp.Token, resp.Id), url)
```
mariusSincovici, line 43: This is not something new, but is it ok if the login token is passed in plain text part of the URL? Aren't there any security issues?

bartoszWojciechO: Yeah, I misunderstood the spec a bit, actually token used in the URL is supposed to be a Trusted Pass token, which is supposed to be safe.

piotrjurkiewicz (approved):

tray/actions.go
```diff
@@ -75,6 +77,17 @@ func (ti *Instance) logout(persistToken bool) bool {
...
 
+func (ti *Instance) notifyServiceExpired(url string, trustedPassURL string) {
+	resp, err := ti.client.TokenInfo(context.Background(), &pb.Empty{})
+	isTokenDataValid := resp.TrustedPassToken != "" && resp.TrustedPassOwnerId != ""
+
+	link := url
+	if err == nil && isTokenDataValid {
+		link = fmt.Sprintf(trustedPassURL, resp.TrustedPassToken, resp.TrustedPassOwnerId)
+	}
+	ti.notify(fmt.Sprintf(cli.ExpiredAccountMessage, link))
```
piotrjurkiewicz, line 88: Do we want to display the same message "Your account has expired. Renew your subscription now to continue enjoying the ultimate privacy and security with NordVPN." for both account expiration and dedicated IP expiration?

piotrjurkiewicz: I think that `NoDedicatedIPMessage = "You don’t have a dedicated IP subscription. To get a personal IP address that belongs only to you, continue in the browser: %s"` should be displayed in the case of dedicated IP expiration.

bartoszWojciechO: Fixed

Savolro (approved):

cli/cli_token.go
```diff
@@ -36,6 +36,8 @@ func (c *cmd) TokenInfo(ctx *cli.Context) error {
 	fmt.Println("Token Information:")
 	fmt.Printf("Token: %s\n", payload.Token)
 	fmt.Println("Expires at:", payload.ExpiresAt)
+	fmt.Printf("Trusted Pass Token: %s\n", payload.TrustedPassToken)
```
mariusSincovici, line 39: is it safe to show the login token here?

tray/actions.go
```diff
@@ -75,6 +77,17 @@ func (ti *Instance) logout(persistToken bool) bool {
 	}
 }
 
+func (ti *Instance) notifyServiceExpired(url string, trustedPassURL string, message string) {
+	resp, err := ti.client.TokenInfo(context.Background(), &pb.Empty{})
+	isTokenDataValid := resp.TrustedPassToken != "" && resp.TrustedPassOwnerId != ""
```
mariusSincovici, line 82: if there is an error, it might be good not to use `resp` because it might be `nil`

tray/actions.go
```diff
@@ -75,6 +77,17 @@ func (ti *Instance) logout(persistToken bool) bool {
...
+
+	link := url
+	if err == nil && isTokenDataValid {
+		link = fmt.Sprintf(trustedPassURL, resp.TrustedPassToken, resp.TrustedPassOwnerId)
+	}
+	ti.notify(message, link)
```
mariusSincovici, line 88: wouldn't make sense in this case to force show notification, in case the user has them disabled?

PR #419, opened 2024-05-21
Improve killswitch

keliramu:


internal/systemd.go
```diff
@@ -0,0 +1,20 @@
+package internal
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// IsShutdown tries to determine if systemd shutdown or reboot is being executed
+func IsShutdown() bool {
```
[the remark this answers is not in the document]
keliramu: Yes, agree

internal/systemd.go
```diff
@@ -0,0 +1,20 @@
+package internal
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// IsSystemShutdown tries to determine if systemd shutdown or reboot is being executed
+func IsSystemShutdown() bool {
+	return FileExists("/run/nologin") || FileExists("/var/run/nologin")
```
mariusSincovici, line 10: it seams there is also a `/etc/nologin` to block the logins. Should it be also that checked?
It was introduced https://github.com/linux-pam/linux-pam/commit/e9e593f6ddeaf975b7fe8446d184e6bc387d450b. 

https://man7.org/linux/man-pages/man8/pam_nologin.8.html

mariusSincovici (approved):

internal/systemd.go
keliramu: Changed `shutdown` detection method to `dbus` monitor for a certain system event

mariusSincovici (approved):

internal/systemd.go
mariusSincovici: yes, but I saw that the files are still checked for non-dbus.

internal/systemd.go
```diff
@@ -0,0 +1,30 @@
...
+
+// IsSystemShutdown detect if system is being shutdown
+func IsSystemShutdown() bool {
+	return FileExists("/run/nologin") || FileExists("/var/run/nologin")
+}
+
+// IsSystemd detect if system is running systemd
+func IsSystemd() bool {
+	// check name of PID1 process: "ps -p 1 -o comm="
+	out, err := exec.Command("ps", "-p", "1", "-o", "comm=").Output()
```
piotrjurkiewicz, line 25: According to systemd documentation. the proper way to check whether the system was booted up using the systemd is to check the presence of ` /run/systemd/system` directory: https://www.freedesktop.org/software/systemd/man/latest/sd_booted.html

internal/systemd.go
```diff
@@ -0,0 +1,30 @@
+package internal
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// IsServiceActive check if given service is active
+func IsServiceActive(service string) bool {
+	out, err := exec.Command(SystemctlExec, "is-active", service).Output()
```
piotrjurkiewicz, line 10: If we enable access to `org.freedesktop.systemd1.Manager` anyway, maybe the better approach is to check the service status directly with dbus message, instead executing the command line binary?

daemon/jobs.go
```diff
@@ -81,13 +83,76 @@ func (r *RPC) StopKillSwitch() error {
 	}
 
 	if cfg.KillSwitch {
-		if err := r.netw.UnsetKillSwitch(); err != nil {
-			return fmt.Errorf("unsetting killswitch: %w", err)
+		// do not unset killswitch rules if system is in shutdown or reboot
+		systemd := internal.IsSystemd()
+		shutdownIsActive := (systemd && r.systemShutdown.Load()) ||
+			(!systemd && internal.IsSystemShutdown())
```
piotrjurkiewicz, line 89: Maybe `internal.IsSystemShutdown()` should be also checked o on systemd, just in the case the dbus monitor not catching the shutdown message?

daemon/jobs.go
```diff
@@ -81,13 +83,77 @@ func (r *RPC) StopKillSwitch() error {
...
 	if cfg.KillSwitch {
-		if err := r.netw.UnsetKillSwitch(); err != nil {
-			return fmt.Errorf("unsetting killswitch: %w", err)
+		// do not unset killswitch rules if system is in shutdown or reboot
+		shutdownIsActive := (r.dbusAvailable.Load() && r.systemShutdown.Load()) ||
+			(!r.dbusAvailable.Load() && internal.IsSystemShutdown())
+		if !shutdownIsActive {
+			if err := r.netw.UnsetKillSwitch(); err != nil {
+				return fmt.Errorf("unsetting killswitch: %w", err)
+			}
```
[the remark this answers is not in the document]
keliramu: I'd say no, because this is during system reboot - after reboot these old KS rules are gone, but app is started and new rules are applied.

daemon/jobs.go
```diff
@@ -81,13 +83,77 @@ func (r *RPC) StopKillSwitch() error {
...
 	}
 	return nil
 }
 
+// StartSystemShutdownMonitor to be run on separate goroutine
+func (r *RPC) StartSystemShutdownMonitor() {
+	// get connection to system dbus
+	conn, err := dbus.SystemBus()
+	if err != nil {
+		fmt.Println(internal.ErrorPrefix, "getting system dbus:", err)
```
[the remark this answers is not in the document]
keliramu: Fixed

snap/snapcraft.yaml
```diff
@@ -85,6 +91,7 @@ apps:
       - network-control
       - network-observe
       - firewall-control
+      - sys-dbus
```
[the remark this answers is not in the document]
keliramu: No, it is not needed.

internal/systemd.go
keliramu: Thx, good to know, indeed, this way is simpler.

internal/systemd.go
keliramu: This is legacy code, just moved it from other file into new file. Will check how to do that via dbus.

daemon/jobs.go
keliramu: Ok, this is maybe a good idea.

daemon/jobs.go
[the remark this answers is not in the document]
mariusSincovici: true, but this code is also executed at normal demon shutdown. And what if on some environment we have an issue in determining `shutdownIsActive` correctly? :smile:

internal/systemd.go
keliramu: Final logic is to check only existence of `/run/nologin` file according to this source: https://www.freedesktop.org/software/systemd/man/latest/shutdown.html
Other files `/etc/nologin` and `/var/run/nologin` looks like have administrative purpose for sys admins to disable login to the system (as described in `pam` subsystem in your reference).

snap/snapcraft.yaml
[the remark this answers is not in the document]
mariusSincovici: then does it need to be also into the template file `snapcraft.yaml.template`?

internal/systemd.go
keliramu: Implemented direct dbus calls - not sure if it is better, but seems more complex in terms of code size.

internal/systemd.go
keliramu: Done

piotrjurkiewicz (approved):

PR #420, opened 2024-05-21
Close HTTP3 round tripper when not needed

mariusSincovici:

When `RoundTripper` is not needed, `Close` function must be called to stop the goroutines.


Savolro (approved):
Typo in commit message. Function is called `Close`

bartoszWojciechO (approved):

PR #434, opened 2024-05-30
Improve user message when starting using app after install

keliramu:


cli/cli.go
```diff
@@ -991,6 +991,10 @@ func (s loaderStream) RecvMsg(m interface{}) error {
 	return s.ClientStream.RecvMsg(m)
 }
 
+var missingUserPermissionsMessage = `Permission denied. Please grant necessary permissions before using the application by executing the following commands:
```
mariusSincovici, line 994: Reboot message was removed, but it needs to be added back. Or at least to specify logout is needed.

keliramu: Fixed

mariusSincovici (approved):

piotrjurkiewicz (approved):

PR #443, opened 2024-06-05
Only start norduser when user has active sessions

bartoszWojciechO:


norduser/group_monitor.go
```diff
@@ -106,82 +123,139 @@ func getUID(username string) (userIDs, error) {
...
+	activeUsers := []string{}
+	for _, match := range matches {
+		activeUsers = append(activeUsers, match[1])
+	}
+
+	return activeUsers
+}
+
+func getSessions() []string {
+	cmd := exec.Command("w", "--no-header")
```
piotrjurkiewicz, line 142: I would prefer another solution than executing an external binary. We already do too many things by calling external command binaries and parsing their output. This is both resource intensive and error prone, because output formats may differ between systems or versions (for example recent bug with parsing uids in `ps` output which resulted in tray icons spam).

norduser/group_monitor.go
```diff
@@ -211,19 +286,22 @@ func (n *NordvpnGroupMonitor) Start() error {
...
+					log.Println(internal.ErrorPrefix, "Failed to get new group members:", err)
 				}
 			}
 		case err, ok := <-watcher.Errors:
 			if !ok {
 				return fmt.Errorf("groupfile monitor error channel closed")
 			}
 			log.Println(internal.ErrorPrefix, "group monitor error:", err)
+		case <-time.After(10 * time.Second):
+			newGroupMembers := updateGroupMembersState(currentGrupMembers, getSessions())
```
piotrjurkiewicz, line 303: Executing an external process every 10 seconds all the time when `nordvpnd` is running is something we certainly should not do.

piotrjurkiewicz: Maybe using fsnotify to monitor the `/var/run/utmp` would be some kind of solution to trigger on session change events without constantly polling in a loop?

keliramu: Here is a `inotify` sample: https://github.com/keliramu/watch-socket/blob/main/server/main.go#L82
without additional dependencies.
Maybe would help.

norduser/group_monitor_test.go
```diff
@@ -95,3 +121,151 @@ func Test_getGroupMembers(t *testing.T) {
 		})
 	}
 }
+
+func Test_parseSessions(t *testing.T) {
```
piotrjurkiewicz, line 125: Is this needed now as we parse utmp binary file directly?

norduser/group_monitor.go
```diff
@@ -106,82 +122,142 @@ func getUID(username string) (userIDs, error) {
 	}, nil
 }
 
-func (n *NordvpnGroupMonitor) handleGroupUpdate(currentGroupMembers userSet, newGroupMembers userSet) {
-	for member := range currentGroupMembers {
-		// member not modified, do nothing
-		if _, ok := newGroupMembers[member]; ok {
-			newGroupMembers[member] = true
+func parseSessions(sessions string) []string {
```
mariusSincovici, line 125: is this used?

norduser/utmp.go
```diff
@@ -0,0 +1,94 @@
+package norduser
+
+/*
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <utmp.h>
```
piotrjurkiewicz, line 7: At least according to the documentation functions in this header are obsolete and new applications should use `<utmpx.h>`, `getutxent()` and `endutxent()`: https://man7.org/linux/man-pages/man3/getutid.3.html

norduser/utmp.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+int get_utmp_user_processes(char*** users) {
+  int index = 0;
+  int size = 0;
+  setutent();
+
+  *users = NULL;
+
+  struct utmp* u;
```
piotrjurkiewicz, line 31: In Python it is possible to define C structs using Python's ctypes library and then make calls to C functions populating those struct directly from Python. The benefit is that you don't have to write and compile C modules and you can use automatic memory management provided by Python, instead manually calling `malloc` etc. Isn't something similar possible in Go?

norduser/utmp.go
```diff
@@ -0,0 +1,94 @@
...
+  while ((u = getutent()) != NULL) {
+    if (u->ut_type != USER_PROCESS) {
+      continue;
+    }
+
+    if (index == size) {
+      char** tmp;
+      tmp = realloc(*users, (size + 1) * sizeof(char*));
```
piotrjurkiewicz, line 39: User name strings can be stored directly in the table:

    tmp = realloc(*users, (size + 1) * UT_NAMESIZE);

That way we would be able to get rid of per-user mallocs, only reallocs on the user table will remain.

norduser/group_monitor.go
```diff
@@ -106,82 +122,142 @@ func getUID(username string) (userIDs, error) {
...
+		norduserd:    service,
+		isSnap:       snapconf.IsUnderSnap(),
+		userIDGetter: osGetter{},
+	}
+}
+
+func (n *NordVPNGroupMonitor) handleGroupUpdate(currentGroupMembers userSet, newGroupMembers userSet) userSet {
+	for member, state := range currentGroupMembers {
+		// member is still in group and session is active, do nothing
+		if newState, ok := newGroupMembers[member]; ok && newState == userActive {
```
mariusSincovici, line 169: there might be some issues here:
* connect with SSH to the computer, before logging in from the graphical interface
* while connected login to the desktop
In this case the tray doesn't appear, because the process is already started for ssh session. Also disconnecting from the ssh would still not display the tray.

norduser/utmp.go
```diff
@@ -0,0 +1,94 @@
...
+      size++;
+    }
+
+    (*users)[index] = malloc(UT_NAMESIZE);
+    if ((*users)[index] == NULL) {
+      free_users_table(users, size);
+      return ERROR_MALLOC_USERNAME;
+    }
+
+    strcpy((*users)[index], u->ut_user);
```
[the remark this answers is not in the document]
piotrjurkiewicz: In `struct utmp` user string name is stored as string in `char ut_user[UT_NAMESIZE]`. So actual user name can be at most `UT_NAMESIZE - 1` because the last position needs to remain for null byte. So yeah, we can both use `strncpy(..., ..., UT_NAMESIZE)` and to be sure also write a null byte to `(*users)[index][UT_NAMESIZE-1]='\0';` There is no need to allocate `UT_NAMESIZE + 1`.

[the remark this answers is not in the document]
mariusSincovici: but the documentations doesn't say it always has a null byte, just only if the name is smaller then the field size, `UT_NAMESIZE=32`. So if the user name has 32 chars, then `[UT_NAMESIZE-1]='\0';` would remove a char.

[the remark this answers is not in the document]
piotrjurkiewicz: > String fields are terminated by a null byte ('\0') if they are shorter than the size of the field.

You are right, I overlooked that fragment.

norduser/utmp.go
```diff
@@ -0,0 +1,94 @@
...
+
+int get_utmp_user_processes(char*** users) {
+  int index = 0;
+  int size = 0;
+  setutent();
+
+  *users = NULL;
+
+  struct utmp* u;
+  while ((u = getutent()) != NULL) {
```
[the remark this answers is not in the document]
bartoszWojciechO: In `utmpx.h` there is no alias for those functions. But we are not calling them in mutlithreaded context, so we should be fine for now(until we want to use these functions in some other place).

norduser/utmp.go
bartoszWojciechO: I couldn't find anything like that. You can call C functions directly in go but you still have to manage the memory.

norduser/utmp.go
bartoszWojciechO: Yeah, but then we would just create a long string. An array of strings is more standard/idiomatic and easier to use on go side.

norduser/utmp.go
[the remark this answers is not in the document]
mariusSincovici: they are defined in `/usr/include/utmp.h`, without `x`.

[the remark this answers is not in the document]
bartoszWojciechO: Also, with normal getutent we don't have to worry about allocation and deallocation of utmp structures, and usage of regular getutent is simpler, making our caller code more streamlined. I'd stick with it for now, if ever have another use case for utmp library we can switch to `getutent_r` or create a more generalized thread safe wrapper.

norduser/utmp.go
```diff
@@ -0,0 +1,99 @@
...
+        endutxent();
+        return ERROR_REALLOC;
+      }
+      *users = tmp;
+
+      size++;
+    }
+
+    (*users)[index] = malloc(__UT_NAMESIZE + 1);
+    (*users)[index][__UT_NAMESIZE]='\0';
```
mariusSincovici, line 53: better move this after the if, otherwise might crash if `(*users)[index] == NULL)`

mariusSincovici (approved):

devzbysiu (approved):

PR #445, opened 2024-06-05
Remove calls to column command

mariusSincovici:
* replace `column` command because it is not always available in the system and for snap it cannot be accessed. 
* try to first get the CLI windows size from stdout fd and if it fails try stty

### DIFF
--- a/cli/terminal.go
+++ b/cli/terminal.go
@@ -1,15 +1,19 @@
 package cli
 
 import (
+	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"log"
 	"os"
+	"os/exec"
 	"strconv"
+	"strings"
 
 	"github.com/NordSecurity/nordvpn-linux/internal"
-
 	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 )
 
 func checkUsernamePasswordIsEmpty(username, password string) error {
@@ -32,7 +36,7 @@ func ReadCredentialsFromTerminal() (string, string, error) {
 		password string
 	)
 
-	if !terminal.IsTerminal(0) || !terminal.IsTerminal(1) {
+	if !term.IsTerminal(0) || !term.IsTerminal(1) {
 		return username, password, fmt.Errorf("Stdin/Stdout should be terminal")
 	}
 	oldState, err := terminal.MakeRaw(0)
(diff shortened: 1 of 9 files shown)


cli/terminal.go
```diff
@@ -118,3 +123,100 @@ func ReadPlanFromTerminal() (int, error) {
 	}
 	return planID, nil
 }
+
+// Columns formats a list of strings to a tidy column representation
+func Columns(data []string) (string, error) {
```
piotrjurkiewicz, line 128: Is there a need to be public?

cli/terminal.go
```diff
@@ -118,3 +123,100 @@ func ReadPlanFromTerminal() (int, error) {
...
+	}
+	if len(row) > 0 {
+		t.AppendRow(row)
+	}
+
+	return t.Render(), nil
+}
+
+// Gets the size of CLI window
+func CliDimensions() (int, int, error) {
```
piotrjurkiewicz, line 199: Is there a need to be public?

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/hako/durafmt v0.0.0-20210608085754-5c1018a4e16b
 	github.com/jbowtie/gokogiri v0.0.0-20190301021639-37f655d3078f
 	github.com/jbowtie/ratago v0.0.0-20200401224626-3140c0a9b186
+	github.com/jedib0t/go-pretty/v6 v6.5.9
```
piotrjurkiewicz, line 27: I think that it should be assessed whether we want to add a new external dependency. How much it increases binary sizes? Do we really need it? And if so, whether we can make use of it in other places.

cli/terminal.go
mariusSincovici: copied like this from the internal package. Changed because it seams to work.

go.mod
mariusSincovici: removed the dependency and used Println instead

bartoszWojciechO (approved):

piotrjurkiewicz (approved):

cli/terminal.go
```diff
@@ -118,3 +122,90 @@ func ReadPlanFromTerminal() (int, error) {
 	}
 	return planID, nil
 }
+
+// formats a list of strings to a tidy column representation
+func columns(data []string) (string, error) {
+	width, _, err := cliDimensions()
+	if err != nil {
+		// workaround for tests because terminal size cannot be calculated
```
devzbysiu, line 130: Hmm :thinking: , wouldn't be better to use strategy pattern and inject different strategy based on if we are in the test or not? So for tests we would inject strategy doing `return strings.Join(data, " "), err` and for regular use we would inject `return formatTable(data, width)`. 
Then the strategy could be also tested separately. What do you think?

devzbysiu (approved):

PR #453, opened 2024-06-12
Improve DNS handling

keliramu:
Improved DNS handling (canceled the idea to use NetworkManager)

### DIFF
--- a/daemon/dns/dns.go
+++ b/daemon/dns/dns.go
@@ -6,9 +6,11 @@ package dns
 import (
 	"errors"
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
 // Setter is responsible for configuring DNS.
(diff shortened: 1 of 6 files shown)


daemon/dns/dns.go
```diff
@@ -73,7 +81,8 @@ func (d *DefaultSetter) Set(iface string, nameservers []string) error {
 		if method.IsAvailable() {
 			d.publisher.Publish("set dns for interface [" + iface + "] using: " + method.Name())
 			if err := method.Set(iface, nameservers); err != nil {
-				return fmt.Errorf("setting dns with %s: %w", method.Name(), err)
+				log.Println(internal.ErrorPrefix, fmt.Errorf("setting dns with %s: %w", method.Name(), err))
```
mariusSincovici, line 84: Shouldn't the used option be marked and use the same one for disconnect? And maybe if that fails loop thru the list with all the commands?
Otherwise wouldn't there be issues if the user changes the system while connected to VPN and at disconnect something else is used?

daemon/dns/dns.go
```diff
@@ -91,7 +100,8 @@ func (d *DefaultSetter) Unset(iface string) error {
 		if method.IsAvailable() {
 			d.publisher.Publish("unset dns for interface [" + iface + "] using: " + method.Name())
 			if err := method.Unset(iface); err != nil {
-				return fmt.Errorf("unsetting dns with %s: %w", method.Name(), err)
+				log.Println(internal.ErrorPrefix, fmt.Errorf("unsetting dns with %s: %w", method.Name(), err))
+				continue
```
mariusSincovici, line 104: Cannot comment at the final return, but if no command is able to unset the DNS shouldn't the function return an error, like `Set`?

mariusSincovici (commented):
it seams that for me if LAN discovery is enabled all the DNS requests are send around the tunnel.

daemon/dns/dns.go
keliramu: Errors is written to the log, if any. If we return with error then user will not be able to disconnect from VPN.

daemon/dns/dns.go
keliramu: I don't think it is necessary, because DNS settings are per-interface in most modern linux distros, and we do created/destroy our virtual interface, or set/reset DNS settings for interface every time when it is needed. Also, it is no problem to set DNS with one method, and unset with another. Except direct file resolv.conf modify which is the only method in docker container.

mariusSincovici: yeah, I was thinking on that with the conf file. because the others the OS would take care of it when the interface is gone.

devzbysiu (approved):

mariusSincovici (approved):

PR #454, opened 2024-06-13
Add CLI command for virtual servers

mariusSincovici:

* Add nordvpn set virtuallocation enable|disable command.
* Display the status of virtual servers in nordvpn settings command


devzbysiu (approved):

cli/cli_settings.go
```diff
@@ -57,6 +57,7 @@ func (c *cmd) Settings(ctx *cli.Context) error {
 		fmt.Printf("DNS: %+v\n", strings.Join(settings.Dns, ", "))
 	}
 	fmt.Printf("LAN Discovery: %+v\n", nstrings.GetBoolLabel(settings.LanDiscovery))
+	fmt.Printf("Virtual location: %+v\n", nstrings.GetBoolLabel(settings.VirtualLocation))
```
piotrjurkiewicz, line 60: We use capitalized words in all other settings names ("Virtual Location")

config/config.go
```diff
@@ -59,6 +59,8 @@ type Config struct {
 	LanDiscovery    bool                `json:"lan_discovery"`
 	RemoteConfig    string              `json:"remote_config,omitempty"`
 	RCLastUpdate    time.Time           `json:"rc_last_update,omitempty"`
+	// Indicates whether the virtual servers are used. True by default
+	VirtualLocation TrueField `json:"virtual_location,omitempty"`
```
piotrjurkiewicz, line 63: Why not `bool` as in other settings?

cmd/daemon/events_linux.go
```diff
@@ -34,6 +34,7 @@ func (*dummyAnalytics) NotifyRequestAPI(events.DataRequestAPI) error   { return
 func (*dummyAnalytics) NotifyRate(events.ServerRating) error           { return nil }
 func (*dummyAnalytics) NotifyHeartBeat(int) error                      { return nil }
 func (*dummyAnalytics) NotifyLANDiscovery(bool) error                  { return nil }
+func (*dummyAnalytics) NotifyVirtualLocation(bool) error               { return nil }
```
piotrjurkiewicz, line 37: Is that also implemented on the libmoose side?

mariusSincovici: Need to check if this setting is supported by moose and if it is will be implemented into a separate PR.

piotrjurkiewicz (approved):

config/config.go
mariusSincovici: The indentation is made automatically and I cannot control it, and it breaks because of the line with the comment.
I'll try to see how it looks if I mode the comment at the end of the line.

config/manager.go
```diff
@@ -174,16 +175,10 @@ func (f *FilesystemConfigManager) load(c *Config) error {
...
 	var data []byte
 
-	if f.fsHandle.FileExists(f.location) {
-		// #nosec G304 -- no input comes from the user
-		data, err = f.fsHandle.ReadFile(f.location)
-		if err != nil {
-			return err
-		}
-	} else {
-		if err := f.fsHandle.CreateFile(f.location, internal.PermUserRW); err != nil {
```
keliramu, line 184: Why this case is removed?

config/manager.go
```diff
@@ -196,24 +191,6 @@ func (f *FilesystemConfigManager) load(c *Config) error {
 		return err
 	}
 
-	if c.FirewallMark == 0 {
-		// without this users need to reset their configs
-		c.FirewallMark = defaultFWMarkValue
```
keliramu, line 201: Is this being handled in other place?

config/manager.go
mariusSincovici: It is deleted because:
* there is a check on line 164 if the file exist, and when it doesn't it returns immediately so this code will never execute
* if the code does executes, then there are other problems because this creates an empty file and decryption will always fail for empty files

config/manager.go
mariusSincovici: The logic changed a bit. Also all this existed also into `newConfig()`, so it was just duplicated code.
Now the `c` is always created with the initial values `*c = *newConfig(f.machineIDGetter)` and then they are just replaced by `json.Unmarshal(decrypted, c)`. In this way when new settings are added can be automatically set to the default value from `newConfig`.

keliramu (approved):

config/config.go
mariusSincovici: I'll let it like this, because adding the comment on the same line will make the line too long.

cmd/daemon/events_linux.go
mariusSincovici: I've created a ticket for this, because at the moment there is no event in moose.

PR #462, opened 2024-06-18
 Display virtual servers

mariusSincovici:

* Delete the servers split from AppData and compute them on the fly
* Virtual servers have different colors for countries, cities and groups commands
* Fix autocomplete for connect and autoconnect to return the suggestions base on user settings
* Make the column function to accept generics
* Fix unit-tests


internal/string.go
```diff
@@ -14,15 +20,10 @@ func StringsToInterfaces(strings []string) []interface{} {
...
-	splits := strings.Split(name, " ")
-	titled := ""
-	for _, v := range splits {
-		if len(v) == 0 {
-			continue
-		}
-		titled += strings.Title(v) + "_"
-	}
-	return strings.TrimRight(titled, "_")
+	name = alphanumeric.ReplaceAllString(name, "")
```
mariusSincovici, line 23: this will remove chars that don't match the regex `alphanumeric`, because this function was used for `groups` but not for countries and cities. And there was added a country that contains `'` in its name.

cli/cli_cities_test.go
```diff
@@ -18,7 +19,7 @@ func TestCitiesList(t *testing.T) {
 
 	tests := []struct {
 		name          string
-		cities        []string
+		cities        []*pb.ServerGroup
```
mariusSincovici, line 22: the tests will be reviewed and more added in the next release. In this PR they were just fixed to compile.

internal/string.go
```diff
@@ -1,10 +1,16 @@
...
 import (
+	"regexp"
 	"strconv"
 	"strings"
+
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
+var alphanumeric = regexp.MustCompile(`[^0-9a-zA-Z \-_]+`)
```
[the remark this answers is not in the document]
mariusSincovici: right, copied from somewhere else :smile: 
I'll rename it.

internal/string.go
mariusSincovici: yes, there is one country that has `'` in its name and with this change will be removed. 
Would it make sense in command line to have `'`? because it just needs the escape char and might look not very nice, plus doesn't work copy/paste.
I'll try to clarify this

cli/terminal.go
```diff
@@ -11,11 +10,32 @@ import (
...
 	"strings"
 
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 	"github.com/NordSecurity/nordvpn-linux/internal"
-	"golang.org/x/crypto/ssh/terminal"
+	"github.com/fatih/color"
 	"golang.org/x/term"
 )
 
+func lengthServerGroup(server *pb.ServerGroup) int {
```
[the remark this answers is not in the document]
mariusSincovici: yeah, the names are quite bad :laughing:
`serverNameLen` sounds nice. and I'll change also for the others to remove `serverGroup` part.

internal/string.go
devzbysiu: Aah, yes. That makes perfect sense now :+1:

cli/loader.go
```diff
@@ -45,7 +43,7 @@ type LoaderConfig struct {
 }
 
 func NewLoader() Loader {
-	if isatty.IsTerminal(os.Stdout.Fd()) {
+	if isStdoutATerminal() {
```
mariusSincovici, line 46: @keliramu I've changed the use the same function everywhere if stdout is a terminal

cli/terminal.go
```diff
@@ -11,11 +10,32 @@ import (
...
+	return len(server.Name)
+}
+
+func displayServerGroup(server *pb.ServerGroup) string {
```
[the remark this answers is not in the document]
mariusSincovici: changed to change to `formatServerName` :smile:

devzbysiu (approved):

piotrjurkiewicz (approved):

PR #463, opened 2024-06-18
Improve connect tests

ggediminass:
De-duplicate tests, re-use code..

### DIFF
--- a/test/qa/lib/network.py
+++ b/test/qa/lib/network.py
@@ -1,4 +1,3 @@
-import os
 import socket
 import time
 from itertools import cycle
@@ -53,10 +52,10 @@ def _capture_packets(connection_settings: (str, str, str)) -> int:
         traffic_filter = TSHARK_FILTER_TCP_OBFUSCATED % server_ip
 
     # If enough packets are captured, do not wait the duration time, exit early, show compact output
-    #tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
-    tshark_result: str = os.popen(f"sudo tshark -i any -T fields -e ip.src -e ip.dst -a duration:3 -a packets:1 -f {traffic_filter}").read()
+    tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
+    #tshark_result: str = os.popen(f"sudo tshark -i any -T fields -e ip.src -e ip.dst -a duration:3 -a packets:1 -f {traffic_filter}").read()
 
-    packets = tshark_result.strip().split("\n")
+    packets = tshark_result.strip().splitlines()
 
     return len(packets)
 
--- a/test/qa/test_combinations.py
+++ b/test/qa/test_combinations.py
@@ -31,32 +31,11 @@ def teardown_function(function):  # noqa: ARG001
     logging.log()
 
 
-@pytest.mark.parametrize(("target_tech", "target_proto", "target_obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.parametrize(("target_tech", "target_proto", "target_obfuscated"), lib.TECHNOLOGIES)
 @pytest.mark.parametrize(("source_tech", "source_proto", "source_obfuscated"), lib.TECHNOLOGIES)
 @pytest.mark.flaky(reruns=2, reruns_delay=90)
 @timeout_decorator.timeout(40)
-def test_reconnect_matrix_standard(
-        source_tech,
-        target_tech,
-        source_proto,
-        target_proto,
-        source_obfuscated,
-        target_obfuscated,
-):
-    lib.set_technology_and_protocol(source_tech, source_proto, source_obfuscated)
-    connect_base_test(ipv6 = False)
-
-    lib.set_technology_and_protocol(target_tech, target_proto, target_obfuscated)
-    connect_base_test(ipv6 = False)
-
-    disconnect_base_test()
-
-
-@pytest.mark.parametrize(("target_tech", "target_proto", "target_obfuscated"), lib.OBFUSCATED_TECHNOLOGIES)
-@pytest.mark.parametrize(("source_tech", "source_proto", "source_obfuscated"), lib.TECHNOLOGIES)
-@pytest.mark.flaky(reruns=2, reruns_delay=90)
-@timeout_decorator.timeout(40)
-def test_reconnect_matrix_obfuscated(
+def test_reconnect_matrix(
         source_tech,
         target_tech,
         source_proto,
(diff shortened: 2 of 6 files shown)


test/qa/lib/network.py
```diff
@@ -53,10 +52,15 @@ def _capture_packets(connection_settings: (str, str, str)) -> int:
         traffic_filter = TSHARK_FILTER_TCP_OBFUSCATED % server_ip
 
     # If enough packets are captured, do not wait the duration time, exit early, show compact output
-    #tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
-    tshark_result: str = os.popen(f"sudo tshark -i any -T fields -e ip.src -e ip.dst -a duration:3 -a packets:1 -f {traffic_filter}").read()
+    tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
+    #tshark_result: str = os.popen(f"sudo tshark -i any -T fields -e ip.src -e ip.dst -a duration:3 -a packets:1 -f {traffic_filter}").read()
```
mariusSincovici, line 56: any reason why switch to the other command?

test/qa/lib/network.py
```diff
@@ -53,10 +52,15 @@ def _capture_packets(connection_settings: (str, str, str)) -> int:
...
 
-    packets = tshark_result.strip().split("\n")
+    if tshark_result.strip() != "":
```
mariusSincovici, line 58: what is the difference between from previously, because if string is empty `len` would return 0?

test/qa/lib/network.py
```diff
@@ -53,10 +52,15 @@ def _capture_packets(connection_settings: (str, str, str)) -> int:
         traffic_filter = TSHARK_FILTER_TCP_OBFUSCATED % server_ip
 
     # If enough packets are captured, do not wait the duration time, exit early, show compact output
-    #tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
-    tshark_result: str = os.popen(f"sudo tshark -i any -T fields -e ip.src -e ip.dst -a duration:3 -a packets:1 -f {traffic_filter}").read()
+    tshark_result: str = sh.tshark("-i", "any", "-T", "fields", "-e", "ip.src", "-e", "ip.dst", "-a", "duration:3", "-a", "packets:1", "-f", traffic_filter)
```
mariusSincovici, line 55: [minor] maybe add here the `strip()`

test/qa/test_combinations.py
```diff
@@ -8,102 +8,31 @@
...
-    output = sh.nordvpn.connect(_tty_out=False)
-    print(output)
-    assert lib.is_connect_successful(output)
-    assert network.is_connected()
-
-    output = sh.nordvpn.disconnect()
-    print(output)
-    assert lib.is_disconnect_successful(output)
-    assert network.is_disconnected()
+    sh.nordvpn.logout("--persist-token")
```
mariusSincovici, line 26: shouldn't this be wrapped in some try/catch in case one of the commands fails, so the test is not marked as failed?

test/qa/test_combinations.py
```diff
@@ -112,50 +41,26 @@ def test_reconnect_matrix_obfuscated(
...
-    assert lib.is_connect_successful(output)
-    assert network.is_connected()
+def test_connect_country_and_city(tech, proto, obfuscated, country, city):
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
 
-    output = sh.nordvpn.connect(country, city, _tty_out=False)
-    print(output)
-    assert lib.is_connect_successful(output)
-    assert network.is_connected()
+    connect_base_test(country, ipv6 = False)
```
mariusSincovici, line 59: shouldn't there be a `disconnect` or is tested server switching?

test/qa/test_connect6.py
```diff
@@ -12,99 +12,81 @@
...
+    logging.log()
+    print(sh.nordvpn.set.ipv6.on())
+
 
+def teardown_function(function):  # noqa: ARG001
+    logging.log(data=info.collect())
+    logging.log()
 
-def teardown_module(module):  # noqa: ARG001
+    print(sh.nordvpn.set.ipv6.off())
```
mariusSincovici, line 29: same here with try/catch?

test/qa/test_connect6.py
```diff
@@ -12,99 +12,81 @@
...
-    logging.log()
+    output = sh.nordvpn.connect(group, _tty_out=False)
+    print(output)
 
+    assert lib.is_connect_successful(output, name, hostname)
+    assert network.is_connected()
 
-@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES_WITH_IPV6[:-1])
+    if ipv6:
+        assert network.is_ipv6_connected(20)
```
mariusSincovici, line 53: why so many retries `20`? If something goes wrong it would take at least 20 seconds to fail.

mariusSincovici (approved):

test/qa/lib/network.py
ggediminass: With the other one, packets were not being captured at all, therefore switched to original command line used before

test/qa/lib/network.py
ggediminass: It was not returning 0, it was actually returning 1, after we split string by newlines, try this:
```
a = ""
b = a.split("\n")
len(b)
```

I changed this logic in here, because if no packets were captured, everybody expects, that `len(packets)` will be 0, but it was not :smile:

test/qa/test_combinations.py
ggediminass: Reconnect combinations are being tested, so `disconnect` is not needed in this case

test/qa/test_connect6.py
ggediminass: I would like to disagree. If these fail, it could indicate a big issue with our app, and we'd miss it, just because we have try/catch block.

Furthermore, even if exception would be caught here, within next test we'd get an exception @ line 22, where we enable IPv6. So in that case, we'd also need to add try/catch block in `setup_function`, which means, that issues for example with IPv6 feature may be unnoticed for longer period of time :smile:

test/qa/test_combinations.py
ggediminass: Answered here:
https://github.com/NordSecurity/nordvpn-linux/pull/463#discussion_r1647108671

test/qa/test_connect6.py
ggediminass: Good point, I will check this out

test/qa/lib/network.py
mariusSincovici: right, has different output if is using custom separator or default space :melting_face: .
you could use `splitlines()` :laughing:

ggediminass: `splitlines()` seems like a better solution, will do that instead 🫡

devzbysiu (approved):

mariusSincovici (approved):

PR #464, opened 2024-06-19
Improve bash scripts options

keliramu:


devzbysiu (approved):

ci/shellcheck.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -euo
+set -euxo
```
mariusSincovici, line 2: isn't `pipefail` needed here too, so it fails if shell is not installed?

mariusSincovici (approved):

ci/shellcheck.sh
keliramu: `sh` does not support this option

PR #465, opened 2024-06-19
Update libmoose

devzbysiu:
Libmoose update

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,8 +1,6 @@
 [submodule "third-party/moose-events"]
 	path = third-party/moose-events
 	url = ../../../../moose/moose-events.git
-	branch = nordVpnApp
 [submodule "third-party/moose-worker"]
 	path = third-party/moose-worker
 	url = ../../../../moose/moose-events.git
-	branch = worker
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,8 +1,6 @@
 [submodule "third-party/moose-events"]
 	path = third-party/moose-events
 	url = ../../../../moose/moose-events.git
-	branch = nordVpnApp
 [submodule "third-party/moose-worker"]
```
mariusSincovici, line 4: aren't the branches needed anymore?

devzbysiu: The branches point to the older version and `.gitmodules` does not support tags, so instead I just moved the submodules to the correct version and commited that change. It's saved in git and then restoring the sumbodules points to the commited version. You can see that change at the bottom of this PR as:
>  Submodule moose-events updated from 884fbe to 7e9dca

and 

>  Submodule moose-worker updated from 8f2caf to be862b

I think we could potentially move the branch to point to the current version and leave it as is but it may be problematic in case we would have to do rollback, because we would have to move the branch again to the previous version.

On the other hand the branches are more explicit :thinking: 

What do you think about that?

Savolro (approved):

mariusSincovici (approved):

PR #466, opened 2024-06-19
Try to restore resolv.conf

keliramu:
Multiple times issue with `resolv.conf` not being restored was reported. To address this issue - on daemon start, check if `/etc/resolv.conf` contains Nordvpn changes and try to restore - if backup exists then from backup, if no backup - generate simple dns settings file.
Note: backup logic was always to check if backup is already exists then do not make new one.


daemon/dns/dns_resolvconf_file.go
```diff
@@ -59,7 +63,7 @@ func resetDNSinResolvconfFile(addresses []string) error {
 	// set DNS
```
mariusSincovici, line 63: I cannot add comment on line 41 `if !internal.FileWritable(resolvconfFilePath) {`, but if the file is marked as immutable then the app still overrides it.

daemon/dns/dns_resolvconf_file.go
```diff
@@ -59,7 +63,7 @@ func resetDNSinResolvconfFile(addresses []string) error {
...
 	_ = internal.FileUnlock(resolvconfFilePath)
 	defer internal.FileLock(resolvconfFilePath)
-	content := "# Generated by NordVPN\n" + strings.Join(addrs, "\n")
+	content := resolvconfFileMark + "\n" + strings.Join(addrs, "\n")
```
mariusSincovici, line 66: [small] might be nice to add a `\n` after the servers, so `cat /ect/resolv.conf` moves the prompt on new line? :smile:

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +98,46 @@ func restoreDNS() error {
...
+			log.Println(internal.ErrorPrefix, fmt.Errorf("restore from backup resolv.conf: %w", err))
+			return
+		}
+		// remove backup
+		internal.FileDelete(resolvconfBackupPath)
+		return
+	}
+	log.Println(internal.WarningPrefix, "/etc/resolv.conf contains our changes - restore with simple dns settings")
+	// there is no backup, but we need to fix dns settings
+	if err := internal.FileWrite(resolvconfFilePath, []byte(resolvconfFileContent), internal.PermUserRWGroupROthersR); err != nil {
```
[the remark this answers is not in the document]
keliramu: This option is not always working - depends if NetworkManager is available and if it is set to handle DNS.

daemon/dns/dns_resolvconf_file.go
```diff
@@ -13,6 +13,10 @@ import (
 const (
 	// resolvconfFilePath defines path to resolv.conf file for DNS
 	resolvconfFilePath = "/etc/resolv.conf"
+	// resolvconfFileMark mark resolv.conf file content to recorgnize
+	resolvconfFileMark = "# Generated by NordVPN"
+	// resolvconfFileContent is simple dns settings to restore as a last option
+	resolvconfFileContent = "#restored\nnameserver 1.1.1.1"
```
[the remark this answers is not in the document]
keliramu: Sure, let's discuss this

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +98,46 @@ func restoreDNS() error {
...
+	}
+	log.Println(internal.WarningPrefix, "/etc/resolv.conf contains our changes - need to fix this")
+	// try to unlock, if file was locked
+	_ = internal.FileUnlock(resolvconfFilePath)
+	// restore from backup if backup file exists
+	if internal.FileExists(resolvconfBackupPath) {
+		log.Println(internal.WarningPrefix, "resolv.conf backup found - restore from it")
+		backup, err := internal.FileRead(resolvconfBackupPath)
+		if err != nil {
+			log.Println(internal.ErrorPrefix, fmt.Errorf("reading backup resolv.conf: %w", err))
```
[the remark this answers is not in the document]
keliramu: Good point

daemon/dns/dns_resolvconf_file.go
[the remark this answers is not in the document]
piotrjurkiewicz: We certainly should not set the DNS to external service without user consent. I think that the best solution in such case would be to set it to `nameserver 127.0.0.53`, which is local resolver address (for example from systemd). And, to handle the case when it might not be present/non working, resolve our API requests internally with our own DNS server. This will allow users to connect to NordVPN even when their system DNS resolution is not working.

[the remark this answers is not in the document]
keliramu: Now 3 options are tried:
1. check if 127.0.0.53:53 is active 
2. find default gateway
3. use 1.1.1.1 (should not happen)

daemon/dns/dns_resolvconf_file.go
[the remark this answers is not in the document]
keliramu: If this case happens, proceed with restoring with simple dns setup

daemon/dns/dns_resolvconf_file.go
keliramu: Assume file was locked by us

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,85 @@ func restoreDNS() error {
...
+		} else {
+			if err := internal.FileWrite(resolvconfFilePath, backup, internal.PermUserRWGroupROthersR); err != nil {
```
mariusSincovici, line 133: shouldn't also check that the backup file doesn't have NordVPN marker? And if it does don't restore from it?

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,85 @@ func restoreDNS() error {
...
+	if !strings.Contains(string(out), resolvconfFileMark) {
+		return
+	}
+
+	log.Println(internal.WarningPrefix, "/etc/resolv.conf contains our changes - need to fix this")
+
+	// try to unlock, if file was locked
+	_ = internal.FileUnlock(resolvconfFilePath)
+	// restore from backup if backup file exists
+	if internal.FileExists(resolvconfBackupPath) {
```
mariusSincovici, line 127: Shouldn't there be some checks for the backup file, at least if is not symlink?
Because, at the moment if backup file exists then the app doesn't do anything, but at restore will just replace `/etc/resolv.conf` with its content.
Or change the logic from setDns and always overwrite the backup file, maybe even mark it as immutable?

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,85 @@ func restoreDNS() error {
...
+		if err != nil {
+			log.Println(internal.ErrorPrefix, fmt.Errorf("reading backup resolv.conf: %w", err))
+		} else {
+			if err := internal.FileWrite(resolvconfFilePath, backup, internal.PermUserRWGroupROthersR); err != nil {
+				log.Println(internal.ErrorPrefix, fmt.Errorf("restore from backup resolv.conf: %w", err))
+			}
+			return
+		}
+		// remove backup
+		internal.FileDelete(resolvconfBackupPath)
```
mariusSincovici, line 139: If the backup file exists maybe should always delete it?

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,85 @@ func restoreDNS() error {
...
+	if err := internal.FileWrite(resolvconfFilePath, []byte(content), internal.PermUserRWGroupROthersR); err != nil {
+		log.Println(internal.ErrorPrefix, fmt.Errorf("writing simple resolv.conf: %w", err))
+	}
+}
+
+func discoverNameserverIp() (netip.Addr, error) {
+	// 1st try: 127.0.0.53:53
+	// then: GW
+	nsIp := netip.MustParseAddr("127.0.0.53")
+	_, err := net.ListenPacket("udp", nsIp.String()+":53")
```
mariusSincovici, line 164: doesn't it work to connect only instead of starting the listerner, `conn, err := net.DialTimeout("udp", address, timeout)` ?
Or optionally use https://pkg.go.dev/net#Resolver to make a real DNS query? But then it might get too complicated for what we need :smile:

mariusSincovici (approved):

daemon/dns/dns_resolvconf_file.go
keliramu: ok, added this extra check, might help but not clear in what situation

daemon/dns/dns_resolvconf_file.go
keliramu: yes, right, this was my mistake/leftover.

daemon/dns/dns_resolvconf_file.go
keliramu: `net.Dial...` does not work on UDP, current approach is simplest and working in most cases.

daemon/dns/dns_resolvconf_file.go
keliramu: Backup file is root protected - no need for extra protection, now on restore I have added check for backup file not to have our mark, backup will be removed. IMO, that is enough.

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,86 @@ func restoreDNS() error {
...
+	_ = internal.FileUnlock(resolvconfFilePath)
+	// restore from backup if backup file exists
+	if internal.FileExists(resolvconfBackupPath) {
+		log.Println(internal.WarningPrefix, "resolv.conf backup found - restore from it")
+		backup, err := internal.FileRead(resolvconfBackupPath)
+		if err != nil {
+			log.Println(internal.ErrorPrefix, fmt.Errorf("reading backup resolv.conf: %w", err))
+		} else {
+			if !strings.Contains(string(out), resolvconfFileMark) {
+				if err := internal.FileWrite(resolvconfFilePath, backup, internal.PermUserRWGroupROthersR); err != nil {
```
mariusSincovici, line 134: I think there might be a missing `return` on success write, otherwise it always goes to the backup logic?
To have the early exit, you could try to move the file deletion after reading it above?

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +101,86 @@ func restoreDNS() error {
...
+	if err := internal.FileWrite(resolvconfFilePath, []byte(content), internal.PermUserRWGroupROthersR); err != nil {
+		log.Println(internal.ErrorPrefix, fmt.Errorf("writing simple resolv.conf: %w", err))
+	}
+}
+
+func discoverNameserverIp() (netip.Addr, error) {
+	// 1st try: 127.0.0.53:53
+	// then: GW
+	nsIp := netip.MustParseAddr("127.0.0.53")
+	_, err := net.ListenPacket("udp", nsIp.String()+":53")
```
piotrjurkiewicz, line 165: The connection `_` needs to be closed to stop listening on that port when there is no error returned.

keliramu: No connection is being created - it's UDP

daemon/dns/dns_resolvconf_file.go
keliramu: Thx, missed that

daemon/dns/dns_resolvconf_file.go
keliramu: Added check if file is set to be immutable and do nothing in that case

daemon/dns/dns_resolvconf_file.go
piotrjurkiewicz: But it opens a socket, puts it into a listen state and returns the connection object. We need to call close on that object to stop listening on the socket.

keliramu: No object is created - variable is `nil`. Ok, looking into that more

piotrjurkiewicz: According to the documentation it should return the connection object when there were no errors.

keliramu: Yes, you are right, added closing the connection

daemon/dns/dns_resolvconf_file.go
```diff
@@ -94,3 +113,93 @@ func restoreDNS() error {
 	}
 	return internal.FileWrite(resolvconfFilePath, out, internal.PermUserRWGroupROthersR)
 }
+
+func tryToRestoreDNS() {
+	// if target is symlink, probably it is managed by other software - do nothing
+	if internal.IsSymLink(resolvconfFilePath) {
```
mariusSincovici, line 119: it might be good to have the same login for VPN disconnect :smile:

piotrjurkiewicz (approved):

daemon/dns/dns_resolvconf_file.go
keliramu: yes, implemented recovery options on regular `restoreDNS`

mariusSincovici (approved):

PR #469, opened 2024-06-20
Update unit-tests for virtual servers

mariusSincovici:


cli/cli.go
```diff
@@ -1186,3 +1186,33 @@ func commandFullName(ctx *cli.Context, args []string) string {
 
 	return strings.Join(fullCommand, " ")
 }
+
+func (c *cmd) printServersForAutoComplete(country string) {
+	if country == "" {
+		resp, err := c.client.Groups(context.Background(), &pb.Empty{})
+		if err != nil {
+			return
```
devzbysiu, line 1194: Don't we want to `log` something here (and in the rest `if err != nil` blocks below)? Would be easier to debug in case of an issue

devzbysiu (approved):

piotrjurkiewicz (approved):

PR #473, opened 2024-06-24
Reset allowlist at network changes

mariusSincovici:

Reset the allowlist at network changes because the OS deletes them when the interface is removed from the system.

### DIFF
--- a/networker/networker.go
+++ b/networker/networker.go
@@ -777,8 +777,8 @@ func (netw *Combined) stopAllowedIPv6Traffic() error {
 }
 
 func (netw *Combined) resetAllowlist() error {
-	// this is done in order to maintain the order of the firewall
-	// rules
+	// this is done in order to maintain the order of the firewall rules
+	log.Println(internal.InfoPrefix, "reset allow list")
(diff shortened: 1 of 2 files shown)


piotrjurkiewicz (approved):

devzbysiu (approved):

networker/networker.go
```diff
@@ -777,8 +777,8 @@ func (netw *Combined) stopAllowedIPv6Traffic() error {
 }
 
 func (netw *Combined) resetAllowlist() error {
-	// this is done in order to maintain the order of the firewall
-	// rules
+	// this is done in order to maintain the order of the firewall rules
+	log.Println(internal.DebugPrefix, "reset allow list")
```
keliramu, line 781: [minor] `Debug` is not appropriate for the release, maybe `InfoPrefix` would be better?

keliramu (approved):

networker/networker.go
mariusSincovici: yes, makes sense. I'll change to info.

PR #474, opened 2024-06-25
Migrate libtelio config

keliramu:
Migrate libtelio config up to v4.3.3 (v5.0.0 not yet included)

### DIFF
--- a/daemon/vpn/nordlynx/libtelio/libtelio.go
+++ b/daemon/vpn/nordlynx/libtelio/libtelio.go
@@ -103,12 +103,19 @@ type Libtelio struct {
 var defaultIP = netip.MustParseAddr("10.5.0.2")
 
 type telioFeatures struct {
-	Lana      *lanaConfig      `json:"lana,omitempty"`
-	Nurse     *nurseConfig     `json:"nurse,omitempty"`
-	Direct    *directConfig    `json:"direct,omitempty"`
-	Derp      *derpConfig      `json:"derp,omitempty"`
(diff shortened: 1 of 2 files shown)


piotrjurkiewicz (approved):

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -183,15 +218,8 @@ func New(prod bool, eventPath string, fwmark uint32,
...
-		defaultTelioConfig := &telioFeatures{}
-		defaultTelioConfig.Lana = &lanaConfig{
-			Prod:      prod,
-			EventPath: eventPath,
-		}
-		defaultTelioConfig.Direct = &directConfig{}
-		defaultTelioConfig.Nurse = &nurseConfig{
-			Fingerprint: deviceID,
-		}
+		defaultTelioConfig := newEmptyTelioConfig()
```
mariusSincovici, line 221: It is not the same behavior as before. With this changes the direct connections and events will not work anymore.

mariusSincovici: also shouldn't the new values be added with some default ones?

keliramu: 1st comment - you are right, 2nd - there was discussion with libtelio and agreement with them is that default values are set/handled by libtelio - consumers from outside should specify empty value (at any level) for libtelio to know to go with defaults.

mariusSincovici (approved):

PR #476, opened 2024-06-26
Set connection attempt event duration

piotrjurkiewicz:


events/events.go
```diff
@@ -62,6 +62,7 @@ type DataConnect struct {
 	Protocol                   config.Protocol
 	ResponseServersCount       int
 	ResponseTime               int
+	Duration                   int
```
mariusSincovici, line 65: Wouldn't make sense to have this as `time.Time` or specify in what the duration is, e.g. seconds or milliseconds?
Might be easier to handle in moose the conversion to what the API call expectes, sometimes seconds other milliseconds.

mariusSincovici (approved):

devzbysiu (approved):

PR #477, opened 2024-06-27
Update libdrop

devzbysiu:
Update to libdrop `v7.0.0`


build/foss/Cargo.toml
```diff
@@ -8,5 +8,7 @@ rust-version = "1.72.1"
 crate-type = ["staticlib"]
 
 [dependencies]
-norddrop = { git = "https://github.com/NordSecurity/libdrop", tag = "v5.4.0_moose_backport" }
-telio = { git = "https://github.com/NordSecurity/libtelio", tag= "v4.3.3" }
+# norddrop = { git = "https://github.com/NordSecurity/libdrop", tag = "v7.0.0" }
+# FIXME: Change it
+norddrop = { path = "../../../libdrop/norddrop" }
```
devzbysiu, line 13: FYI: It's just to make the integration easier because libdrop moved entirely to `cdylib`, so I added a `lib` to `crate-type` locally to postpone dynamic linking until I have everything working.

fileshare/transfer.go
```diff
@@ -159,197 +161,338 @@ func GetTransferFileStatus(file *pb.File, in bool) (status string) {
...
-		Id:       in.ID,
-		Path:     in.RelativePath,
-		FullPath: filepath.Join(in.BasePath, in.RelativePath),
-		Size:     in.TotalSize,
-		// This only shows the amount from the last status change.
-		// The only up to date source for data transferred are events.
-		// To show the correct value we track ongoing transfers in event manager,
-		// and event manager overwrites values of transferred fields if needed.
-		Transferred: 0,
+func determineTransferPath(allFiles []*pb.File) string {
```
devzbysiu, line 455: FYI: That's probably too easy implementation, I'll verify that while working on tests and playing with fileshare

fileshare/transfer_test.go
```diff
@@ -1,289 +1,278 @@
 package fileshare
```
devzbysiu, line 1: FYI: Working on the tests

fileshare/transfer.go
```diff
@@ -159,197 +161,338 @@ func GetTransferFileStatus(file *pb.File, in bool) (status string) {
...
+func directionFromTransferInfo(ti *norddrop.TransferInfo) pb.Direction {
+	var direction pb.Direction
+	switch ti.Kind.(type) {
+	case norddrop.TransferKindOutgoing:
+		direction = pb.Direction_OUTGOING
+	case norddrop.TransferKindIncoming:
+		direction = pb.Direction_INCOMING
+	default:
+		log.Printf("%s unknown direction found when parsing libdrop transfers: %s", internal.WarningPrefix,
+			reflect.TypeOf(ti.Kind))
```
mariusSincovici, line 191: same here with `reflect`

fileshare/transfer.go
```diff
@@ -159,197 +161,338 @@ func GetTransferFileStatus(file *pb.File, in bool) (status string) {
...
-type LibdropFileState struct {
-	CreatedAt     uint64 `json:"created_at"`
-	State         string `json:"state"`
-	BytesSent     uint64 `json:"bytes_sent"`
-	BytesReceived uint64 `json:"bytes_received"`
-	BasePath      string `json:"base_dir"`
-	FinalPath     string `json:"final_path"`
-	StatusCode    int    `json:"status_code"`
+func filesFromTransferInfo(ti *norddrop.TransferInfo) []*pb.File {
+	switch ti.Kind.(type) {
```
mariusSincovici, line 198: same here with switch and var?

fileshare/transfer.go
```diff
@@ -159,197 +161,338 @@ func GetTransferFileStatus(file *pb.File, in bool) (status string) {
...
+
+	transferPath := findLongestCommonPrefix(filePaths)
+	if transferPath == "" || transferPath == "/" {
+		transferPath = "multiple files"
+	}
+
+	return transferPath
+}
+
+func findLongestCommonPrefix(filePaths []string) string {
```
mariusSincovici, line 478: Isn't there too many iterations over the list? :smile: 
Also this should prob go in the `string.go` file, maybe we ca reuse it?

didn't tested the bellow impl, but just an idea? :laughing: 
```
func commonPrefix(strs []string) string {
	if len(strs) == 0 {
		return ""
	}

	// Start with the first string as the initial prefix
	prefix := strs[0]

	for _, str := range strs[1:] {
		l := min(len(prefix), len(str))
		for i := 0; i < l; i++ {
			if prefix[i] != str[i] {
				prefix = prefix[:i]
				break
			}
		}
	}

	return prefix
}
```

fileshare/transfer.go
```diff
@@ -159,197 +161,338 @@ func GetTransferFileStatus(file *pb.File, in bool) (status string) {
...
+	return status
+}
 
-	return out
+func getTransferStatusByFiles(files []*pb.File) pb.Status {
+	allCanceled := true
+	allFinished := true
+	hasNoErrors := true
+	hasStarted := false
+	for _, file := range files {
```
mariusSincovici, line 426: [minor] if you want you can simplify the bellow `if`s  `allCanceled = allCanceled && file.Status != pb.Status_CANCELED`

fileshare/event_manager.go
```diff
@@ -125,45 +128,32 @@ func (em *EventManager) DisableNotifications() error {
...
-	switch genericEvent.Type {
-	case requestReceived:
-		em.handleRequestReceivedEvent(genericEvent.Data)
-	case requestQueued:
-	case transferStarted:
-	case transferProgress:
-		em.handleTransferProgressEvent(genericEvent.Data)
-	case transferFinished:
-		em.handleTransferFinishedEvent(genericEvent.Data)
+		log.Printf("DROP EVENT: %s", reflect.TypeOf(event.Kind))
```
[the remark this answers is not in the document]
devzbysiu: Aah, yes! Perfect :bow:

fileshare/event_manager.go
```diff
@@ -125,45 +128,32 @@ func (em *EventManager) DisableNotifications() error {
...
+	}
+
+	switch event.Kind.(type) {
```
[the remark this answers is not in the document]
devzbysiu: Yesss! I *knew* golang has something like that. My brain tried `switch ev := event.Kind.(type); ev {` and that's not the way to go. 
Much better :+1:

fileshare/event_manager.go
```diff
@@ -125,45 +128,32 @@ func (em *EventManager) DisableNotifications() error {
...
+	case norddrop.EventKindRequestReceived:
+		ev := event.Kind.(norddrop.EventKindRequestReceived)
+		em.handleRequestReceivedEvent(ev)
+	case norddrop.EventKindFileStarted:
+		ev := event.Kind.(norddrop.EventKindFileStarted)
+		em.handleFileStartedEvent(ev)
+	case norddrop.EventKindFileProgress:
```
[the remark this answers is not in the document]
devzbysiu: Yes, sorry I forgot to put the comment here. There are much more cases, because currently, we have one big handler for `transferFinished` and inside more events are handled. With new libdrop I have to split it into multiple handlers and I'm adjusting tests first to minimize the regression.

So yes, the transferFinished handler is not used in this PR and that's why we have bunch of "unused xyz" linter errors below

fileshare/event_manager.go
```diff
@@ -125,45 +128,32 @@ func (em *EventManager) DisableNotifications() error {
...
+		ev := event.Kind.(norddrop.EventKindFileProgress)
+		em.handleFileProgressEvent(ev)
 	default:
-		log.Printf("Unknown libdrop event: %s", eventJSON)
+		// TODO: Verify the rest of the events
+		log.Printf("Unsupported libdrop event: %s", reflect.TypeOf(event.Kind))
```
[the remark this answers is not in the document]
devzbysiu: Yep, I'm fixing it :+1:

fileshare/event_manager.go
```diff
@@ -189,45 +179,69 @@ func (em *EventManager) handleRequestReceivedEvent(eventJson json.RawMessage) {
...
-	err := json.Unmarshal(eventJSON, &event)
+	transfer, err := em.getLiveTransfer(event.TransferId)
 	if err != nil {
-		log.Printf("unmarshalling drop event: %s", err)
+		log.Print(err)
 		return
 	}
 
-	transfer, err := em.getLiveTransfer(event.TransferID)
+	file, ok := transfer.Files[event.FileId]
```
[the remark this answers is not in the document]
devzbysiu: Hmm, not sure about the sync :thinking: , I'll verify it :+1:

fileshare/fileshare_startup/startup.go
```diff
@@ -23,9 +23,10 @@ import (
 	meshpb "github.com/NordSecurity/nordvpn-linux/meshnet/pb"
 )
 
+var ErrLAddressAlreadyInUse = errors.New("address already in use")
+
 // Values set when building the application
 var (
-	version   = "0.0.0"
```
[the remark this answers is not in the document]
devzbysiu: It was just a stub for the API and now API does not need the version :smile:

fileshare/libdrop/libdrop.go
```diff
@@ -2,80 +2,107 @@
...
+func (dks DefaultKeyStore) Privkey() []byte {
+	return []byte(dks.privKey)
+}
+
+type DefaultLogger struct {
+	logLevel norddrop.LogLevel
+}
+
+func (dl DefaultLogger) OnLog(level norddrop.LogLevel, msg string) {
+	log.Println(logLevelToPrefix(level), "DROP("+norddrop.Version()+"): "+msg)
```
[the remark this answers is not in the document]
devzbysiu: What do you mean by "too big" :thinking: - like too long in the code?

fileshare/libdrop/libdrop.go
```diff
@@ -179,34 +194,30 @@ func (f *Fileshare) Accept(transferID, dstPath string, fileID string) error {
 func (f *Fileshare) Cancel(transferID string) error {
 	f.mutex.Lock()
 	defer f.mutex.Unlock()
-
-	res := f.norddrop.CancelTransfer(transferID)
-	return toError(res)
+	return f.norddrop.FinalizeTransfer(transferID)
```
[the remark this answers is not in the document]
devzbysiu: Yes, no more cancel. But `FinalizeTransfer` is `CancelTransfer` but renamed :smile:

fileshare/transfer.go
devzbysiu: I actually had to get rid of the idea entirely. Files from the transfer are not enough to determine the path because the transfer which is not yet accepted has no concept of a path :cry:

fileshare/transfer.go
devzbysiu: Done :+1:

fileshare/transfer.go
devzbysiu: Yep, can't determine the path this way because not accepted transfer has no concept of a path

fileshare/event_manager.go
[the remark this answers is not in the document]
bartoszWojciechO: Do you mean some thread safety by sync? `transfers` is a collection local to this function(created from our own storage and libdrop storage). We also lock a mutex at the top function(in this case `EventFunc`), so there shouldn't be any races.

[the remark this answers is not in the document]
mariusSincovici: Ok, missed that. So there should be no changes to `Files` from another thread while we're here. :+1:

fileshare/libdrop/libdrop.go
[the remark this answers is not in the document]
mariusSincovici: to long line into the log files :smile: 
but if it helps debugging it is ok.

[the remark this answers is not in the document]
devzbysiu: I think it won't be too long, from what I can see currently it gives prefix + some hash and it's around 50 chars long

fileshare/fileshare.go
```diff
@@ -5,14 +5,10 @@ import (
 	"net/netip"
 	"time"
 
+	norddrop "github.com/NordSecurity/libdrop-go/v7"
```
Savolro, line 8: This will make it impossible to compile without `libdrop.so` available

fileshare/fileshare_startup/startup.go
```diff
@@ -23,9 +23,10 @@ import (
 	meshpb "github.com/NordSecurity/nordvpn-linux/meshnet/pb"
 )
 
+var ErrLAddressAlreadyInUse = errors.New("address already in use")
```
Savolro, line 26: I see it is being checked but nothing produces it. Is it used?

fileshare/libdrop/symbols/cgo.go
```diff
@@ -0,0 +1,9 @@
+package symbols
```
Savolro, line 1: Any reason to export this to a separate package when it's always included with
``` go
_ "github.com/NordSecurity/nordvpn-linux/fileshare/libdrop/symbols" // required for linking process
```
in the upper package? I could see the benefit of this somewhere in `main` configuration though as it would allow different linking configuration.

fileshare/server.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/netip"
 	"strings"
 
+	"github.com/NordSecurity/nordvpn-linux/fileshare/libdrop"
```
Savolro, line 10: This will cause problems when compiling without `libdrop.so` available

internal/libdrop_event_to_string.go
```diff
@@ -0,0 +1,35 @@
+package internal
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+
+	norddrop "github.com/NordSecurity/libdrop-go/v7"
+	_ "github.com/NordSecurity/nordvpn-linux/fileshare/libdrop/symbols" // this is required to make cgo symbols available during linking
```
Savolro, line 9: Will cause problems when compiling without `libdrop.so` available. As this is specific to libdrop, whole code could be migrated under `libdrop` package

internal/libdrop_event_to_string_test.go
```diff
@@ -0,0 +1,101 @@
...
+					Size: 1000,
+				},
+			},
+		},
+	}
+
+	expected := "norddrop.EventKindRequestQueued"
+
+	assert.Equal(t, expected, eventToString(event, brokenMarshaler{}))
+}
```
Savolro, line 95: This could be a single parametrized table driven test with 2 cases.

fileshare/fileshare.go
```diff
@@ -5,14 +5,10 @@ import (
...
 	"github.com/NordSecurity/nordvpn-linux/fileshare/pb"
 )
 
-const (
-	DirDepthLimit     = 5
-	TransferFileLimit = 1000
```
Savolro, line 13: These constants define the feature (fileshare), not the implementation (libdrop), therefore it is OK for them to stay in the `fileshare` package. Migration caused usage of `libdrop` package in multiple places, therefore a requirement for native dependency availability (`libdrop.so`).

It is OK for specific implementation (libdrop) to know about the feature it implements (fileshare) in this case in order to avoid introducing additional abstraction layers.

Savolro (changes requested):

fileshare/fileshare.go
devzbysiu: Moved it back. I moved it here, because of circular dependency issue:
```
package github.com/NordSecurity/nordvpn-linux/cmd/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare/libdrop
        imports github.com/NordSecurity/nordvpn-linux/fileshare: import cycle not allowed
```
but I moved the `EventToString` helper from libdrop package to fileshare package, so the part 
```
imports github.com/NordSecurity/nordvpn-linux/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare/libdrop
```
is no longer valid.
This helper method is only used in `even_manager.go` so we can have it there I think

internal/libdrop_event_to_string_test.go
devzbysiu: I was thinking about that as well, but the assertion is different, because in one case I'm comparing plain string, but in other I'm comparing JSON strings. I could to some check in the assertion part and when it's not a JSON then compare plain strings, but it's bit ugly and introduces not needed complexity IMO

internal/libdrop_event_to_string.go
devzbysiu: Done :+1: , it's now in `event_manager.go`. I had it under `libdrop` package initially, but this introduces circular dependency and this helper is not used anywhere else than `event_manager.go`

devzbysiu: Done :+1: , this file `libdrop_event_to_string.go` no longer exists. I moved the helper method to `event_manager.go` as it's needed only there and could not put it in `libdrop` package because there was a circular dependency cycle:
```
package github.com/NordSecurity/nordvpn-linux/cmd/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare/libdrop
        imports github.com/NordSecurity/nordvpn-linux/fileshare: import cycle not allowed
```
but I moved the `EventToString` helper from libdrop package to fileshare package, so the part 
```
imports github.com/NordSecurity/nordvpn-linux/fileshare
        imports github.com/NordSecurity/nordvpn-linux/fileshare/libdrop
```
is no longer valid.

fileshare/libdrop/symbols/cgo.go
devzbysiu: I moved it to separate package, because there was a dependency cycle: fileshare needs libdrop to get symbols for linking and libdrop needs fileshare because of `DirDepthLimit`  and `TransferFileLimit` consts

fileshare/fileshare_startup/startup.go
devzbysiu: It was produced by the `toError` helper func used for translating erorr codes coming from libdrop to our error representation. Now, libdrop provides actual errors so I deleted this `toError`. And this `ErrLAddressAlreadyInUse` was defined there ([here](https://github.com/NordSecurity/nordvpn-linux/pull/477/files#diff-72526275c385cbaa1d36587d90c8494c6ace084277d70c5fe235f24c0e3b2a8cL10))
It should be actually removed. I will use libdrop err [here](https://github.com/NordSecurity/nordvpn-linux/pull/477/files#diff-dcde5f61b97deab08ccf85fccbfeb0dd6308d47b30db676202080055d75bfe10L184)

norduser/fileshare.go
```diff
@@ -68,17 +68,18 @@ func startFileshare(fileshareProcessManager *childprocess.GRPCChildProcessManage
 	case childprocess.CodeFailedToCreateUnixScoket:
 		fallthrough
 	case childprocess.CodeAlreadyRunning:
-		log.Println(internal.InfoPrefix+" fileshare started, final result: ", result)
+		log.Println(internal.InfoPrefix, "fileshare started, final result:", result)
 		return true
 	}
 
-	log.Printf(internal.ErrorPrefix+" failed to start fileshare: %d, will retry", result)
+	log.Printf(internal.ErrorPrefix+" failed to start fileshare: %d, will retry\n", result)
```
keliramu, line 75: Would be nice to have code style consistency i.e. see line #71 above

devzbysiu: Done :+1:

fileshare/libdrop/libdrop.go
```diff
@@ -2,114 +2,247 @@
...
+}
+
+func libdropEventToInternalEvent(nev norddrop.Event) fileshare.Event {
+	return fileshare.Event{
+		Kind:      toInternalEventKind(nev.Kind),
+		Timestamp: nev.Timestamp,
+	}
+}
+
+func toInternalEventKind(kind norddrop.EventKind) fileshare.EventKind {
```
mariusSincovici, line 87: maybe now is too late, but lots of the structs from here are almost identical with the ones from `norddrop`.
does it make sense to have this conversion, wouldn't be easier to have just an alias for the `norddrop` types?

fileshare/libdrop/libdrop.go
```diff
@@ -167,46 +293,46 @@ func (f *Fileshare) Accept(transferID, dstPath string, fileID string) error {
...
 }
 
 // CancelFile id in a transfer
 func (f *Fileshare) CancelFile(transferID string, fileID string) error {
 	f.mutex.Lock()
 	defer f.mutex.Unlock()
 
-	res := f.norddrop.RejectFile(transferID, fileID)
-	if err := toError(res); err != nil {
+	if err := f.norddrop.RejectFile(transferID, fileID); err != nil {
```
mariusSincovici, line 315: can just `return f.norddrop.RejectFile(transferID, fileID)`?

fileshare/libdrop/libdrop.go
devzbysiu: The issue is that if I introduce the aliases for those types, I need to have a reference to `norddrop` types in those aliases, like:
```
type EventKindRequestReceived = norddrop.EventKindRequestReceived
```
This will create circular dependency issue, which can be dealt with, but most importantly, this will pollute the fileshare package with implementation details. Right now, I'm using `libdrop` package only in `fileshare/main` and injecting it where needed and fileshare has no idea about libdrop details

fileshare/events.go
```diff
@@ -2,58 +2,189 @@ package fileshare
...
+	StatusCodeConnectionClosedByPeer StatusCode = 22
+	// Peer’s DDoS protection kicked in.
+	// Transfer should be resumed after some cooldown period.
+	StatusCodeTooManyRequests StatusCode = 23
+	// This error code is intercepted from the OS errors. Indicate lack of
+	// privileges to do certain operation.
+	StatusCodePermissionDenied StatusCode = 24
+)
+
+var defaultMarshaler = jsonMarshaler{}
```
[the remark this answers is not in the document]
devzbysiu: Hmm, I don't think so :thinking:, `jsonMarshaller` is stateless, immutable, and `json.Marshal` does not rely on global state or anything like that. Golang race detector does not report anything fishy here

fileshare/libdrop/libdrop.go
devzbysiu: Yup. Fixed (I'll push the fixes for it and the rest of the comments tomorrow)

cmd/fileshare/main.go
```diff
@@ -67,56 +81,159 @@ func main() {
...
+		os.Exit(int(childprocess.CodeFailedToEnable))
+	}
+
+	meshPrivKey, err := base64.StdEncoding.DecodeString(privKeyResponse.GetPrivateKey())
+	if err != nil || len(meshPrivKey) != 32 {
+		log.Println(internal.ErrorPrefix, "failed to decode mesh private key")
+		os.Exit(int(childprocess.CodeFailedToEnable))
+	}
+
+	configDirPath, _ := internal.GetConfigDirPath(homeDir)
```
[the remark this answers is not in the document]
devzbysiu: Hmm, I don't actually know why the error here was ignored (on main, it does assign to `err`, but does nothing with it). Yes, I think it makes sense to exit

cmd/fileshare/main.go
```diff
@@ -67,56 +81,159 @@ func main() {
...
+
+	resp, err := meshClient.IsEnabled(context.Background(), &meshpb.Empty{})
+	if err != nil || !resp.GetStatus().GetValue() {
+		log.Println(internal.ErrorPrefix, "meshnet not enabled:", err)
+		os.Exit(int(childprocess.CodeMeshnetNotEnabled))
+	}
+
+	defaultDownloadDirectory, err := fileshare.GetDefaultDownloadDirectory()
+	if err != nil {
+		log.Println(internal.WarningPrefix, "failed to find default download directory:", err)
```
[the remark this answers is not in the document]
devzbysiu: As far as I can see, `fileshare.GetDefaultDownloadDirectory` returns `""` on error, but I won't hurt to set it explicitly here as well

cmd/fileshare/main.go
```diff
@@ -2,23 +2,44 @@
...
 	"golang.org/x/net/netutil"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 )
 
-// Values set when building the application
-var Environment = ""
+var (
+	ErrMeshNotEnabled          = errors.New("meshnet not enabled")
+	ErrMeshAddressAlreadyInUse = errors.New("meshnet address is already in use, probably another fileshare instance is already running")
```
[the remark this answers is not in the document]
devzbysiu: I removed them. The `ErrMeshNotEnabled` was returned from fileshare startup, but the code was moved here, so the error is not returned, but the error case is handled here.

The `ErrMeshAddressAlreadyInUse` was returned from fileshare startup as well and it was returned when out translation code (from libdrop status codes to our errors) returned the error below - `ErrLAddressAlreadyInUse`.
Now our error is returned [here](https://github.com/NordSecurity/nordvpn-linux/pull/477/files#diff-7c18b6bf7cdc8e29e3b8b617b4c59e730dfd0005843db5ac42c08673e7837e0bR229) and handled below [here](https://github.com/NordSecurity/nordvpn-linux/pull/477/files#diff-b935e7de49c31ee7389e1feec410b4bfee251f4bf2edba0a98ec532d93b8d61aR220)

Savolro (approved):

bartoszWojciechO (approved):

PR #478, opened 2024-06-27
Change string functions used for servers keys

mariusSincovici:

* Each server has a list of strings computed to faster pick the server when connecting, aka keys. Change functions used to convert the string for keys to be in sync with what is used for user input.
* Keep `string.SnakeCase` function in sync with `string.Title` to remove non-alphanumeric chars from the input.

For testing:
`Lao People's Democratic Republic` contains a `'` in its name. It is displayed `Lao_Peoples_Democratic_Republic ` and connect must work with the displayed name, but also with `Lao_People\'s_Democratic_Republic` so it is backward compatible.

### DIFF
--- a/test/qa/lib/settings.py
+++ b/test/qa/lib/settings.py
@@ -20,11 +20,6 @@ class Settings:
         key = key.lower()
         return self.settings.get(key, "")
 
-    # convenient to get a value from settings only once
-    @staticmethod
-    def get_value(key: str) -> str:
-        return Settings().get(key)
-
 
 def get_server_ip() -> str:
     """Returns str with IP Address of the server from `nordvpn status`, that NordVPN client is currently connected to."""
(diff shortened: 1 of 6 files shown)


test/qa/lib/settings.py
```diff
@@ -1,86 +1,104 @@
...
+    def get_value(key: str) -> str:
+        return Settings().get(key)
+
+
 def get_server_ip() -> str:
     """Returns str with IP Address of the server from `nordvpn status`, that NordVPN client is currently connected to."""
     return sh.nordvpn.status().split('\n')[3].replace('IP: ', '')
 
 
 def get_current_connection_protocol():
```
piotrjurkiewicz, line 34: Is there a good reason for such a static method? It only allows to call `Settings.get_value("xxx")` instead of `Settings().get("xxx")`, which actually requires more characters to write.

test/qa/lib/settings.py
```diff
@@ -1,86 +1,104 @@
...
+                self.settings[previous_key] += line.strip() + " "
+
+    def get(self, key: str) -> str:
+        key = key.lower()
+        return self.settings.get(key, "")
+
+    # convenient to get a value from settings only once
+    @staticmethod
+    def get_value(key: str) -> str:
+        return Settings().get(key)
```
piotrjurkiewicz, line 26: Is there a good reason for such a static method? It only allows to call Settings.get_value("xxx") instead of Settings().get("xxx"), which actually requires more characters to write.

mariusSincovici: yes, you're right :) 
I'll remove it, for some reason I had at first the impression it is shorter :laughing:

piotrjurkiewicz (approved):

devzbysiu (approved):

PR #483, opened 2024-07-03
Improve dns handling on reconnect

keliramu:


networker/networker.go
```diff
@@ -1482,6 +1492,54 @@ func (netw *Combined) allowFileshare(publicKey string, address netip.Addr) error
...
+	return nil
+}
+
+func (netw *Combined) denyDNS() error {
+	ruleName := "deny-private-dns"
+	rules := []firewall.Rule{{
+		Name:           ruleName,
+		Direction:      firewall.Outbound,
+		Protocols:      []string{"udp", "tcp"},
+		Ports:          []int{53},
```
mariusSincovici, line 1518: what should it happen if user has port `53` in allowed list?

keliramu: Allowlist FW rule will be added down after deny-DNS-private-LAN-ranges - it will work but only for non-private LAN ranges.

keliramu: Rearranged order of rules: ACCEPT rules now comes first, so, if user allowlist port 53 - that rule will be added in front of DENY rules which prevent leaks. User gets possibility to use port 53 for his custom DNS handling setup.

networker/networker.go
```diff
@@ -997,11 +1047,12 @@ func (netw *Combined) unsetAllowlist() error {
 	}
 
 	for _, rule := range []string{
-		"allowlist_subnets",
-		"allowlist_subnets_forward",
-		"allowlist_forward_related",
-		"allowlist_ports_tcp",
-		"allowlist_ports_udp",
+		"allowlist_subnets_in",
+		"allowlist_subnets_out",
```
mariusSincovici, line 1051: missing to remove `allowlist_forward_related` and `allowlist_subnets_forward` from FW

keliramu: Fixed.

bartoszWojciechO (approved):

networker/networker.go
keliramu: Also added condition check if port 53 is whitelist then do not add drop-dns rules

networker/networker.go
```diff
@@ -968,10 +962,19 @@ func (netw *Combined) setAllowlist(allowlist config.Allowlist) error {
 			}
 		}
 	}
-
 	if err := netw.fw.Add(rules); err != nil {
 		return err
 	}
+
+	// if port 53 is whitelisted - do not add drop-dns rules
+	if !allowlist.Ports.TCP[53] && !allowlist.Ports.UDP[53] {
```
mariusSincovici, line 970: Maybe not to problematic, but enabling one protocol type would allow also the other, e.g. allowing UDP only would allow also TCP.

networker/networker.go
```diff
@@ -1505,6 +1513,54 @@ func (netw *Combined) allowFileshare(publicKey string, address netip.Addr) error
...
+func (netw *Combined) denyDNS() error {
+	ruleName := "deny-private-dns"
+	rules := []firewall.Rule{{
+		Name:           ruleName,
+		Direction:      firewall.Outbound,
+		Protocols:      []string{"udp", "tcp"},
+		Ports:          []int{53},
+		PortsDirection: firewall.Destination,
+		RemoteNetworks: []netip.Prefix{
+			netip.MustParsePrefix("10.0.0.0/8"),
```
mariusSincovici, line 1542: [for future] we should define those ranges in one place, because they are redefined in several places.

mariusSincovici (approved):

Savolro (approved):

PR #484, opened 2024-07-03
Fix lan routing rules

keliramu:


devzbysiu (approved):

networker/networker.go
```diff
@@ -900,20 +900,23 @@ func (netw *Combined) setAllowlist(allowlist config.Allowlist) error {
 			continue
```
mariusSincovici, line 900: Cannot comment on line 895, but
does it make sense to have the `netw.gateway.Retrieve(subnet, netw.policyRouter.TableID())` when TableID is 0? Because that retrieves a route to an address ignoring the given table, main table when 0.
Actually does it make sense to execute this entire `for` in that case?

mariusSincovici (approved):

PR #485, opened 2024-07-04
Prevent using RoundTripper after it was closed

mariusSincovici:

The inner `RoundTripper` can be used by multiple requests in the same time. It can be closed on error or on VPN connection. This 2 operation(execute reques + close) can overlap resulting in a crash because requests are using a closed `RoundTripper`.
To prevent the crash protect access to `RoundTripper` with a `RWLock`. Plus add an `atomic.Bool` to prevent recreating RoundTripper by 2 concurrent requests on the same error.

### DIFF
--- a/request/quic_transport.go
+++ b/request/quic_transport.go
@@ -1,26 +1,34 @@
 package request
 
 import (
+	"log"
 	"net/http"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/quic-go/quic-go/http3"
 )
 
 // QuicTransport is modified/enhanced RoundTripper.
 // Thread safe.
 type QuicTransport struct {
-	inner    *atomic.Pointer[http3.RoundTripper]
-	createFn func() *http3.RoundTripper
+	// for protects the access to inner and to shouldRecreate.
+	// This will allow multiple requests in the same time for inner, but only one can recreate it
+	mu sync.RWMutex
+	// when executing the requests set this to true on error and when inner is recreated is set to false.
+	// This is used to ensure only one failed request will recreate inner value on error
+	shouldRecreate atomic.Bool
+	inner          *http3.RoundTripper
+	createFn       func() *http3.RoundTripper
 }
 
 func NewQuicTransport(fn func() *http3.RoundTripper) *QuicTransport {
-	p := &atomic.Pointer[http3.RoundTripper]{}
-	p.Store(fn())
 	return &QuicTransport{
-		inner:    p,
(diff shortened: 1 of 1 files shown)


request/quic_transport.go
```diff
@@ -1,26 +1,34 @@
...
 // Thread safe.
 type QuicTransport struct {
-	inner    *atomic.Pointer[http3.RoundTripper]
-	createFn func() *http3.RoundTripper
+	// for protects the access to inner and to shouldRecreate.
+	// This will allow multiple requests in the same time for inner, but only one can recreate it
+	mu sync.RWMutex
+	// when executing the requests set this to true on error and when inner is recreated is set to false.
+	// This is used to ensure only one failed request will recreate inner value on error
+	shouldRecreate atomic.Bool
```
mariusSincovici, line 23: if you think that having also this makes too complicated the logic, then it can be removed. 
This only purpose is to prevent recreation of the `inner` multiple times, if more requests fail in the same time.

request/quic_transport.go
```diff
@@ -1,26 +1,34 @@
...
+	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/quic-go/quic-go/http3"
 )
 
 // QuicTransport is modified/enhanced RoundTripper.
 // Thread safe.
 type QuicTransport struct {
-	inner    *atomic.Pointer[http3.RoundTripper]
-	createFn func() *http3.RoundTripper
+	// for protects the access to inner and to shouldRecreate.
```
keliramu, line 18: Comment language should be reviewed. Last part says that lock is needed to protect atomic variable?

request/quic_transport.go
```diff
@@ -29,28 +37,52 @@ func (m *QuicTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		m.recreateRoundTrip(false)
+		m.mu.RLock()
+		response, err = m.inner.RoundTrip(req)
+		m.mu.RUnlock()
+	}
+
+	return response, err
+}
+
+func shouldRecreate(err error) bool {
```
keliramu, line 72: I recommend to rename this fn as it is the same as variable and creates unnecessary code reading confusion, maybe it could be renamed to `checkShouldRecreate`

request/quic_transport.go
mariusSincovici: yeah, it is correct.

The atomic is set to `true` when a request returns an error. In that case the RWMutex is locked for `read`.
Then the atomic is set to `false` when `inner` is recreated, in which case the mutex is locked for `writing`. So nobody can change the value of the atomic unless it has the lock in read or write. It cannot be boolean because there are multiple "readers" that can change its value in the same time.

keliramu (approved):

devzbysiu (approved):

PR #487, opened 2024-07-05
Restart norduser/tray when user switches from GUI interface to text only

bartoszWojciechO:


cmd/daemon/main.go
```diff
@@ -487,10 +487,16 @@ func main() {
...
 	}
 
-	norduserMonitor := norduser.NewNordVPNGroupMonitor(norduserService)
+	norduserMonitor := norduser.NewNorduserProcessMonitor(norduserService)
 	go func() {
-		if err := norduserMonitor.Start(); err != nil {
-			log.Println("Error when starting norduser monitor: ", err.Error())
+		if snapconf.IsUnderSnap() {
+			if err := norduserMonitor.StartSnap(); err != nil {
+				log.Println("Error when starting norduser monitor for snap: ", err.Error())
```
devzbysiu, line 494: I would suggest to add `internal.XYZPrefix` to the log. Maybe `internal.WarningPrefix` here since it's not changing the control flow?
Also, you don't need space after ":" when you are using `log.Println`, it will be added by `Println`

norduser/group.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if groupMembers == "" {
+		return []string{}
+	}
+
+	return strings.Split(groupMembers, ",")
+}
+
+func getNordVPNGroupMembers() ([]string, error) {
+	const nordvpnGroupName = "nordvpn"
```
mariusSincovici, line 52: There is a `internal.NordvpnGroup`, can't that be used instead?

norduser/process_monitor.go
```diff
@@ -0,0 +1,246 @@
...
+// appropriate new state based on result of those actions
+// Desired state transitions:
+//   - notActive 	=> 	loginGUI - start application, update state to runningGUI
+//   - notActive 	=> 	loginText - start application, update state to runningText
+//   - runningGUI 	=> 	notActive - stop application, update state to notActive
+//   - runningText 	=> 	notActive - stop application, update state to notActive
+//   - runningGUI 	=> 	loginText - restart application, update state to runningText
+//   - runningText 	=> 	loginGUI - update state to runningGUI
+//
+// Other sate transitions should result in a noop.
```
mariusSincovici, line 41: `sate` -> state ?

norduser/process_monitor.go
```diff
@@ -0,0 +1,246 @@
...
+// More on runningGUI to loginText transition:
+// Due to library limitations, when user doesn't have any GUI sessions, tray will be disabled. In order to enable it on
+// subsequent GUI logins, we need to restart the application.
+//
+// Such actions are not necessary in case of transitioning from runningText to loginGUI, since in this case tray was
+// not started.
+func (s *norduserState) changeState(newState norduserState,
+	username string,
+	userIDGetter userIDGetter,
+	norduserSrevice service.NorduserService) {
```
mariusSincovici, line 52: `norduserSrevice` -> `Service`

norduser/utmp.go
```diff
@@ -49,15 +40,8 @@ int get_utmp_user_processes(char*** users) {
...
-    if ((*users)[index] == NULL) {
-      free_users_table(users, size);
-      endutxent();
-      return ERROR_MALLOC_USERNAME;
-    }
-    (*users)[index][__UT_NAMESIZE]='\0';
-
-    strncpy((*users)[index], u->ut_user, __UT_NAMESIZE);
+	(*users)[index].login_pid = u->ut_pid;
+    strncpy((*users)[index].username, u->ut_user, __UT_NAMESIZE + 1);
```
mariusSincovici, line 44: I think this will still be problematic if the `u->ut_user` has `__UT_NAMESIZE` chars, because it will take the byte after `u->ut_user`. And if the next byte is not `0` by accident, then `(*users)[index].username` will contain some garbage char and no `\0`.

norduser/utmp.go
```diff
@@ -75,25 +59,51 @@ import (
...
-	var usersCArray **C.char
+type userData map[string]norduserState
+
+// getActiveUsers returns a map of [username]userType where type can be text or gui. If any of the given users login
+// processes will be detected to have a gui(done based on the environment), user type will be gui. Otherwise user type
+// will be text.
+func getActiveUsers() (userData, error) {
+	var usersCArray *C.user
 	size := C.get_utmp_user_processes(&usersCArray)
+	defer C.free(unsafe.Pointer(usersCArray))
```
mariusSincovici, line 70: if `size` is an error it will double free the pointer.

norduser/utmp.go
```diff
@@ -75,25 +59,51 @@ import (
...
+		username := C.GoString(usernameC)
+
+		// userType was determined to be gui for any of the users login processes, so we skip this user.
+		if userType, ok := users[username]; ok && userType == loginGUI {
+			continue
+		}
+
+		loginPID := uint32(userC.login_pid)
+
+		desktopSession, err := findEnvVariableForPID(loginPID, "XDG_CURRENT_DESKTOP")
```
mariusSincovici, line 94: I think we need to define a list of supported desktop env for the app documentation. Because not all have this defined.

bartoszWojciechO: Added tech debt task

mariusSincovici (approved):

PR #491, opened 2024-07-09
For virtual server have Virtual in server name

mariusSincovici:

While connecting and disconnecting from a virtual server include Virtual into the server name.


tray/menu.go
```diff
@@ -89,10 +89,7 @@ func addVpnSection(ti *Instance) {
 	mStatus.Disable()
 
 	if ti.state.vpnStatus == ConnectedString {
-		vpnServerName := ti.state.vpnName
-		if vpnServerName == "" {
```
bartoszWojciechO, line 93: Not needed anymore?

bartoszWojciechO (approved):

tray/menu.go
mariusSincovici: I've created a function `serverName()` that calculates the server name, because there are too many `if` to put in each place :smile: 
This code must be into that function.

keliramu (approved):

PR #497, opened 2024-07-10
 ggImplement .so download

devzbysiu:


ci/download_dependencies.sh
```diff
@@ -0,0 +1,131 @@
...
+libmoose_worker_version="v10.0.4-worker"
+libmoose_worker_url="${LIBMOOSE_RELEASES_URL}/${libmoose_worker_version}/downloads/linux.zip"
+libmoose_worker_zipfile="${temp_dir}/libmoose-worker-${libmoose_worker_version}.zip"
+libmoose_worker_dst="${temp_dir}/libmoose-worker-${libmoose_worker_version}"
+
+mkdir -p "${temp_dir}"
+
+function fetch_gitlab_release() {
+  if [[ $# -ne 2 ]]; then
+    echo "Two parameters are required for fetch_gitlab_release function."
```
bartoszWojciechO, line 33: [minor] This probably won't come up, as this script will not be called directly by people, but maybe just to be safe we could print how many parameters were provided and what parameters are required? Would help with debugging if something ever breaks here.

ci/download_dependencies.sh
```diff
@@ -0,0 +1,131 @@
...
+    # ARCHS_SO_REVERSE contains architecture names used in all three libraries,
+    # so I check if the .so file exists.
+    if [[ -e ${so_path} ]]; then
+      cp "${so_path}" "${lib_root}/${ARCHS_SO_REVERSE[${arch}]}/latest"
+    fi
+  done
+}
+
+# Release zips are pretty big. Skip downloading if it was already done.
+if [[ -e "${lib_root}/checkout-completed-flag" ]]; then
```
bartoszWojciechO, line 82: Is this optimization for local/mage builds? I assume that in pipelines this script will be ran once at the beginning stages. Would be nice to have an option to run a clean build(always download fresh artifacts). We have the same approach for building the deb package, there are mage targets with `Fast` in the name which skip building if package is already present and targets without `Fast` always rebuild. But maybe just adding(in a separate task) something like a `Clean` target that would remove all the artifacts would be better.

bartoszWojciechO (approved):

ci/download_dependencies.sh
devzbysiu: Oh, that's a good idea :+1:, I'll change it

ci/download_dependencies.sh
devzbysiu: Yes, that's optimization for local builds. In the pipeline the build will always be fresh one and it will always download those dependencies with each pipeline run.

Regarding the `Fast` and not-`Fast` builds, I need to think about this, because currently, those release zips and .so files stuff all happen inside `./bin` which gets cleared with `mage clean`. And by default all builds are `Fast` (so they are not re-downloading the release files) after first build. Then the not-`Fast` mode is basically `mage clean && mage build:XYZ` but it forces us to skip everything - currently it's just data build which is additionally skipped because dependencies download happens early, so I'm not sure adding separate `Fast` mode is worth it :thinking: , but on the other hand I like consistency :smile:. I'm leaning towards your suggestion, but I'm not sure. I'll ask the team and we'll see :smile:

devzbysiu: On the other hand, the current solution is consistent with openvpn download :sweat_smile: - it also doesn't have `Fast` mode and skips the download if it already happened and the cleaning happens via `mage clean` because it's also inside `./bin`

ci/download_dependencies.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
+set -euxo pipefail
+
+source "${WORKDIR}/ci/archs.sh"
+
+temp_dir="${WORKDIR}/bin/deps/releases"
+lib_root="${WORKDIR}/bin/deps/lib"
+
+libtelio_version="v5.0.0-rc2"
```
piotrjurkiewicz, line 9: Are those a single places where version numbers will be defined? Or they have to be specified somewhere else also?

devzbysiu: Ahh, yes, that's the single place. I should probably put that somewhere more accessible :thinking: , maybe `.env`?

devzbysiu: Update: this will need to change anyway because of the libtelio integration in their pipelines - there is a need to be able to control the version used, so I'll definitely change it

piotrjurkiewicz: I think it is ok to have it here. But wouldn't it need to be synchronized with submodules commit?

devzbysiu: :thinking:, yes, you are right. So there are more things which need to be synchronized.

For internal build:
- version/tag/sha of the release so we have correct .so files for internal builds
- version/tag/sha of the go bindings for all libs

For public build
- version/tag/sha of the libdrop and libtelio for the public builds
- version/tag/sha of the go bindings for all libs

And those need to be controlled by env variables.
Ok, so this needs more work

devzbysiu: Ok I did some updates. Now, the library versions are defined in `lib-versions.env`. For now it's just the versions used to download .so files. I left the comment there to make sure that when someone changes those versions to also take into account the bindings, but for now I won't implement setting the same version for bindings, because it would require dirty hacks with patch files (same way we have with adding moose to go.mod). I'm not yet convinced we need this, but I'll think about that more.

Regarding the `export_lib_versions.sh` - it will be used by the CI to export the versions defined there but still allows to override them by the pipeline - so we have a default, but we can override it still. This will be helpful for libtelio team's integration pipeline

ci/archs.sh
```diff
@@ -52,18 +52,21 @@ declare -A ARCHS_GO=(
...
 
-# Key is one of ARCHS
-declare -A ARCHS_RUST=(
-    [i386]=i686
-    [amd64]=x86_64
-    [armel]=armv5_eabi
-    [armhf]=armv7_eabihf
-    [aarch64]=aarch64
+# for .so files comming from libtelio, libdrop and libmoose
+declare -A ARCHS_SO_REVERSE=(
```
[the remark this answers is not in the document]
devzbysiu: So this is how the libtelio, libdrop and libmoose produce artifacts in their releases:
| lib | arch |
| --- | --- |
| libdrop: | aarch64, armv5     , armv7, i686, x86_64 |
| libmooose-nordvpnapp: | aarch64, armv5_eabi, armv7_eabihf, i686, x86_64 |
| libmooose-worker: | aarch64, armv5_eabi, armv7_eabihf, i686, x86_64 |
| libtelio: | aarch64, armv5, armv7hf, i686, x86_64 |

So this dict is just a union of all the architectures from the libs, and it's mapped to what golang expects in cgo bindings. 

(I restored `ARCHS_RUST` because I'll need it later but, currently it's not used anywhere. At least I didn't find any use of it either in nordvpn-linux or nordvpn-linux-build :thinking: )

Did I answer you question? :smile:

ci/download_dependencies.sh
devzbysiu: We discussed this on the meeting and we are going with the current implementation. There will be work, in a separate ticket, to make it consistent

mariusSincovici (approved):

ci/archs.sh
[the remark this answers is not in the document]
mariusSincovici: this will be clarified later if libraries have only different names or also different arch.

.env.sample
```diff
@@ -14,3 +14,11 @@ IDEMPOTENT_DOCKER=1
...
 # Test credentials to be used in tests
 NA_TESTS_CREDENTIALS={"default": {"email": "name@example.com", "token": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"}}
+
+# Access token used to download pre-build libtelio, libdrop and libmoose binaries
+PERSONAL_ACCESS_TOKEN="<your-token-here>"
+
+# libtelio, libdrop, libmoose release pages
+LIBTELIO_RELEASES_URL="<libtelio-releases-url-here>"
+LIBDROP_RELEASES_URL="<libdrop-releases-url-here>"
+LIBMOOSE_RELEASES_URL="<libmoose-releases-url-here>"
```
Savolro, line 24: Not relevant for OSS builds. Can be migrated to internal knowledge base

ci/download_dependencies.sh
```diff
@@ -0,0 +1,135 @@
...
+libmoose_nordvpnapp_zipfile="${temp_dir}/libmoose-nordvpnapp-${LIBMOOSE_NORDVPNAPP_VERSION}.zip"
+libmoose_nordvpnapp_dst="${temp_dir}/libmoose-nordvpnapp-${LIBMOOSE_NORDVPNAPP_VERSION}"
+
+libmoose_worker_artifact_url="${LIBMOOSE_RELEASES_URL}/${LIBMOOSE_WORKER_VERSION}/downloads/linux.zip"
+libmoose_worker_zipfile="${temp_dir}/libmoose-worker-${LIBMOOSE_WORKER_VERSION}.zip"
+libmoose_worker_dst="${temp_dir}/libmoose-worker-${LIBMOOSE_WORKER_VERSION}"
+
+mkdir -p "${temp_dir}"
+
+function fetch_gitlab_release() {
```
Savolro, line 27: naming: As artifacts downloaded can be any uploaded artifacts, not just releases, this should be named `feetch_gitlab_artifact`. `release_url` needs to be renamed as well

.env.sample
```diff
@@ -14,3 +14,11 @@ IDEMPOTENT_DOCKER=1
 
 # Test credentials to be used in tests
 NA_TESTS_CREDENTIALS={"default": {"email": "name@example.com", "token": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"}}
+
+# Access token used to download pre-build libtelio, libdrop and libmoose binaries
+PERSONAL_ACCESS_TOKEN="<your-token-here>"
```
Savolro, line 19: Naming: This is not personal in most cases. Also, some indication about what it authenticates to would be nice. E. g. `GL_ACCESS_TOKEN`

magefiles/docker.go
```diff
@@ -1,5 +1,9 @@
 package main
 
+// NOTE: Do not import code from the nordvpn-linux app packages here.
+// It will force `mage` to look for *.so dependencies needed for the app
```
Savolro, line 4: This is only true when building packages that contain code which is using native dependencies. For most of packages in the repo this is not true.

magefiles/mage.go
```diff
@@ -1,5 +1,9 @@
 package main
 
+// NOTE: Do not import code from the nordvpn-linux app packages here.
+// It will force `mage` to look for *.so dependencies needed for the app
```
Savolro, line 4: Same as above

magefiles/scripts.go
```diff
@@ -1,19 +1,53 @@
 package main
 
+// NOTE: Do not import code from the nordvpn-linux app packages here.
+// It will force `mage` to look for *.so dependencies needed for the app
```
Savolro, line 4: Same as above

magefiles/scripts.go
```diff
@@ -47,3 +89,40 @@ func getGitInfo() (*gitInfo, error) {
...
+	return prefixPath(p, "PREFIX_STATIC")
+}
+
+func prefixPath(p string, envKey string) string {
+	dir := os.Getenv(envKey)
+	if dir == "" {
+		dir = "/"
+	}
+	return filepath.Clean(dir + p)
+}
```
Savolro, line 128: Can be re-used from `internal`

Savolro (changes requested):

.env.sample
```diff
@@ -1,7 +1,6 @@
 # Salt used for encrypting the configuration.
 SALT=f1nd1ngn3m0
 # Source of binaries of app dependencies.
-CI_API_V4_URL=http://example.com
```
devzbysiu, line 4: I'll update internal knowledge base after this is merged to main

magefiles/docker.go
devzbysiu: Done :+1:

magefiles/mage.go
devzbysiu: Done :+1:

magefiles/scripts.go
devzbysiu: Done :+1:

.env.sample
devzbysiu: Done and moved out of `.env.sample` as this is not needed for OSS build as well. I'll update internal knowledge base after this gets merged to main

ci/download_dependencies.sh
devzbysiu: Done :+1:

.env.sample
devzbysiu: True. I removed it. I'll update internal knowledge base after this gets merged to main

Savolro (approved):

PR #498, opened 2024-07-10
Publish initial settings after subscribe to analytics

piotrjurkiewicz:


cmd/daemon/main.go
```diff
@@ -312,6 +311,7 @@ func main() {
 		}
 	}
 	daemonEvents.Subscribe(analytics)
+	daemonEvents.Settings.Publish(cfg)
```
mariusSincovici, line 314: wouldn't this create generate events at every daemon restart, for the same user? or this is what is intended?

mariusSincovici (approved):

Savolro (approved):

keliramu (approved):

PR #503, opened 2024-07-15
Get dedicated IP servers from services data

mariusSincovici:

The list of DIP servers available for a user are part of the DIP service data. At connect and autoconnect use this list instead of calling the servers API.
* Cache the services list for a user to prevent to often API calls [NOT]
* Because the servers list changes in time add a validity duration for DIP service
* If user has DIP subscription but no servers call the API

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"os"
 	"sync"
 	"time"
 
@@ -16,8 +15,10 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
-// How much time dedicated IP servers list is valid, before making a new API call
-const cacheValidityForDIPServer = 10 * time.Minute
+type DedicatedIPService struct {
+	ExpiresAt string
+	ServerID  int64
+}
 
 // Checker provides information about current authentication.
 type Checker interface {
@@ -25,10 +26,8 @@ type Checker interface {
 	IsLoggedIn() bool
 	// IsVPNExpired is used to check whether the user is allowed to use VPN
 	IsVPNExpired() (bool, error)
-	// IsDedicatedIPExpired is used to check whether the user is allowed to use dedicated IP servers
-	IsDedicatedIPExpired() (bool, error)
-	// Get the service data if the user has it and nil otherwise
-	ServiceData(serviceID int64) (*config.ServiceData, error)
+	// Get all available dedicated IP services
+	GetDedicatedIPServices() ([]DedicatedIPService, error)
 }
 
 const (
@@ -36,16 +35,38 @@ const (
 	DedicatedIPServiceID = 11
 )
 
+type expirationChecker interface {
+	// isExpired checks if date in '2006-01-02 15:04:05' format has passed
+	isExpired(date string) bool
+}
+
+type systemTimeExpirationChecker struct{}
+
+// isTokenExpired reports whether the token is expired or not.
+func (systemTimeExpirationChecker) isExpired(expiryTime string) bool {
+	if expiryTime == "" {
+		return true
+	}
+
+	expiry, err := time.Parse(internal.ServerDateFormat, expiryTime)
+	if err != nil {
+		return true
+	}
+
+	return time.Now().After(expiry)
+}
+
 // RenewingChecker does both authentication checks and renewals in case of expiration.
 type RenewingChecker struct {
-	cm    config.Manager
-	creds core.CredentialsAPI
-	mu    sync.Mutex
+	cm         config.Manager
+	creds      core.CredentialsAPI
+	expChecker expirationChecker
+	mu         sync.Mutex
 }
 
 // NewRenewingChecker is a default constructor for RenewingChecker.
 func NewRenewingChecker(cm config.Manager, creds core.CredentialsAPI) *RenewingChecker {
-	return &RenewingChecker{cm: cm, creds: creds}
+	return &RenewingChecker{cm: cm, creds: creds, expChecker: systemTimeExpirationChecker{}}
 }
 
 // IsLoggedIn reports user login status.
@@ -81,8 +102,8 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
 	}
 
 	data := cfg.TokensData[cfg.AutoConnectData.ID]
-	if isTokenExpired(data.ServiceExpiry) {
-		if err := r.fetchServices(cfg.AutoConnectData.ID, &data); err != nil {
+	if r.expChecker.isExpired(data.ServiceExpiry) {
+		if err := r.fetchSaveServices(cfg.AutoConnectData.ID, &data); err != nil {
 			return true, fmt.Errorf("updating service expiry token: %w", err)
 		}
 		if err := r.cm.SaveWith(saveVpnExpirationDate(cfg.AutoConnectData.ID, data)); err != nil {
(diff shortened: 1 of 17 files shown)


config/config.go
```diff
@@ -100,3 +100,30 @@ type meshnet struct {
 func (d *NCData) IsUserIDEmpty() bool {
 	return d.UserID == uuid.Nil
 }
+
+// TODO: the bellow structures need to be in core package,
```
mariusSincovici, line 104: there will be a ticket to address this

config/config.go
```diff
@@ -100,3 +100,30 @@ type meshnet struct {
...
+// but it would create circular dep. for now put them here are refactor project later
+type Services struct {
+	CachedDate   time.Time     `json:"cached_date,omitempty"`
```
mariusSincovici, line 107: Maybe not store this in config and on start make a request

auth/auth.go
```diff
@@ -74,44 +105,56 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
...
+	noDIPServers := true
+	noDIPServices := true
+	dipServices := []DedicatedIPService{}
+	for _, service := range services {
+		if service.Service.ID == DedicatedIPServiceID && !r.expChecker.isExpired(service.ExpiresAt) {
+			noDIPServices = false
+			serversLen := len(service.Details.Servers)
+			if serversLen != 1 {
+				log.Println(internal.ErrorPrefix,
+					"unexpected number of dedicated ip servers in service, expected 1, is", serversLen)
```
keliramu, line 138: Maybe just take 1st item and proceed with this service?

auth/auth.go
```diff
@@ -74,44 +105,56 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
...
+				continue
+			}
+			noDIPServers = false
+			dipServices = append(dipServices,
```
keliramu, line 142: Maybe we should exit on first match? what is the point of having multiple DIP services? it's only data structure is like that, but in Nordaccount.com it is not possible to buy multiple services, afaik.

auth/auth.go
```diff
@@ -222,7 +265,8 @@ func (r *RenewingChecker) renewVpnCredentials(data *config.TokenData) error {
 	return nil
 }
 
-func (r *RenewingChecker) updateVpnExpirationDate(data *config.TokenData) error {
+// fetchSaveServices fetches services and updates data appropriately
+func (r *RenewingChecker) fetchSaveServices(userId int64, data *config.TokenData) error {
```
keliramu, line 269: I dont see use results of this function - it fetches services and saves, but saved services are not used, this data is fetched again regardless of what is saved?

daemon/servers.go
```diff
@@ -402,3 +409,110 @@ func selectFilter(tag string, group config.ServerGroup, obfuscated bool) core.Pr
...
+
+		if len(dedicatedIPServices) == 0 {
+			return nil, false, internal.NewErrorWithCode(internal.CodeDedicatedIPRenewError)
+		}
+
+		index := slices.IndexFunc(dedicatedIPServices, func(s auth.DedicatedIPService) bool {
+			return s.ServerID == server.ID
+		})
+		if index == -1 {
+			log.Println(internal.ErrorPrefix, "server not into the DIP servers list")
```
keliramu, line 476: IMO, error text should be "server is not in the DIP servers list"

daemon/rpc_connect_test.go
```diff
@@ -165,39 +174,6 @@ func TestRpcConnect(t *testing.T) {
 			checker: &workingLoginChecker{vpnErr: errors.New("test error")},
 			resp:    internal.CodeTokenRenewError,
 		},
-		{
```
keliramu, line 168: Would be good at least one test for DIP expired to check

daemon/rpc_account.go
```diff
@@ -12,6 +14,31 @@ import (
...
+		return "", fmt.Errorf("no dip services found")
+	}
+
+	layout := "2006-01-02 15:04:05"
+	latest, err := time.Parse(layout, dipServices[0].ExpiresAt)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse initial expiration date")
+	}
+
+	for _, dipService := range dipServices {
```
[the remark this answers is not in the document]
keliramu: Indeed we can - confirmed by API team.

auth/auth.go
keliramu: It is possible to have multiple DIP services - confirmed by API team.

keliramu (approved):

auth/auth.go
bartoszWojciechO: It's still used for fetching regular(not dedicated) VPN service.

daemon/rpc_connect_test.go
bartoszWojciechO: Expiration checks are done on the `RenewingChecker` side, there are some tests there: https://github.com/NordSecurity/nordvpn-linux/pull/503/files#diff-b06746870e62f95017714121be6134cffdc4bd1d46dd21f8cae8c484e3e2829dR253

daemon/rpc_account.go
```diff
@@ -12,6 +14,31 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
+func findLatestDIPExpirationData(dipServices []auth.DedicatedIPService) (string, error) {
+	if len(dipServices) == 0 {
+		return "", fmt.Errorf("no dip services found")
+	}
+
+	layout := "2006-01-02 15:04:05"
```
mariusSincovici, line 22: Isn't this same with `internal.ServerDateFormat`?

daemon/rpc_account.go
```diff
@@ -30,14 +57,21 @@ func (r *RPC) AccountInfo(ctx context.Context, _ *pb.Empty) (*pb.AccountResponse
 		accountInfo.Type = internal.CodeSuccess
 	}
 
-	dedicatedIPExpired, err := r.ac.IsDedicatedIPExpired()
+	accountInfo.DedicatedIpStatus = internal.CodeSuccess
+	dipServices, err := r.ac.GetDedicatedIPServices()
 	if err != nil {
-		log.Println(internal.ErrorPrefix, "checking dedicated IP expiration: ", err)
+		log.Println(internal.ErrorPrefix, "getting dedicated ip services: %w", err)
 		return &pb.AccountResponse{Type: internal.CodeTokenRenewError}, nil
```
mariusSincovici, line 64: If the user doesn't have DIP subscription it will always return and error.

daemon/rpc_account.go
```diff
@@ -30,14 +57,21 @@ func (r *RPC) AccountInfo(ctx context.Context, _ *pb.Empty) (*pb.AccountResponse
...
-	} else if dedicatedIPExpired {
+	}
+
+	if len(dipServices) == 0 {
 		accountInfo.DedicatedIpStatus = internal.CodeNoService
```
mariusSincovici, line 68: If the user has the service, but doesn't have a server selected would display no-service. is this ok?

daemon/rpc_account.go
```diff
@@ -30,14 +57,21 @@ func (r *RPC) AccountInfo(ctx context.Context, _ *pb.Empty) (*pb.AccountResponse
...
-	} else {
-		accountInfo.DedicatedIpStatus = internal.CodeSuccess
+	}
+
+	dedicatedIPExpirationDate, err := findLatestDIPExpirationData(dipServices)
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "getting latest dedicated ip expiration date: ", err)
+		return &pb.AccountResponse{Type: internal.CodeTokenRenewError}, nil
```
mariusSincovici, line 74: same here, if user has no DIP subscription the command fails.

daemon/rpc_account.go
bartoszWojciechO: If user doesn't have DIP subscription `GetDedicatedIPServices` should return an empty list. Only if user has DIP subscriptions but no DIP servers for any of them it should return an error. Unless something is implemented wrong. :thinking:

mariusSincovici: sorry the error is from `findLatestDIPExpirationData` a few lines bellow. The function returns error when no servers are available. And since DIP subscription is missing user has no servers

mariusSincovici (commented):
+1
Cannot approve it, just to comment :smile:

auth/auth.go
```diff
@@ -14,31 +14,61 @@ import (
...
-	IsDedicatedIPExpired() (bool, error)
+	// GetDedicatedIPServices returns all available server IDs, if server is not selected by the user it will set
+	// ServerID for that service to NoServerSelected
+	GetDedicatedIPServices() ([]DedicatedIPService, error)
 }
 
 const (
 	VPNServiceID         = 1
 	DedicatedIPServiceID = 11
+	NoServerSelected     = -1
```
Savolro, line 37: It should be OK to use 0

Savolro (approved):

PR #505, opened 2024-07-16
Fix forwarding

keliramu:
VPN connected gateway will only work when LAN-discovery is ON.    
When LAN-discovery is OFF - during VPN connect, firewall will block INPUT, OUTPUT and FORWARD traffic.

### DIFF
--- a/daemon/firewall/rule.go
+++ b/daemon/firewall/rule.go
@@ -39,6 +39,8 @@ const (
 	Outbound
 	// TwoWay defines that rule is applicable for both incoming and outgoing packets
 	TwoWay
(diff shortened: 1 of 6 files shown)


daemon/firewall/rule.go
```diff
@@ -39,6 +39,8 @@ const (
 	Outbound
 	// TwoWay defines that rule is applicable for both incoming and outgoing packets
 	TwoWay
+	// Forward new case, as Firewall does not only In/Out but alsow Forwarding control
```
devzbysiu, line 42: [small] Typo `alsow`

devzbysiu (approved):

PR #510, opened 2024-07-19
Fix log handling in QA tests

keliramu:


test/qa/lib/daemon.py
```diff
@@ -1,20 +1,19 @@
...
+from . import logging, ssh
 
 
 def _rewrite_log_path():
     project_root = os.environ["WORKDIR"].replace("/", "\\/")
     pattern = f"s/^LOGFILE=.*/LOGFILE={project_root}\\/dist\\/logs\\/daemon.log/"
-    #sh.sudo.sed("-i", pattern, "/etc/init.d/nordvpn")
-    os.popen(f"sudo sed -i {pattern} /etc/init.d/nordvpn").read()
+    sh.sudo.sed("-i", pattern, "/etc/init.d/nordvpn")
+    #os.popen(f"sudo sed -i {pattern} /etc/init.d/nordvpn").read()
```
devzbysiu, line 16: Did this just stop working? So we are switching to a different way of doing the same thing?

devzbysiu (approved):

test/qa/lib/daemon.py
keliramu: Originally there was `sh.sudo.sed` but not so long time ago it was replaced with `os.popen` - it was part of snap adoption, but that change was unnecessary, because this function runs only in docker without `systemd`

PR #519, opened 2024-07-24
Add connect tests for virtual servers

bartoszWojciechO:


test/qa/test_connect.py
```diff
@@ -450,3 +452,25 @@ def test_status_connected(tech, proto, obfuscated):
...
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(40)
+def test_connect_to_virtual_server(tech, proto, obfuscated):
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+    sh.nordvpn.set("virtual-location", "on")
+    output = sh.nordvpn.countries().stdout.decode("utf-8")
+
+    # This pattern captures all substring starting with \x1b\[94m[ that are single words. It should capture all of the
+    # virtual server names, as in the terminal output they are colored blue.
+    pattern = r"\x1b\[94m\w+\x1b\[0m"
```
mariusSincovici, line 467: It might be nice to have also a test that there is no color when `_tty_out=False` when calling the command.

test/qa/test_connect.py
```diff
@@ -450,3 +452,25 @@ def test_status_connected(tech, proto, obfuscated):
         assert time_passed - 1 <= time_connected <= time_passed + 1
 
     disconnect_base_test()
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
```
mariusSincovici, line 458: I think we should not add anymore the retries. If something is broke should stop at the first attempt.
There is also a ticket to remove all this, for all the tests.

test/qa/test_connect.py
```diff
@@ -24,11 +26,11 @@ def teardown_function(function):  # noqa: ARG001
     daemon.stop()
 
 
-def connect_base_test(connection_settings, group=(), name="", hostname=""):
+def connect_base_test(connection_settings, group=(), name="", hostname="", virtual=False):
```
mariusSincovici, line 29: by default `Virtual-location` is true. Is is in sync with this param value?

test/qa/test_connect.py
```diff
@@ -450,3 +452,25 @@ def test_status_connected(tech, proto, obfuscated):
...
 
     disconnect_base_test()
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(40)
+def test_connect_to_virtual_server(tech, proto, obfuscated):
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+    sh.nordvpn.set("virtual-location", "on")
```
mariusSincovici, line 462: might be good to also have tests for virtual location `off`. That no virtual servers are returned and user cannot connect to one.

test/qa/test_connect.py
```diff
@@ -450,3 +452,25 @@ def test_status_connected(tech, proto, obfuscated):
         assert time_passed - 1 <= time_connected <= time_passed + 1
 
     disconnect_base_test()
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.flaky(reruns=2, reruns_delay=90)
+@timeout_decorator.timeout(40)
+def test_connect_to_virtual_server(tech, proto, obfuscated):
```
mariusSincovici, line 460: might be good to have also a test for autoconnect, because there the logic changed and now it is successful set only if a server is found with the user settings.

test/qa/test_connect.py
bartoszWojciechO: This is to denote that server/group we are connecting to is a virtual, not the setting value.

test/qa/test_connect.py
bartoszWojciechO: There is a separate task to add tests related to settings, so I guess it covers this case.

mariusSincovici (approved):

keliramu (approved):

PR #524, opened 2024-07-26
Fix connect for country and city name using local servers list

mariusSincovici:
* Fix how server tags are generated when combining country and city name.
* Update unit-tests to check connect when using the API servers and also for local cached servers list


daemon/rpc_connect_test.go
```diff
@@ -165,84 +166,150 @@ func TestRpcConnect(t *testing.T) {
 			checker: &workingLoginChecker{vpnErr: errors.New("test error")},
 			resp:    internal.CodeTokenRenewError,
 		},
+		// TODO: add a test for DIP
```
mariusSincovici, line 169: waiting for DIP PR to be merged https://github.com/NordSecurity/nordvpn-linux/pull/503

mariusSincovici:
Moving back to draft because the tests can randomly fail for DIP.

test/qa/test_meshnet_routing.py
```diff
@@ -93,8 +93,6 @@ def test_killswitch_exitnode(lan_discovery: bool, local: bool):
     assert network.is_available()
 
 
-@pytest.mark.flaky(reruns=2, reruns_delay=90)
```
mariusSincovici, line 96: removed this because tests were failing and wanted to get the reason.

daemon/servers.go
```diff
@@ -126,6 +126,11 @@ func getServers(
 			serverGroup,
 			obfuscated,
 		)
+
+		// remove all DIP servers from the list if the search wasn't made for a server name and there is more then 1 server found
```
keliramu, line 130: Small typo `then` -> `than`

daemon/servers.go
```diff
@@ -126,6 +126,11 @@ func getServers(
...
+		if err == nil && serverTag.Action != core.ServerByName && len(ret) > 1 {
```
keliramu, line 131: If there is just 1 servers selected and it is DIP servers but not the one which is bought by user - user will get failure on connect later?

keliramu (approved):

daemon/servers.go
mariusSincovici: yes, user will get later the connection error when also the subscription is checked + the DIP server associated with the account.

devzbysiu (approved):

PR #527, opened 2024-07-29
Improve allowlist routing

keliramu:
Changed allow subnet routing rules to be added as pre-routing and make route decision based on `main` routing table, instead of adding them to table 205 and trying to guess how to route them.

### DIFF
--- a/daemon/routes/iprule/iprule.go
+++ b/daemon/routes/iprule/iprule.go
@@ -17,12 +17,12 @@ import (
 
 // Router uses `ip rule` under the hood
 type Router struct {
-	rpFilterManager  routes.RPFilterManager
-	ifgroupManager   ifgroup.Manager
-	tableID          uint
-	fwmark           uint32
-	prevAllowSubnets []string
-	mu               sync.Mutex
+	rpFilterManager routes.RPFilterManager
+	ifgroupManager  ifgroup.Manager
+	tableID         uint
+	fwmark          uint32
+	allowSubnets    []string
+	mu              sync.Mutex
 }
 
 // NewRouter is a default constructor for Router
@@ -119,7 +119,7 @@ func (r *Router) SetupRoutingRules(
 		}
 
 		// cleanup previous allow subnets
-		removeAllowSubnetRules(r.prevAllowSubnets, ipv6)
+		removeAllowSubnetRules(r.allowSubnets, ipv6)
 
 		// on top, add allowlisted subnet routing rules
 		for _, subnet := range allowSubnets {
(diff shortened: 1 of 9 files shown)


daemon/routes/iprule/iprule.go
```diff
@@ -341,6 +370,38 @@ func removeFwmarkRule(fwMarkVal uint32, ipv6 bool) error {
...
+
+// removeAllowSubnetRules remove all allow subnet rules
+func removeAllowSubnetRules(subnets []string, ipv6 bool) {
+	for _, subnet := range subnets {
+		_, subnetIPNet, err := net.ParseCIDR(subnet)
+		if err != nil {
+			continue
+		}
+		if err := removeAllowSubnetRule(subnetIPNet, ipv6); err != nil {
+			log.Println(internal.DeferPrefix, err)
```
mariusSincovici, line 389: [small] `internal.DeferPrefix` isn't this `Error`?

internal/string.go
```diff
@@ -72,3 +72,7 @@ func IntsToStrings(numbers []int) []string {
 	}
 	return strs
 }
+
+func CopyStringSlice(src []string) []string {
+	return append(make([]string, 0), src...)
```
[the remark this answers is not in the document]
keliramu: What reallocation could be prevented? In case of len(src)==0 we cannot return src - we still need to make/allocate empty  slice to be it separate copy and not reference to the same empty slice.

daemon/routes/iprule/iprule.go
```diff
@@ -341,6 +368,38 @@ func removeFwmarkRule(fwMarkVal uint32, ipv6 bool) error {
 	return nil
 }
 
+// addAllowSubnetRule create/add allow subnet rule
+func addAllowSubnetRule(prioID uint, subnet *net.IPNet, ipv6 bool) error {
+	if err := netlink.RuleAdd(allowSubnetRule(int(prioID), subnet, ipv6)); err != nil {
+		return fmt.Errorf("adding allow subnet rule: %s", err)
```
devzbysiu, line 374: `%w` instead of `%s`?

daemon/routes/iprule/iprule.go
```diff
@@ -16,11 +17,12 @@ import (
...
-	rpFilterManager routes.RPFilterManager
-	ifgroupManager  ifgroup.Manager
-	tableID         uint
-	fwmark          uint32
-	mu              sync.Mutex
+	rpFilterManager  routes.RPFilterManager
+	ifgroupManager   ifgroup.Manager
+	tableID          uint
+	fwmark           uint32
+	prevAllowSubnets []string
```
mariusSincovici, line 24: [minor] regarding the naming, shouldn't it be `allowSubnets`, because it contains the list with the current allow list?

mariusSincovici (approved):

daemon/routes/iprule/iprule.go
keliramu: probably you are right, just wanted member variable to be different from parameter name for better readability, but `prev` does not fit

PR #528, opened 2024-07-29
Implement moose first_open event

piotrjurkiewicz:


events/moose/moose.go
```diff
@@ -234,21 +234,16 @@ func (s *Subscriber) NotifyIpv6(data bool) error {
...
-		moose.NordvpnappUserInterfaceItemTypeButton,
-		data.Server,
-		fmt.Sprintf("%d", data.Rate),
-	))
-}
-
-func (s *Subscriber) NotifySendInvitation(any) error {
+func (s *Subscriber) NotifyUiItemsClick(data events.UiItemsAction) error {
+	itemType := moose.NordvpnappUserInterfaceItemTypeButton
+	if data.ItemType == "textbox" {
```
mariusSincovici, line 239: wouldn't be possible to have `"textbox"` defined into a constant or an enum, since it is used also in the server.go file?

daemon/rpc_rate.go
```diff
@@ -10,7 +11,7 @@ import (
 
 func (r *RPC) RateConnection(ctx context.Context, in *pb.RateRequest) (*pb.Payload, error) {
 	if r.netw.LastServerName() != "" {
-		r.events.Service.Rate.Publish(events.ServerRating{Rate: int(in.GetRating()), Server: r.netw.LastServerName()})
+		r.events.Service.UiItemsClick.Publish(events.UiItemsAction{ItemName: "server_speed_rating", ItemType: "button", ItemValue: r.netw.LastServerName(), FormReference: fmt.Sprintf("%d", int(in.GetRating()))})
```
mariusSincovici, line 14: same for `button`?

config/manager.go
```diff
@@ -235,6 +237,11 @@ func (f *FilesystemConfigManager) newKey() error {
 		return err
 	}
 
+	err = f.fsHandle.CreateFile(f.vault, internal.PermUserRW)
```
mariusSincovici, line 240: Wouldn't make sense instead of having this to add `chmod` into the `WriteFile`, to fix all the places the permissions? if this was made to set the current permissions on the file.
Because the same problem is with `f.save(*c)`. If the file exists it will not change the permissions.

mariusSincovici (approved):

events/moose/moose.go
piotrjurkiewicz: These are defined by enumerated constants, but in file which is conditionally imported only for moose-enabled builds. So I think it is better to just use strings, instead of copying that enum to our code and later to having to keep it manually in sync with the moose library.

keliramu (approved):

PR #531, opened 2024-07-31
Add wait for iptables

mariusSincovici:
In some cases iptables are locked and the command fails with `Another app is currently holding the xtables lock. Perhaps you want to use the -w option?`. To prevent this add `-w` into the param list for `iptables` Add also for how long to wait for the lock to be free, otherwise the daemon gets stuck.

### DIFF
--- a/daemon/firewall/iptables/iptables.go
+++ b/daemon/firewall/iptables/iptables.go
@@ -21,8 +21,8 @@ const (
 
 const (
 	accept   ruleTarget = "ACCEPT"
-	drop                = "DROP"
-	connmark            = "CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff"
+	drop     ruleTarget = "DROP"
+	connmark ruleTarget = "CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff"
 )
 
 // ruleTarget specifies what can be passed as an argument to `-j`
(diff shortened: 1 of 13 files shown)


piotrjurkiewicz:
There is also an iptables call in `daemon/firewall/iptables/iptables_test.go:676`

piotrjurkiewicz (approved):

devzbysiu (approved):

internal/constants.go
```diff
@@ -82,6 +82,8 @@ const (
 	FileshareLogFileName = "nordfileshare" + LogFileExtension
 
 	LogFileExtension = ".log"
+
+	SecondsToWaitForIptablesLock = "5" // use string to easier concatenate into the command args
```
keliramu, line 86: IMO, this parameters should live in `firewall` top package

daemon/firewall/iptables/iptables.go
```diff
@@ -96,7 +96,7 @@ func (ipt *IPTables) applyRule(rule firewall.Rule, add bool) error {
 		}
 		for _, ipTableRule := range ipTablesRules {
 			// -w does not accept arguments on older iptables versions
-			args := fmt.Sprintf("%s %s -w", flag, ipTableRule)
+			args := fmt.Sprintf("%s %s -w "+internal.SecondsToWaitForIptablesLock, flag, ipTableRule)
```
keliramu, line 99: Let's be consistent: `-w` parameter should be passed via `%d` same
```
fmt.Sprintf("%s %s -w %d", internal.SecondsToWaitForIptablesLock, flag, ipTableRule)
```

keliramu (approved):

internal/constants.go
keliramu: It should be integer, I guess

daemon/firewall/iptables/iptables.go
keliramu: Or there is some reason you are doing this way?

keliramu: Sorry, I see your comment one line above, ok

PR #532, opened 2024-08-01
Fix settings file creation

piotrjurkiewicz:


config/manager.go
```diff
@@ -140,6 +140,11 @@ func (f *FilesystemConfigManager) save(c Config) error {
 		return err
 	}
 
+	err = f.fsHandle.CreateFile(f.location, internal.PermUserRW)
+	if err != nil {
+		return err
+	}
+
 	return f.fsHandle.WriteFile(f.location, encrypted, internal.PermUserRW)
```
mariusSincovici, line 148: Shouldn't `WriteFile` be changed to create the folders parents?

mariusSincovici (approved):

config/manager.go
piotrjurkiewicz: Maybe

piotrjurkiewicz: @bartoszWojciechO What do you think?

bartoszWojciechO: I have only implemented the abstraction so that filesystem access can be mocked. Both ways are fine by me.

bartoszWojciechO (approved):

PR #542, opened 2024-08-07
Publish setting changes

bartoszWojciechO:


protobuf/state/state.proto
```diff
@@ -12,6 +15,44 @@ enum ConnectionState {
...
+	THREATPROTECTIONLITE = 1;
+	OBFUSCATE = 2;
+	FIREWALL = 3;
+	ROUTING = 4;
+	NOTIFY = 5;
+	MESHNET = 6;
+	IPV6 = 7;
+	DEFAULTS = 8;
+	LAN_DISCOVERY = 9;
+	VIRTUA_LOCATION = 10;
```
keliramu, line 29: Typo `VIRTUA_LOCATION`

state/state.go
```diff
@@ -44,11 +50,17 @@ func (s *StatePublisher) notify(e interface{}) {
...
+	defer s.mu.Unlock()
+	s.notify(SettingEvent{setting: settingType, enabled: enabled})
+}
+
 func (s *StatePublisher) NotifyConnect(e events.DataConnect) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	log.Printf(internal.DebugPrefix+" notifying about connect event: %+v", e)
+	log.Println(internal.DebugPrefix, "notifying about connect event: %+v", e)
```
devzbysiu, line 63: Will this formatting string here work with `log.Println`?

protobuf/state/state.proto
```diff
@@ -12,6 +15,44 @@ enum ConnectionState {
     CONNECTED = 2;
 }
 
+enum Setting {
+    AUTOCONNECT = 0;
+	THREATPROTECTIONLITE = 1;
```
devzbysiu, line 20: Formatting? Or is this indentation intentional? :smile:

devzbysiu (approved):

protobuf/state/state.proto
```diff
@@ -22,8 +63,23 @@ message ConnectionStatus {
...
+}
+
+message LoginEvent {
+    LoginEventType type = 1;
+}
+
 message AppState {
     oneof state {
         ConnectionStatus connection_status = 1;
+        SettingChange setting_change = 2;
```
keliramu, line 78: Meshnet state to be added later?

bartoszWojciechO: For this task there's a notification for meshnet setting change(on/off). I don't see any task for more complicated meshnet data, such as new peers appearing or changing permissions, maybe it should be added. :thinking:

config/manager.go
```diff
@@ -85,19 +89,22 @@ type FilesystemConfigManager struct {
...
+	configPublisher ConfigPublisher
 	mu              sync.Mutex
 }
 
 // NewFilesystemConfigManager is constructed from a given location and salt.
 func NewFilesystemConfigManager(location, vault, salt string,
 	machineIDGetter MachineIDGetter,
-	fsHandle FilesystemHandle) *FilesystemConfigManager {
+	fsHandle FilesystemHandle,
+	conifgPublisher ConfigPublisher) *FilesystemConfigManager {
```
mariusSincovici, line 100: [typo] `conifgPublisher`

config/manager.go
```diff
@@ -114,7 +121,11 @@ func (f *FilesystemConfigManager) SaveWith(fn SaveFunc) error {
 	}
 
 	c = fn(c)
-	return f.save(c)
+	err := f.save(c)
+	if err == nil && f.configPublisher != nil {
+		f.configPublisher.Publish(&c)
```
mariusSincovici, line 126: Shouldn't this not be called while the mutex is locked?

bartoszWojciechO: Why?

mariusSincovici (approved):

config/manager.go
mariusSincovici: The only concern is not to have a deadlock,  if some component will listening on this events and then tries to read the settings from here.

PR #543, opened 2024-08-07
  Update public build for dynamic linking changes

devzbysiu:


ci/docker/ruster/Dockerfile
```diff
@@ -32,18 +36,33 @@ RUN set -eux; \
...
+    arm-unknown-linux-gnueabi
+
+RUN set -eux; \
+  mkdir -p /usr/lib/i386-linux-gnu \
+    /usr/lib/x86_64-linux-gnu \
+    /usr/lib/arm-linux-gnueabihf \
+    /usr/lib/arm-linux-gnueabi \
+    /usr/lib/aarch64-linux-gnu
+
+RUN --mount=type=secret,id=gl_access_token export GL_ACCESS_TOKEN="$(cat /run/secrets/gl_access_token)"; \
```
mariusSincovici, line 56: doesn't this mean that if the token expires the image cannot be used anymore?

devzbysiu: The `--mount-type=secret(...)` is used only at build time and the secret is available only for the `RUN` command it's used in. After the command is executed, the secret is unmounted so it's not stored in the image. So this is build-only feature and if the token expires later it doesn't matter because it's not stored in the image and it's not used at runtime

mariusSincovici (approved):

magefiles/mage.go
```diff
@@ -143,6 +143,31 @@ func Clean() error {
 			return err
 		}
 	}
+
+	// cleanup rust for public builds
+	env, err := getEnv()
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(env["FEATURES"], "internal") {
```
keliramu, line 152: I dont think it is necessary (or even totally correct) to check for `internal` - IMO, it's sufficient to check for dir presence - if it is present - clean it. Just add these dirs to the `buildFolders` list on line #106

keliramu (approved):

build/foss/build.sh
Savolro, line 1: as `foss` is removed, I think this could be simply moved to `ci/build_rust.sh`

devzbysiu: I'm cloning `libtelio` and `libdrop` inside `build/foss` for public build

build/foss/build.sh
```diff
@@ -1,21 +1,104 @@
...
+  fi
+  repo_url=$1
+  version=$2
+  dst_arch_dir=$3
+  repo_name="${repo_url##*/}"
+  repo_name="${repo_name%.git}"
+
+  if [[ ! -d "${dst_arch_dir}/${repo_name}" ]]; then
+    pushd "${dst_arch_dir}"
+    git clone --branch "${version}" "${repo_url}"
```
Savolro, line 32: Please check if this is doable with branches, tags and commit checksums

magefiles/mage.go
devzbysiu: Addressed here https://github.com/NordSecurity/nordvpn-linux/pull/575/files

build/foss/build.sh
devzbysiu: Done here https://github.com/NordSecurity/nordvpn-linux/pull/575/files

build/foss/build.sh
devzbysiu: This didn't work, but I updated it to allow checkout to any revision

PR #547, opened 2024-08-08
Fix autoconnect to not turn off obfuscate

keliramu:


devzbysiu (approved):

daemon/rpc_set_autoconnect.go
```diff
@@ -74,7 +74,7 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 				ServerTag:            in.GetServerTag(),
 				Protocol:             cfg.AutoConnectData.Protocol,
 				ThreatProtectionLite: cfg.AutoConnectData.ThreatProtectionLite,
-				Obfuscate:            in.GetObfuscate(),
+				Obfuscate:            cfg.AutoConnectData.Obfuscate,
```
mariusSincovici, line 77: why do we modify so many things in there since only the `Obfuscate`  and `ServerTag` actually changes?

why `pb.SetAutoconnectRequest` has so many members, but in the end just 2 are needed, and to populate the rest into cli_set_autoconnect.go the app fetches first the settings, just to pass them back? e.g. allow list and before was obfuscation.

keliramu: I see there is `obfuscate` and `allowlist` parameters in `AutoconnectRequest` but their usage is under question now. This is legacy. Maybe that was needed when some user settings were kept individually for each user - now we do not have user side settings persistence, only deamon side. E.g. `allowlist` is being sent in gRPC request - for now it looks unnecessary. But I would leave it as is and maybe need to register a technical debt task for this area.

bartoszWojciechO (approved):

PR #549, opened 2024-08-09
Add internet monitoring

mariusSincovici:


keliramu (approved):

test/qa/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+import time
+
+import pytest
+import sh
+
+from lib import logging, network
+
+_CHECK_FREQUENCY=5
+
+#TODO: check connectivity outside VPN tunnel!
```
mariusSincovici, line 14: is it still a TODO?

test/qa/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+        except sh.ErrorReturnCode as e:
+            print(f"~~~_check_connection_to_ip: Failed to connect to {ip_address}: {e}.")
+        time.sleep(_CHECK_FREQUENCY)
+
+
+def _check_connection_to_ip_outside_vpn(ip_address):
+    while True:
+        try:
+            print(f"~~~_check_connection_to_ip_outside_vpn: {ip_address}")
+            "icmp_seq=" in sh.ping("-c", "1", "-w", "1", "-I", "eth0", ip_address) # noqa: B015
```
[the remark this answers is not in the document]
keliramu: Yes, it should be blocked during vpn connect, need fwmark

test/qa/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+    dns_resolver_thread.start()
+
+    yield
+
+
+def _check_connection_to_ip(ip_address):
+    while True:
+        try:
+            print(f"~~~_check_connection_to_ip: {ip_address}")
+            "icmp_seq=" in sh.ping("-c", "1", "-w", "1", ip_address) # noqa: B015
```
devzbysiu, line 62: Can we go with higher count for ping? In tests we have `-c 2` and on nightly pipelines, I observed that we sometimes have not 100% but 50% packet loss so some pings went through

test/qa/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+        except sh.ErrorReturnCode as e:
+            print(f"~~~_check_connection_to_ip_outside_vpn: Failed to connect to {ip_address}: {e}.")
+        time.sleep(_CHECK_FREQUENCY)
+
+
+def _check_dns_resolution(domain):
+    while True:
+        try:
+            print(f"~~~_check_dns_resolution: {domain}")
+            socket.gethostbyname(domain)
```
Savolro, line 81: This will not indicate no-nets as it retrieves hostname from DNS cache if it was available. Which will be true in this case.

test/qa/conftest.py
keliramu: Removed

Savolro (approved):

test/qa/conftest.py
keliramu: Increased ping count to 3

test/qa/conftest.py
keliramu: Replaced with dns.Resolver and using external DNS server directly

test/qa/conftest.py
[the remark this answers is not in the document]
keliramu: Added mark to ping and removed selection of network interface

devzbysiu (approved):

test/qa/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+        time.sleep(_CHECK_FREQUENCY)
+
+
+def _check_dns_resolution(domain):
+    while True:
+        try:
+            print(f"~~~_check_dns_resolution: {domain}")
+            resolver = dns.resolver.Resolver()
+            resolver.nameservers = ['8.8.8.8']
+            resolver.resolve(domain, 'A')  # 'A' for IPv4
```
Savolro, line 85: How would this pass our firewall?

keliramu: To go arround tunnel with DNS query, need to play with sockets - had experiments but no success yet as it requires more effort. To be done as separate task.

PR #560, opened 2024-08-14
Add process monitoring

devzbysiu:


internal/filesystem.go
```diff
@@ -593,3 +593,46 @@ func OpenOrCreateRegularFile(fileName string, flags int, permission fs.FileMode)
...
+	for _, dir := range procDirs {
+		if _, err := strconv.Atoi(dir.Name()); err != nil {
+			continue
+		}
+		exePath := filepath.Join("/proc", dir.Name(), "exe")
+		resolvedPath, err := readlink(exePath)
+		if err != nil {
+			continue
+		}
+		if resolvedPath == executablePath {
```
piotrjurkiewicz, line 633: Maybe paths should be converted into canonical forms before comparing them

meshnet/server.go
```diff
@@ -3069,6 +3069,22 @@ func (s *Server) getPeerWithIdentifier(id string, peers mesh.MachinePeers) *mesh
...
+	var cfg config.Config
+	if err := s.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, "failed to read configuration when listing peers:", err)
+		return nil, fmt.Errorf("reading configuration when listing peers: %w", err)
+	}
+
+	token := cfg.TokensData[cfg.AutoConnectData.ID].Token
+	peers, err := s.reg.List(token, cfg.MeshDevice.ID)
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "failed to list peers:", err)
```
mariusSincovici, line 3082: It might be better not to print logs here.
Since the function returns an error, the caller should decide what to do with the error. Otherwise there will be duplicate messages into logs.
Also above.

internal/filesystem.go
```diff
@@ -593,3 +593,46 @@ func OpenOrCreateRegularFile(fileName string, flags int, permission fs.FileMode)
...
+		log.Println(WarningPrefix, "failed to check if process is running, returning false:", err)
+		return false
+	}
+	return isRunning
+}
+
+func isProcessRunning(executablePath string, readdir readdirFunc, readlink readlinkFunc) (bool, error) {
+	procDirs, err := readdir("/proc")
+	if err != nil {
+		log.Printf(ErrorPrefix+" error reading /proc directory: %v\n", err)
```
mariusSincovici, line 620: If there is an error the same error message would be printed 2 times, here and in `IsProcessRunning(execPath string) bool`

mariusSincovici (approved):

meshnet/jobs.go
```diff
@@ -30,3 +37,40 @@ func JobRefreshMeshnet(s *Server) func() error {
 		return nil
 	}
 }
+
+func JobMonitorFileshareProcess(s *Server) func() error {
+	wasFileshareUp := false
+	return func() error {
```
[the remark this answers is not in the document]
devzbysiu: It's a little bit more than `BlockFileshare` vs `AllowFileshare` because there is also `wasFileshareUp` state. And the `Block..` vs `Allow..` are called only on state change, so I can't just pull everything out of `if` blocks and determine only `Block..` vs `Allow..` in an if, but I can do something like this:
```
func JobMonitorFileshareProcess(s *Server) func() error {
	state := fileshareState{
		wasUp: false,
	}
	return func() error {
		state.isUp = internal.IsProcessRunning(internal.FileshareBinaryPath)

		var change func(UniqueAddress) error = nil
		switch state {
		case fileshareState{wasUp: true, isUp: false}:
			// fileshare process was running but it stopped - block fileshare ports
			change = s.netw.BlockFileshare
			state.wasUp = false
		case fileshareState{wasUp: false, isUp: true}:
			// fileshare process was down but now it is running - allow fileshare ports
			change = s.netw.AllowFileshare
			state.wasUp = true
		default:
			// No change is needed - either:
			// - fileshare was running and is running, or
			// - fileshare was down and is down
			// In both cases there is no state change.
		}

		if change != nil {
			peers, err := s.listPeers()
			if err != nil {
				return err
			}
			for _, peer := range peers {
				if err := change(UniqueAddress{UID: peer.PublicKey, Address: peer.Address}); err != nil {
					log.Println(internal.WarningPrefix, "failed to apply fileshare port change:", err)
				}
			}
		}

		return nil
	}
}
```
Is this better?

internal/filesystem.go
devzbysiu: I like this, I'll do that :+1:

bartoszWojciechO (approved):

meshnet/server.go
devzbysiu: Done :+1:

internal/filesystem.go
devzbysiu: Done

internal/filesystem.go
devzbysiu: Done :+1:

meshnet/jobs.go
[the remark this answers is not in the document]
devzbysiu: Improved

PR #562, opened 2024-08-19
Return available DIP servers and MFA status in the account RPC

bartoszWojciechO:


protobuf/daemon/account.proto
```diff
@@ -4,9 +4,15 @@ package pb;
...
 
-message DedidcatedIPServices {
-  int64 dedicated_ip_status = 5;
-  string dedicated_ip_expires_at = 6;
+message DedidcatedIPService {
+  int64 server_id = 1;
+  string dedicated_ip_expires_at = 2;
+}
+
+enum MFAStatus {
```
mariusSincovici, line 12: wouldn't make sense to move this enum in `common.proto` rename to something more generic, because it might be useful in the future for other responses?
possible names :melting_face:  `TriState`, `State`, `TriBool`...?

mariusSincovici (approved):

auth/auth.go
```diff
@@ -93,6 +95,25 @@ func (r *RenewingChecker) IsLoggedIn() bool {
...
+	}
+
+	data := cfg.TokensData[cfg.AutoConnectData.ID]
+
+	resp, err := r.creds.MultifactorAuthStatus(data.Token)
+	if err != nil {
+		return false, fmt.Errorf("querying MFA status: %w", err)
+	}
+
+	return resp.Status == "enabled", nil
```
keliramu, line 114: Would be good to have constant defined and used on both ends

cli/cli_account.go
```diff
@@ -68,10 +68,19 @@ func (c *cmd) Account(ctx *cli.Context) error {
 
 	if err := displayServiceStatus("Dedicated IP",
 		payload.DedicatedIpStatus,
-		payload.DedicatedIpExpiresAt); err != nil {
+		payload.LastDedicatedIpExpiresAt); err != nil {
 		return err
 	}
 
+	mfa := "enabled"
```
keliramu, line 75: Would be good to have constants defined and used on both ends

auth/auth.go
keliramu: Ok, here it is from API and might be different thing than in `cli_account.go`

keliramu: But maybe string constants could be defined in `pb` and used consistently instead of having numeric constants in `pb` but converting to/from strings on both ends?

keliramu (approved):

devzbysiu (approved):

cli/cli_account.go
mariusSincovici: If you are referring on passing the string `"enabled"` in the gRPC response, then I think it is better to pass the enum instead of strings, as it is now. 
Because the string is only useful if it is displayed as it is, otherwise will be string comparison vs int comparison. In that case I think int comparison is nicer :smiley: .

Issue #550, opened 2024-08-10
Build not working: cargo not found?


Wikpi:
Hello, I was trying to build the nordvpn-linux app and ran into a problem.

I was reading [BUILD.md](https://github.com/NordSecurity/nordvpn-linux/blob/main/BUILD.md) and executed the build command:

`mage build:binariesDocker`

Which then gave the following:

```Running target: Build:BinariesDocker
Running dependency: Download
exec: ci/check_dependencies.sh 
Running dependency: main.Build.RustDocker
creating docker client
creating ruster docker container
starting ruster docker container
waiting for ruster docker container to finish
allexport      	off
braceexpand    	on
emacs          	off
errexit        	on
errtrace       	off
functrace      	off
hashall        	on
histexpand     	off
history        	off
ignoreeof      	off
interactive-comments	on
keyword        	off
monitor        	off
noclobber      	off
noexec         	off
noglob         	off
nolog          	off
notify         	off
nounset        	on
onecmd         	off
physical       	off
pipefail       	off
posix          	off
privileged     	off
verbose        	off
vi             	off
xtrace         	off
+ source /opt/ci/archs.sh
++ [[ ! -n amd64 ]]
++ ARCHS_REVERSE=([i386]=i386 [i686]=i386 [amd64]=amd64 [x86_64]=amd64 [armel]=armel [armv5l]=armel [armv5_eabi]=armel [armhf]=armhf [armhfp]=armhf [armv7_eabihf]=armhf [arm64]=aarch64 [aarch64]=aarch64)
++ declare -A ARCHS_REVERSE
++ ARCHS_DEB=([i386]=i386 [amd64]=amd64 [armel]=armel [armhf]=armhf [aarch64]=arm64)
++ declare -A ARCHS_DEB
++ ARCHS_RPM=([i386]=i386 [amd64]=x86_64 [armel]=armv5l [armhf]=armhfp [aarch64]=aarch64)
++ declare -A ARCHS_RPM
++ ARCHS_GO=([i386]=386 [amd64]=amd64 [armel]=arm [armhf]=arm [aarch64]=arm64)
++ declare -A ARCHS_GO
++ ARCHS_RUST=([i386]=i686 [amd64]=x86_64 [armel]=armv5_eabi [armhf]=armv7_eabihf [aarch64]=aarch64)
++ declare -A ARCHS_RUST
++ export ARCHS
++ export ARCHS_REVERSE
++ export ARCHS_DEB
[...]

PR #563, opened 2024-08-19
Upgrade rust to 1.80.1 in ruster image

devzbysiu:
Fixes build issue with rust version mentioned [here](https://github.com/NordSecurity/nordvpn-linux/issues/550#issuecomment-2281670976)

### DIFF
--- a/ci/docker/ruster/Dockerfile
+++ b/ci/docker/ruster/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH \
-    RUST_VERSION=1.72.1
+    RUST_VERSION=1.80.1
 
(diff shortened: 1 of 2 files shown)


ci/docker/ruster/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH \
-    RUST_VERSION=1.72.1
+    RUST_VERSION=1.80.0
```
devzbysiu, line 7: FYI: confirmed with lib leads, that we should be safe with this version

keliramu (approved):

mariusSincovici (approved):

matszczygiel:
Hey, just wanted to note that Rust 1.80 was found to have miscompillation issue https://blog.rust-lang.org/2024/08/08/Rust-1.80.1.html. You may want to use 1.80.1 instead

PR #568, opened 2024-08-20
Improve snap build

keliramu:


devzbysiu (approved):

ci/build_snap.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+source "${WORKDIR}/ci/env.sh"
+
+cd "${WORKDIR}"
+
+if [ "${ENVIRONMENT}" = "prod" ]; then
+    STRIP="$(which eu-strip 2>/dev/null)"
+    BASEDIR="bin/${ARCH}"
```
mariusSincovici, line 10: maybe this should be revisit at some point, because it will be the 3rd place where the list of apps is defined.

mariusSincovici (approved):

PR #573, opened 2024-08-21
Return current version from Ping RPC command

piotrjurkiewicz:


cli/cli_version.go
```diff
@@ -0,0 +1,50 @@
...
+	"strings"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+func (c *cmd) Version(ctx *cli.Context) error {
+	fmt.Printf("CLI: ")
```
mariusSincovici, line 15: I don't think this is needed because CLI is part of the same package as the daemon, so they will always have the same version.

mariusSincovici (approved):

cli/cli_version.go
keliramu: The output:
```
$ nv version
CLI: NordVPN Version 3.18.4+52a128f1 - dev (52a128f1)
Daemon: NordVPN Version 3.18.4+52a128f1
```
Is it possible that CLI and DAEMON would differ in versions? Does it make sense to show two versions?

piotrjurkiewicz: I added this to have an example that it works. Maybe in some circumstances it would be possible to end up with two different versions and such a message would be helpful for debugging.

mariusSincovici: Unless the user replaces the CLI/daemon manually, for now they cannot be different.
regarding debug, if this is for debug purpose the maybe it should be some `if(isDebug)` somewhere.

but there is a bit of inconsistency. `nordvpn --version` returns something, while `nordvpn version` returns something else.
Maybe if you want to test what the daemon version is, then just print the response from ping and no CLI. In this way you have `nordvpn --version` = CLI version and `nordvpn version` = daemon version?

bartoszWojciechO (approved):

PR #584, opened 2024-08-28
Add servers RPC

bartoszWojciechO:


daemon/rpc_servers_test.go
```diff
@@ -0,0 +1,204 @@
...
+				{
+					Identifier: core.VirtualLocation,
+					Values: []struct {
+						Value string "json:\"value\""
+					}{
+						{Value: "True"},
+					},
+				},
+			},
+			Groups: core.Groups{
```
mariusSincovici, line 57: for this group add also the standard group, to be closer to the response from the API.

daemon/rpc_servers.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return pb.Technology_UNKNWON
+}
+
+func groupFilter(groups core.Groups) config.ServerGroup {
+	filter := []config.ServerGroup{
+		// standard VPN
+		config.ServerGroup_STANDARD_VPN_SERVERS,
+		// P2P(the servers part of this group have from API also standard group which needs to be ignored)
+		config.ServerGroup_P2P,
```
mariusSincovici, line 34: Will there ever by P2P servers? because Standard VPN group is checked first and all P2P have also defined.

Maybe move `config.ServerGroup_STANDARD_VPN_SERVERS` to be the last one checked, in this way if a server is part of any other specialty group will be better identified?

mariusSincovici (approved):

devzbysiu (approved):

PR #590, opened 2024-08-29
Include connection parameters in the status RPC response

bartoszWojciechO:


protobuf/daemon/status.proto
```diff
@@ -6,6 +6,20 @@ option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
 
 import "config/protocol.proto";
 import "config/technology.proto";
+import "config/group.proto";
+
+enum ConnectionSource {
```
keliramu, line 11: Is there defined/agreed a purpose of connection source? It is not being logged, so maybe a frontend is planning to show it to the user?

bartoszWojciechO: Yeah, it is a GUI requirement.

devzbysiu (approved):

keliramu (approved):

PR #611, opened 2024-09-10
Dynamic linking

devzbysiu:


bartoszWojciechO (approved):

ci/compile.sh
```diff
@@ -46,6 +46,15 @@ export CGO_ENABLED=1
 GOARCH="${ARCHS_GO["${ARCH}"]}"
 export GOARCH="${GOARCH}"
 
+if [[ "${#}" -ge 1 ]]; then
+  if [[ "${1}" == "docker" ]]; then
+    # Inside container, go tried to put cache under HOME, and it's owned
```
Savolro, line 51: Is this still true? Aren't build containers using `root`?

ci/docker/ruster/Dockerfile
```diff
@@ -32,18 +36,33 @@ RUN set -eux; \
...
+    /usr/lib/aarch64-linux-gnu
+
+RUN --mount=type=secret,id=gl_access_token export GL_ACCESS_TOKEN="$(cat /run/secrets/gl_access_token)"; \
+  set -eux; \
+  curl --header "PRIVATE-TOKEN: ${GL_ACCESS_TOKEN}" -o /usr/lib/i386-linux-gnu/libsqlite3.so -L ${SQLITE_DOWNLOAD_URL_PREFIX}/linux_i686_sqlite3.so; \
+  curl --header "PRIVATE-TOKEN: ${GL_ACCESS_TOKEN}" -o /usr/lib/x86_64-linux-gnu/libsqlite3.so -L ${SQLITE_DOWNLOAD_URL_PREFIX}/linux_x86_64_sqlite3.so; \
+  curl --header "PRIVATE-TOKEN: ${GL_ACCESS_TOKEN}" -o /usr/lib/arm-linux-gnueabihf/libsqlite3.so -L ${SQLITE_DOWNLOAD_URL_PREFIX}/linux_armv7_eabihf_sqlite3.so; \
+  curl --header "PRIVATE-TOKEN: ${GL_ACCESS_TOKEN}" -o /usr/lib/arm-linux-gnueabi/libsqlite3.so -L ${SQLITE_DOWNLOAD_URL_PREFIX}/linux_armv5_eabi_sqlite3.so; \
+  curl --header "PRIVATE-TOKEN: ${GL_ACCESS_TOKEN}" -o /usr/lib/aarch64-linux-gnu/libsqlite3.so -L ${SQLITE_DOWNLOAD_URL_PREFIX}/linux_aarch64_sqlite3.so; \
+  apt-get remove -y --auto-remove curl
```
Savolro, line 63: Any reason https://docs.docker.com/build/building/multi-stage/ was not used?

devzbysiu: As far as I remember we didn't want to do that in dynamic linking, because our docker images need cleanup in general anyway and this was going to be done with the cleanup later? Something like that, it was some time ago, not sure if I remember correctly

ci/compile.sh
devzbysiu: You're right, it's not needed anymore. Removed

Savolro (approved):

PR #612, opened 2024-09-11
Filter servers by tech/protocol/obfuscation in Servers RPC

bartoszWojciechO:


protobuf/daemon/servers.proto
```diff
@@ -6,6 +6,11 @@ option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
 
 import "config/group.proto";
 
+enum ServersError {
+    GET_CONFIG_ERROR = 0;
```
keliramu, line 10: Good/common practice is to have one/first item as `no value` or `NO_ERROR` in this case

mariusSincovici (approved):

PR #614, opened 2024-09-12
Add FeatureDirect config with default values

devzbysiu:


keliramu (approved):

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -162,6 +162,7 @@ func New(prod bool, eventPath string, fwmark uint32,
 		log.Println(internal.ErrorPrefix, "failed to get telio config:", err)
 
 		defaultTelioConfig := teliogo.GetDefaultFeatureConfig()
+		defaultTelioConfig.Direct = &teliogo.FeatureDirect{}
```
mariusSincovici, line 165: Maybe check when it is nil and only then set it? in case the behavior changes to not to replace the default one.

PR #618, opened 2024-09-13
Move .so files to /usr/lib/nordvpn

devzbysiu:
Changes:
- nfpm now creates a package with `.so` files under `/usr/lib/nordvpn` instead of `/usr/lib`
- additional linker configuration is required to take `/usr/lib/nordvpn` into account when searching for `.so`


Savolro (approved):

contrib/scriptlets/deb/postinst
```diff
@@ -24,7 +24,10 @@ userModMessage="To allow other users to control the application run 'usermod -aG
 userDelMessage="To limit access exclusively to the root user, remove all users from the '$NORDVPN_GROUP' group."
 helpMessage="If you need help using the app, use the command 'nordvpn --help'."
 
-ldconfig # update shared library cache
+# update configuration and shared library cache for linker to find .so files
+echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
+ldconfig
```
mariusSincovici, line 29: shouldn't `ldconfig` be also executed at package uninstall?

mariusSincovici (approved):

ci/nfpm/template.yaml
```diff
@@ -124,7 +124,7 @@ contents:
       mode: 0744
 
   - src: ${WORKDIR}/bin/deps/lib/${ARCH}/latest
-    dst: /usr/lib/
+    dst: /usr/lib/${NAME}
```
piotrjurkiewicz, line 127: I don't know how its on Red Hat, but on Debian-based distributions shared libraries should be placed within architecture directory, for example `/usr/lib/x86_64-linux-gnu/nordvpn`.

devzbysiu: Yes, it's a bit of a mess. Usually shared libs should be put in `/usr/lib64/..` or as you mentioned in `/usr/lib/<arch>/..` but this is for system-wide libs. `libtelio.so` etc are not considered system wide and also we currently don't have a mechanism to detect the arch in scriptlets to properly refresh linker cache, so I ended up putting it in `/usr/lib/nordvpn`. I'll create a ticket to improve this

contrib/scriptlets/deb/postinst
devzbysiu: I'll check that :+1:

piotrjurkiewicz: If these libraries are supposed to be private for the package the global linker configuration should not be changed.

devzbysiu: The alternative is rpath, but the issue here is that rpath will be set only on our binaries, not on the dependencies, so for example `libtelio.so` won't find `libsqlite3.so`. I would have to move `libsqlite3.so` somewhere available for the linker risking overriding user's version, or I would have to put it somewhere not commonly available and modify the linker configuration anyway

PR #629, opened 2024-09-25
Fix MFA check

keliramu:


bartoszWojciechO (approved):

daemon/rpc_connect.go
```diff
@@ -53,9 +53,6 @@ func (r *RPC) connect(
 		return internal.ErrNotLoggedIn
 	}
 
-	// get user's current mfa status (here user is logged-in)
-	r.checkMfaStatus()
```
Savolro, line 57: This became a duplicated function now. Calling this will result in publishing MFA status twice. This can easily be replaced with `RenewnigChecker.IsMFAEnabled`

Savolro (changes requested):

PR #632, opened 2024-09-25
Use sudo to execute ldconfig for deb postinst script

devzbysiu:
The whole script is run with Effective User ID 0, but the sudo was required to be used to refresh linker cache in deb postinstall script. It was not needed for rpm package.

### DIFF
--- a/contrib/scriptlets/deb/postinst
+++ b/contrib/scriptlets/deb/postinst
@@ -26,7 +26,8 @@ helpMessage="If you need help using the app, use the command 'nordvpn --help'."
 
 # update configuration and shared library cache for linker to find .so files
 echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
-ldconfig
+# NOTE: sudo is required here even though the script is run with EUID 0
+sudo ldconfig
 
 mkdir -m 0750 -p "$LOG_DIR"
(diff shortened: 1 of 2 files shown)


contrib/scriptlets/deb/postinst
```diff
@@ -26,7 +26,8 @@ helpMessage="If you need help using the app, use the command 'nordvpn --help'."
 
 # update configuration and shared library cache for linker to find .so files
 echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
-ldconfig
+# NOTE: sudo is required here even though the script is run with EUID 0
+sudo ldconfig
```
mariusSincovici, line 30: Shouldn't be executed with `sudo` also in postrm?

devzbysiu: Didn't notice anything wrong on postrm. I can add it "just in case"

Savolro, line 30: What if this is installed on sudoless environment?

devzbysiu: Hmm, good point. So it can be installed on sudoless env only when run with root. I can check if sudo is available and then use it, if it's not available, then it's run with root. I need to check if there are no issues there

devzbysiu:
Closing. This isn't sufficient solution

PR #636, opened 2024-09-30
Provide a way to override dependency versions

bartoszWojciechO:


ci/set_bindings_version.sh
```diff
@@ -0,0 +1,35 @@
...
+
+declare -A LIB_NAME_TO_VERSION
+[[ -n "${LIBTELIO_VERSION:-}" ]] && LIB_NAME_TO_VERSION[libtelio]=$LIBTELIO_VERSION
+[[ -n "${LIBDROP_VERSION:-}" ]] && LIB_NAME_TO_VERSION[libdrop]=$LIBDROP_VERSION
+
+declare -A LIB_NAME_TO_TAG_VERSION
+[[ -n "${LIBTELIO_TAG_VERSION:-}" ]] && LIB_NAME_TO_TAG_VERSION[libtelio]=$LIBTELIO_TAG_VERSION
+[[ -n "${LIBDROP_TAG_VERSION:-}" ]] && LIB_NAME_TO_TAG_VERSION[libdrop]=$LIBDROP_TAG_VERSION
+
+lib_name=$1
```
mariusSincovici, line 19: Maybe check if param exists or add an error message when `lib_name` not found. Or both? :smile:

mariusSincovici (approved):

ci/set_bindings_version.sh
```diff
@@ -0,0 +1,35 @@
...
+    [libtelio]=github.com/NordSecurity/libtelio-go
+    [libdrop]=github.com/NordSecurity/libdrop-go
+)
+
+declare -A LIB_NAME_TO_VERSION
+[[ -n "${LIBTELIO_VERSION:-}" ]] && LIB_NAME_TO_VERSION[libtelio]=$LIBTELIO_VERSION
+[[ -n "${LIBDROP_VERSION:-}" ]] && LIB_NAME_TO_VERSION[libdrop]=$LIBDROP_VERSION
+
+declare -A LIB_NAME_TO_TAG_VERSION
+[[ -n "${LIBTELIO_TAG_VERSION:-}" ]] && LIB_NAME_TO_TAG_VERSION[libtelio]=$LIBTELIO_TAG_VERSION
```
devzbysiu, line 16: Is there a need to have `XYZ_TAG_VERSION`? Can we treat the version like git revision and have one env var and then set it like `full_package_path=$repo_path@$version`? So for libtelio `LIBTELIO_VERSION` will decide if this is a tag or a sha and similarly for libdrop. Would this work?

bartoszWojciechO: `TAG_VERSION` is actually the version of the bindings(I think I'll change it to `BINDINGS_VERSION` because it's misleading). `XYZ_VERSION` determines library version and bindings version if `XYZ_TAG_VERSION` is not set, but there are use cases for setting a different bindings version, so we need a separate variable to control it.

ci/set_bindings_version.sh
bartoszWojciechO: Should already fail thanks to `set -euxo pipefail` I think.

ci/set_bindings_version.sh
devzbysiu: Aah, ok if this is for bindings then it makes sense

devzbysiu (approved):

PR #649, opened 2024-10-16
Always revoke NC token when user logs out

bartoszWojciechO:


daemon/rpc_logout.go
```diff
@@ -99,6 +95,10 @@ func (r *RPC) Logout(ctx context.Context, in *pb.LogoutRequest) (payload *pb.Pay
 		}
 	}
 
+	if r.ncClient.Revoke(internal.IsDevEnv(string(r.environment))) {
```
devzbysiu, line 98: Shouldn't it be `if !r.ncClient.Revoke...`?

keliramu (approved):

PR #652, opened 2024-10-17
Make fileshare monitoring job to react always

devzbysiu:
Changes:
- fileshare monitoring job reacts always - previously it reacted only when the fileshare process state change from running to stopped and from stopped to running, now it will react as long as process is running and a long as process is stopped


devzbysiu:
Can't make the  requests for peers list every 5 seconds. Closing

PR #653, opened 2024-10-17
Improve HeartBeat events

Savolro:
* HeartBeat events are now sent once meshnet setting updates
* `is_on_vpn` value is always updated after connect and disconnect

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -346,8 +346,8 @@ func (s *Subscriber) NotifyUiItemsClick(data events.UiItemsAction) error {
 	))
 }
 
-func (s *Subscriber) NotifyHeartBeat(timePeriodMinutes int) error {
-	return s.response(moose.NordvpnappSendServiceQualityStatusHeartbeat(int32(timePeriodMinutes)))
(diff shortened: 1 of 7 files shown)


events/moose/moose.go
```diff
@@ -610,8 +622,11 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
 			connectionTime, // seconds
 			"",
 			int32(-1),
-		))
+		)); err != nil {
+			return err
+		}
 	}
+	return s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateIsOnVpnValue(false))
```
mariusSincovici, line 629: why is this needed if for connect was not?

mariusSincovici (approved):

events/moose/moose.go
Savolro: I simply followed the logic that this has to be `true` only after regular VPN connection but after disconnect this value always has to be `false`. Yes, it would likely be correct if this was in the regular disconnect branch as it is never set to `true` from meshnet connections but I kept it here to cover more area for recover in case some misconfiguration happened.

bartoszWojciechO (approved):

PR #657, opened 2024-10-17
Display post-quantum status in output of the settings command

bartoszWojciechO:


cli/cli_status.go
```diff
@@ -60,6 +61,9 @@ func Status(resp *pb.StatusResponse) string {
 		b.WriteString(
 			fmt.Sprintf("Current protocol: %s\n", resp.Protocol.String()),
 		)
+		b.WriteString(
+			fmt.Sprintf("Post-quantum VPN: %s\n", nstrings.GetBoolLabel(resp.PostQuantum)),
```
mariusSincovici, line 65: [minor] Should this be printed for OpenVPN, since it cannot be enabled?

mariusSincovici (approved):

keliramu (approved):

cli/cli_status.go
bartoszWojciechO: This is an acctuall PQ status, technically it's still possible to be PQ connected while tech is set to OpenVPN.

1. Connect to PQ server.
2. Disable PQ.
3. Set tech to OpenVPN.
You'll get warrning messages for 2 and 3 but untill you disconnect or re-connect, the connection is still PQ.

PR #658, opened 2024-10-17
Use sudo when checking DNS outside VPN tunnel

Savolro:


devzbysiu (approved):

Savolro:
This will not work as DNS resolving is done by the systemd, therefore it is not affected by fwmark.

PR #663, opened 2024-10-22
Add Post-Quantum docs to manpage

Savolro:


contrib/manual/mantemplate
Savolro, line 1: Note: compatibility with DIP sub-section is still missing

mariusSincovici (approved):

contrib/manual/mantemplate
Savolro: Those are expected to work together.

PR #669, opened 2024-10-25
Fix libtelio/drop override in moose builds

bartoszWojciechO:


ci/compile.sh
```diff
@@ -71,21 +74,19 @@ if [[ $tags == *"moose"* ]]; then
 		-X 'main.EventsDomain=${events_domain:-""}' \
 		-X 'main.EventsSubdomain=${EVENTS_SUBDOMAIN:-""}'"
 
-	# Apply moose patch in case compiling with moose
-	git apply "${WORKDIR}"/contrib/patches/add_moose.diff || \
```
devzbysiu, line 75: This path is also used in `test.sh`, can we remove it from there also?
Then I think we could remove the whole patch from the repo

devzbysiu (approved):

PR #670, opened 2024-10-28
Add flag to fail fast when downloading .so during build

devzbysiu:
Changes:
- add `--fail-with-body` to fail fast when downloading dependencies

### DIFF
--- a/ci/check_dependencies.sh
+++ b/ci/check_dependencies.sh
@@ -43,6 +43,7 @@ function fetch_gitlab_artifact() {
   curl \
     --retry 3 \
     --retry-delay 2 \
+    --fail \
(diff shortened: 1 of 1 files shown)


keliramu (approved):

ci/check_dependencies.sh
```diff
@@ -43,6 +43,7 @@ function fetch_gitlab_artifact() {
   curl \
     --retry 3 \
     --retry-delay 2 \
+    --fail-with-body \
```
mariusSincovici, line 46: Is it needed to save the document at failure? Or this is just for debugging?

mariusSincovici (approved):

PR #676, opened 2024-11-14
Add DNF5 support

keliramu:


test/qa/install.sh
```diff
@@ -94,8 +94,8 @@ install_apt() {
         get_install_opts_for_apt
         install_opts="$RETVAL"
         # Ensure apt is set up to work with https sources
-        $SUDO apt-get "${install_opts}" update
-        $SUDO apt-get "${install_opts}" install apt-transport-https
+        $SUDO apt-get ${install_opts:-} update
```
mariusSincovici, line 97: why is this needed? isn't `install_opts` always set to `""`?

keliramu: if no cli options are given `set -u` option would break script, and if external enclosing double quotes are used - `apt-get` would complain on "" parameter.

devzbysiu (approved):

PR #681, opened 2024-11-18
Notify about account changes

devzbysiu:
Changes:
- mechanism to notify about partial account changes
- notification about update of expiry date for an account
- update of tests


daemon/state/state.go
```diff
@@ -1,46 +1,53 @@
...
 	for _, sub := range s.subscribers {
+		timeout, cancel := context.WithTimeout(context.Background(), notifyTimeout)
+		defer cancel()
 		select {
 		case <-sub.stopChan:
 			close(sub.stateChan)
 		case sub.stateChan <- e:
 			newSubs = append(newSubs, sub)
-		default:
+		case <-timeout.Done():
```
devzbysiu, line 48: Making this event sending mechanism more robust will be addressed in a separate ticket

daemon/events/events.go
```diff
@@ -222,12 +223,14 @@ func (s *SettingsEvents) Publish(cfg config.Config) {
 	s.Ipv6.Publish(cfg.IPv6)
 	s.Technology.Publish(cfg.Technology)
 	s.Obfuscate.Publish(cfg.AutoConnectData.Obfuscate)
-	s.Notify.Publish(!(cfg.UsersData.NotifyOff != nil && len(cfg.UsersData.NotifyOff) > 0))
+	s.Notify.Publish(cfg.UsersData.NotifyOff == nil || len(cfg.UsersData.NotifyOff) <= 0)
```
mariusSincovici, line 226: [small] it might be enough `len(cfg.UsersData.NotifyOff) == 0`

bartoszWojciechO (approved):

PR #690, opened 2024-11-21
Fix OpenVPN 2.6 issues

Savolro:


daemon/vpn/openvpn/openvpn.go
```diff
@@ -225,6 +241,10 @@ func (ovpn *OpenVPN) stop() error {
 	ovpn.manager = nil
 	ovpn.tun = nil
 	ovpn.active = false
+	if ovpn.stopCh != nil {
```
mariusSincovici, line 244: if there is an error before this, then the `ovpn.stopCh` will not be closed.

mariusSincovici (approved):

PR #691, opened 2024-11-21
Prevent binary data in log

keliramu:


events/logger/logger.go
```diff
@@ -118,11 +119,16 @@ func dataRequestAPIToString(
 		b.WriteString(fmt.Sprintf("Error: %s\n", data.Error))
 	}
 	if data.Response != nil {
+		tmpBody := ""
+		// do not print binary data
+		if !slices.Contains(data.Response.Header.Values("Content-Type"), "application/octet-stream") {
```
mariusSincovici, line 124: [small] if we'll still have issues or the responses are not correctly marked as binary, maybe `utf8.validate` can be used.

mariusSincovici (approved):

events/logger/logger.go
```diff
@@ -118,11 +119,16 @@ func dataRequestAPIToString(
 		b.WriteString(fmt.Sprintf("Error: %s\n", data.Error))
 	}
 	if data.Response != nil {
+		tmpBody := ""
```
devzbysiu, line 122: [small] maybe let's set it to `"(binary data)" just to indicate that there is something but it's binary?

devzbysiu (approved):

events/logger/logger.go
keliramu: How expensive is `utf8.validation`? Extra validation/processing wastes resources - is it worth doing it here? How many cases you know with not correctly marked binary in http response?

events/logger/logger.go
keliramu: good idea, yes, I'll add this

PR #695, opened 2024-11-25
feat: idempotent requests

gbaranski:


gbaranski:
Tested it manually, by making it to refresh every time that `nordvpn account` is called, and it did make an idempotent request.

Question is - should I write integration tests for that? The current testing functions don't necessarily allow for custom headers & sending multiple requests to verify that the 2nd one returns the expected header.

gbaranski:
also another concern I have is if that's not going to break the already existing configurations? As this code is changing the structure of TokenData. 
I used `omitempty` and made it optional, so in theory it shouldn't be an issue, but once I ran the code I received this, followed by a daemon crash:
```
2024/11/25 13:39:22 [Info] Daemon has started
2024/11/25 13:39:22 cipher: message authentication failed
2024/11/25 13:39:22 cipher: message authentication failed
```

core/core_test.go
```diff
@@ -127,7 +128,7 @@ func TestDefaultAPI_TokenRenew(t *testing.T) {
 				http.DefaultClient,
 				response.NoopValidator{},
 			)
-			_, err := api.TokenRenew("refresh me")
+			_, err := api.TokenRenew("refresh me", uuid.New())
```
mariusSincovici, line 131: maybe you could create some tests to see that it is added into the requests header?
And that the value remains the same between requests, but changes at logout?

auth/auth.go
```diff
@@ -255,14 +263,15 @@ func (r *RenewingChecker) renew(uid int64, data config.TokenData) error {
...
-	resp, err := r.creds.TokenRenew(data.RenewToken)
+	resp, err := r.creds.TokenRenew(data.RenewToken, *data.IdempotencyKey)
 	if err != nil {
 		return err
 	}
 
 	data.Token = resp.Token
 	data.RenewToken = resp.RenewToken
 	data.TokenExpiry = resp.ExpiresAt
+	data.IdempotencyKey = nil
```
mariusSincovici, line 274: I don't think this is needed.

auth/token_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/daemon/response"
+	"github.com/NordSecurity/nordvpn-linux/request"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockRt struct {
```
keliramu, line 22: Let's have more meaning in names e.g. `mockRt` -> `mockRenewResp`?

keliramu: or `RoundTripper`...

keliramu (approved):

devzbysiu (approved):

core/core_test.go
```diff
@@ -109,11 +110,27 @@ func TestDefaultAPI_CurrentUser(t *testing.T) {
...
 	tests := []testCase{
 		testNewCase(t, http.StatusCreated, TokenRenewURL, "core_token_renew", nil),
 		testNewCase(t, http.StatusBadRequest, TokenRenewURL, "core_token_renew", ErrBadRequest),
 		testNewCase(t, http.StatusNotFound, TokenRenewURL, "core_token_renew", ErrNotFound),
 		testNewCase(t, http.StatusInternalServerError, TokenRenewURL, "", ErrServerInternal),
+
+		{
+			name: "Idempotent request",
+			handler: func(rw http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, idempotencyKey.String(), r.Header.Get("Idempotency-Key"))
```
mariusSincovici, line 123: Couldn't this go into the `testNewCase` and then it is checked for all the requests?

gbaranski: @mariusSincovici 

It looks like `testNewCase()` is being used for other requests as well, which are not necessarily idempotent, thus testing for the existence of idempotency-key won't work.

It'd need a modification in `testNewCase()`, which I think doesn't make sense, and would result in even more code and complexity, than this simple `testCase{}` that is manually created.

auth/token_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (m *mockApi) NotificationCredentials(token string, appUserID string) (core.NotificationCredentialsResponse, error) {
+	return core.NotificationCredentialsResponse{
+		Endpoint: "",
+		Username: "",
+		Password: "",
+	}, nil
+}
+
+func TestTokenRenewWithBadConnection(t *testing.T) {
```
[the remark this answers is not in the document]
gbaranski: didn't use table testing, as it'd be a little bit confusing, as the state should not not reset between the tests.

I've used `t.Run()` for more clarity.

mariusSincovici (approved):

PR #699, opened 2024-11-29
feat: non-blocking libdrop event handling

gbaranski:
~~I saw some ideas to implement a channel, but is it actually necessary?~~

~~The implementation of OnEvent() is using a mutex anyways, so no data races should be happening, and Go-routines seem to be light enough to handle this without overusing resources.~~

### DIFF
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -264,13 +264,8 @@ func (em *EventManager) handleFileProgressEvent(event EventKindFileProgress) {
 		return
 	}
 
-	fmt.Printf("total transferred b4: %d\n", transfer.TotalTransferred)
 	transfer.TotalTransferred += event.Transferred - file.Transferred // add only delta
 	file.Transferred = event.Transferred
-	fmt.Printf("event.Transferred: %d\n", event.Transferred)
-	fmt.Printf("file.Transferred: %d\n", file.Transferred)
-	fmt.Printf("transfer.TotalTransferred: %d\n", transfer.TotalTransferred)
-	fmt.Printf("\n\n\n")
 
 	if progressCh, ok := em.transferSubscriptions[transfer.ID]; ok {
 		var progressPercent uint32
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -115,9 +115,9 @@ func (em *EventManager) process() {
 	}
 }
 
-// sends an event to the event manager
-// the event manager will process it asynchronously and return immediately
+// AsyncEvent sends an event to the event manager in an asynchronous manner
 //
+// This function should return immediately,
 // unless the asyncEvents channel is full, in which case it will block until there is space
 func (em *EventManager) AsyncEvent(event ...Event) {
 	select {
@@ -128,8 +128,7 @@ func (em *EventManager) AsyncEvent(event ...Event) {
 	}
 }
 
-// sends an event to the event manager
-// the event manager will process is synchronously and return after the event is fully processed
+// SyncEvent sends an event to the event manager in a synchronous manner, and waits for the event to be fully processed
 func (em *EventManager) SyncEvent(event ...Event) {
 	em.syncEvents <- event
 	<-em.syncDoneCh
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -110,7 +110,7 @@ func (em *EventManager) Event(event ...Event) {
 	select {
 	case em.events <- event:
 	default:
-		log.Println(internal.WarningPrefix, " async events channel is full. Event() will block until there is space")
+		log.Println(internal.WarningPrefix, "async events channel is full. Event() will block until there is space")
 		em.events <- event
 	}
 }
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -115,6 +115,10 @@ func (em *EventManager) process() {
 	}
 }
 
+// sends an event to the event manager
+// the event manager will process it asynchronously and return immediately
+//
+// unless the asyncEvents channel is full, in which case it will block until there is space
 func (em *EventManager) AsyncEvent(event ...Event) {
 	select {
 	case em.asyncEvents <- event:
@@ -124,6 +128,8 @@ func (em *EventManager) AsyncEvent(event ...Event) {
 	}
 }
 
+// sends an event to the event manager
+// the event manager will process is synchronously and return after the event is fully processed
 func (em *EventManager) SyncEvent(event ...Event) {
 	em.syncEvents <- event
 	<-em.syncDoneCh
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -116,7 +116,12 @@ func (em *EventManager) process() {
 }
 
 func (em *EventManager) AsyncEvent(event ...Event) {
-	em.asyncEvents <- event
+	select {
+	case em.asyncEvents <- event:
+	default:
+		log.Println(internal.WarningPrefix, " async events channel is full. AsyncEvent() will block until there is space")
+		em.asyncEvents <- event
+	}
 }
 
 func (em *EventManager) SyncEvent(event ...Event) {
--- a/fileshare/event_manager.go
+++ b/fileshare/event_manager.go
@@ -57,9 +57,7 @@ type EventManager struct {
 	notificationManager   *NotificationManager
 	defaultDownloadDir    string
 
-	syncEvents  chan []Event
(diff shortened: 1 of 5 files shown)


gbaranski:
I've implemented both `SyncEvent()` and `AsyncEvent()`.
Right now `SyncEvent()` is something that will be used only in tests.
`AsyncEvent()` is the method used in production code.

`OnEvent()` method has been changed to `onEvent()`, in order to prevent data races, which can happen if one uses OnEvent, and the other uses the channel.

fileshare/event_manager.go
```diff
@@ -127,10 +174,7 @@ func (em *EventManager) DisableNotifications() error {
 }
 
 // OnEvent processes events and handles live transfer state.
-func (em *EventManager) OnEvent(event Event) {
-	em.mutex.Lock()
-	defer em.mutex.Unlock()
-
+func (em *EventManager) handleEvent(event Event) {
```
devzbysiu, line 177: Doc comment still contains old name

fileshare/event_manager.go
```diff
@@ -66,17 +70,60 @@ func NewEventManager(
...
+			em.syncDoneCh <- struct{}{}
+		}
 	}
 }
 
+func (em *EventManager) AsyncEvent(event ...Event) {
+	em.asyncEvents <- event
+}
+
+func (em *EventManager) SyncEvent(event ...Event) {
```
devzbysiu, line 122: So this is used only in tests?

fileshare/event_manager.go
```diff
@@ -66,17 +70,60 @@ func NewEventManager(
...
+			em.handleEvent(e)
+		}
+		em.mutex.Unlock()
+	}
+
+	for {
+		select {
+		case e, ok := <-em.asyncEvents:
+			if !ok {
+				log.Println(internal.WarningPrefix + " asyncEvents channel closed")
```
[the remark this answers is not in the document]
gbaranski: Copilot actually suggested me this, and I guess it's because of this: 
https://github.com/NordSecurity/nordvpn-linux/blob/0e0e77eb68695f98469ee6dfe8dc21478d923ebc/fileshare/event_manager.go#L179

Will fix

fileshare/event_manager.go
```diff
@@ -601,7 +659,7 @@ func (em *EventManager) Subscribe(id string) <-chan TransferProgressInfo {
 	em.mutex.Lock()
 	defer em.mutex.Unlock()
 
-	em.transferSubscriptions[id] = make(chan TransferProgressInfo)
+	em.transferSubscriptions[id] = make(chan TransferProgressInfo, 32) // use buffered channels, because we don't want to block the event processing
```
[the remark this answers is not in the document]
gbaranski: well I decided that 32 is big enough to probably cover all the needs.
 
Maybe for further optimizations it'd be cool to actually measure what's the need during a normal file upload.

But considering that before it was an unbuffered channel, anything higher than 1 would be enough.

fileshare/event_manager.go
```diff
@@ -66,17 +70,60 @@ func NewEventManager(
...
 		isProd:                isProd,
 		liveTransfers:         map[string]*LiveTransfer{},
 		transferSubscriptions: map[string]chan TransferProgressInfo{},
 		meshClient:            meshClient,
 		osInfo:                osInfo,
 		filesystem:            filesystem,
 		defaultDownloadDir:    defaultDownloadDir,
+		syncEvents:            make(chan []Event),
+		syncDoneCh:            make(chan struct{}),
+		asyncEvents:           make(chan []Event, 2048),
```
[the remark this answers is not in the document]
gbaranski: that's valid point. I set it to 2048 for tests.
I'll change it to 32, 2048 seems like way too much.

fileshare/event_manager.go
gbaranski: Basically yes, but I don't want to put it to test-only file, because it can be used in production code as well.

fileshare/event_manager.go
```diff
@@ -66,15 +70,69 @@ func NewEventManager(
...
+				log.Println(internal.WarningPrefix, "syncEvents channel closed")
+				return
+			}
+			fn(e)
+			em.syncDoneCh <- struct{}{}
+		}
+	}
+}
+
+// sends an event to the event manager
```
devzbysiu, line 118: Make sure the code is aligned with go doc comments convention - they should start with the name of the symbol being documented

fileshare/event_manager.go
devzbysiu: I think I would go with YAGNI rule here :smile:, because we don't have any signals that synchronous processing will be needed. IMHO it's better to not try to foresee the future, it works in very rare cases

What do you think?

gbaranski: @devzbysiu but then what do you do with the need for having synchronous processing in tests? Maybe I can just make the `SyncEvent()` function private?

gbaranski:
I've removed `SyncEvent` and renamed `AsyncEvent` to `Event`. The tests are now using mockNotifier for synchronizing, and getting know if the event was processed.

fileshare/event_manager_test.go
```diff
@@ -41,12 +41,35 @@ type mockNotification struct {
...
+
+func (mn *mockNotifier) Wait() {
+	timeout := make(chan bool, 1)
+	go func() {
+		time.Sleep(1 * time.Second)
+		timeout <- true
+	}()
+
+	select {
+	case <-timeout:
```
devzbysiu, line 55: Can we use `<-time.After(1 * time.Second)` here? I think you won't need to spawn separate goroutine then

fileshare/event_manager.go
```diff
@@ -206,6 +240,29 @@ func (em *EventManager) handleRequestReceivedEvent(event EventKindRequestReceive
...
+		progress := TransferProgressInfo{
+			TransferID:  transferID,
+			Transferred: transferred,
+			Status:      status,
+		}
+		select {
+		case ch <- progress:
+		default:
+			log.Println(internal.WarningPrefix, " progress channel is full. removing oldest item and sending")
+			<-ch
```
devzbysiu, line 260: So this will drop progress events in case the channel is full. I assume this will happen only when we are too slow with processing events from libdrop right?

gbaranski: yes, that's right. I made an assumption that the progress events are not that important, and dropping an event is not a big deal, because you'll have other 32 events waiting for you anyways.

devzbysiu (approved):

fileshare/events.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 type EventCallback interface {
-	OnEvent(event Event)
+	Event(event ...Event)
```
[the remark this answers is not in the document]
gbaranski: yes, it is. Look at the tests.

fileshare/event_manager_test.go
```diff
@@ -1363,3 +1374,127 @@ func TestAutoaccept(t *testing.T) {
...
+		},
+		{
+			name:       "Event() events should be in order & all received",
+			progChSize: int(numEvents * 2),
+			fn: func(t *testing.T, em *EventManager, progCh chan TransferProgressInfo) {
+				for _, chunk := range chunks {
+					em.Event(chunk...)
+				}
+				lastProgress := uint32(0)
+				for i := uint64(0); i < numEvents-2; i++ {
```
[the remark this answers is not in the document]
gbaranski: -1 because of table counting starts at 0, -1 because I want to go one number higher than the limit.

mariusSincovici (approved):

fileshare/event_manager_test.go
```diff
@@ -1363,3 +1374,132 @@ func TestAutoaccept(t *testing.T) {
...
+			Kind: EventKindFileStarted{
+				TransferId: transferID,
+				FileId:     fileID,
+			},
+		},
+	}
+
+	// subtract by the existing RequestQueued and FileStarted events
+	// to make sure we have the correct number of events
+	// otherwise it's not chunked properly, as the size is not disivable by chunkSize(1026 % 8 = 2)
```
mariusSincovici, line 1417: [spelling] `disivable`
Add an assert into the test if the `events` length is not exactly divisible with `8`.
So if the tests is changed in the future to know sooner why it would fail.

fileshare/event_manager_test.go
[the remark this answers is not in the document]
gbaranski: nevermind, that was not the reason.

it was because the data is chunked improperly.

now it's fixed, and comments were added.

PR #700, opened 2024-11-29
chore: add requirement of glibc >= 2.29

gbaranski:


gbaranski:
```
$ rpm -qpR dist/app/rpm/nordvpn-3.19.2+ccf4335-1.x86_64.rpm

iptables
iproute
procps
ca-certificates
libxml2
libidn2
zlib
(libnl3 or libnl3-200)
glibc >= 2.29
```

gbaranski:
Screenshot from 2024-12-12 10-05-54

I confirmed that it works as intented(fails to install) under Ubuntu 18.04, with glibc 2.27.

gbaranski:
I confirmed that it works as intented(fails to install) under Fedora 28, with glibc 2.27.

Screenshot from 2024-12-12 14-27-40

gbaranski:
When running with snap it looks like Snap is using all the packages from `base22`, so Ubuntu Core 22.04. It means that glibc >=2.29 is included. I ran `ldd --version` and it includes glibc 2.35.

[Line where the base is defined
(https://github.com/NordSecurity/nordvpn-linux/blob/015838037d578fd5e43e2b53c2218a3f9a266d15/snap/snapcraft.yaml#L3-L4)

I confirmed that it works properly(does not crash) on Ubuntu 18.04 with glibc 2.27.

Screenshot from 2024-12-12 14-48-38

keliramu (approved):

devzbysiu (approved):

PR #706, opened 2024-12-05
Use hardware info when generating machine ID

mariusSincovici:


config/machine_id.go
```diff
@@ -0,0 +1,152 @@
...
+	log.Println(internal.ErrorPrefix, "failed to generate machine ID", err)
+	id, err = uuid.NewRandom()
+	if err == nil {
+		return id
+	}
+
+	// Fallback to manually generating a UUID
+	log.Println(internal.ErrorPrefix, "failed to generate random UUID", err)
+	var fallbackUUID uuid.UUID
+	_, err = rand.Read(fallbackUUID[:])
```
keliramu, line 50: What value has `fallbackUUID` variable? as this type is [16]byte - is it initialized with 0 ?

config/machine_id.go
```diff
@@ -0,0 +1,152 @@
...
+func (getter MachineID) calculateMachineID() (uuid.UUID, error) {
+	sourceFiles := []string{
+		"/etc/machine-id",
+		"/var/lib/dbus/machine-id",
+		"/sys/class/dmi/id/product_uuid",
+	}
+
+	hostname, err := getter.hostNameReader()
+	if err != nil {
+		return uuid.UUID{}, fmt.Errorf("failed to read hostname %w", err)
```
keliramu, line 96: Maybe `nil` should be here return 1st param

config/machine_id.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, fileName := range sourceFiles {
+		data, err := getter.fileReader(fileName)
+		if err == nil && len(data) > 0 {
+			machineUUID, err := uuid.Parse(string(data))
+			if err == nil {
+				return uuid.NewSHA1(machineUUID, []byte(hostname)), nil
+			}
+		}
+	}
+	return uuid.UUID{}, fmt.Errorf("failed to get device UUID")
```
keliramu, line 108: Maybe `nil` should be here return 1st param

config/machine_id.go
mariusSincovici: cannot return nil, unless the function return type is changed to `*uuid.UUID`

config/machine_id.go
```diff
@@ -0,0 +1,237 @@
...
+	productUUID   = 1 << 6
+)
+
+type MachineID struct {
+	hostNameReader HostNameReader
+	fileReader     FileReader
+
+	sync.Mutex
+	machineID    *uuid.UUID // store computed device ID
+	usedInfoMask int16      // stores the bitmask to identify how the ID was generate
```
devzbysiu, line 41: [small] typo `generate`

config/machine_id.go
```diff
@@ -0,0 +1,237 @@
...
+func NewMachineID(fileReader FileReader, hostNameReader HostNameReader) *MachineID {
+	return &MachineID{
+		hostNameReader: hostNameReader,
+		fileReader:     fileReader,
+		machineID:      nil,
+		usedInfoMask:   -1,
+	}
+}
+
+func (getter *MachineID) GetMachineID() (ret uuid.UUID) {
```
devzbysiu, line 53: Why named return value here? :thinking:

config/machine_id.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	defer func() {
+		if getter.machineID == nil {
+			getter.machineID = &ret
+			if IsUnderSnap() {
+				getter.usedInfoMask |= isSnapEnv
+			}
+		}
+	}()
```
devzbysiu, line 84: Hmm, isn't this a side effect? Besides retuning machine id, it's also setting the `machineID` field while the method is named `getMachineID`. Then in `GetUsedInformationMask` we have call to `getMachineID`, but it's not assigning result to anything because it relies on the side-effect of setting the internal field:
```
if getter.machineID != nil {
	getter.getMachineID()
}
```

If this is more comfortable and simpler this way, then maybe just a name change will be fine?

config/machine_id.go
```diff
@@ -0,0 +1,237 @@
...
+
+	getter.usedInfoMask = 0
+
+	id, err := getter.generateID()
```
devzbysiu, line 88: [small] Do we want to log the errors here? It may be tricky to figure out that something is wrong if we'll always return the fallback id and we could miss that the original methods are failing

config/machine_id.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (getter *MachineID) readCPUSerial() ([]byte, error) {
+	data, err := getter.fileReader("/proc/cpuinfo")
+	if err != nil {
+		return nil, err
+	}
+
+	if len(data) == 0 {
+		return nil, fmt.Errorf("cpuinfo file is empty")
```
devzbysiu, line 180: [small] no need for `fmt.Errorf`

devzbysiu (approved):

keliramu (approved):

PR #712, opened 2024-12-09
fix: with `io.ReadAll()` use `io.LimitReader()` for DoS prevention

gbaranski:
All occurences of `io.ReadAll()` now use `io.LimitReader()` before passing data to the former.
The limit for the reader has been set in [core.go](https://github.com/NordSecurity/nordvpn-linux/blob/a58ca400e7936c264438efa59413dbd455a85481/core/core.go#L24), after discussions the value has been set to `10MB`.

However, it is with some exceptions:
- [daemon/dns/hosts.go](https://github.com/NordSecurity/nordvpn-linux/blob/a58ca400e7936c264438efa59413dbd455a85481/daemon/dns/hosts.go#L88). Reading `/etc/hosts` should not be a subject to any attacks. As editing the file on it's own requires `root` access.
- [events/logger/logger.go NotifyRequestAPIVerbose()](https://github.com/NordSecurity/nordvpn-linux/blob/a58ca400e7936c264438efa59413dbd455a85481/events/logger/logger.go#L58). Marked as "Do not use in production builds"

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -568,6 +568,14 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
 
 const maxBytesLimit int64 = 1024 * 1024 * 10 // 10MB
 
+type ErrMaxBytesLimit struct {
+	Limit int64
+}
+
+func (err *ErrMaxBytesLimit) Error() string {
+	return fmt.Sprintf("input exceeded the max limit of %d bytes", err.Limit)
+}
+
 // MaxBytesReadAll is a wrapper around io.ReadAll that limits the number of bytes read from the reader.
 //
 // If the reader exceeds the maxBytesLimit, the function returns an error.
--- a/core/core.go
+++ b/core/core.go
@@ -568,7 +568,9 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
 
 const maxBytesLimit int64 = 1024 * 1024 * 10 // 10MB
 
-// re-implementation of io.ReadAll with max limit
+// MaxBytesReadAll is a wrapper around io.ReadAll that limits the number of bytes read from the reader.
+//
+// If the reader exceeds the maxBytesLimit, the function returns an error.
 func MaxBytesReadAll(r io.Reader) ([]byte, error) {
 	limitedReader := &io.LimitedReader{
 		R: r,
--- a/core/core.go
+++ b/core/core.go
@@ -570,26 +570,17 @@ const maxBytesLimit int64 = 1024 * 1024 * 10 // 10MB
 
 // re-implementation of io.ReadAll with max limit
 func MaxBytesReadAll(r io.Reader) ([]byte, error) {
-	b := make([]byte, 0, 512)
-	for {
-		// ---
-		// my modification
-		if int64(len(b)) > maxBytesLimit {
-			return b, fmt.Errorf("input exceeded the max limit")
-		}
-		// ---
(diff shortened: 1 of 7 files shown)


core/core.go
```diff
@@ -565,3 +565,22 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
 
 	return data, nil
 }
+
+const maxBytesLimit int64 = 1024 * 1024 * 10 // 10MB
+
+// re-implementation of io.ReadAll with max limit
+func MaxBytesReadAll(r io.Reader) ([]byte, error) {
+	limitedReader := &io.LimitedReader{
```
devzbysiu, line 573: There is also `io.LimitReader(r io.Reader, n int64)` which wraps reader in `LimitedReader` on it's own

core/core.go
```diff
@@ -565,3 +565,22 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
...
+		R: r,
+		N: maxBytesLimit,
+	}
+	data, err := io.ReadAll(limitedReader)
+	if err != nil {
+		return nil, err
+	}
+	if limitedReader.N == 0 {
```
devzbysiu, line 581: Isn't this off-by-one here? So you can't read 10MB of data now, because you'll have `N` set to 0 and it will fail, right? :thinking:

core/core.go
```diff
@@ -565,3 +565,22 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
...
+		return nil, fmt.Errorf("input exceeded the max limit")
```
devzbysiu, line 582: [small] Maybe let's add the limit value to the error msg?

bartoszWojciechO (approved):

core/core.go
```diff
@@ -565,3 +565,22 @@ func getData[T any](api *DefaultAPI, token string, url string) (T, error) {
 
 	return data, nil
 }
+
+const maxBytesLimit int64 = 1024 * 1024 * 10 // 10MB
+
+// re-implementation of io.ReadAll with max limit
```
devzbysiu, line 571: I think doc comments in go should start with the method name

core/core_test.go
```diff
@@ -224,3 +226,30 @@ func TestDefaultAPI_ServiceCredentials(t *testing.T) {
 		})
 	}
 }
+
+func TestMaxBytes(t *testing.T) {
```
devzbysiu, line 230: I would suggest to add tests for edge conditions. This would detect the off-by-one mentioned in other comment. Generally a lot of bugs happen at the edge conditions that's why it's super useful to have edge conditions tested if possible

core/core.go
gbaranski: that's right, fixed

core/core.go
gbaranski: that's right, there was a bug related to off-by-one, but not there.

I've created tests that prove that this is working now with input length of -1 +0 or +1 in difference to the limit

core/core_test.go
gbaranski: done

devzbysiu (approved):

core/core_test.go
```diff
@@ -224,3 +226,34 @@ func TestDefaultAPI_ServiceCredentials(t *testing.T) {
...
```
mariusSincovici, line 230: Please switch to table testing.

mariusSincovici (changes requested):

mariusSincovici (approved):

PR #713, opened 2024-12-09
Add fileshare process monitoring

devzbysiu:
Changes:
- `JobMonitorFileshareProcess` replaced with `NetlinkProcessMonitor` - based on netlink Process Events Connector API:
  - monitor listens for EXIT and EXEC events from the kernel
  - EXEC event handler verifies that the process created is a fileshare process by getting PID from the event and reading `/proc` VFS to check the process path
  - if it's confirmed that the created process is a fileshare process, then we save its PID and allow fileshare port
  - on EXIT, we compare the PID and if it does match then we are blocking fileshare because it means that fileshare process exited - no matter how (was killed or stopped)
- the monitor is enabled only when meshnet is on


meshnet/monitor.go
devzbysiu, line 1: I'm going to cover this with tests in the meantime

protobuf/meshnet/service_response.proto
```diff
@@ -12,6 +12,8 @@ enum ServiceErrorCode {
 	NOT_LOGGED_IN = 0;
 	API_FAILURE = 1;
 	CONFIG_FAILURE = 2;
+	// Failed to set up fileshare process monitor
+	MONITOR_FAILURE = 3;
```
devzbysiu, line 16: FYI: We'll discuss this on the daily tomorrow

meshnet/monitor.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/vishvananda/netlink"
+)
+
+type (
+	PID     uint32
+	SetupFn func() (MonitorChannels, error)
+)
+
+type MonitorChannels struct {
+	EventCh chan netlink.ProcEvent
```
bartoszWojciechO, line 17: Maybe channels here can be limited to receive only?

meshnet/monitor.go
```diff
@@ -0,0 +1,99 @@
...
+		return MonitorStopper{}, err
+	}
+
+	stopCh := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case ev := <-channels.EventCh:
+				pm.handleProcessEvent(&ev)
+			case <-stopCh:
```
bartoszWojciechO, line 62: Maybe context would be useful in this case? I.e `Start` takes a context instead of returning `MonitorStopper` and passes it to `setup` function.

meshnet/monitor_event_handler.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	return nil
+}
+
+func isFileshareProcess(pid PID) bool {
+	// ignore older processes, fileshare is always
+	// younger than the daemon so it has higher PID
+	if pid < PID(os.Getpid()) {
```
keliramu, line 113: Is this condition always true? e.g. main daemon crashed/killed then restarted and fileshare daemon was running - will it be restarted as well?

meshnet/monitor_event_handler.go
```diff
@@ -0,0 +1,139 @@
...
+	if eh.filesharePID != 0 {
+		// fileshare already started and we noted the PID, no need to
+		// process next events anymore until the PID gets reset in [EventHandler.OnProcessStopped]
+		return
+	}
+	if !isFileshareProcess(ev.PID) {
+		return
+	}
+	eh.filesharePID = ev.PID
+	go eh.allowFileshare()
```
keliramu, line 47: Maybe protection from data races is needed here?

meshnet/monitor_event_handler.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return peers, nil
+}
+
+func (eh *FilesharePortAccessController) OnProcessStopped(ev ProcEvent) {
+	if eh.filesharePID != ev.PID {
+		return
+	}
+	eh.filesharePID = 0
+	go eh.blockFileshare()
```
keliramu, line 90: Maybe protection from data races is needed here?

meshnet/monitor.go
devzbysiu: Good point. But I only changed `ErrCh` to read-only because I'm using others to simulate processing in tests

meshnet/monitor.go
devzbysiu: Yes, it's better with context :+1:, updated

meshnet/monitor_event_handler.go
devzbysiu: When the daemon get's killed, `nordfileshare` also dies. I'll check the crash case to make sure that it isn't graceful shutdown or something similar which causes fileshare to die :+1:

bartoszWojciechO (approved):

meshnet/monitor_event_handler.go
devzbysiu: Added synchronization. It does not have significant impact on the processing time of the events

meshnet/monitor_event_handler.go
```diff
@@ -0,0 +1,161 @@
...
+	if eventHandler.filesharePID != 0 {
+		// fileshare already started and we noted the PID, no need to
+		// process next events anymore until the PID gets reset in [EventHandler.OnProcessStopped]
+		return
+	}
+	if !eventHandler.processChecker.isFileshareProcess(ev.PID) {
+		return
+	}
+	eventHandler.filesharePID = ev.PID
+	go eventHandler.allowFileshare()
```
mariusSincovici, line 51: can't there be a problem if multiple routines are starting in the "same" time? which will be the last, here or the `blockFileshare`?

meshnet/server.go
```diff
@@ -223,6 +227,18 @@ func (s *Server) EnableMeshnet(ctx context.Context, _ *pb.Empty) (*pb.MeshnetRes
 		}, nil
 	}
 
+	monitorCtx, cancelMonitor := context.WithCancel(context.Background())
+	err = s.fileshareProcMonitor.Start(monitorCtx)
```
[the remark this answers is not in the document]
devzbysiu: Good idea :+1: . I moved the starting code to `startFileshareMonitor` and it's cancelling the old monitor if there is any, before starting new one

meshnet/server.go
```diff
@@ -39,21 +39,23 @@ var (
...
+	nameservers          dns.Getter
+	pub                  events.Publisher[error]
+	subjectPeerUpdate    events.Publisher[[]string]
+	daemonEvents         *daemonevents.Events
+	lastPeers            string
+	lastConnectedPeer    string
+	norduser             service.NorduserFileshareClient
+	scheduler            gocron.Scheduler
+	fileshareProcMonitor NetlinkProcessMonitor
+	cancelMonitor        context.CancelFunc
```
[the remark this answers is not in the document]
devzbysiu: Now the action of starting the monitor is also happening in the `StartMeshnet`, not only `EnableMeshnet`. The first one happens when users enable meshnet in the terminal and then for example restart the daemon. The meshnet was already enabled so the daemon calls `StartMeshnet`, so I added the monitor start also there and now it for sure needs synchronization. The start is now in `startFileshareMonitor` and there is also `stopFileshareMonitor`

meshnet/monitor_event_handler.go
devzbysiu: When daemon gets killed, the fileshare process also dies, but for the crash case there is no cleanup, so the process is still running. In such case daemon will be younger after restart. I changed the implementation and now, when the event handler gets created, I'm checking first if the fileshare process is already running, if it is - then we save it's PID and this condition won't be executed because if we already have fileshare PID, then the whole `OnProcessStarted` skips the processing.
If, however the PID is not yet saved, then it means that fileshare was not running, so in such case it will always be younger. I added comment describing this [here](https://github.com/NordSecurity/nordvpn-linux/pull/713/files#diff-f91276f6430f8c75462b7aae0000f9b0d50f1d52777c13eb72fc3cd08456211aR57-R62)

meshnet/monitor_event_handler.go
devzbysiu: Added synchronization. It does not have significant impact on the processing time of the events

meshnet/monitor.go
```diff
@@ -0,0 +1,83 @@
...
+		handler: handler,
+		setup:   setup,
+	}
+}
+
+// Start starts listening for process events using Process Events Connector netlink API.
+//
+// It recreates the source of the events by calling [SetupFn]
+// every time [NetlinkProcessMonitor.Start] is called.
+func (pm *NetlinkProcessMonitor) Start(ctx context.Context) error {
```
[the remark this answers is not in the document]
devzbysiu: Yes, I implemented the check. Also added tests for this mechanism

daemon/routes/netlink/retriever.go
```diff
@@ -163,7 +163,7 @@ func routeCmp(a netlink.Route, b netlink.Route) int {
 // isSubnet returns true if network contains whole range of subnet. Otherwise, returns false.
 func isSubnet(network, subnet *net.IPNet) bool {
 	// nil is treated as an IP with 0 prefix
-	if network == nil {
+	if network == nil || network.IP.Equal(net.IPv4zero) || network.IP.Equal(net.IPv6zero) {
```
devzbysiu, line 166: Changes in `netlink` lib

keliramu (approved):

meshnet/monitor_event_handler.go
```diff
@@ -0,0 +1,85 @@
...
+		return
+	}
+
+	// NOTE: at this point, we can ignore older processes. It's because
+	// we checked above that the [eventHandler.filesharePID] is not set
+	// which means that nordfileshare process was not running at the time
+	// of creation of [FilesharePortAccessController] - constructor checks
+	// if nordfilshare is already running - so we know that nordfileshare
+	// PID will be higher than the daemon PID.
+	if ev.PID < eventHandler.processChecker.CurrentPID() {
```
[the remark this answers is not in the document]
devzbysiu: Good point. Not sure how likely this. I'll measure the performance without this if check, because it will read `/proc` for all new processes then

devzbysiu:
The Process Event Connector mechanism is not working on docker because of process and user namespaces mismatch and it's not reporting errors to user-space. I'm dropping this approach.

PR #723, opened 2024-12-18
Repair register new user command

keliramu:


daemon/rpc_login.go
```diff
@@ -135,14 +141,24 @@ func (r *RPC) loginCommon(customCB customCallbackType) (payload *pb.LoginRespons
 }
 
 // LoginOAuth2 is called when logging in with OAuth2.
-func (r *RPC) LoginOAuth2(in *pb.Empty, srv pb.Daemon_LoginOAuth2Server) error {
+func (r *RPC) LoginOAuth2(in *pb.LoginOAuth2Request, srv pb.Daemon_LoginOAuth2Server) error {
```
mariusSincovici, line 144: This is not sending the login success/fail event, is it sent from somewhere else?

protobuf/daemon/login.proto
```diff
@@ -4,6 +4,21 @@ package pb;
 
 option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
 
+enum LoginType {
+  LoginType_UNKNOWN = 0;
```
mariusSincovici, line 8: [small] can it ever be `UNKNOWN`?

keliramu: it's grpc common/recommended practice to have `None` value e.g. when creating variable of that type to not get assigned some valid value which is not intended.

daemon/rpc_login.go
keliramu: Yet event duration has to be calculated correctly

mariusSincovici (approved):

devzbysiu (approved):

PR #724, opened 2024-12-19
Address flaky update tests

ggediminass:


test/qa/lib/meshnet.py
```diff
@@ -666,3 +667,28 @@ def is_connect_successful(output:str, peer_hostname: str):
...
+    """
+
+    for refresh_count in range(retry):
+        local_peer_list = sh.nordvpn.mesh.peer.list(_tty_out=False)
+        remote_peer_list = ssh_client.exec_command("nordvpn mesh peer list")
+
+        if "Status: connected" in local_peer_list and \
+            "Status: connected" in remote_peer_list:
+            logging.log(f"peer list refresh count: {refresh_count}")
+            break
```
mariusSincovici, line 689: [small] I think if this is replaced with `return` +  `logging.log(f"===` is moved outside of `for`. then you don't need the next `if` :smile:

test/qa/lib/meshnet.py
```diff
@@ -666,3 +667,28 @@ def is_connect_successful(output:str, peer_hostname: str):
...
+    Verifies if local and remote NordVPN mesh peers see each other as connected in peer list.
+
+    Args:
+        ssh_client (ssh.Ssh): SSH client to execute commands on the remote system.
+
+    Raises:
+        pytest.fail: If peers are not connected after `retry` attempts.
+    """
+
+    for refresh_count in range(retry):
```
mariusSincovici, line 682: Shouldn't there be a delay not to have issues with the API?

mariusSincovici (approved):

test/qa/lib/meshnet.py
ggediminass: Good point, I did not think about this. Would you agree, that a delay of a second or two should be enough?

mariusSincovici: yeah, 1-2 seconds should be fine

test/qa/lib/meshnet.py
ggediminass: My preference is to print peer lists only if failure occurs :smile:

devzbysiu (approved):

PR #725, opened 2024-12-20
Deduplicate fileshare code

ggediminass:


test/qa/lib/__init__.py
```diff
@@ -333,3 +333,18 @@ def get_virtual_countries() -> list[str]:
         countries.append(country)
 
     return countries
+
+
+class CommandExecutor:
+    def __init__(self, ssh_client = None):
+        self.ssh_client = ssh_client
+
+    def __call__(self, command: str = "-1"):
```
mariusSincovici, line 342: why `"-1"` :smiley:

test/qa/lib/fileshare.py
```diff
@@ -60,14 +62,13 @@ def create_directory(file_count: int, name_suffix: str = "", parent_dir: str | N
         transfer_paths.append(path.removeprefix("/tmp/"))
         filenames.append(filename)
 
-        disallowed_filesize = ["G", "T", "P", "E", "Z", "Y"]
-        for size in disallowed_filesize:
-            if size in file_size:
-                raise ValueError("Specified file size is too big. Specify either (K)ilobytes or (M)egabytes")
+        # same size files generated with fallocate are not unqiue, so adding random factor
```
mariusSincovici, line 65: [spelling] "unqiue"

test/qa/lib/__init__.py
```diff
@@ -333,3 +333,18 @@ def get_virtual_countries() -> list[str]:
...
+        """
+        Executes `command` locally, if `ssh_client` parameter was not provided to constructor.
+
+        Otherwise, `command` is executed on a remote SSH client.
+        """
+        if self.ssh_client is None:
```
mariusSincovici, line 348: maybe add a check to be sure that the parameter is a string?
```
        if not isinstance(command, str):
            raise TypeError(f"Expected a string, got {type(command).__name__}")
```

test/qa/lib/fileshare.py
```diff
@@ -34,24 +36,24 @@ def create_directory(file_count: int, name_suffix: str = "", parent_dir: str | N
...
+        file_size (int, optional): The size of each file to be created, in megabytes. Defaults to `1`.
     Returns:
         Directory: A Directory object containing:
             - dir_path: Path to the created directory.
             - paths: Full paths to the created files.
             - transfer_paths: File paths with leading directories removed.
             - filenames: Names of the created files.
     """
+
+    exec_command = CommandExecutor(ssh_client) if ssh_client is not None else CommandExecutor()
```
mariusSincovici, line 48: is the `if` needed?

test/qa/lib/fileshare.py
```diff
@@ -34,24 +36,24 @@ def create_directory(file_count: int, name_suffix: str = "", parent_dir: str | N
         name_suffix (str, optional): A suffix to append to the filenames. Defaults to an empty string.
         parent_dir (str | None, optional): The parent directory where the temporary directory will be created.
                                            If None, the system default temporary directory is used. Defaults to None.
-        file_size (str, optional): The size of each file to be created, specified using typical file size notation
-                                   (e.g., "1K", "128M"). Defaults to "1K".
+        file_size (int, optional): The size of each file to be created, in megabytes. Defaults to `1`.
```
mariusSincovici, line 39: Add a check for max size.

test/qa/lib/__init__.py
```diff
@@ -333,3 +333,23 @@ def get_virtual_countries() -> list[str]:
         countries.append(country)
 
     return countries
+
+
+class CommandExecutor:
```
devzbysiu, line 338: Is there a plan to use this across all the code in `qa/lib` or it's just for fileshare?

test/qa/lib/fileshare.py
```diff
@@ -34,24 +37,28 @@ def create_directory(file_count: int, name_suffix: str = "", parent_dir: str | N
...
             - paths: Full paths to the created files.
             - transfer_paths: File paths with leading directories removed.
             - filenames: Names of the created files.
     """
+
+    if file_size > MAX_FILE_SIZE:
+        ex = f"Maximum allowed file size is {MAX_FILE_SIZE} MB"
+        raise ValueError(ex)
+
+    exec_command = CommandExecutor(ssh_client)
```
devzbysiu, line 53: Could we name this something like `ssh_command` or `ssh_exec` or something similar? Now if you want to figure out if it's executed locally or via ssh, you need to find the declaration and check if `ssh_client` was passed in or not

test/qa/lib/fileshare.py
```diff
@@ -331,29 +337,18 @@ def files_from_transfer_exist_in_filesystem(transfer_id: str, dir_list: list[Dir
...
-
-        files_in_transfer = [line.split()[0] for line in sh.nordvpn.fileshare.list(transfer_id, tty_out=False).split("\n") if "downloaded" in line]
-        for file in files_in_transfer:
-            try:
-                hash_util = sh.Command(FILE_HASH_UTILITY)
-                file_hash = hash_util(f"{download_location}/{file}").strip().split()[0]
-                assert any(file_hash in directory.filehashes for directory in dir_list)
-            except RuntimeError:
-                return False
+    exec_command = CommandExecutor(ssh_client) if ssh_client is not None else CommandExecutor()
```
devzbysiu, line 340: This is a good case to check out: we can just do `exec_command = CommandExecutor(ssh_client)` because if `ssh_client` is `None` then we'll get local execution anyway, BUT this may be confusing because now, when reading the code, you'll have to check if `ssh_client` is `None` to make sure if this is local or ssh execution. Or you need to make sure that the name of the var reflects if it's local or ssh execution which is generally not enforced in any way so it may be easy to miss with this design

Maybe it would be easier to read if we would have separate methods to exec locally and via ssh? Something like `executor.local_exec(...)` and `executor.ssh_exec(...)` (didn't spend much time thinking about naming here :smile: ). 

The issue with this approach would be that we would have to fail at runtime when you didn't pass ssh_client to the constructor, but called `executor.ssh_exec(...)` so it's not ideal :thinking:

mariusSincovici (approved):

devzbysiu (approved):

PR #727, opened 2024-12-31
Fix nc uts

bartoszWojciechO:


nc/nc_test.go
```diff
@@ -192,62 +191,81 @@ func TestStartStopNotificationClient(t *testing.T) {
...
+
+		t.Run(test.name, func(t *testing.T) {
+			ctx, cancelFunc := context.WithCancel(context.Background())
+			connectedChan := make(chan interface{})
+			go func() {
+				notificationClient.connect(&mockMqttClient, false, ctx, make(chan<- interface{}), make(chan<- mqtt.Client))
+				connectedChan <- true
+			}()
+
+			cancelFunc()
```
mariusSincovici, line 262: since the context is always canceled, possible before starting the goroutine, will the `connect` code ever be executed or it will stop at the first `select`?

nc/nc_test.go
```diff
@@ -192,62 +191,81 @@ func TestStartStopNotificationClient(t *testing.T) {
...
+
```
mariusSincovici, line 263: maybe add a test to check what happens if the context is not immediately canceled, but after some delay?

nc/nc_test.go
bartoszWojciechO: `connect` code should always be executed full, with different path based on the tests parameters. It will always eventually block in https://github.com/NordSecurity/nordvpn-linux/pull/727/files#diff-3ef8346631bf7e381d05cd8c7af5211833424bae0c8a08f8b79831acf2ff0755R334, as nothing is listening on the `connectedChan`. `cancelFunc` timing might affect the path, from what I saw when running the tests generally context cancellation happens when we are blocked in the final select, but it would be hard to ensure that it happens every time.

mariusSincovici (approved):

devzbysiu (approved):

mariusSincovici (approved):

PR #728, opened 2025-01-02
Nc token revoke

bartoszWojciechO:


daemon/rpc_logout.go
```diff
@@ -99,6 +95,10 @@ func (r *RPC) Logout(ctx context.Context, in *pb.LogoutRequest) (payload *pb.Pay
 		}
 	}
 
+	if !r.ncClient.Revoke(internal.IsDevEnv(string(r.environment))) {
```
mariusSincovici, line 98: The revoke function should always be called with `true` for purge.
Also the logged message maybe it should be `error` instead of `warning`?

mariusSincovici (approved):

devzbysiu (approved):

PR #734, opened 2025-01-08
show actual ip address of the user

gbaranski:
solved by calling the insights API, and displaying the `ip` from the insights data

### DIFF
--- a/core/models.go
+++ b/core/models.go
@@ -554,6 +554,7 @@ type Pivot struct {
 }
 
 type Insights struct {
+	IP          *string `json:"ip"`
 	City        string  `json:"city"`
 	Country     string  `json:"country"`
 	Isp         string  `json:"isp"`
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -242,6 +242,17 @@ func (r *RPC) connect(
 		return nil
 	}
 
+	// update the insights with new data from the API
+	// it's needed so that `rpc_status.go` returns an updated IP address
+	newInsights, err := r.api.Insights()
+	if err != nil || newInsights == nil {
+		log.Println(internal.ErrorPrefix, err)
+	} else {
+		if err := r.dm.SetInsightsData(*newInsights); err != nil {
+			log.Println(internal.ErrorPrefix, err)
+		}
+	}
+
 	// If server has at least one IPv6 address
 	// regardless if IPv4 or IPv6 is used to connect
 	// to the server - DO NOT DISABLE IPv6.
--- a/daemon/rpc_status.go
+++ b/daemon/rpc_status.go
@@ -49,11 +49,19 @@ func (r *RPC) Status(context.Context, *pb.Empty) (*pb.StatusResponse, error) {
 		postQuantum = connectionParameters.PostQuantum
 	}
 
+	insights := r.dm.GetInsightsData()
+	var ip string
+	if insights.Insights.IP != nil {
+		ip = *insights.Insights.IP
+	} else {
+		ip = status.IP.String()
+	}
+
 	return &pb.StatusResponse{
 		State:           string(status.State),
 		Technology:      status.Technology,
 		Protocol:        status.Protocol,
-		Ip:              status.IP.String(),
+		Ip:              ip,
 		Hostname:        status.Hostname,
 		Name:            status.Name,
 		Country:         status.Country,


core/models.go
```diff
@@ -554,6 +554,7 @@ type Pivot struct {
 }
 
 type Insights struct {
+	IP          *string `json:"ip"`
```
keliramu, line 557: Why regular string is no enough here? Why need pointer?

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+		}
+	}
+
+}
+
+func JobActualIP(dm *DataManager, api core.InsightsAPI) func(context.Context, bool) error {
+	return func(ctx context.Context, isConnected bool) error {
+		var newIP netip.Addr
+		defer func() {
```
devzbysiu, line 44: The actual IP will be set to zero value if there was an error in the job, even if the IP was set to something else previously. Is this intentional? I guess it's better to not show the wrong IP if we don't have data on the new IP (or confirmation that the current IP is still correct), so maybe put a comment here that explains it, so it won't be confusing in the future

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+		insights, err := insightsUntilSuccess(ctx, api)
+		if err != nil {
+			return err
+		}
+
+		insightsIP, err := netip.ParseAddr(insights.IP)
+		if err != nil {
+			return err
+		}
+		if insightsIP.IsValid() {
```
devzbysiu, line 61: [small] Is it possible to have false here after successful `ParseAddr`?

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/netip"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/network"
+)
+
+func insightsUntilSuccess(ctx context.Context, api core.InsightsAPI) (core.Insights, error) {
+	for i := 0; ; i++ {
```
devzbysiu, line 15: Is there a reason that this is infinite? Shouldn't we have some max retries?
Also - if the `api.Insights()` will keep failing, we'll spam logs, with the error, indefinitely

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+		select {
```
devzbysiu, line 16: [small] instead of double-select, you could also check `ctx.Err()` at the beginning of the  loop and return if it was cancelled. This way you can have just one select in the loop, maybe it would be easier to follow :shrug:, if not then you can ignore this comment

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+		case <-ctx.Done():
+			return core.Insights{}, ctx.Err()
+		default:
+			insights, err := api.Insights()
+			if err == nil && insights != nil {
+				return *insights, nil
```
devzbysiu, line 22: Is there a need to dereference here? It will copy whole insights struct. Maybe pointer will do? Or maybe we could just return the IP from here?

daemon/jobs.go
```diff
@@ -75,6 +75,44 @@ func (r *RPC) StartJobs(
 		}
 	}
 
+	actualIP := JobActualIP(r.dm, r.api)
+
+	go func() {
+		call := func(ctx context.Context, isConnected bool) {
+			err := actualIP(ctx, isConnected)
+			if err != nil {
+				log.Println(internal.ErrorPrefix, "actual ip job error")
```
devzbysiu, line 84: Can we log the `err` here as well?

daemon/jobs.go
```diff
@@ -75,6 +75,44 @@ func (r *RPC) StartJobs(
...
+			case events.DataConnect:
+			case events.DataDisconnect:
+				if cancel != nil {
+					cancel()
+				}
+
+				var ctx context.Context
+				ctx, cancel = context.WithCancel(context.Background())
+
+				if _, ok := ev.(events.DataConnect); ok {
```
devzbysiu, line 102: So this code has a switch on `events` and processes all cases the same way, but inside the case you are checking event type anyway. Maybe it would be easier to follow with something like:
```
switch ev.(type) {
case events.DataConnect:
  go call(ctx, true)
case events.DataDisconnect:
  call(ctx, false)
}
```
And then handle the cancellation outside of the switch?

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,67 @@
...
+				// Continue to the next retry
+			case <-ctx.Done():
+				return core.Insights{}, ctx.Err() // Exit if context is canceled during sleep
+			}
+		}
+	}
+
+}
+
+func JobActualIP(dm *DataManager, api core.InsightsAPI) func(context.Context, bool) error {
```
devzbysiu, line 41: I see some small inconsistency with the `Job*` name. Other `Job*` functions are actually run by `r.scheduler.NewJob(...)`, but this is not. I was slightly confused why you'll be running infinite loop inside a scheduled job :smile:, but it's not a scheduled job. Maybe would be nicer to have it as a plain function which will be run directly instead of returning function and having `actualIP = JobActualIP(...)` later?

daemon/job_actual_ip_test.go
```diff
@@ -0,0 +1,192 @@
...
+			insightFunc := tt.insightsAPI()
+			api := &mockInsights{
+				insightsFunc: insightFunc,
+			}
+
+			ip, err := insightsIPUntilSuccess(ctx, api, func(i int) time.Duration {
+				return time.Millisecond * 10
+			})
+
+			if (err != nil) != (tt.expectedErr != nil) {
```
mariusSincovici, line 118: `assert.ErrorIs` or something similar?

daemon/job_actual_ip_test.go
```diff
@@ -0,0 +1,192 @@
...
+				t.Fatalf("unexpected error: got %v, want %v", err, tt.expectedErr)
+			}
+
+			if err == nil && ip.String() != tt.expectedIP {
```
mariusSincovici, line 122: same here, if err == nil asser.Equal?

daemon/job_actual_ip.go
```diff
@@ -0,0 +1,135 @@
...
+	result := make(chan Result)
+	for i := 0; ; i++ {
+		if ctx.Err() != nil {
+			return netip.Addr{}, ctx.Err()
+		}
+
+		// this goroutine is used so that this function is immediately stopped once the context is cancelled
+		go func() {
+			ip, err := tryInsightsIP(api)
+			result <- Result{ip, err}
```
mariusSincovici, line 46: if the context is canceled before writing into the channel, this will remain blocked

daemon/actual_ip_resolver.go
```diff
@@ -0,0 +1,131 @@
...
+	if err == nil && insights != nil {
+		ip, err := netip.ParseAddr(insights.IP)
+		if err == nil {
+			return ip, nil
+		} else {
+			return netip.Addr{}, err
+		}
+	} else {
+		return netip.Addr{}, err
+	}
```
devzbysiu, line 29: This can be simplified - use DeMorgan's law to invert the condition and return early, this way you can get rid of nesting. It creates a precondition for what happens next and it's easier to reason about the code, because at the top you see required conditions and then when conditions are met, you see the meat of the function

So instead of
```
if err == nil && insights != nil {
...
} else {
...
}
```
you can have:
```
if err != nil || insights == nil {
  // return early
}
// no else needed - less nesting
```

It can be done also for internal `if` block

daemon/actual_ip_resolver.go
```diff
@@ -0,0 +1,131 @@
...
+			} else {
+				log.Println(internal.ErrorPrefix, "insights ip attempt failed: ", r.error)
+			}
+		case <-ctx.Done():
+			return netip.Addr{}, ctx.Err()
+		}
+
+		select {
+		case <-time.After(backoff(i)): // wait before retrying
+		// PS: there's no fallthrough
```
devzbysiu, line 64: :smile: I get kicked by this from time to time in golang :smile:

daemon/actual_ip_resolver.go
```diff
@@ -0,0 +1,131 @@
...
+		error
+	}
+	result := make(chan Result, 1)
+	for i := 0; ; i++ {
+		if ctx.Err() != nil {
+			return netip.Addr{}, ctx.Err()
+		}
+
+		// this goroutine is used so that this function is immediately stopped once the context is cancelled
+		go func() {
```
devzbysiu, line 44: Do we need to have this in a separate goroutine? :thinking: What if the `tryInsightsIP` is slow? Won't this cause lot of goroutines to be spawned?

daemon/actual_ip_resolver.go
```diff
@@ -0,0 +1,131 @@
...
+			ip, err := tryInsightsIP(api)
+			if ctx.Err() == nil {
```
devzbysiu, line 46: Won't this cause a race? :thinking: if the context gets cancelled after this `if` but before sending data via `result` channel, it's possible that the loop will exit, but you'll still send something to `result` and there is no one left to read it I think. That's probably not an issue though, but I think that's additional reason for not making this check in a separate routine

daemon/actual_ip_resolver.go
gbaranski: no, it won't call `tryInsightsIP` unless the previous call returned.

daemon/actual_ip_resolver.go
gbaranski: yeah...

daemon/actual_ip_resolver.go
gbaranski: i don't see any way to make it to not be in a separate goroutine.
It has to stop when the context is cancelled.
if you send to channel, but no one is listening, simply nothing will happen.

mariusSincovici (approved):

devzbysiu (approved):

PR #741, opened 2025-01-15
Introduce FileSize class

ggediminass:


test/qa/lib/fileshare.py
```diff
@@ -11,6 +11,50 @@
...
+    Attributes:
+        size (int): The size of the file in KB. Default is 1 KB.
+    """
+
+    def __init__(self, size: int = 1):
+        """
+        Initialize the FileSize object.
+
+        Args:
+            size (int, optional): The initial size of the file in KB. Default is 1 KB.
```
keliramu, line 27: Why need default of 1kb? Why not 0?

test/qa/lib/fileshare.py
```diff
@@ -22,13 +66,14 @@
 MSG_HISTORY_CLEARED = "File transfer history cleared."
 MSG_CANCEL_TRANSFER = "File transfer canceled."
 
-DEFAULT_FILE_SIZE = 1
-MAX_FILE_SIZE = 1024
+DEFAULT_FILE_SIZE: FileSize = FileSize(1)
+MAX_FILE_SIZE: FileSize = FileSize()
```
mariusSincovici, line 70: might be easier to use if you have the `from_x` function as class functions and then to use `FileSize.from_mb(1024)` and this returns an `FileSize` object

PR #747, opened 2025-01-20
Add virtual state to connected response

devzbysiu:
Changes:
- bump grpcio to accept generated protobufs
- set manual connection source in rpc_connect - this resets the source after it was set to auto when auto-connection is happening
- add connecton parameters and is_virtual info to the state response
- e2e tests for is_virtual and connection params in grpc stream response


gbaranski (commented):
`isVirtual` has a lot of ambiguity, what about renaming it to `isVirtualLocation`? Like it is in `ServerData`?

bartoszWojciechO (approved):

PR #748, opened 2025-01-22
LVPN-5407: Change notification settings only for members of the NordVPN group

bartoszWojciechO:


daemon/rpc_set_notify.go
```diff
@@ -10,6 +10,19 @@ import (
 )
 
 func (r *RPC) SetNotify(ctx context.Context, in *pb.SetNotifyRequest) (*pb.Payload, error) {
+	isInNordVPNGroup, err := internal.IsInAllowedGroup(uint32(in.Uid))
```
mariusSincovici, line 13: wouldn't it be easier to just remove the Uid from the request and take it from the gRPC connection? So it wouldn't be possible for a user to disable notifications for another user.

mariusSincovici (approved):

keliramu (approved):

PR #758, opened 2025-02-05
Update killswitch rules when network change is detected

bartoszWojciechO:


networker/vpn_test.go
```diff
@@ -135,6 +137,102 @@ func TestRefreshVPN_VPNFailure(t *testing.T) {
 	assert.NotEmpty(t, combined.fw.(*workingFirewall).rules) // We want to keep rules to avoid leaking
 }
 
+func TestRefreshVPN_KillswitchNewInterface(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	// check if nics contains all of the nicNames, return names of all interfaces that were not found
+	containsInterfacesFunc := func(nics []net.Interface, nicNames ...string) []string {
```
devzbysiu, line 144: [small] you could put the `nics` in a set and then just check if the the `nicName` is in this set without the need to iterate over `nics`

devzbysiu (approved):

gbaranski (approved):

networker/vpn.go
```diff
@@ -92,6 +92,17 @@ func (netw *Combined) refreshVPN(ctx context.Context) (err error) {
 	isVPNStarted := netw.isVpnSet
 	isMeshStarted := netw.isMeshnetSet
 
+	if netw.isKillSwitchSet {
+		// reset killswitch to account for new network configuration(new interface)
+		if err := netw.unsetKillSwitch(); err != nil {
```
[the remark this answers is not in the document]
bartoszWojciechO: Do you remember where/which function? Killswitch is added through our `iptables` wrapper, which doesn't support rules update afaik.

networker/vpn_test.go
bartoszWojciechO: Would that require changing code outside of the tests? `iptables` wrapper stores NICs in a list. We could convert the list to a map, but that would still require two loops(one to convert, one to search for the NICs in the map). Unless you meant some other way?

devzbysiu: No, you're right. I didn't see that the rules are always read before calling this helper func, so it would require creating map every time anyway, so that doesn't make much sense :smile:

networker/networker.go
```diff
@@ -1130,11 +1130,11 @@ func (netw *Combined) unsetNetwork() error {
 func (netw *Combined) SetKillSwitch(allowlist config.Allowlist) error {
 	netw.mu.Lock()
 	defer netw.mu.Unlock()
-	return netw.setKillSwitch(allowlist)
+	return netw.setKillSwitch(allowlist, false)
 }
 
-func (netw *Combined) setKillSwitch(allowlist config.Allowlist) error {
-	if !netw.isNetworkSet {
+func (netw *Combined) setKillSwitch(allowlist config.Allowlist, force bool) error {
```
mariusSincovici, line 1136: [small] if rules update works, would there be an issue if are always applied, so `force` is removed?
Or there could be issues with the order?

mariusSincovici (approved):

networker/networker.go
bartoszWojciechO: There are some issues, for example `nordvpn c` -> `nordvpn set killswitch on` would return an error, because killswitch rules are already present. For now it's safer to have to explicitly force the update. Maybe when we have a smarter firewall implementation we could change it.

PR #759, opened 2025-02-06
Block ARP requests to tunnel interface

bartoszWojciechO:


arptables/arptables.go
```diff
@@ -0,0 +1,124 @@
...
+
+type CommandFunc func(cmd string, args ...string) (string, error)
+
+func RunCommand(cmd string, args ...string) (string, error) {
+	output, err := exec.Command(cmd, args...).CombinedOutput()
+	return string(output), err
+}
+
+func isRulePresent(commandFunc CommandFunc, targetRule string, chain string) (bool, error) {
+	currentRules, err := commandFunc("arptables", "-L", chain)
```
keliramu, line 35: Is binary `arptables` available by default? Maybe we need add deb/rpm/snap package dependency?

keliramu: We do have `iptables` dependency added

daemon/vpn/quench/libquench.go
```diff
@@ -204,6 +204,10 @@ func (q *Quench) Start(ctx context.Context, creds vpn.Credentials, server vpn.Se
 	q.tun = tun
 	q.server = server
 
+	if err := q.tun.BlockARP(); err != nil {
```
keliramu, line 207: Maybe this `blockARP` should be done right after `tun.Up`? otherwise there will be some time period when IP could be leaked

daemon/vpn/nordlynx/kernel_space.go
```diff
@@ -112,6 +112,10 @@ func (k *KernelSpace) Start(
 		return err
 	}
 
+	if err := tun.BlockARP(); err != nil {
```
keliramu, line 115: Does it work before `tun.Up`? Maybe it should be common pattern for all: `blockARP` right after `tun.Up`

arptables/arptables.go
```diff
@@ -0,0 +1,124 @@
...
+	args := append([]string{modeArg, nicChainArg}, strings.Split(rule, " ")...)
+	if _, err := commandFunc("arptables", args...); err != nil {
+		return fmt.Errorf("executing arptables command: %w", err)
+	}
+
+	return nil
+}
+
+func UnblockARP(commandFunc CommandFunc, nicName string) error {
+	if err := cmdARPTables(commandFunc, Delete, INPUT, nicName); err != nil {
```
mariusSincovici, line 89: If the user had it configured before connecting to VPN, then the application would delete the rules.

PR #761, opened 2025-02-07
Only log RC version field when new RC was fetched

bartoszWojciechO:


config/remote/firebase.go
```diff
@@ -224,7 +226,11 @@ func (rc *RConfig) getRemoteConfigByVersion(prefix string, stringVersion string)
 	if err != nil {
 		return "", err
 	}
-	log.Println("remote config version field:", versionField)
+
+	// only log new RC version field when it has changed
+	if newRCFetched {
+		log.Println("remote config version field:", versionField)
```
keliramu, line 232: Log prefix would be nice to add

keliramu (approved):

devzbysiu (approved):

PR #763, opened 2025-02-10
Hide set ipv6 command

bartoszWojciechO:


keliramu (approved):

cli/cli.go
```diff
@@ -184,19 +184,20 @@ func NewApp(version, environment, hash, salt string,
...
-				Description: fmt.Sprintf(
-					MsgSetBoolDescription,
-					SetIpv6UsageText,
-					"ipv6",
-					"ipv6",
-				),
-				BashComplete: cmd.SetBoolAutocomplete,
-			},
+			// TODO: uncomment once IPv6 support is enabled in the backend
+			// {
```
mariusSincovici, line 188: Shouldn't it also be removed from the `status` command?

What about `(f *FilesystemConfigManager) load` in case the user has it enabled, to be set to false?

gbaranski (approved):

cli/cli.go
```diff
@@ -184,19 +184,20 @@ func NewApp(version, environment, hash, salt string,
...
-				ArgsUsage: MsgSetBoolArgsUsage,
-				Description: fmt.Sprintf(
-					MsgSetBoolDescription,
-					SetIpv6UsageText,
-					"ipv6",
-					"ipv6",
-				),
-				BashComplete: cmd.SetBoolAutocomplete,
-			},
+			// TODO: uncomment once IPv6 support is enabled in the backend
```
devzbysiu, line 187: Will we ever have the IPv6 support? If I understood correctly, we are getting rid of this at all so maybe we don't need to comment this out and just remove?

devzbysiu (approved):

PR #764, opened 2025-02-10
Fix allowlist port range validation

bartoszWojciechO:


daemon/rpc_set_allowlist.go
```diff
@@ -42,7 +42,7 @@ func arePortsValid(portRangeStart int64, portRangeEnd int64, currentPorts config
 	}
 
 	if portRangeEnd != 0 &&
-		((portRangeStart < internal.AllowlistMinPort || portRangeStart > internal.AllowlistMaxPort) ||
+		((portRangeEnd < internal.AllowlistMinPort || portRangeEnd > internal.AllowlistMaxPort) ||
 			portRangeEnd < portRangeStart) {
```
keliramu, line 46: Last part of this `if` should be inverted, because now is `portRangeEnd < portRangeStart` which is OK, but this `if` checks for NOT-OK situations

bartoszWojciechO: `portRangeEnd` in this case is the upper bound, for example in `[1 50]` range 1 is `portRangeStart` and 50 is `portRangeEnd`. So in this case we want `portRangeEnd` to be greater than `portRangeStart`.

keliramu (approved):

devzbysiu (approved):

PR #766, opened 2025-02-11
Include specialty servers in servers list respose

mariusSincovici:

* Include Double VPN and Onion over VPN servers into the list.
* To prevent possible race condition on the members of ServersData, when data is changed from DataManager the entire object is replaced instead of only the members.

### DIFF
--- a/daemon/rpc_servers_test.go
+++ b/daemon/rpc_servers_test.go
@@ -446,7 +446,7 @@ func TestServers(t *testing.T) {
 		},
 		{
 			name:        "failure because of filter error",
-			serversList: core.Servers{}, // servers will return an error because it will fails to find available servers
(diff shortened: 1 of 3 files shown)


daemon/rpc_servers_test.go
```diff
@@ -446,7 +446,7 @@ func TestServers(t *testing.T) {
 		},
 		{
 			name:        "failure because of filter error",
-			serversList: core.Servers{}, // servers will retrun an error because it will fails to find available servers
+			serversList: core.Servers{}, // servers will return an error because it will fails to find available servers
```
keliramu, line 449: [small] `will fails` -> just `fails` or `will fail`

keliramu (approved):

bartoszWojciechO (approved):

devzbysiu (approved):

PR #768, opened 2025-02-11
LVPN-6714: Remove mesh private key once its not in use

bartoszWojciechO:


cmd/daemon/main.go
```diff
@@ -109,6 +109,25 @@ const (
...
+	}
+
+	if !cfg.Mesh && cfg.MeshPrivateKey != "" {
+		err := configManager.SaveWith(func(c config.Config) config.Config {
+			c.MeshPrivateKey = ""
+			return c
+		})
+		if err != nil {
+			log.Println(internal.ErrorPrefix, "failed to clean up mesh private key on shutdown:", err)
+		}
```
devzbysiu, line 127: Would it make sense to move this to separate method? It's almost the same as the changes in `rpc_disconnect.go`, only log message differs

cmd/daemon/main.go
devzbysiu, line 1: Would it be possible to have e2e tests for those changes?

daemon/rpc_disconnect.go
```diff
@@ -9,6 +10,20 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
+func MeshPrivateKeyCleanup(cfg config.Config, configManager config.Manager) error {
```
keliramu, line 13: Not an exported fn

keliramu: sorry, it is

daemon/rpc_disconnect.go
```diff
@@ -9,6 +10,20 @@ import (
...
+	if !cfg.Mesh && cfg.MeshPrivateKey != "" {
```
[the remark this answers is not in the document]
keliramu: Or this conditions guards from shutdown with Mesh=On and then restart correctly with Mesh=On?

test/qa/test_meshnet_other.py
```diff
@@ -1,8 +1,17 @@
...
 import sh
+import grpc
 
 import lib
-from lib import daemon, meshnet, settings, ssh
+import sys
+import os
+from lib import daemon, meshnet, settings, ssh, logging
+
+sys.path.append(os.path.abspath(os.path.join(
```
keliramu, line 10: is it needed to have this code in the `import` section?

keliramu (approved):

daemon/rpc_disconnect.go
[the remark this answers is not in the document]
bartoszWojciechO: Yeah, the key is regenerated when user enables meshnet. If we clean it up without this condition, then meshnet would not function properly after startup.

test/qa/test_meshnet_other.py
bartoszWojciechO: Yes, because we need to have this directory in path before `lib.protobuf.meshnet` imports.

cmd/daemon/main.go
```diff
@@ -619,4 +619,10 @@ func main() {
 	if err := rpc.StopKillSwitch(); err != nil {
 		log.Println(internal.ErrorPrefix, "stopping KillSwitch:", err)
 	}
+
+	if err := fsystem.Load(&cfg); err != nil {
```
mariusSincovici, line 623: If the key must exist only while the application is running, does it make sense to save it in files? :thinking: 
Because eventually it just needs more work to clean it.

meshnet/register.go
```diff
@@ -1,34 +1,40 @@
...
+	ClearMeshPrivateKey()
+}
+
 // Checker provides information about meshnet.
 type Checker interface {
 	// IsRegistrationInfoCorrect returns true when device has been registered to meshnet.
 	IsRegistrationInfoCorrect() bool
 	// Register the device
 	Register() error
+	// GetMeshPrivateKey returns returns meshnet private key and a boolean value indicating if it has been generated
```
mariusSincovici, line 26: `returns` 2 times

meshnet/register.go
```diff
@@ -91,58 +97,55 @@ func (r *RegisteringChecker) Register() error {
...
-		peer, err = r.reg.Register(token, cmesh.Machine{
-			HardwareID:      cfg.MachineID,
-			PublicKey:       r.gen.Public(privateKey),
-			OS:              cmesh.OperatingSystem{Name: "linux", Distro: distroName},
-			SupportsRouting: true,
-		})
-	}
 	if err != nil {
-		return err
+		return fmt.Errorf("registering new mesh machine: %w", err)
```
mariusSincovici, line 125: in case of a failure shouldn't `r.meshPrivateKey` be cleared?
or would it work to move `r.meshPrivateKey = r.gen.Private()` ad the end of the function?

meshnet/server.go
```diff
@@ -171,10 +172,24 @@ func (s *Server) EnableMeshnet(ctx context.Context, _ *pb.Empty) (*pb.MeshnetRes
 		}, nil
 	}
 
+	meshPK, ok := s.mc.GetMeshPrivateKey()
+	if !ok {
+		log.Println(internal.ErrorPrefix, "Meshnet private key not stored for the enable operation.")
```
mariusSincovici, line 177: this file uses the publisher `s.pub.Publish`

mariusSincovici (commented):
cleanup for set defaults is missing

test/qa/test_meshnet_other.py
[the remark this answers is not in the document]
bartoszWojciechO: Apparently python looks in `qa/lib` but not in `qa/lib/<subdirectory>`. So we need to add it manually. Same approach is used for other grpc tests.

meshnet/server.go
```diff
@@ -171,10 +171,24 @@ func (s *Server) EnableMeshnet(ctx context.Context, _ *pb.Empty) (*pb.MeshnetRes
...
+		return &pb.MeshnetResponse{
+			Response: &pb.MeshnetResponse_MeshnetError{
+				MeshnetError: pb.MeshnetErrorCode_NOT_REGISTERED,
+			},
+		}, nil
+	}
+
+	// There is a delay in the backend between registering a new key and when that key is recognized, so we need to wait
+	// some time, otherwise connection will fail.
+	time.Sleep(5 * time.Second)
```
devzbysiu, line 186: Is there a way to avoid this sleep? I'm afraid this may add to our flakiness in tests

test/qa/test_meshnet_other.py
```diff
@@ -122,3 +131,50 @@ def test_set_meshnet_on_post_quantum_enabled():
...
+    sh.nordvpn.connect()
+    sh.nordvpn.set.meshnet("off")
+
+    with grpc.insecure_channel(lib.NORDVPND_SOCKET) as channel:
+        stub = service_pb2_grpc.MeshnetStub(channel)
+        response = stub.GetPrivateKey(empty_pb2.Empty())
+        assert response.private_key != "", "Meshnet private key should not be removed mid-VPN connection."
+
+    daemon.restart()
+    def is_meshnet_pk_removed() -> bool:
```
devzbysiu, line 166: Maybe this could be reused above and then assert on the bool result?

meshnet/server.go
bartoszWojciechO: There is no way to detect if servers started to recognize if key is recognized by the servers afaik. Maybe we could use some retry mechanism, but we don't get any information from the connect failure, just a long timeout, so it would probably be worse than the sleep.

devzbysiu: Yeah, that's what I was afraid of. Do you know if other apps are using sleeps also?

meshnet/register.go
```diff
@@ -91,58 +97,56 @@ func (r *RegisteringChecker) Register() error {
...
 func (r *RegisteringChecker) register(cfg *config.Config) error {
-	privateKey := cfg.MeshPrivateKey
-	if privateKey == "" {
-		privateKey = r.gen.Private()
-	}
+	r.meshPrivateKey = r.gen.Private()
 	token := cfg.TokensData[cfg.AutoConnectData.ID].Token
 	distroName, err := distro.ReleaseName()
 	if err != nil {
 		return err
```
mariusSincovici, line 116: here too? `r.meshPrivateKey = ""`

mariusSincovici (approved):

PR #770, opened 2025-02-11
Notify about NordWhisper connection state changes

bartoszWojciechO:


daemon/vpn/quench/libquench.go
```diff
@@ -46,14 +47,26 @@ type Observer struct {
 	currentState              vpn.State
```
keliramu, line 47: Only now found that `Observer` is not used outside - should not be public/exported.
Also would very nice to have some comments about the true purpose of this struct and related logic.

keliramu: Or does it implement some interface? Should be described in comments

daemon/vpn/quench/libquench.go
```diff
@@ -41,20 +42,32 @@ func (l *Logger) Log(logLevel quenchBindigns.LogLevel, message string) {
...
-		eventsChan: nil,
+func newObserver(eventNotifier *vpn.Events) *observer {
+	return &observer{
+		eventsChan:    nil,
+		eventNotifier: eventNotifier,
 	}
 }
 
-func (o *Observer) SubscribeToEvents(ctx context.Context) <-chan vpn.State {
+func (o *observer) SetServerData(server vpn.ServerData) {
```
keliramu, line 63: Should not be public/exported?

daemon/vpn/quench/libquench.go
```diff
@@ -41,20 +42,32 @@ func (l *Logger) Log(logLevel quenchBindigns.LogLevel, message string) {
...
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	o.currentServer = server
+}
+
+func (o *observer) SubscribeToEvents(ctx context.Context) <-chan vpn.State {
```
keliramu, line 70: Should not be public/exported?

daemon/vpn/quench/libquench.go
bartoszWojciechO: I think we have this approach in some other places. `observer` but it's thread safe methods are still UpperCase and non-thread safe methods are lowerCase.

Savolro: In this particular case it does not matter that much as `observer` type is unexported. In general, it's good practice to not over-expose things regardless if method is thread safe or not

Savolro (approved):

devzbysiu (approved):

mariusSincovici (approved):

PR #772, opened 2025-02-14
Cleanup meshnet gRPC API

Savolro:
Combining common status codes coming from gRPC (ServiceError, MeshnetError) into a general meshnet error allows preparation code reuse for most of peer modifying endpoints.


protobuf/meshnet/peer.proto
```diff
@@ -57,6 +56,23 @@ message UpdatePeerRequest {
 	string identifier = 1;
 }
 
+
+// Error defines a generic meshnet error that could be returned by most meshnet endpoints
+message Error {
```
mariusSincovici, line 61: [small] Wouldn't be better to be MeshnetError?

mariusSincovici (approved):

protobuf/meshnet/peer.proto
```diff
@@ -167,21 +179,17 @@ enum DisableAutomaticFileshareErrorCode {
 message AllowRoutingResponse {
 	oneof response {
 		Empty empty = 1;
-		UpdatePeerErrorCode update_peer_error_code = 2;
 		AllowRoutingErrorCode allow_routing_error_code = 3;
-		ServiceErrorCode service_error_code = 4;
-		MeshnetErrorCode meshnet_error_code = 5;
+		UpdatePeerError update_peer_error = 6;
```
devzbysiu, line 183: Is the gap with field numbers intentional here?

protobuf/meshnet/peer.proto
```diff
@@ -57,6 +56,23 @@ message UpdatePeerRequest {
...
+
+// Error defines a generic meshnet error that could be returned by most meshnet endpoints
+message Error {
+	oneof error {
+		ServiceErrorCode service_error_code = 1;
+		MeshnetErrorCode meshnet_error_code = 2;
+	}
+}
+
+// npdatePeerError can be either generic meshnet error or generic peer update error
```
devzbysiu, line 68: typo `npdate`

protobuf/meshnet/peer.proto
Savolro: Yes, this is for client and server schema mismatches considering no extra measures are done. E. g. client has old version of schema but server has a new one, will try to find type with id `6`, will not be able to do it will produce an appropriate error. Otherwise, if this was `4`, it would try to parse the given data to `ServiceErrorCode` and would produce error that it was unable to do so.

PR #775, opened 2025-02-17
Fix incorrectly set connection parameters

devzbysiu:
Changes:
- fixes connection parameters incorrectly set to auto-connect data

### DIFF
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -183,13 +183,8 @@ func (r *RPC) connect(
 	event.TargetServerIP = subnet.Addr().String()
 	event.DurationMs = max(int(time.Since(connectingStartTime).Milliseconds()), 1)
 
-	r.ConnectionParameters.SetConnectionParameters(
-		pb.ConnectionSource_MANUAL,
-		ServerParameters{
-			Country: cfg.AutoConnectData.Country,
-			City:    cfg.AutoConnectData.City,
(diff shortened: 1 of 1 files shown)


keliramu (approved):

daemon/rpc_connect.go
```diff
@@ -183,12 +183,21 @@ func (r *RPC) connect(
...
+		)
+	}
+
 	r.ConnectionParameters.SetConnectionParameters(
 		pb.ConnectionSource_MANUAL,
 		ServerParameters{
-			Country: cfg.AutoConnectData.Country,
-			City:    cfg.AutoConnectData.City,
-			Group:   cfg.AutoConnectData.Group,
+			Country: country.Name,
```
mariusSincovici, line 198: I think this is not correct, it will set the server info, but maybe the user just enter the country name or just the group. :thinking: 
Shouldn't they be [like](https://github.com/NordSecurity/nordvpn-linux/commit/a2583a849f2662464b7698275d49c2306b879676#diff-3da421935f58cd839392b57009a4e6e072b980515cce4abb39fb0de8abb1a6c4R211) .
The params are detected from the input.

devzbysiu: Yes, you're right, totally forgot that it was already done and removed by mistake

mariusSincovici (approved):

PR #786, opened 2025-02-28
fix: for Incorrect conversion between integer types #2

odaysec:
## Ticket 🎟️ #785 

To fix the problem, we need to ensure that the integer value parsed from the string does not exceed the bounds of the `uint32` type. This can be achieved by using `strconv.ParseUint` with a specified bit size of 32, which directly parses the string into a `uint32` value. This approach avoids the need for additional bounds checking and ensures that the value is within the valid range for a `uint32`.

### DIFF
--- a/cmd/norduser/main.go
+++ b/cmd/norduser/main.go
@@ -257,11 +257,12 @@ func startSnap() {
 		log.Println(internal.ErrorPrefix, "Failed to add autostart:", err)
 	}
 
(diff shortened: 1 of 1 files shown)


mariusSincovici:
Since there is no response for this PR to address the issues, I'll close it.
Feel free to reopen it and also please sign also the CLA clause.
Thanks

PR #791, opened 2025-03-03
LVPN-7106: Fix allowlist firewall tests

edvyno1:
Removing nord dns entries from firewall as they don't impact firewall corectness, only fail tests when the tests check firewall rules before the dns entry happens to be deleted from the firewall.

### DIFF
--- a/test/qa/lib/firewall.py
+++ b/test/qa/lib/firewall.py
@@ -3,7 +3,7 @@ import re
 
 import sh
 
-from . import Port, Protocol, daemon, logging
+from . import Port, Protocol, daemon, logging, dns
 
 IP_ROUTE_TABLE = 205
 
@@ -121,14 +121,14 @@ def __rules_connmark_chain_input(interface: str):
         ]
 
 
-def __rules_connmark_chain_forward(interface: str):
+def __rules_block_forwarding(interface: str):
     return \
         [
             f"-A FORWARD -o {interface} -m comment --comment nordvpn -j DROP",
         ]
 
 
-def __rules_connmark_chain_output(interface: str):
+def __rules_block_dns_port():
     return \
         [
             "-A OUTPUT -d 169.254.0.0/16 -p tcp -m tcp --dport 53 -m comment --comment nordvpn -j DROP",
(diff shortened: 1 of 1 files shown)


test/qa/lib/firewall.py
```diff
@@ -377,7 +360,13 @@ def _get_iptables_rules() -> list[str]:
     # TODO: add full ipv6 support, separate task #LVPN-3684
     print("Using iptables")
     fw_lines = os.popen("sudo iptables -S").read()
-    return fw_lines.split('\n')[3:-1]
+    fw_list = fw_lines.split('\n')[3:-1]
+    fw_list_duped = fw_list.copy()
+    dns_full = dns.DNS_NORD + dns.DNS_TPL
+    for rule in fw_list_duped:
```
mariusSincovici, line 366: (suggestion)  `[rule for rule in fw_list if not any(dns in rule for dns in dns_full)]`?

test/qa/lib/firewall.py
```diff
@@ -344,13 +325,16 @@ def is_active(ports: list[Port] | None = None, subnets: list[str] | None = None)
     print(sh.ip.route())
 
     expected_rules = _get_firewall_rules(ports, subnets)
+    expected_rules = list(set(expected_rules))
```
mariusSincovici, line 328: are there any duplicates returned by `_get_firewall_rules`?

test/qa/lib/firewall.py
```diff
@@ -228,6 +199,13 @@ def __rules_allowlist_port_chain_input(interface: str, ports_udp: list[Port], po
 
     return result
 
+def __rules_allowlist_port_chain_forward(interface: str):
```
mariusSincovici, line 202: Might be better `__rules_block_forwarding`  or similar because this is blocking the forward on the interface and is not related to the whitelisted ports.

mariusSincovici (approved):

test/qa/lib/firewall.py
edvyno1: fixed

test/qa/lib/firewall.py
edvyno1: applied

test/qa/lib/firewall.py
edvyno1: removed duplication and added `__rules_block_dns_port` func

devzbysiu (approved):

PR #799, opened 2025-03-07
LVPN-6130: Report full connection status in state events

bartoszWojciechO:


daemon/rpc_state.go
```diff
@@ -109,11 +113,14 @@ func statusStream(stateChan <-chan interface{},
...
 				if err := srv.Send(
 					&pb.AppState{State: &pb.AppState_ConnectionStatus{
-						ConnectionStatus: &pb.ConnectionStatus{
-							State:  pb.ConnectionState_DISCONNECTED,
-							ByUser: e.ByUser,
-						},
-					}}); err != nil {
+						ConnectionStatus: &pb.StatusResponse{
+							State:      pb.ConnectionState_DISCONNECTED,
+							ByUser:     true,
```
devzbysiu, line 118: I think this should be set to `e.ByUser`. For example in `libtelio.go` we can have the disconnected event caused by node disconnection or manual disconnection and `ByUser` is true only for manual disconnection

daemon/vpn/quench/libquench.go
```diff
@@ -117,7 +130,7 @@ func (o *observer) Connecting() {
 	}
 
 	o.notifyConnectionStateChange(vpn.ConnectingState)
-	o.eventNotifier.Connected.Publish(getConnectEvent(events.StatusAttempt, o.currentServer))
+	o.eventNotifier.Connected.Publish(getConnectEvent(events.StatusAttempt, o.currentServer, o.nicName))
```
devzbysiu, line 133: [small] `getConnectEvent` could be a method of the `observer`, this way you won't need to pass `o.currentServer` and `o.nicName` to it as an arguments, similarly as you did in `openvpn.go`

events/events.go
```diff
@@ -84,6 +84,8 @@ type DataConnect struct {
 	TargetServerName           string
 	Error                      error
 	IsVirtualLocation          bool
+	Upload                     uint64
+	Download                   uint64
```
devzbysiu, line 88: Why was this added here and in `DataDisconnect` now? Is this part of some analytics task? :thinking:

protobuf/daemon/status.proto
```diff
@@ -14,6 +14,12 @@ enum ConnectionSource {
   AUTO = 2;
 }
 
+enum ConnectionState {
+    DISCONNECTED = 0;
```
keliramu, line 18: gRPC best practices recommend having first enum value as neutral value for initialization e.g. UNKNOWN=0.
Is current 1st value really ok to be default?

devzbysiu (approved):

keliramu (approved):

PR #814, opened 2025-03-20
LVPN-7483: Store previous internal library versions

Savolro:
As artifacts from internal repositories are quite heavy, this behaviour is added to speed up local builds when switching application or library versions.


daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -34,6 +34,7 @@ const (
 )
 
 type lib interface {
+	// ?
```
mariusSincovici, line 37: Does it need a fix?

ci/nfpm/template.yaml
```diff
@@ -50,6 +50,7 @@ depends:
   - zlib1g
   - libnl-genl-3-200
   - libc6 (>= 2.29)
+  - libsqlite3-0
```
mariusSincovici, line 53: Is this needed, Isn't this part of the package?

ci/build_rust.sh
```diff
@@ -27,16 +30,19 @@ function clone_if_absent() {
...
-
-    pushd "${repo_name}"
-    git checkout "${version}"
-    popd
-
-    popd
   fi
+
+  pushd "${repo_name}"
+  # In case checkout does not work, try to fetch the desired version and check it out again
```
bartoszWojciechO, line 39: Why can't we just always fetch the tag? Fro simplicity and in edge case it changes on remote(though for lib version that's not so probable).

Savolro: This would do a remote call on every build. Yes, edge case of git ref updated on the remote will remain. However, adding 2-3 seconds on local builds when in most cases nothing will change is not worth it.

daemon/vpn/nordlynx/libtelio/cgo_internal.go
```diff
@@ -2,10 +2,10 @@
 
 package libtelio
 
-// #cgo amd64 LDFLAGS: -L${SRCDIR}/../../../../bin/deps/lib/amd64/latest -ltelio -lsqlite3
-// #cgo 386 LDFLAGS: -L${SRCDIR}/../../../../bin/deps/lib/i386/latest -ltelio -lsqlite3
-// #cgo arm LDFLAGS: -L${SRCDIR}/../../../../bin/deps/lib/armel/latest -ltelio -lsqlite3
-// #cgo arm LDFLAGS: -L${SRCDIR}/../../../../bin/deps/lib/armhf/latest -ltelio -lsqlite3
-// #cgo arm64 LDFLAGS: -L${SRCDIR}/../../../../bin/deps/lib/aarch64/latest -ltelio -lsqlite3
-// #cgo LDFLAGS: -ldl -lm
+// #cgo amd64 LDFLAGS: -L${SRCDIR}/../..//../../bin/deps/lib/libmoose-nordvpnapp/current/amd64
```
PostDen, line 5: why libmoose mention is here?

Savolro: This is used only when `internal` tag is provided. `moose` always contains `libsqlite.so` and is always included in internal builds, therefore it is the place to take `libsqlite3.so` from.

magefiles/docker.go
```diff
@@ -108,20 +108,40 @@ func runDocker(
...
+			Target: "/lib/modules",
 		},
+	}
+
+	if env["MOUNT_HOST_GOMODCACHE"] == "1" {
+		// `GOMODCACHE` is not accessible from go code directly during build, therefore it
+		// has to be acquired in a separate process.
+		out, err := exec.Command("go", "env", "GOMODCACHE").CombinedOutput()
+		if err != nil {
+			fmt.Println("Error while executing 'go env GOMODCACHE':", err)
```
devzbysiu, line 129: Does it make sense to continue here in case of an error? Similarly in the line 133. If we have issues with getting `GOMODCACHE` then why still add the mount?

devzbysiu (approved):

PR #819, opened 2025-03-26
LVPN-7376: Address flaky fileshare tests

ggediminass:


test/qa/test_fileshare.py
```diff
@@ -47,10 +47,17 @@ def setup_module(module):  # noqa: ARG001
     ssh_client.exec_command("nordvpn set notify off")
     ssh_client.exec_command("nordvpn set mesh on")
 
-    ssh_client.exec_command("nordvpn mesh peer refresh")
-    sh.nordvpn.mesh.peer.refresh()
-    peer = meshnet.PeerList.from_str(sh.nordvpn.mesh.peer.list()).get_internal_peer()
-    assert meshnet.is_peer_reachable(ssh_client, peer)
+    for _ in range(3):
```
mariusSincovici, line 50: There should be some sleep between refreshes, at least 1 second.

mariusSincovici: Or maybe the test should be blocked here until the peers have connected status. If it is not happening until test timeout then the test can fail because of this. Because the `3` can introduce flaky and maybe it works after `5` retries :smile: .

test/qa/lib/meshnet.py
```diff
@@ -656,30 +656,36 @@ def get_clean_peer_list(peer_list: str):
...
-            time.sleep(1)
-            i += 1
-    print(sh_no_tty.nordvpn.mesh.peer.list())
-    output = ssh_client.exec_command("nordvpn mesh peer list")
-    print(output)
+    if ssh_client is None:
+        i = 0
+        while i < retry:
+            try:
+                return "icmp_seq=" in sh.ping("-c", "1", peer_hostname)
```
mariusSincovici, line 673: I don't think it is ok to use pings to test. There can be issues with them. You can use the same command as in `is_internet_reachable` but change the ports the same as for ssh client, bellow.
________
for ssh client can be used `nc -zv -w 1 IP 22`  if it exists on the image.
Or with ssh:`ssh -o BatchMode=yes -o ConnectTimeout=1 IP exit`.
when it is not able to connect returns `ssh: connect to host IP port 22: Connection timed out` . Otherwise success or `Host key verification failed.`(which is also ok)

test/qa/lib/meshnet.py
```diff
@@ -656,30 +656,30 @@ def get_clean_peer_list(peer_list: str):
...
-    print(sh_no_tty.nordvpn.mesh.peer.list())
-    output = ssh_client.exec_command("nordvpn mesh peer list")
-    print(output)
+    if ssh_client is None:
+        return network.is_internet_reachable(peer_hostname, 22)
+    else:  # noqa: RET505
+        i = 0
+        while i < retry:
+            try:
+                ssh_client.exec_command(f"ssh -o BatchMode=yes -o ConnectTimeout=1 {peer_hostname}")
```
bartoszWojciechO, line 675: Why not ping?

bartoszWojciechO (approved):

bartoszWojciechO (approved):

PR #822, opened 2025-03-31
LVPN-7513: Cache account data

bartoszWojciechO:


daemon/data_models.go
```diff
@@ -153,3 +157,45 @@ func (data *VersionData) save() error {
 	}
 	return nil
 }
+
+func isCacheValid(added time.Time) bool {
```
keliramu, line 161: Naming: it is about Account Data Cache - but you cannot tell it from this name, IMO, should be more specific: `isAccountCacheValid`

daemon/data_manager.go
```diff
@@ -385,3 +387,22 @@ func (dm *DataManager) CountryCodeToCountryName(code string) string {
...
+	dm.accountData.set(accountData)
+}
+
+func (dm *DataManager) GetAccountData() (*pb.AccountResponse, bool) {
+	dm.mu.Lock()
+	defer dm.mu.Unlock()
+	return dm.accountData.get()
+}
+
+func (dm *DataManager) InvalidateAccountdata() {
```
devzbysiu, line 404: There is small inconsistency between `Accountdata` here and `AccountData` in other methods

daemon/data_manager.go
```diff
@@ -30,6 +30,7 @@ type DataManager struct {
 	insightsData     InsightsData
 	serversData      ServersData
 	versionData      VersionData
+	accountData      AccountData
```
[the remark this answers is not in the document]
bartoszWojciechO: I considered this, but `CachedValue` doesn't fit this use case. Besides update function, it is also a key/value store, etc. It would work in the future if we'd update the cache to be global.

Is individual lock needed? `AccountData` could be extracted out of data manager(same as connection paramteters).

mariusSincovici (approved):

devzbysiu (approved):

PR #828, opened 2025-04-03
LVPN-6468: Fix heartbeat context

Savolro:


events/moose/moose.go
```diff
@@ -552,6 +559,9 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
 			return err
 		}
 		if data.EventStatus == events.StatusSuccess {
+			if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateThreatProtectionLiteEnabledValue(data.ThreatProtectionLite)); err != nil {
```
mariusSincovici, line 562: Wouldn't this mean that the attempt and failed events will not have TP set, but in settings is `on`?

events/moose/moose.go
```diff
@@ -552,6 +559,9 @@ func (s *Subscriber) NotifyConnect(data events.DataConnect) error {
...
+				return err
+			}
```
mariusSincovici, line 564: shouldn't `...StateIsOnVpnValue` be moved  at the start of the function, so that at least that is set correctly if something else fails?
 Same for disconnect.

cmd/daemon/events_moose.go
```diff
@@ -23,11 +23,7 @@ func newAnalytics(
 	subAPI core.SubscriptionAPI,
 	ver, env, id string) *moose.Subscriber {
 	_ = os.Setenv("MOOSE_LOG_FILE", "Stdout")
-	logLevel := "error"
-	if !internal.IsProdEnv(env) {
-		logLevel = "debug"
-	}
-	_ = os.Setenv("MOOSE_LOG", logLevel)
```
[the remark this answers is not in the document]
Savolro: No, it won't. Default is `None`.

events/moose/moose.go
Savolro: Yes, exactly. This sets the value in `current_state`, not `user_preferences`. This means that it has to be only set after a successful VPN connection.

keliramu (approved):

events/moose/moose.go
```diff
@@ -198,6 +198,15 @@ func (s *Subscriber) Init() error {
 	return nil
 }
 
+func (s *Subscriber) Stop() error {
+	if err := s.response(moose.MooseNordvpnappFlushChanges()); err != nil {
+		return fmt.Errorf("stopping moose worker: %w", err)
+	}
+	if err := s.response(worker.Stop()); err != nil {
```
devzbysiu, line 205: Is this empty on purpose? Maybe just go with `_ = ...`?

devzbysiu (approved):

PR #829, opened 2025-04-04
LVPN-6714: Hide private key in GetPrivateKey RPC when meshnet is not enabled

bartoszWojciechO:


meshnet/server.go
```diff
@@ -1963,9 +1963,14 @@ func (s *Server) GetPrivateKey(ctx context.Context, _ *pb.Empty) (*pb.PrivateKey
 		}, nil
 	}
 
+	privateKey := cfg.MeshPrivateKey
+	if !cfg.Mesh {
+		privateKey = ""
```
mariusSincovici, line 1968: wouldn't be better to return error like for other gRPC calls? `return ErrMeshnetNotEnabled`

keliramu (approved):

devzbysiu (approved):

PR #832, opened 2025-04-07
LVPN-7456: Trigger connecting state immediately

devzbysiu:
Now, when connection is started, the clients get notification about "Connecting" state immediately - not only after the server was selected, which caused delays in displaying the state.

Changes:
- new `ConnectionInfo` that stores the `ConnectionStatus` - it's passed where it's either modified or read
- connection triggered just after initiated, without specifying the details of connection target
- when connection was cancelled and we set the state to "Disconnected", ignore the notification in tray when we don't have disconnect target set


networker/networker.go
```diff
@@ -235,37 +200,36 @@ func NewCombined(
...
+func (netw *Combined) Start(
+	ctx context.Context,
+	creds vpn.Credentials,
+	serverData vpn.ServerData,
+	allowlist config.Allowlist,
+	nameservers config.DNS,
+	enableLocalTraffic bool,
+) (err error) {
+	netw.mu.Lock()
+	defer netw.updateConnectionStatusAfterStart()
```
mariusSincovici, line 217: is it ok if this is executed after the lock is released?

networker/networker.go
```diff
@@ -549,7 +487,7 @@ func (netw *Combined) restart(
 // Stop VPN connection and clean up network after it stopped.
 func (netw *Combined) Stop() error {
 	netw.mu.Lock()
-	defer netw.updateConnectionStatus(ResetStartTime)
+	defer netw.updateConnectionStatusAfterStop()
```
mariusSincovici, line 490: same here

networker/networker.go
devzbysiu: No, it's wrong.. It's LIFO order execution so it will run after mutex is released. Fixing

daemon/rpc_connect.go
```diff
@@ -233,6 +240,8 @@ func (r *RPC) connect(
 			t = internal.CodeDisconnected
 			event.EventStatus = events.StatusCanceled
 			event.Error = nil
+			r.connectionInfo.SetStatus(state.ConnectionStatus{State: pb.ConnectionState_DISCONNECTED, StartTime: nil})
```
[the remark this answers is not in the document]
devzbysiu: In case of an error from networker, we are handling it gracefully, and `retErr` is not being set

networker/networker.go
devzbysiu: Fixed

networker/networker.go
devzbysiu: Fixed

daemon/rpc_state.go
```diff
@@ -80,8 +80,12 @@ func statusStream(stateChan <-chan interface{},
 			switch e := ev.(type) {
 			case events.DataConnect:
 				state := pb.ConnectionState_CONNECTING
-				if e.EventStatus == events.StatusSuccess {
+				//exhaustive:ignore
```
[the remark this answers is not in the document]
devzbysiu: Yes, good idea. Fixed

daemon/rpc_connect.go
```diff
@@ -235,6 +242,8 @@ func (r *RPC) connect(
 			event.Error = nil
 		}
 		r.events.Service.Connect.Publish(event)
+		r.connectionInfo.SetStatus(state.ConnectionStatus{State: pb.ConnectionState_DISCONNECTED, StartTime: nil})
```
Savolro, line 245: This call always goes before the event publishing. Maybe RPC does not need to know about `connectionInfo` and we can leave it for a subscriber to set the status properly based on received events?

Savolro (approved):

PR #834, opened 2025-04-07
LVPN-5901: Fix group flag handling in connect and autoconnect commands

bartoszWojciechO:
Previously, due to library limitiations, --group flag was handled/parsed incorrectly for both connect and set autoconnect commands. Group would be ignored under certain conditions(primarily when command parameters would be provided in an incorrect order).


cli/cli.go
```diff
@@ -1282,8 +1283,6 @@ func (c *cmd) printServersForAutoComplete(country string, hasGroupFlag bool, gro
 			output += server.Name + "\n"
 		}
 
-		fmt.Print(output)
```
mariusSincovici, line 1285: autocomplete is broken

bartoszWojciechO: Should be fixed now.

cli/cli.go
devzbysiu, line 1: Can those be simplified after your changes? 
- https://github.com/NordSecurity/nordvpn-linux/blob/main/daemon/rpc_set_autoconnect.go#L54-L57
- https://github.com/NordSecurity/nordvpn-linux/blob/main/daemon/rpc_set_autoconnect.go#L79-L83

cli/cli.go
```diff
@@ -1312,3 +1313,37 @@ func (c *cmd) printServersForAutoComplete(country string, hasGroupFlag bool, gro
...
 	}
 }
+
+func parseConnectArgs(ctx *cli.Context) (string, string, error) {
+	args := ctx.Args()
+
+	// generate server tag from given args
+	var serverTag string
+	var serverGroup string
+	if args.Len() > 0 {
```
devzbysiu, line 1323: [small] you can minimize nesting by inverting the condition here and returning early

cli/cli_set_autoconnect.go
devzbysiu, line 1: Maybe it would be good to include this new flag in examples section of the command description at the top of this file? Similarly for connect

devzbysiu (approved):

cli/cli.go
```diff
@@ -1312,3 +1313,38 @@ func (c *cmd) printServersForAutoComplete(country string, hasGroupFlag bool, gro
 		}
 	}
 }
+
+func parseConnectArgs(ctx *cli.Context) (string, string, error) {
```
mariusSincovici, line 1317: [small] could you add a comment why this function was created?

mariusSincovici (commented):
There are 2 cases that I've found that maybe should be improved:
* `nv set autoconnect 1 de`
```
Auto-connect country: Germany
Auto-connect city: 
Auto-connect group: UNDEFINED  
```
Would it be possible in this case to have standard VPN at the server matching? And also at display if is undefined to display standard?

* `nv set autoconnect 1 --group Europe us`
this works and connects to US, so it seams like the group is ignored in this case?

mariusSincovici (approved):

cli/cli_set_autoconnect.go
bartoszWojciechO: Added a tech debt task, UX team is working on the new messages.

PR #844, opened 2025-04-14
Fix DNS leaks when enabling meshnet

bartoszWojciechO:


networker/networker.go
```diff
@@ -1618,26 +1619,28 @@ func (netw *Combined) allowFileshareAll() error {
...
-
-	if ruleIndex == -1 {
+	if !netw.dnsDenied {
+		log.Println(internal.DebugPrefix, "attemtp to deny dns when it was not previously denied")
 		return nil
 	}
 
-	if err := netw.fw.Delete([]string{ruleName}); err != nil {
+	if err := netw.fw.Delete([]string{"deny-private-dns"}); err != nil {
 		return err
```
keliramu, line 1628: imo, should be similar to line #1658 i.e. add extra context to error.

networker/networker.go
```diff
@@ -1618,26 +1619,28 @@ func (netw *Combined) allowFileshareAll() error {
...
-	ruleIndex := slices.Index(netw.rules, ruleName)
-
-	if ruleIndex == -1 {
+	if !netw.dnsDenied {
+		log.Println(internal.DebugPrefix, "attemtp to deny dns when it was not previously denied")
 		return nil
 	}
 
-	if err := netw.fw.Delete([]string{ruleName}); err != nil {
+	if err := netw.fw.Delete([]string{"deny-private-dns"}); err != nil {
```
keliramu, line 1627: I would create constant for rule name and use it to Delete, Create the rule, also to add extra context info for error

networker/networker.go
```diff
@@ -1618,26 +1619,28 @@ func (netw *Combined) allowFileshareAll() error {
...
 func (netw *Combined) denyDNS() error {
-	ruleName := "deny-private-dns"
+	if netw.dnsDenied {
+		log.Println(internal.DebugPrefix, "attemtp to deny dns when it was already denied")
+		return nil
+	}
+
 	rules := []firewall.Rule{{
-		Name:           ruleName,
+		Name:           "deny-private-dns",
```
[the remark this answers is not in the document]
bartoszWojciechO: Yeah, would require deep refactoring. I think we have some tasks for this in the backlog but it's probably not planned for yet.

keliramu (approved):

networker/networker.go
```diff
@@ -1618,26 +1620,28 @@ func (netw *Combined) allowFileshareAll() error {
...
-	if ruleIndex == -1 {
+	if !netw.dnsDenied {
+		log.Println(internal.DebugPrefix, "attemtp to undeny dns when it was not previously denied")
 		return nil
 	}
 
-	if err := netw.fw.Delete([]string{ruleName}); err != nil {
-		return err
+	if err := netw.fw.Delete([]string{denyPrivateDNSRule}); err != nil {
+		return fmt.Errorf("deleting deny-private-dns dns rule: %w", err)
```
devzbysiu, line 1629: We should probably use `denyPrivateDNSRule` here also

networker/networker.go
```diff
@@ -1618,26 +1620,28 @@ func (netw *Combined) allowFileshareAll() error {
...
+
+	netw.dnsDenied = false
 
 	return nil
 }
 
 func (netw *Combined) denyDNS() error {
-	ruleName := "deny-private-dns"
+	if netw.dnsDenied {
+		log.Println(internal.DebugPrefix, "attemtp to deny dns when it was already denied")
```
devzbysiu, line 1639: Typo `attemtp`

devzbysiu (approved):

networker/networker.go
bartoszWojciechO: IMO having this hardcoded has some advantages. When this message pops up in the logs, it would be easier to find the source by grepping the code. Otherwise we'd have `"deleting %s dns rule:"` which would be harder to find.

devzbysiu: Yeah, ok. Makes sense :+1:

PR #846, opened 2025-04-15
LVPN-7765: Skip invalid records from server list

keliramu:


core/models.go
```diff
@@ -496,6 +496,28 @@ func (s *Server) UnmarshalJSON(b []byte) error {
...
+func (s *Server) IsValid() bool {
+	return s.Name != "" &&
+		s.Station != "" &&
+		s.Hostname != "" &&
+		len(s.Technologies) > 0 &&
+		len(s.Locations) > 0
+}
+
+// CleanServersList allocation-free list cleaning
+func CleanServersList(servers *Servers) {
```
bartoszWojciechO, line 509: Would it be useful to add some logging in this case? For example `x invalid servers were found during cleanup`?

bartoszWojciechO (approved):

daemon/job_servers.go
```diff
@@ -128,6 +128,9 @@ func JobServers(dm *DataManager, api core.ServersAPI, validate bool) func() erro
 func generateKeys(server core.Server) []string {
 	loweredHostnameID := strings.ToLower(strings.Split(server.Hostname, ".")[0])
 	country := server.Country()
+	if country == nil {
```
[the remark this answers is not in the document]
keliramu: Here this check was removed, here it relies on list of valid server records

devzbysiu (approved):

core/models.go
keliramu: This function was removed.

core/core.go
```diff
@@ -392,6 +398,11 @@ func (api *DefaultAPI) RecommendedServers(filter ServersFilter, longitude, latit
 		return nil, nil, err
 	}
 
+	// if at least one record is not valid - reject whole list, asuming something wrong is with whole list
```
devzbysiu, line 401: [small] Typo `asuming`

daemon/rpc_test.go
```diff
@@ -543,6 +551,11 @@ func serversList() core.Servers {
 		},
 	}
 
+	// if at least one record is not valid - reject whole list, asuming something wrong is with whole list
```
devzbysiu, line 554: [small] Also here `asuming`

devzbysiu (approved):

core/core.go
DoYouEvenCpp: nitpick: here it also should be `something is wrong`

core/core.go
```diff
@@ -330,6 +330,12 @@ func (api *DefaultAPI) Servers() (Servers, http.Header, error) {
 	if err = json.NewDecoder(resp.Body).Decode(&ret); err != nil {
 		return nil, nil, err
 	}
+
+	// if at least one record is not valid - reject whole list, assuming something wrong is with whole list
```
DoYouEvenCpp, line 334: nitpick: should be `something is wrong`

core/models.go
```diff
@@ -496,6 +496,24 @@ func (s *Server) UnmarshalJSON(b []byte) error {
...
+// IsValid returns true if server record is valid.
+func (s *Server) IsValid() bool {
+	return s.Name != "" &&
+		s.Station != "" &&
+		s.Hostname != "" &&
+		len(s.Technologies) > 0 &&
+		len(s.Locations) > 0
+}
+
+func (s *Servers) Validate() error {
```
DoYouEvenCpp, line 508: nitpick: my understanding of `Validate` is that if some data set isn't quite right then it gets fixed, by undertaking some specific actions.
I don't think this is the case here, right?
Is `IsValid` used in any other contexes?

Maybe then simply remove the `IsValid`, snap the check in-place in the method over the range, and rename it from `Validate` to `IsValid`?

DoYouEvenCpp: Also, it would be nice to add a comment also to the `Validate` - just to keep coherent way of commenting what's the purpose of a given method.

daemon/rpc_test.go
DoYouEvenCpp: And also the order, here isn't quite right - `something is wrong`

DoYouEvenCpp (approved):
LGTM!

core/core.go
keliramu: Sorry, a bit too late, I got it merged. But thx :)

PR #850, opened 2025-04-18
LVPN-6261: Maintain custom dns order

edvyno1:
Currently the order of the custom DNS that are set is disregarded, this commit changes so that we sort a copy of the DNS array instead of the DNS array itself, when performing the check

### DIFF
--- a/daemon/rpc_set_dns.go
+++ b/daemon/rpc_set_dns.go
@@ -26,9 +26,11 @@ func (r *RPC) SetDNS(ctx context.Context, in *pb.SetDNSRequest) (*pb.SetDNSRespo
 		}, nil
 	}
 
-	slices.Sort(nameservers)
-	slices.Sort(cfg.AutoConnectData.DNS)
-	if slices.Equal(nameservers, cfg.AutoConnectData.DNS) {
+	nameserverCheck := slices.Clone(nameservers)
+	autoConnectDataCheck := slices.Clone(cfg.AutoConnectData.DNS)
+	slices.Sort(nameserverCheck)
+	slices.Sort(autoConnectDataCheck)
+	if slices.Equal(nameserverCheck, autoConnectDataCheck) {
 		return &pb.SetDNSResponse{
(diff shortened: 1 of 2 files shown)


daemon/rpc_set_dns.go
```diff
@@ -26,9 +26,11 @@ func (r *RPC) SetDNS(ctx context.Context, in *pb.SetDNSRequest) (*pb.SetDNSRespo
 		}, nil
 	}
 
-	slices.Sort(nameservers)
-	slices.Sort(cfg.AutoConnectData.DNS)
-	if slices.Equal(nameservers, cfg.AutoConnectData.DNS) {
+	nameserverCheck := append([]string{}, nameservers...)
+	AutoConnectDataCheck := append([]string{}, cfg.AutoConnectData.DNS...)
```
keliramu, line 30: for common style, local variable name should start with lower letter

keliramu (approved):

devzbysiu (approved):

test/qa/test_dns.py
```diff
@@ -319,3 +320,15 @@ def test_custom_dns_already_disabled_connected(tech, proto, obfuscated):
...
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
+def test_custom_dns_order_is_kept(tech, proto, obfuscated):
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+    nameserver_list = ["8.8.8.8", "1.1.1.1"]
+    sh.nordvpn.set.dns(nameserver_list)
+    with lib.Defer(sh.nordvpn.disconnect):
+        sh.nordvpn.connect()
+        resolver = dnspy.Resolver()
+        assert nameserver_list == resolver.nameservers
```
mariusSincovici, line 333: I think this works only when resolv.conf file is used and not with systemd? :thinking: 
Maybe it is sufficient to only check the `nordvpn settings` output?

mariusSincovici (approved):

PR #852, opened 2025-04-18
LVPN-5901: Fix autoconnect to an undefined group

bartoszWojciechO:


cli/cli.go
```diff
@@ -1314,6 +1314,19 @@ func (c *cmd) printServersForAutoComplete(country string, hasGroupFlag bool, gro
 	}
 }
 
+// removeGroupFlagFromArgs removes flag and its value from args
+// The function assumes that flag value occurs after flag name, i.e --<flag> <value>
+func removeFlagFromArgs(args []string, flag string) []string {
+	for index, arg := range args {
+		if arg == "--"+flag {
+			// return the args slice sans the flag and its argument
+			return append(args[:index], args[index+2:]...)
```
devzbysiu, line 1323: [small] you can go with `slices.Delete(args, index, index+2)` instead of `append`

devzbysiu, line 1323: Btw. won't this panic when there is no value given? :thinking:

devzbysiu (approved):

Savolro (approved):

PR #854, opened 2025-04-23
LVPN-7830 firebase removal

Savolro:


cli/cli.go
```diff
@@ -292,7 +286,7 @@ func NewApp(version, environment, hash, salt string,
 				Action:       cmd.SetTechnology,
 				BashComplete: cmd.SetTechnologyAutoComplete,
 				ArgsUsage:    SetTechnologyArgsUsageText,
-				Description:  setTechnologyDescription,
+				Description:  fmt.Sprintf(SetTechnologyDescription, SupportedValuesWithNordWhisper),
```
bartoszWojciechO, line 289: Will we ever go back to dynamically enabling/disabling whisper based on remote config? If not, maybe we could just get rid of `SupportedValuesWithNordWhisper`/`SupportedValuesWithoutNordWhisper` and have it hardcoded in `SetTechnologyDescription` like before.

bartoszWojciechO (approved):

devzbysiu (approved):

PR #856, opened 2025-04-23
LVPN-7837: Fix wrong connection status

devzbysiu:
The issue here was that the connection status is being reset on go error, but we have the cases where we are not returning error, but just responding to client with error status code and this case was not handled in defer.

Changes:

- connect function is now returning error and optional payload
- then based on those values I'm resetting the state to "Disconnected"

### DIFF
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -33,7 +33,7 @@ func (r *RPC) Connect(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer) (retEr
 
 func (r *RPC) connectWithContext(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer, source pb.ConnectionSource) error {
 	var err error
-	var payload *pb.Payload
+	var didFail bool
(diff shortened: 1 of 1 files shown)


daemon/rpc_connect.go
```diff
@@ -43,21 +44,36 @@ func (r *RPC) connectWithContext(in *pb.ConnectRequest, srv pb.Daemon_ConnectSer
...
+		r.connectionInfo.SetStatus(state.ConnectionStatus{State: pb.ConnectionState_DISCONNECTED, StartTime: nil})
+		r.vpnEvents.Connected.Publish(events.DataConnect{EventStatus: events.StatusFailure})
+	}
+
+	if err != nil {
+		return err
+	}
+
+	if payload != nil {
+		return srv.Send(payload)
```
bartoszWojciechO, line 63: Is there only one `Send` per the entire execution? Maybe this RPC doesn't need to be a stream at all?

bartoszWojciechO (approved):

Savolro (approved):

PR #857, opened 2025-04-24
LVPN-7510: log caller name for config update

bartoszWojciechO:
Filename and line number of the call to SaveFile will be appended to config update debug log, to make tracing easier in case of log spam.

### DIFF
--- a/daemon/state/state.go
+++ b/daemon/state/state.go
@@ -111,7 +111,7 @@ func (s *StatePublisher) NotifyConfigChanged(e config.DataConfigChange) error {
 
 	log.Println(internal.DebugPrefix, "notifying about config change:", e.Caller)
 
-	s.notify(e)
+	s.notify(e.Config)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


Savolro (approved):

daemon/state/state.go
```diff
@@ -105,11 +105,12 @@ func (s *StatePublisher) NotifyMFA(bool) error {
...
 
-func (s *StatePublisher) NotifyConfigChanged(e *config.Config) error {
+func (s *StatePublisher) NotifyConfigChanged(e config.DataConfigChange) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	log.Println(internal.DebugPrefix, "notifying about config change")
+	log.Println(internal.DebugPrefix, "notifying about config change:", e.Caller)
+
 	s.notify(e)
```
devzbysiu, line 114: Here you are passing `config.DataConfigChange`, but later in `rpc_state.go` you are matching type with `*config.Config`. Won't this cause the config update to be ignored in `rpc_state.go`?

bartoszWojciechO: Good point, fixed.

devzbysiu (approved):

PR #865, opened 2025-05-06
LVPN-7579: Fix possible data race in NotificationManager

bartoszWojciechO:
The race occurs because creating a notification for the user and storing this notification ID so that it can be handled later is not handled atomically.

### DIFF
--- a/fileshare/notification_manager.go
+++ b/fileshare/notification_manager.go
@@ -223,8 +223,8 @@ type NotificationManager struct {
 	fileshare          Fileshare
 	openFileFunc       func(string)
 	defaultDownloadDir string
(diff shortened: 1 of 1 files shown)


devzbysiu (approved):

fileshare/notification_manager.go
```diff
@@ -223,6 +223,8 @@ type NotificationManager struct {
 	fileshare          Fileshare
 	openFileFunc       func(string)
 	defaultDownloadDir string
+	// fileLock guards OpenFile and NotifyFile operations, as they cannot be performed at the same time
+	fileLock sync.Mutex
```
mariusSincovici, line 227: [small] maybe rename to easier understand what it does(I don't have too many ides :smile:), something like `downloadedFilesLock`?

PR #866, opened 2025-05-06
LVPN-7958: Avoid data race in monitorFileshareProcessJob

keliramu:
Job is executed every 1sec, job itself is quick ~1..10ms, but at the `set mesh on` time (or right after it) it's taking longer than 1sec and overlapping with next job run - and gocron scheduler was not configured properly.

### DIFF
--- a/meshnet/jobs.go
+++ b/meshnet/jobs.go
@@ -2,7 +2,6 @@ package meshnet
 
 import (
 	"log"
-	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


bartoszWojciechO (approved):

bartoszWojciechO:
Have you managed to find out what was the cause of the data race? Why was the monitor job function called twice at the same time?

PR #867, opened 2025-05-07
LVPN-6859: Transfer killswitch firewall rules from filter to mangle

edvyno1:


daemon/firewall/firewall.go
```diff
@@ -60,7 +61,7 @@ func (fw *Firewall) Add(rules []Rule) error {
 			}
 			fw.publisher.Publish(fmt.Sprintf("replacing existing rule %s", rule.Name))
 		}
-
+		log.Printf("ADDING RULE: %v", rule)
```
mariusSincovici, line 64: Add some level for the message.

daemon/firewall/forwarder/fw.go
```diff
@@ -195,6 +196,7 @@ func removeAllowlistRules(commandFunc runCommandFunc) error {
 func addAllowlistRules(commandFunc runCommandFunc, interfaceNames []string, allowlistedSubnets []string) error {
 	for _, iface := range interfaceNames {
 		for _, subnet := range allowlistedSubnets {
+			log.Printf("========adding %s subnet\n", subnet)
```
mariusSincovici, line 199: Leftover?

daemon/firewall/rule.go
```diff
@@ -84,6 +84,7 @@ type Rule struct {
 	SourcePorts      []int  `json:"source_ports"`
 	DestinationPorts []int  `json:"destination_ports"`
 	Comment          string `json:"comment"`
+	Physical         bool   `json:"physical"`
```
mariusSincovici, line 87: What is physical representing?

config/allowlistlist.go
```diff
@@ -16,24 +17,19 @@ func NewAllowlist(udpPorts []int64, tcpPorts []int64, subnets []string) Allowlis
...
+		Subnets: subnets,
 	}
 }
 
 // Allowlist is a collection of ports and subnets
 type Allowlist struct {
-	Ports   Ports   `json:"ports"`
-	Subnets Subnets `json:"subnets"`
+	Ports   Ports    `json:"ports"`
+	Subnets []string `json:"subnets"` // TODO change to netip.Prefix and refactor
```
mariusSincovici, line 32: Isn't this change breaking the application settings and backward compatibility?

daemon/firewall/allowlist/allowlist_routing.go
```diff
@@ -61,7 +61,7 @@ func routePortsToIPTables(commandFunc runCommandFunc, port string, protocol stri
 		// already set or error happened
 		return err
 	}
-	// iptables -t mangle -I PREROUTING -p tcp --dport 22 -j MARK --set-mark 0xe1f1 -m comment --comment "nordvpn"
+	// iptables -t mangle -I PREROUTING -p tcp --dport 22 -j MARK --set-mark 0xe1f1 -m comment --comment "nordvpn_allowlist"
```
mariusSincovici, line 64: Not from this PR, but I think there is a problem with `nordvpn_allowlist` command in `generateFlushRules`. The regex from there will not match this and will not delete this rules.

config/allowlistlist.go
devzbysiu, line 32: Previously this was set with amortized `O(1)` lookup, but now it's `O(n)` lookup. I'm guessing this won't have huge impact here right?

Savolro: > Previously this was set with amortized `O(1)` lookup, but now it's `O(n)` lookup. I'm guessing this won't have huge impact here right?

It should not as data set is not that big and lookup is never on the hot path. Ordering of elements is important in this case

Savolro: > Isn't this change breaking the application settings and backward compatibility?

In the removed code you could see that it was converting from map to slice during marshaling and now it is just a slice on its own

daemon/firewall/allowlist/allowlist_routing.go
```diff
@@ -106,26 +106,13 @@ func getCleanupIPTablesRules(commandFunc runCommandFunc, chain string) error {
...
-	return nil
-}
-
-func clearRouting(commandFunc runCommandFunc) error {
-	chains := []string{"PREROUTING", "OUTPUT"}
-	for _, chain := range chains {
-		err := getCleanupIPTablesRules(commandFunc, chain)
-		if err != nil {
-			return err
+			return clearRouting(commandFunc)
```
mariusSincovici, line 115: Is it still needed to have the recursive implementation? Doesn't work to iterate over all the rules?

daemon/firewall/rule.go
mariusSincovici: Could you add a comment to explain this? :smile:

mariusSincovici (approved):

devzbysiu (approved):

Savolro (approved):

PR #868, opened 2025-05-08
LVPN-7966: Cancellable HTTP context for moose-worker

bartoszWojciechO:
moose-worker needs to send out all outstanding events before it is stopped. Sometimes this may block the daemon shutdown for longer than necessary. A custom HTTP RoundTripper was added, which allows our moose wrapper to cancel all ongoing HTTP requests on shutdown. This fix was needed primarily for qa tests, as they disconnect/reconnect often, causing issues for the ongoing connections.

### DIFF
--- a/request/cancellable.go
+++ b/request/cancellable.go
@@ -5,19 +5,22 @@ import (
 	"net/http"
 )
 
-type CancellableRoundTripper struct {
+// ContextRoundTripper adds a common context to requests that pass through it.
+type ContextRoundTripper struct {
 	inner http.RoundTripper
 	ctx   context.Context
 }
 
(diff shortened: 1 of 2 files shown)


request/cancellable.go
```diff
@@ -0,0 +1,23 @@
+package request
+
+import (
+	"context"
+	"net/http"
+)
+
+type CancellableRoundTripper struct {
```
Savolro, line 8: Naming: Implementation itself does not imply anything regarding canceling, only regarding context. Could we have a better name?

Also, docstring explaining what this is this would be nice

Savolro (changes requested):

devzbysiu (approved):

mariusSincovici (approved):

Savolro (approved):

PR #871, opened 2025-05-14
LVPN-8012: Handle meshnet key registration delay

bartoszWojciechO:
There is a delay on the backed between registering a new meshnet/NordLynx key and the key being recognized by the server. It may lead to connection failures, for example when user enables meshnet(registering a new key) and then immediately tries to connect usint NordLynx(key is not recognized yet). A sleep was added to ensure a wait time after new key is registered.

### DIFF
--- a/meshnet/register.go
+++ b/meshnet/register.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"sync"
+	"time"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
 	"github.com/NordSecurity/nordvpn-linux/core"
@@ -14,6 +15,13 @@ import (
 
 	"github.com/google/uuid"
 )
+
+// DelayFunc blocks the app for a duration of time
+type DelayFunc func(duration time.Duration)
+
+func delayFunc(duration time.Duration) {
+	time.Sleep(duration)
+}
 
 // Checker provides information about meshnet.
 type Checker interface {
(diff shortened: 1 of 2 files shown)


meshnet/register.go
```diff
@@ -37,7 +46,7 @@ func NewRegisteringChecker(
 	gen KeyGenerator,
 	reg cmesh.Registry,
 ) *RegisteringChecker {
-	return &RegisteringChecker{cm: cm, gen: gen, reg: reg}
+	return &RegisteringChecker{cm: cm, gen: gen, reg: reg, delayFunc: delayFunc}
```
mariusSincovici, line 49: [small] wouldn't it work to pass directly `time.Sleep` instead of defining `delayFunc`?

mariusSincovici (approved):

meshnet/register_test.go
```diff
@@ -138,11 +162,14 @@ func TestIsRegistered_AlreadyRegistered(t *testing.T) {
...
+	delayChecker := delayChecker{}
 	rc := NewRegisteringChecker(cm, &generator{}, &registry{})
+	rc.delayFunc = delayChecker.Delay
 	ok := rc.IsRegistrationInfoCorrect()
 	assert.True(t, ok)
 	// Registration should not be done, values should not change
 	assert.NotEqual(t, privateKey, cm.Cfg.MeshPrivateKey)
 	assert.NotEqual(t, registryUUID, cm.Cfg.MeshDevice.ID.String())
 	assert.NotEqual(t, registryIP, cm.Cfg.MeshDevice.Address.String())
+	assert.False(t, delayChecker.called, "App blocked when no new mesh key was registered.")
```
DoYouEvenCpp, line 174: Do we care about the very sleep interval being used in the production code?
If so, then maybe a slightly different approach here could be applied, eg. a spy, a mock maybe?

bartoszWojciechO: Imo as long as it's a simple constant delay and not calculated dynamically it would be enough to just know that it was called, to make thing simpler. I think we use similar approach for such use cases in other places in the code.

devzbysiu (approved):

PR #872, opened 2025-05-14
LVPN-5163: Re-create HTTP1 transport on failure or VPN conn

Savolro:


request/http_retransport.go
```diff
@@ -1,41 +1,102 @@
...
+	response, err := inner.RoundTrip(req)
+
+	if err != nil {
+		// Re-create the RoundTripper inner and retry one more time to execute the same
+		// request.
+		m.recreateRoundTrip()
+	}
+
+	// Check the errors if inner RoundTripper needs to be recreated.
+	if m.shouldRetryFunc != nil && m.shouldRetryFunc(err) {
```
mariusSincovici, line 84: Wouldn't make sense to move this into the above `if` and execute only when err != nil?

mariusSincovici (approved):

request/http_retransport.go
```diff
@@ -1,41 +1,102 @@
...
-func NewHTTPReTransport(fn func() http.RoundTripper) *HTTPReTransport {
+// RoundTripperCreateFunc is a function used to create a new instance of round tripper when it needs
+// to be recreated.
+type RoundTripperCreateFunc func() http.RoundTripper
+
+// ShouldRetryFunc defines a function that determines whether an HTTP request should be retried
+// based on the given error.
+type ShouldRetryFunc func(err error) bool
+
+// NewHTTPReTransport is a default constructor function for a HTTPReTransport.
```
DoYouEvenCpp, line 38: absolutely petty nitpick:
```suggestion
// NewHTTPReTransport is a default constructor function for an HTTPReTransport.
```

request/http_retransport.go
```diff
@@ -1,41 +1,102 @@
...
 )
 
-// HTTPReTransport is std RoundTripper enhanced to reconnect after vpn connect is done
+// HTTPReTransport is modified/enhanced RoundTripper.
 // Thread safe.
 type HTTPReTransport struct {
-	inner    http.RoundTripper
-	createFn func() http.RoundTripper
-	mu       sync.RWMutex
+	// for protects the access to inner and to shouldRecreate. This will allow multiple
```
DoYouEvenCpp, line 16: Shouldn't this be a bit rephrased?

IIUC, the `shouldRecreate` refers to the overall logic of recreation an HTTP request. If that's the case then what about:

```suggestion
	// protects access to inner and to recreation logic. This will allow multiple
```

devzbysiu (approved):

bartoszWojciechO (approved):

keliramu (approved):

request/http_retransport.go
```diff
@@ -1,41 +1,105 @@
...
+
+	response, err := inner.RoundTrip(req)
+
+	if err != nil {
+		// Check whether inner RoundTripper was updated while request was executed.
+		m.mu.RLock()
+		innerIsSame := counter == m.counter
+		m.mu.RUnlock()
+		if innerIsSame {
+			m.recreateRoundTrip()
```
mariusSincovici, line 82: Might be a bit of edge case, but I think this will still not solve the problem if 2 or more requests fail in the same time. Each will call recreateRoundTrip.
e.g.
`req1` and `req2` both fail, both take the RLock and have `innerIsSame==true`. Both call `recreateRoundTrip`, but one after the other.
___
Possible fix would be to add param to `recreateRoundTrip(int)` and then called `recreateRoundTrip(counter)`  outside.
And recreateRoundTrip to check after Lock for what value of counter was this called.
```
func (m *HTTPReTransport) recreateRoundTrip(counter) {
	m.mu.Lock()
	defer m.mu.Unlock() 
         if m.counter != counter { return;}
         m.counter++
	m.inner = m.createFn()
```

PR #877, opened 2025-05-19
Add TargetServerGroup in connect events

dfetti:


daemon/rpc_connect.go
```diff
@@ -190,6 +190,7 @@ func (r *RPC) connect(
 	event.TargetServerCountry = country.Name
 	event.TargetServerDomain = server.Hostname
 	event.TargetServerIP = subnet.Addr().String()
+	event.TargetServerGroup = in.GetServerGroup()
```
mariusSincovici, line 193: This will not work.
The `in.GetServerGroup()` is not always correct, e.g. `nordvp c p2p` => group is empty.

Try to use the result from bellow, `parameters := GetServerParameters(`. That should give better results.

daemon/rpc_connect.go
```diff
@@ -193,6 +193,7 @@ func (r *RPC) connect(
 	event.DurationMs = max(int(time.Since(connectingStartTime).Milliseconds()), 1)
 
 	parameters := GetServerParameters(in.GetServerTag(), in.GetServerGroup(), r.dm.GetCountryData().Countries)
+	event.TargetServerGroup = parameters.Group.String()
```
devzbysiu, line 196: The `.String()` is coming from generated protobuf code, I'm just wondering if this is stable. Is it possible that the implementation will change when we'll upgrade the package? :thinking: 
Also, the `EnumStringOf` called there has this docs: 
```
// EnumStringOf returns the enum value as a string, either as the name if
// the number is resolvable, or the number formatted as a string.
```
Could it result in number instead of enum name in our case?

dfetti: Reworked it to use the string received from the API. I hope I did it right.

dfetti:
Done by https://github.com/NordSecurity/nordvpn-linux/pull/889

PR #880, opened 2025-05-20
LVPN-7646: Make connection status originated from a single source

DoYouEvenCpp:
Introduce a single, unified, way of providing status updates.

Previously, in order to update the connection status not only a respective event must have been published, but also an manual update to the `ConnectionStatus` needed to be triggered.
With this PR it's now done just by producing Connect/Disconnect event.

Scope of the changes done under this PR:
- status updates are now originated solely trough `events.DataConnect` and `events.DataDisconnect`, with no need for further `SetStatus` calls
- added a new event `DataConnectChangeNotif` - it's used internally to send notification once connection status is updated
- extended `event.DataConnect` structure with a few fields, so all the data needed to produce connection status update is available in a single event
- moved `AddAddrs` and `DelAddrs` under  `tunnel.T` interface so testing is more flexible (eg. to leverage DI)
- removed any dependencies to `ConnectionStatus` or `ConnectionInfo` from `Networker`


daemon/rpc_connect.go
```diff
@@ -51,8 +50,7 @@ func (r *RPC) connectWithContext(in *pb.ConnectRequest, srv pb.Daemon_ConnectSer
 
 	// set connection status to "Disconnected"
 	if didFail || err != nil {
-		r.connectionInfo.SetStatus(state.ConnectionStatus{State: pb.ConnectionState_DISCONNECTED, StartTime: nil})
-		r.vpnEvents.Connected.Publish(events.DataConnect{EventStatus: events.StatusFailure})
+		r.vpnEvents.Disconnected.Publish(events.DataDisconnect{EventStatus: events.StatusFailure})
```
devzbysiu, line 53: Maybe the `EventStatus` is not needed here? The `DataDisconnect` event is ignored anyway in `ConnectionStatusNotifyDisconnect`. Now it may be confusing, because it looks like you are saying that the Disconnect failed :smile: , previously it was `DataConnect` with failure

devzbysiu: I think it's fine with publishing `DataDisconnect` :thinking: , the details why we have disconnected state are not used anywhere anyway. We can extend this in the future if this will be needed

daemon/state/connection_info.go
```diff
@@ -1,64 +1,119 @@
...
+		return 0, 0
+	}
+	return transferStats.Tx, transferStats.Rx
+}
+
+// StatusWithTransferRates returns the current connection status with updated transfer rates
+func (cs *ConnectionInfo) StatusWithTransferRates() types.ConnectionStatus {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+	cs.status.Tx, cs.status.Rx = cs.getTransferRatesForTunnel(cs.status.TunnelName)
```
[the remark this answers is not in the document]
mariusSincovici: ok, the older values are also never used in any case.

daemon/state/connection_info.go
```diff
@@ -1,64 +1,120 @@
...
+// StatusWithTransferRates returns the current connection status with updated transfer rates
+func (cs *ConnectionInfo) StatusWithTransferRates() types.ConnectionStatus {
+	cs.mu.RLock()
+	status := cs.status
+	cs.mu.RUnlock()
+	status.Tx, status.Rx = cs.getTransferRatesForTunnel(status.TunnelName)
+	return status
+}
+
+func (cs *ConnectionInfo) Status() types.ConnectionStatus {
```
mariusSincovici: [suggestion] remove this and keep only the above,  having `status`  name, to prevent confusion.

mariusSincovici (approved):

PR #881, opened 2025-05-21
LVPN-8079: Block telio events when event is being handled

bartoszWojciechO:
Change the telio event handling function so that it will block when sending the event downstream when connection is active. Previously, the function would default to dropping the event immediately if receiver was not listening, which would cause some events to be ignored.

### DIFF
--- a/daemon/vpn/nordlynx/libtelio/libtelio.go
+++ b/daemon/vpn/nordlynx/libtelio/libtelio.go
@@ -88,11 +88,6 @@ func newTelioCallbackHandler(statesChan chan<- state) *telioCallbackHandler {
 }
 
 func (t *telioCallbackHandler) handleEvent(e teliogo.Event) *teliogo.TelioError {
-	t.mu.Lock()
-	defer func() {
-		t.mu.Unlock()
-	}()
-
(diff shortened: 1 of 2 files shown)


daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -66,41 +66,78 @@ func (cb eventCb) Event(payload teliogo.Event) *teliogo.TelioError {
...
+	monitoringContext context.Context
+}
+
+func newTelioCallbackHandler(statesChan chan<- state) *telioCallbackHandler {
+	return &telioCallbackHandler{statesChan: statesChan}
+}
+
+func (t *telioCallbackHandler) handleEvent(e teliogo.Event) *teliogo.TelioError {
+	t.mu.Lock()
+	defer func() {
```
devzbysiu, line 92: No need for anonymous `func`, you can just do `defer t.mu.Unlock()`

devzbysiu (approved):

mariusSincovici (approved):

naujalyte (approved):

PR #884, opened 2025-05-23
Automate NordWhisper tests

naujalyte:


keliramu (approved):

devzbysiu (approved):

test/qa/test_connect.py
```diff
@@ -85,7 +85,7 @@ def test_connect_to_server_random_by_name(tech, proto, obfuscated):
 
 
 @pytest.mark.parametrize("group", lib.ADDITIONAL_GROUPS)
-@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES[:-1])
```
edvyno1, line 88: is this test incompatible with nordwhisper or why is it omitted?

naujalyte: seems logical, will introduce new variables :+1:

naujalyte: done :heavy_check_mark:

test/qa/lib/__init__.py
```diff
@@ -72,6 +84,12 @@
     "P2P",
 ]
 
+# Used for test parametrization with NordWhisper, since other additional groups are not supported with this technology.
+ADDITIONAL_GROUPS_NORDWHISPER = [
```
mariusSincovici, line 88: [small] why not just `NORDWHISPER_GROUPS`? :smile:

test/qa/lib/__init__.py
```diff
@@ -41,6 +50,9 @@
 TECHNOLOGIES_BASIC2 = [
     ("openvpn", "udp", "off"),
 ]
+TECHNOLOGIES_BASIC3 = [
```
mariusSincovici, line 53: [small] same here, except of being in sync with the other(which also doesn't make to code too readable) why not `Nordwhisper_technology`? Shouldn't also contain the protocol type in the struct, `webtunnel`?

test/qa/test_combinations.py
```diff
@@ -75,6 +75,8 @@ def test_status_change_technology_and_protocol(
 
     if source_tech == "openvpn":
         assert source_proto.upper() in sh.nordvpn.status()
+    elif source_tech == "nordwhisper":
+        assert "Webtunnel" in sh.nordvpn.status()
```
mariusSincovici, line 79: Each line makes a gRPC call to get the status.
It might be nicer to store status in a var and just check for the needed info.
Or use the same logic like [here](https://github.com/NordSecurity/nordvpn-linux/pull/884/files#diff-a400e6833169415e735e636e57542a07db4b18fdc1045d7c42c146f5e5900388R404)

test/qa/test_routing.py
```diff
@@ -55,7 +55,7 @@ def test_routing_enabled_connect(tech, proto, obfuscated):
     assert SUBNET_3 in policy_rules
 
     policy_routes = sh.ip.route.show.table(firewall.IP_ROUTE_TABLE)
-    network_interface = "nordtun" if tech == "openvpn" else "nordlynx"
+    network_interface = "nordtun" if tech == "openvpn" else "qtun" if tech == "nordwhisper" else "nordlynx"
```
mariusSincovici, line 58: This should go into a function since the same code is so many times duplicated.

test/qa/lib/__init__.py
naujalyte: Because NordWhisper can also connect to "STANDARD_GROUPS" so just want to be in sync with what's already in the code and separate standard groups from additional :smile:

test/qa/lib/__init__.py
naujalyte: updated according to suggestions

PR #885, opened 2025-05-23
LVPN-7864: Publish first_open event after we have the location info

devzbysiu:
Changes:
- new struct which subscribes to `DeviceLocation` events
- when we get the location and the application was started for the first time, we publish `first_open` event, but only once
- tests


events/firstopen/firstopen_event.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// registerNotifier is the internal constructor that lets you inject a custom guard.
+func registerNotifier(
+	cm *config.FilesystemConfigManager,
+	deviceLocation events.Subscriber[core.Insights],
+	uiItemsClick events.Publisher[events.UiItemsAction],
+	guard *sync.Once,
+) {
+	notifier := FirstOpenNotifier{
```
keliramu, line 45: [small] `notifier` is only used when below condition is `true` - so, it can be created in that conditional block

devzbysiu: True :+1: , fixing

keliramu (approved):

PR #890, opened 2025-05-27
LVPN-7941: Fix autoconnect for obfuscated servers

mariusSincovici:

* When group name is not after --group argument, then the group enum value is stored into the autoconnect tag. For obfuscated server the enum name is different than the API group name.
* Store into the autoconnect the country code and not country name
* Add tests for autoconnect

### DIFF
--- a/config/server_groups.go
+++ b/config/server_groups.go
@@ -13,3 +13,14 @@ var GroupMap = map[string]ServerGroup{
 	"africa_the_middle_east_and_india": ServerGroup_AFRICA_THE_MIDDLE_EAST_AND_INDIA,
 	"obfuscated_servers":               ServerGroup_OBFUSCATED,
 }
+
+// GroupTitleForId converts group ID to group lowercase title
+func GroupTitleForId(group ServerGroup) string {
+	for k, v := range GroupMap {
+		if v == group {
+			return k
+		}
+	}
+
+	return ""
+}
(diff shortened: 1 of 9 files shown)


daemon/rpc_set_autoconnect.go
```diff
@@ -74,7 +74,7 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 			c.AutoConnectData = config.AutoConnectData{
 				ID:                   cfg.AutoConnectData.ID,
 				ServerTag:            serverTag,
-				Country:              parameters.Country,
+				Country:              parameters.CountryCode,
```
mariusSincovici, line 77: Switch to store country codes instead of country names

config/server_groups.go
```diff
@@ -13,3 +13,14 @@ var GroupMap = map[string]ServerGroup{
 	"africa_the_middle_east_and_india": ServerGroup_AFRICA_THE_MIDDLE_EAST_AND_INDIA,
 	"obfuscated_servers":               ServerGroup_OBFUSCATED,
 }
+
+// Convert group IP to group lowercase title
```
devzbysiu, line 17: Also it probably should be `ID`, not `IP` right? :thinking:

bartoszWojciechO (approved):

PR #891, opened 2025-05-28
LVPN-8027: fix ip rules for /0 networks

bartoszWojciechO:
Any allowlisted subnet with a prefix /0 translates to 'from all lookup main' rule. Before this commit, on disconnect/ip rules matching an allowlisted subnet would be deleted. This would delete every 'from all lookup main', causing no-net. With this commit, app tracks the ip rules that were added, and removes them based on priority and destination subnet, so the original 'from all lookup main' will be omitted during cleanup(because priority doesn't match).

### DIFF
--- a/daemon/routes/iprule/iprule.go
+++ b/daemon/routes/iprule/iprule.go
@@ -17,12 +17,12 @@ import (
 
 // Router uses `ip rule` under the hood
 type Router struct {
-	rpFilterManager routes.RPFilterManager
-	ifgroupManager  ifgroup.Manager
-	tableID         uint
-	fwmark          uint32
-	allowSubnets    []string
-	mu              sync.Mutex
+	rpFilterManager      routes.RPFilterManager
+	ifgroupManager       ifgroup.Manager
+	tableID              uint
+	fwmark               uint32
+	subnetToRulePriority map[string]uint
+	mu                   sync.Mutex
 }
 
 // NewRouter is a default constructor for Router
@@ -32,9 +32,10 @@ func NewRouter(
 	fwmark uint32,
 ) *Router {
 	return &Router{
-		rpFilterManager: rpFilterManager,
-		ifgroupManager:  ifgroupManager,
-		fwmark:          fwmark,
+		rpFilterManager:      rpFilterManager,
+		ifgroupManager:       ifgroupManager,
+		fwmark:               fwmark,
+		subnetToRulePriority: make(map[string]uint),
 	}
 }
 
@@ -73,7 +74,7 @@ func (r *Router) SetupRoutingRules(
 			if err := removeFwmarkRule(r.fwmark, ipv6); err != nil {
 				log.Println(internal.DeferPrefix, err)
 			}
-			removeAllowSubnetRules(allowSubnets, ipv6)
+			r.removeAllowSubnetRules(ipv6)
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


Savolro (changes requested):
Could a tests for this case be added with `category.Route`?

daemon/routes/iprule/iprule_test.go
```diff
@@ -183,3 +186,62 @@ func TestCustomTable(t *testing.T) {
...
+	rulesAfterAdd, err := netlink.RuleList(netlink.FAMILY_V4)
+
+	assert.NoError(t, err, "Failed to save new state for ip rules: %s", err)
+	// ignore rules that existed before the test for validation purposes
+	rulesAfterAdd = slices.DeleteFunc(rulesAfterAdd, func(rule netlink.Rule) bool {
+		idx := slices.Index(rulesBeforeAdd, rule)
+		return idx != -1
+	})
+
+	_, subnet2, _ := net.ParseCIDR("103.238.215.35/24")
```
Savolro: https://pkg.go.dev/github.com/projectcalico/libcalico-go/lib/net#MustParseCIDR can be used in tests. Or, you can even construct the struct directly without a need to parse it

daemon/routes/iprule/iprule_test.go
```diff
@@ -183,3 +186,62 @@ func TestCustomTable(t *testing.T) {
...
+	rulesBeforeAdd, err := netlink.RuleList(netlink.FAMILY_V4)
+	assert.NoError(t, err, "Failed to save previous state for ip rules: %s", err)
+
+	router := Router{
+		subnetToRulePriority: make(map[string]uint),
+	}
+
+	allowlist := []string{
+		// two /0 subnets should be simplified to a single default route
+		"35.74.174.235/0",
```
Savolro, line 202: Are those real IP addresses?

Savolro (approved):

daemon/routes/iprule/iprule_test.go
bartoszWojciechO: Randomly generated.

daemon/routes/iprule/iprule.go
```diff
@@ -118,27 +119,42 @@ func (r *Router) SetupRoutingRules(
...
+	r.removeAllowSubnetRules(ipv6)
+
+	// on top, add allowlisted subnet routing rules
+	for _, subnet := range subnets {
+		_, subnetIPNet, err := net.ParseCIDR(subnet)
+		if err != nil {
+			return fmt.Errorf("parsing subnet: %w", err)
+		}
+
+		if _, ok := r.subnetToRulePriority[subnetIPNet.String()]; ok {
```
keliramu, line 144: Whole logic in this function is complex - need comment on explaining decisions why it is done so

daemon/routes/iprule/iprule.go
```diff
@@ -118,27 +119,42 @@ func (r *Router) SetupRoutingRules(
...
-	// remember what allow subnets are in use to be able to cleanup
-	r.allowSubnets = internal.CopyStringSlice(allowSubnets)
+	return nil
+}
+
+func (r *Router) addAllowlistRules(subnets []string, ipv6 bool) error {
+	// cleanup previous allow subnets
+	r.removeAllowSubnetRules(ipv6)
+
+	// on top, add allowlisted subnet routing rules
```
keliramu: This comment is misleading

devzbysiu (approved):

keliramu (approved):

PR #893, opened 2025-05-28
Send analytics evnets related to quality of service both for attempt …

DoYouEvenCpp:
…and OK/ERR cases

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -114,9 +114,10 @@ type DataRequestAPI struct {
 	// Note: Never use `Request.Body`, use `Request.GetBody` instead
 	Request *http.Request
 	// Note: In case you read `Response.Body`, make sure it is set to what it was before
-	Response *http.Response
-	Duration time.Duration
-	Error    error
+	Response  *http.Response
+	Duration  time.Duration
+	Error     error
+	IsAttempt bool
 }
 
 // Analytics analytics handling engine interface
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -715,7 +715,9 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
 	}
 
 	var eventStatus moose.NordvpnappEventStatus
-	if data.Error != nil {
+	if data.IsAttempt {
+		eventStatus = moose.NordvpnappEventStatusAttempt
+	} else if data.Error != nil {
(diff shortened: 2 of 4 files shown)


events/events.go
```diff
@@ -117,6 +117,8 @@ type DataRequestAPI struct {
 	Response *http.Response
 	Duration time.Duration
 	Error    error
+	// IsAttempt indicates if whether the event represents an attempt
```
devzbysiu, line 120: [small] probably shouldn't be `if whether`

events/moose/moose.go
```diff
@@ -715,13 +717,23 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
 	}
```
mariusSincovici: I cannot add comment above, but now there are too many messages logged for endpoints not relevant. Maybe it would make sense not to propagate the error and return `nil` instead? :thinking:

events/moose/moose.go
```diff
@@ -715,13 +717,23 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
 	} else {
 		eventStatus = moose.NordvpnappEventStatusFailureDueToRuntimeException
 	}
+
+	//for attempt events duration shall be set to 0
+	duration := int32(0)
+	if !data.IsAttempt {
+		duration = int32(data.Duration.Milliseconds())
+	}
 	return s.response(fn(
```
mariusSincovici: This reports some empty fields, shouldn't also those be updated and return the correct data?

request/publishing.go
```diff
@@ -24,12 +24,19 @@ func NewPublishingRoundTripper(
 
 func (rt *PublishingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	startTime := time.Now()
+	rt.publisher.Publish(events.DataRequestAPI{
```
Savolro, line 27: This will also cause logger subscriber to print almost the same message twice. Do we want to log attempts as well?

DoYouEvenCpp: Good point. Will introduce an early-exit if logger is handling attempt events, so the output shouldn't get cluttered.

events/moose/moose.go
DoYouEvenCpp: A follow-up ticket has been created to track that one.

Savolro (approved):

PR #889, opened 2025-05-27
LVPN-7653: Adding additional reporting of server info

thegravi:
Changes: 
- Moose events now use **CountryCode** instead of **CountryName** for server connection/disconnection. 
- Implemented **ServerSelectionRule** and **ServerGroup** analytics fields.

PR #898, opened 2025-06-02
LVPN-7964: Country related fields must use ISO 3166 1 alpha2 standard

thegravi:
Changes:
- Integrated final `context_application_config_currentState_serverCountry_value` analytics values
- Refactored existing functions in `moose.go` file to eliminate DRY and for better readability:
    `NotifyLogin`, `NotifyLogout`, `NotifyProtocol`, `NotifyTechnology`, `NotifyConnect`, `NotifyDisconnect`.
- Eliminated use of `moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateIsOnVpnValue` analytics for meshnet.

Depends on #889 

While working on two recent tasks, I ran into an issue with **moose** event error handling. The problem is that every time an event is submitted, we immediately check if it was submitted without any errors. If an error happens, the rest of the code stops running — this causes later events to be skipped entirely.

To improve this, I’d like to suggest a potential change to the `Publisher` pattern: instead of handling just a single error, we could collect a list of errors. This way, all the moose event-functions could add their errors to the list, allowing the flow to continue even if some events fail.
Another option would be to concatenate the errors into a single long string.
Changes would help avoid situations where one error blocks the rest of the process.

However, given the current context, our error reporting does not provide sufficient information—only the basic error message. Therefore, we cannot determine exactly where they originated from. To enhance our understanding of the error's origin, we would also need to log the calling function. There are methods to address this.

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -526,7 +526,7 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
 		return err
 	}
 
-	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateServerCountryValue("n/a")); err != nil {
+	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateServerCountryValue(moose.UnavailableEventParameterValue)); err != nil {
 		return err
 	}
 
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -526,7 +526,7 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
 		return err
 	}
 
-	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateServerCountryValue(moose.UnavailableEventParameterValue)); err != nil {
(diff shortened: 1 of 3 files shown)


events/moose/moose.go
```diff
@@ -608,95 +485,51 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
...
-			"",
-			errToExceptionCode(data.Error),
-			nil,
-		)); err != nil {
-			return err
-		}
-		if err := s.response(moose.NordvpnappUnsetContextApplicationNordvpnappConfigCurrentStateThreatProtectionLiteEnabledValue()); err != nil {
-			return err
-		}
+	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateServerCountryValue("n/a")); err != nil {
```
mariusSincovici: Maybe define a const with n/a because I see there are other places where are used in moose.

mariusSincovici (approved):

bartoszWojciechO (approved):

nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=898) <br/>All committers have signed the CLA.

PR #900, opened 2025-06-03
Remove leftover rules from LVPN-6859 PR

edvyno1:
This commit will remove the unused rules that came with the firewall rework in https://github.com/NordSecurity/nordvpn-linux/pull/867


networker/networker.go
```diff
@@ -888,7 +888,7 @@ func (netw *Combined) setAllowlist(allowlist config.Allowlist) error {
 				Name:       "allowlist_ports_" + pair.name,
 				Interfaces: ifaces,
 				Protocols:  []string{pair.name},
-				Direction:  firewall.TwoWay,
+				Direction:  firewall.Inbound,
```
keliramu, line 885: Would be good to explain this change with a comment, and also add comment to the above logic where subnet rules are added, and why it is different.

mariusSincovici (approved):

devzbysiu (approved):

networker/networker.go
edvyno1: added comment

PR #902, opened 2025-06-04
LVPN-8091: Add RC JSON schema

keliramu:


contrib/rc-json/schema_v1.json
```diff
@@ -0,0 +1,91 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "nordvpn-linux-app-remote-config",
+  "title": "nordvpn-linux-app-remote-config v1",
+  "description": "nordvpn-linux-app-remote-config v1",
+  "$comment": "",
+  "type": "object",
```
devzbysiu, line 7: Should we go strict with `"additionalProperties": false`  for object types?

contrib/rc-json/schema_v1.json
```diff
@@ -0,0 +1,91 @@
...
+        "settings"
+      ],
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Name of feature config or parameter name, must be unique",
+          "minLength": 1,
+          "pattern": "([A-Za-z0-9_-]+)"
+        },
+        "value_type": {
```
devzbysiu, line 38: `value` prop has a definition of type as list of types and here `value_type` is also list of types, but those two are disconnected, right? So I could do something like:
```
"value_type": "string"
```
and then
```
"value": 12345
```
Because the type specified for `value_type`, `string`, is on its allowed types list and also the `value` is integer which is also on `value`'s allowed types list so everything matches. 

Do we want to go with strict schema like:

```
...
"if": {
  "properties": { "value_type": { "const": "string" } }
},
"then": {
  "properties": { "value": { "type": "string" } }
}
...
```
or it's too much?

contrib/rc-json/schema_v1.json
keliramu: Thanks for idea, but we are not trying to have `ideal` validation, just `good enough` - balancing effort and flexibility.

contrib/rc-json/schema_v1.json
keliramu: As for above feedback, thanks for idea, but we are not trying to have `ideal` validation, just `good enough` - balancing effort and flexibility.

devzbysiu (approved):

contrib/rc-json/schema_v1.json
```diff
@@ -0,0 +1,91 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
mariusSincovici, line 2: Shouldn't be https://json-schema.org/draft/2020-12/schema, because from here https://json-schema.org/specification seams to be the latest?

PR #903, opened 2025-06-05
LVPN-8135: Modify daemon startup

devzbysiu:
Changes:
- consent checker which prepares the deamon on start:
  - try to determine user location to figure our the consent mode
  - in case of failures or KS is on, fallback to GDPR mode
  - if user is in GDPR mode country and is logged in, do light logout
  - if user is in standard mode country, set consent to true
- login endpoints now return `LoginStatus_CONSENT_MISSING` if consent was not completed by user 
- reset to defaults skips analytics consent field
- `nordvpn settings` shows 'not set' when analytics consent was not completed
- tests

NOTE: For those changes, analytics consent is set to false by default to not break the app. It will be addressed in `LVPN-8137`


config/config.go
```diff
@@ -43,8 +38,12 @@ type Config struct {
 	Firewall     bool       `json:"firewall"` // omitempty breaks this
 	FirewallMark uint32     `json:"fwmark"`
 	Routing      TrueField  `json:"routing"`
-	Analytics    TrueField  `json:"analytics"`
-	Mesh         bool       `json:"mesh"`
+	// Deprected: In v4, this field is no longer used. Use `AnalyticsConsent` field instead.
+	Analytics TrueField `json:"analytics"` // remove in 4.2
```
Savolro: This should be safe to remove. JSON decode simply ignores excessive fields when parsing

config/config.go
```diff
@@ -43,8 +38,12 @@ type Config struct {
...
+	// AnalyticsConsent describes user decision about extra analytics.
+	// If nil, the consent flow was not yet completed by user.
+	AnalyticsConsent *bool `json:"analyticsConsent"`
```
Savolro: naming. The rest of the fields are in `lower_snake_case`

config/config.go
```diff
@@ -63,6 +62,21 @@ type Config struct {
 	VirtualLocation TrueField `json:"virtual_location,omitempty"`
 }
 
+// WithLoginData makes a copy of current configuration
+// with login data values from `other` configuration.
+func (c Config) WithLoginData(other *Config) Config {
```
Savolro: Is there a need for it to be exported method?

config/config.go
```diff
@@ -63,6 +62,21 @@ type Config struct {
...
+	c.AutoConnectData.ID = other.AutoConnectData.ID
+	c.TokensData = other.TokensData
+	return c
+}
+
+// WithAnalyticsConsent makes a copy of current configuration
+// with analytics consent values from `other` configuration.
+func (c Config) WithAnalyticsConsent(value bool) Config {
```
Savolro: Is there a need for it to be exported method?

protobuf/daemon/settings.proto
```diff
@@ -30,7 +30,7 @@ message Settings {
   bool meshnet = 6;
   bool routing = 7;
   uint32 fwmark = 8;
-  bool analytics = 9;
+  optional bool analytics = 9;
```
Savolro: Don't we want to rename this?

config/manager.go
```diff
@@ -171,28 +171,34 @@ func (f *FilesystemConfigManager) Reset(preserveLoginData bool) (retErr error) {
...
+	retErr = f.load(&current)
+	if retErr != nil {
+		return fmt.Errorf("loading old config: %w", retErr)
+	}
+
+	newCfg = *newConfig(f.machineIDGetter)
+
+	// never reset analytics consent to default if it was set already
+	if current.AnalyticsConsent != nil {
+		newCfg = newCfg.WithAnalyticsConsent(*current.AnalyticsConsent)
```
mariusSincovici: [small] any reason not to change the original cfg? same for the login data

config/config.go
mariusSincovici: [small] suggestion to have a class for this that has 3 states and better understand what each value means.

daemon/rpc_login.go
```diff
@@ -29,6 +27,12 @@ var lastLoginAttemptTime time.Time
 
 // Login the user with given token
 func (r *RPC) LoginWithToken(ctx context.Context, in *pb.LoginWithTokenRequest) (*pb.LoginResponse, error) {
+	if !IsConsentFlowCompleted(r.cm) {
```
mariusSincovici: this is in 3 places, could it go as a function that is called in all the places? Later maybe more things can be added into that function, if needed extra checks.

daemon/userconsent.go
```diff
@@ -0,0 +1,173 @@
...
+
+	mode := modeForCountryCode(countryCode(strings.ToLower(insights.CountryCode)))
+	log.Printf(internal.DebugPrefix+" consent mode for country code '%s': %s", insights.CountryCode, mode)
+	return mode
+}
+
+func (acc *AnalyticsConsentChecker) doLightLogout() error {
+	return acc.cm.SaveWith(func(c config.Config) config.Config {
+		delete(c.TokensData, c.AutoConnectData.ID)
+		c.AutoConnectData.ID = 0
```
mariusSincovici, line 180: what about meshnet?

protobuf/daemon/login.proto
```diff
@@ -24,14 +24,24 @@ message LoginResponse {
...
   SUCCESS = 0;
   UNKNOWN_OAUTH2_ERROR = 1;
   ALREADY_LOGGED_IN = 2;
   NO_NET = 3;
+  CONSENT_MISSING = 4;
 }
 
 message LoginOAuth2Response {
-  LoginOAuth2Status status = 1;
+  LoginStatus status = 1;
```
mariusSincovici, line 36: [small] Shouldn't this be one of?

protobuf/daemon/login.proto
```diff
@@ -24,14 +24,24 @@ message LoginResponse {
...
   string url = 2;
 }
+
+message LoginOAuth2CallbackResponse {
+  LoginStatus status = 1;
+}
+
+message IsLoggedInResponse {
+  bool is_logged_in = 1;
```
mariusSincovici, line 45: [small] Same here, does it make sense to have the bool if status can already give  the information? :thinking:

protobuf/daemon/settings.proto
mariusSincovici: Shouldn't this be removed and something else be returned that has 3 states?

tray/actions.go
```diff
@@ -37,14 +37,16 @@ func (ti *Instance) login() {
 		return
 	}
 
+	// will be addressed in LVPN-8137
+	//exhaustive:ignore
```
mariusSincovici: [small] add the missing cases with a panic :smile:? or log an error message

cmd/daemon/main.go
```diff
@@ -559,6 +559,10 @@ func main() {
 
 	pb.RegisterDaemonServer(s, rpc)
 	meshpb.RegisterMeshnetServer(s, meshService)
+
+	consentChecker := daemon.NewConsentChecker(fsystem, defaultAPI, authChecker)
```
devzbysiu: NOTE: the class is probably not needed, check if it won't be annoying to get rid of it and just have a func

config/config.go
devzbysiu: Done

daemon/userconsent_oss.go
Savolro, line 1: This is not necessarily for OSS builds but for any builds without moose. This can be named in a same way as telio and quench, e.g . `<...>_no_moose.go`

daemon/userconsent.go
Savolro, line 1: Some suffix would be nice to be sure that this is an alternative for `_oss.go` version.

daemon/userconsent.go
```diff
@@ -0,0 +1,173 @@
+//go:build moose
```
Savolro: I'd avoid adding build flags here and rather move them to the `main` package as this does not actually depend on moose but rather is a new checker that can be used with or without moose.

daemon/userconsent_test.go
```diff
@@ -0,0 +1,323 @@
+//go:build moose
```
Savolro: As per previous comment, this can be removed. If this is here, this will only be executed with `cgo` tests for now without a need to do so.

daemon/userconsent_test.go
```diff
@@ -0,0 +1,323 @@
...
+type fakeInsights struct {
+	insightsResult *core.Insights
+	err            error
+}
+
+func (f *fakeInsights) Insights() (*core.Insights, error) {
+	return f.insightsResult, f.err
+}
+
+type fakeAuthChecker struct {
```
Savolro: Why not to use https://github.com/NordSecurity/nordvpn-linux/blob/main/test/mock/auth/auth.go#L5?

daemon/userconsent_test.go
```diff
@@ -0,0 +1,323 @@
...
+			} else {
+				assert.Nil(t, cm.cfg.AnalyticsConsent)
+			}
+			assert.Equal(t, len(cm.cfg.TokensData), tt.expectedTokensLen)
+			assert.Equal(t, cm.cfg.AutoConnectData.ID, tt.expectedAutoConnID)
+		})
+	}
+}
+
+type fakeConfigManager struct {
```
Savolro: Why not to use https://github.com/NordSecurity/nordvpn-linux/blob/main/test/mock/config_manager.go?

daemon/userconsent_test.go
```diff
@@ -0,0 +1,323 @@
...
+	f.cfg = fn(f.cfg)
+	f.saved = true
+	return nil
+}
+
+func (f *fakeConfigManager) Reset(preserveLoginData bool) error {
+	return nil
+}
+
+type fakeInsights struct {
```
Savolro: Could this be moved to `test/mock`?

daemon/userconsent_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+func TestModeForCountryCode_CaseInsensitive(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	codes := []string{"US", "Us", "uS", "us"}
+	for _, codeStr := range codes {
+		t.Run(codeStr, func(t *testing.T) {
+			cc := countryCode(strings.ToLower(codeStr))
```
Savolro: This makes the test invalid as it works because `strings.ToLower()` is called in the test and not in the lookup function.

daemon/userconsent.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Println(internal.WarningPrefix, "insigts data is nil, falling back to GDPR mode")
+		return consentModeGDPR
+	}
+
+	mode := modeForCountryCode(countryCode(strings.ToLower(insights.CountryCode)))
+	log.Printf(internal.DebugPrefix+" consent mode for country code '%s': %s", insights.CountryCode, mode)
+	return mode
+}
+
+func (acc *AnalyticsConsentChecker) doLightLogout() error {
```
Savolro, line 177: Some documentation on this would be nice.

Savolro (changes requested):

config/manager.go
devzbysiu: By "original" you mean the `current`?
The `current` only holds the old data which we need in some cases like login or analytics consent. And because we reset settings here I'm using this clean, new config as a base and only borrowing old settings from `current` if needed
Or did you mean something else?

daemon/userconsent.go
devzbysiu: Do you mean doc comments on the method or implementation documentation? 
I updated the specs we have with information how light logout is performed and also created a ticket to write implementation docs with low level details

daemon/userconsent.go
devzbysiu: Meshnet won't start now, because it also verifies if the user is logged in and with the changes here, `IsLoggedIn` will report that user is not logged in. The config still has meshnet enabled and mesh private key stays intact so this is less intrusive way.
Do you thing it would be better to disable meshnet also? And potentially remove the key?

config/config.go
devzbysiu: Removed

config/config.go
devzbysiu: No, need. Fixed

config/config.go
devzbysiu: No, need. Fixed

protobuf/daemon/settings.proto
devzbysiu: I renamed it and used separate type to represent three states

config/config.go
devzbysiu: Done

protobuf/daemon/login.proto
devzbysiu: It's actually more comfortable to use in the code like that, because you can switch on the status and do something with URL on success, so the `status` field is the single source of truth, you don't need to figure out if `no status == url set == it's ok`
Also it's not mutually exclusive to have status and URL, because on SUCCESS you want to use the URL.
If we want to use `oneof` then we probably should structure the response to have either "success response" or "error response" and then error response would have status and maybe some message etc and the success would have only url. Something like;

```
message LoginOAuth2Response {
  oneof result {
    OAuth2Success success = 1;
    OAuth2Error   error   = 2;
  }
}

message OAuth2Success {
  string url = 1;
}

message OAuth2Error {
  LoginStatus status = 1;  
  string  message = 2;
}
```
But I'm not sure we should have this kind of changes in this release :thinking:

protobuf/daemon/login.proto
devzbysiu: Similarly here. Those are not mutually exclusive fields and for example the `status == success` reports that API responded as expected and then you still can have either `is_logged_in == true` or `is_logged_in == false`

tray/actions.go
devzbysiu: I added the notification that it's not yet implemented. This is just a temporary change which is being addressed in  LVPN-8137. Without it, the tray won't work and LVPN-8137 can't be finished, so we won't forget about this. It's just to allow merging of this PR. Additionally the consent is set to some value in this PR to not break the app, so this case won't be triggered

daemon/userconsent_oss.go
devzbysiu: I did it as we discussed, so there is no more `userconsent_oss.go`. It's just `userconsent.go`, there is an interface that this and no-op version implements and factories are behind the flag: [here](https://github.com/NordSecurity/nordvpn-linux/pull/903/files#diff-141584c0cf0b461a6ab59248e613d87674e2fae1316f1965971a719c8c8dc142) and [here](https://github.com/NordSecurity/nordvpn-linux/pull/903/files#diff-15b1649da6fdf4f4c24940f80f3d6b58a4ad98786da6e9404a590a9624ad4729)

daemon/userconsent.go
devzbysiu: After the changes mentioned [here](https://github.com/NordSecurity/nordvpn-linux/pull/903#discussion_r2135203405) this is not needed anymore

daemon/userconsent.go
devzbysiu: Done

daemon/userconsent_test.go
devzbysiu: Done

daemon/userconsent_test.go
devzbysiu: True, done

daemon/userconsent_test.go
devzbysiu: True, done

daemon/userconsent_test.go
devzbysiu: Done

daemon/userconsent_test.go
devzbysiu: This is by design actually, but I expressed it in a wrong way. What I wanted to say with this test is that "only lowercase country codes will match here"
Anyway I changed it a bit and now when the `CountryCode` is created, constructor sets the code to lowercase, so no need to remember to use `string.ToLower()` anywhere

config/config.go
```diff
@@ -63,6 +59,21 @@ type Config struct {
 	VirtualLocation TrueField `json:"virtual_location,omitempty"`
 }
 
+// WithLoginData makes a copy of current configuration
+// with login data values from `other` configuration.
+func (c Config) withLoginData(other *Config) Config {
+	c.AutoConnectData.ID = other.AutoConnectData.ID
```
DoYouEvenCpp, line 65: Just to be sure - there shouldn't be any races in here, cause the synchronisation is guaranteed by the callee (the manager), correct?

core/models_test.go
```diff
@@ -978,3 +978,15 @@ func TestNordWhisperPort_UnmarshalJSON(t *testing.T) {
 		})
 	}
 }
+
+func TestNewCountryCode_SetsCoutryCodeToLowercase(t *testing.T) {
```
DoYouEvenCpp: nit: there's a typo in here - `/s/SetsCoutry/SetsCountry` ;)

config/manager.go
mariusSincovici: It's ok.  I was just referring not to have a the function to create a copy of the param with the changed field.
```
newCfg = *newConfig(f.machineIDGetter)
newCfg.AnalyticsConsent = value
```

daemon/userconsent.go
mariusSincovici: If it is not set to false, it will fail to connect because of login.
But in settings will still be true. After login it is still true, but not working. User would need to disable and enable for it to work.

daemon/userconsent.go
devzbysiu: Added doc comments and updated our internal documentation

config/config.go
devzbysiu: Yes, that's correct. Those are just helpers and the consumer needs to make sure those are used in a correct way

core/models_test.go
devzbysiu: Nice catch :smile: 
Fixed

daemon/userconsent.go
devzbysiu: Change the light logout and it disabled meshnet also

keliramu (approved):

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+	// fallback to strict mode in case of an issue with API
+	insights, err := acc.insightsAPI.Insights()
+	if err != nil {
+		log.Println(internal.WarningPrefix, "insights api error, falling back to GDRP mode:", err)
+		return consentModeGDPR
+	}
+
+	// fallback to strict mode in case of nil response
+	if insights == nil {
+		log.Println(internal.WarningPrefix, "insigts data is nil, falling back to GDPR mode")
```
thegravi: typo insights

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+		return consentModeGDPR
+	}
+
+	cc := insights.CountryCode
+	// allow override of country code in dev mode
+	if acc.isDevEnv {
+		if envVarCC, exists := os.LookupEnv("NORDVPN_USER_CC"); exists {
+			log.Println(internal.DebugPrefix, "overriding user's coutry code to", envVarCC)
```
thegravi: typo coutry

core/models.go
```diff
@@ -585,3 +585,16 @@ type Insights struct {
...
 }
+
+type CountryCode struct {
+	cc string
+}
+
+func NewCountryCode(code string) CountryCode {
+	cc := strings.ToLower(code)
+	return CountryCode{cc}
+}
```
thegravi, line 596: Don't you want to enforce/check that the code consists of only 2 letters and no digits?

```go
func (c CountryCode) IsValid() bool {
    if len(c.cc) != 2 {
        return false
    }
    
    for _, char := range c.cc {
         // ensures only ASCII letters
        if !unicode.IsLetter(char) || char > 127 {
            return false
        }
    }
    
    return true
}
```

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+//
+// It uses country code and list of lowercase county codes in standard mode to
+// check it. Countries not on the standard mode list fall into GDPR mode.
+func modeForCountryCode(cc core.CountryCode) consentMode {
+	mode, ok := countryCodeToConsentMode[cc]
+	if !ok {
+		return consentModeGDPR
+	}
+	return mode
+}
```
thegravi: Do we really need a map for few options, since all return the same mode?

Why don't use somethis like this?
```go
func modeForCountryCode(cc core.CountryCode) consentMode {
    switch cc.String() {
    case "us", "ca", "jp", "au":
        return consentModeStandard
    default:
        return consentModeGDPR
    }
}
```

cmd/daemon/userconsent_moose.go
```diff
@@ -0,0 +1,19 @@
+//go:build moose
```
thegravi, line 1: Just a thought - why do we tie ourselves to the names of libraries? Why not use as "functionality" instead of the library's name?

e.g.  `userconsent_moose` -> `userconsent_analytics/userconsent_metrics`

config/config.go
```diff
@@ -43,8 +37,10 @@ type Config struct {
 	Firewall     bool       `json:"firewall"` // omitempty breaks this
 	FirewallMark uint32     `json:"fwmark"`
 	Routing      TrueField  `json:"routing"`
-	Analytics    TrueField  `json:"analytics"`
-	Mesh         bool       `json:"mesh"`
+	// AnalyticsConsent describes user decision about extra analytics.
+	// If nil, the consent flow was not yet completed by user.
```
thegravi: How can this be nil? Would there be room for `ConsentMode_Selecting`? This could help avoid misunderstanding the `none` case. Additionally, we could have a concrete state regarding whether the user is in progress to determine whether to accept consent or if it hasn't started yet. This would be useful for the GUI.

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+			log.Println(internal.WarningPrefix, "failed to perform light logout:", err)
+		}
+		return
+	}
+
+	// standard mode has analytics enabled by default and no required
+	// consent flow, so update the config with `AnalyticsConsent := true`
+	if consentMode == consentModeStandard {
+		if err := acc.setConsentAllowed(); err != nil {
+			log.Println(internal.WarningPrefix, "failed to save analytics consent", err)
```
thegravi: Shall we explain more? `"failed to save analytics consent when user in standard-mode and expecting consent allowed:", err`

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+
+// IsConsentFlowCompleted reads configuration file and
+// checks if `AnalyticsConsent` field is set.
+func (acc *AnalyticsConsentChecker) IsConsentFlowCompleted() bool {
+	var cfg config.Config
+	if err := acc.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, "failed to load config when checking consent flow", err)
+		return false
+	}
+	return cfg.AnalyticsConsent != config.ConsentMode_NONE
```
thegravi: Why not make future-proof by explicitly checking for concrete modes?

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	consentMode := acc.consentModeFromUserLocation()
+
+	// logout user if in GDPR consent mode so that any client trying to log in
+	// will get the status informing it that user needs to complete analytics
+	// consent flow
+	if consentMode == consentModeGDPR && acc.authChecker.IsLoggedIn() {
+		if err := acc.doLightLogout(); err != nil {
+			log.Println(internal.WarningPrefix, "failed to perform light logout:", err)
```
thegravi: A little bit more explicit: `"failed to perform light logout when user in gdpr-mode:", err`

daemon/userconsent.go
```diff
@@ -0,0 +1,198 @@
...
+	// fallback to strict mode in case of nil response
+	if insights == nil {
+		log.Println(internal.WarningPrefix, "insigts data is nil, falling back to GDPR mode")
+		return consentModeGDPR
+	}
+
+	cc := insights.CountryCode
+	// allow override of country code in dev mode
+	if acc.isDevEnv {
+		if envVarCC, exists := os.LookupEnv("NORDVPN_USER_CC"); exists {
```
thegravi, line 156: `NORDVPN_USER_CC` could be set incorrecty, I suggest validating

cmd/daemon/main.go
```diff
@@ -508,6 +516,7 @@ func main() {
 		rcConfig,
 		internalVpnEvents,
 		connectionInfo,
+		consentChecker,
```
thegravi, line 519: Why do we put everything in the same basket (BBOM)? Or are there tech dept tickets waiting?

core/models.go
devzbysiu: Yeah, I was thinking about this initially. Ideally we would want that, probably in the constructor following "parse, don't validate" rule and we would have to validate this according to ISO for country codes the backend is using, so it may be more complex validation. 

But on the other hand we are not full on DDD in the codebase and from the functional perspective the result with and without validation is the same for those changes - we are falling into GDPR mode, so IMO we would have more complexity with not much gain currently. So I would not do this know, but this is for sure something to keep in mind when we expand the use of this model.

What do you think?

daemon/userconsent.go
devzbysiu: Yeah, looks good. I can do it this way :+1:

cmd/daemon/userconsent_moose.go
devzbysiu: I don't know if there was a reason to use lib names for build flags. Probably this was just a natural choice when we want to link with `moose` to have build flag as `moose` and those libraries are not changing - so we don't have a substitute for moose which would require build flags update

config/config.go
devzbysiu: `nil` is a leftover from previous model when I used `*bool`, updated.

Regarding the `ConsentMode_Selecting` there is no such use case now that we need information about consent flow being in progress, and from functional perspective it's not giving us anything, because if you are in the `selecting` state, every client behaves the same way as in `none` state until consent is set.

This kind of `*ing` state is used for connecting so far, because we need to display loading indicator in GUI and allow cancelling when the connection process is in progress, but for consent screen we don't have such case.

daemon/userconsent.go
devzbysiu: Done

daemon/userconsent.go
devzbysiu: Done

daemon/userconsent.go
devzbysiu: Done

daemon/userconsent.go
devzbysiu: Improved

daemon/userconsent.go
devzbysiu: Improved

daemon/userconsent.go
devzbysiu: This is available in dev only mode for testing purposes. It's dev/qa responsibility to set it properly, but it and also allows setting it to invalid values to verify how the feature will behave if we get unexpected response from the API

dfetti (approved):

daemon/userconsent.go
devzbysiu: Done

cmd/daemon/main.go
devzbysiu: Yes, we have tech debt tickets, I'll send you in PM

PR #907, opened 2025-06-05
LVPN-7774: Fix incorrect connection status label when disconnected from speciality groups

DoYouEvenCpp:
This PR fixes regression introduced with https://github.com/NordSecurity/nordvpn-linux/pull/880, namely when disconnected connection params were still provided within the `StatusResponse`.

It also fixes incorrect connection status when a connection to speciality groups is cancelled

### DIFF
--- a/daemon/rpc_state_test.go
+++ b/daemon/rpc_state_test.go
@@ -1,0 +1,176 @@
+package daemon
+
+import (
+	"context"
+	"net/netip"
+	"sync"
+	"testing"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/daemon/state/types"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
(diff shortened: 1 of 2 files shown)


daemon/rpc_state_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"sync"
+	"testing"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/daemon/state/types"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/assert"
```
keliramu: (1) `github.com/stretchr/testify/assert` vs (2) `gotest.tools/v3/assert`: 
1) assert continues on fail - allow to see all failures;
2) assert breaks on first fail;
We mostly use (1). Just to know the difference.

keliramu (approved):

daemon/rpc_state_test.go
DoYouEvenCpp: Switched to testify/assert then!

naujalyte (approved):

PR #904, opened 2025-06-05
LVPN-8134: Terminal consent flow

bartoszWojciechO:

PR #911, opened 2025-06-09
LVPN-8137: Execute consent flow in tray

devzbysiu:
Changes:
- login action in tray opens terminal with consent flow if consent was not yet completed, then it continues with login
- the action of triggering the flow is done via D-Bus and `nordvpn://consent` URI 

NOTE: This PR is the last PR for the consent flow in CLI and it enables the feature. Should be merged last

### DIFF
--- a/cli/cli_set_analytics.go
+++ b/cli/cli_set_analytics.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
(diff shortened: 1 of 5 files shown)


cli/cli_set_analytics.go
```diff
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"os"
```
devzbysiu, line 6: Ignore the changes in this file. The actual consent flow is implemented [in that PR](https://github.com/NordSecurity/nordvpn-linux/pull/904) and this PR should be merged after it

PR #913, opened 2025-06-10
LVPN-8096: Split VPN tunnel and general connect events

Savolro:
Instead of requiring internal VPN state reporters to have full context on VPN server data, leave it only for regular VPN connection state reporters and make `state.ConnectionInfo` stateful so it combines the known context with the internal events and produces updates of the listeners.


daemon/vpn/nordlynx/libtelio/libtelio.go.orig
DoYouEvenCpp, line 1: I guess this should not have reached here? :wink:

daemon/vpn/nordlynx/kernel_space.go
```diff
@@ -134,7 +121,7 @@ func (k *KernelSpace) Start(
 func (k *KernelSpace) Stop() error {
 	k.Lock()
 	defer k.Unlock()
-	k.eventsPublisher.Disconnected.Publish(events.DataDisconnect{})
+	k.eventsPublisher.Disconnected.Publish(events.StatusAttempt)
```
DoYouEvenCpp: Why `StatusAttempt` in here?
In the other place (`ConnectionStatusNotifyInternalDisconnect`) you mentioned: 
```
// Currently only StatusSuccess is being reported in case disconnect fails internally

File: daemon/vpn/nordlynx/libtelio/libtelio_test.go [hunk 2]
```diff
@@ -396,7 +396,7 @@ func NewSubscriber(eventsReceivedWG *sync.WaitGroup) subscriber {
 	}
 }
 
-func (s *subscriber) ConnectionStatusNotifyConnect(data events.DataConnect) error {
+func (s *subscriber) ConnectionStatusNotifyInternalConnect(_ events.TypeEventStatus) error {
```
DoYouEvenCpp: nit: just remove `_` completely?

daemon/rpc_connect.go
```diff
@@ -203,47 +180,48 @@ func (r *RPC) connect(
...
+
+	event := events.DataConnect{
+		APIHostname:                r.api.Base(),
+		Auto:                       false,
+		Protocol:                   cfg.AutoConnectData.Protocol,
+		Technology:                 cfg.Technology,
+		ThreatProtectionLite:       cfg.AutoConnectData.ThreatProtectionLite,
+		ResponseServersCount:       1,
+		ResponseTime:               0,
+		DurationMs:                 getElapsedTime(connectingStartTime),
```
DoYouEvenCpp: Now I come to think about it - maybe we could (should?) get rid of DurationMs from the DataConnect completely now?
- it's not used in other places where we emit that event
- it's used anyway in the ConnectionInfo, eventually

(guess I should have done it with my previous PR :smile: already!)

DoYouEvenCpp (changes requested):

daemon/vpn/nordlynx/kernel_space.go
Savolro: Removed this line as we don't do anything like this in other implementations and stop is always called on the rpc_disconnect.go

daemon/rpc_connect.go
Savolro: DurationMs is used by some of the subscribers.

daemon/rpc_connect.go
```diff
@@ -50,7 +50,7 @@ func (r *RPC) connectWithContext(in *pb.ConnectRequest, srv pb.Daemon_ConnectSer
 
 	// set connection status to "Disconnected"
 	if didFail || err != nil {
-		r.vpnEvents.Disconnected.Publish(events.DataDisconnect{})
+		r.events.Service.Disconnect.Publish(events.DataDisconnect{})
```
mariusSincovici, line 53: I think for canceling connection is not working. the status remains connecting.
Because for cancel connection no error is returned and did fail is also false.

daemon/state/connection_info.go
```diff
@@ -104,11 +112,44 @@ func (c *ConnectionInfo) ConnectionStatusNotifyDisconnect(events.DataDisconnect)
...
+) error {
+	c.mu.RLock()
+	fullyConnected := c.fullyConnected
+	c.mu.RUnlock()
+	if !fullyConnected {
+		return nil
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.status.State = state
```
mariusSincovici, line 165: This blocks also the readers until `c.internalNotif.Publish` is finished. Wouldn't be better to unlock the mutex after it is assigned?

daemon/vpn/nordlynx/libtelio/libtelio.go
mariusSincovici, line 1: meshnet routing seams to have the status broken.

daemon/state/connection_info.go
```diff
@@ -55,60 +57,110 @@ func (cs *ConnectionInfo) Status() types.ConnectionStatus {
...
+	// Currently only StatusSuccess is being reported in case disconnect fails internally
+	return c.notifyInternalState(pb.ConnectionState_DISCONNECTED, "")
+}
+
+func (c *ConnectionInfo) notifyInternalState(
+	state pb.ConnectionState,
+	tunnelName string,
+) error {
+	c.mu.Lock()
+	// Always set tunnelNmae as internal event may be shot before the real event
```
devzbysiu: typo `tunnelNmae`

daemon/vpn/nordlynx/libtelio/libtelio_test.go
```diff
@@ -396,7 +396,7 @@ func NewSubscriber(eventsReceivedWG *sync.WaitGroup) subscriber {
 	}
 }
 
-func (s *subscriber) ConnectionStatusNotifyConnect(data events.DataConnect) error {
+func (s *subscriber) ConnectionStatusNotifyInternalConnect(events.TypeEventStatus) error {
```
devzbysiu: Looks like this has wrong type

daemon/state/connection_info.go
```diff
@@ -55,60 +57,110 @@ func (cs *ConnectionInfo) Status() types.ConnectionStatus {
...
+	c.setStatus(status, false)
+	c.internalNotif.Publish(events.DataConnectChangeNotif{Status: status})
 }
 
 func (c *ConnectionInfo) ConnectionStatusNotifyConnect(e events.DataConnect) error {
 	var startTime *time.Time = nil
-	var Rx uint64 = 0
-	var Tx uint64 = 0
 
 	//invariant: for DataConnect possible values of EvenStatus are either connected or connecting
```
DoYouEvenCpp: nit: this comment got invalid with your changes, right?

mariusSincovici (approved):

pvraketa (approved):

Issue #20, opened 2023-04-28
Strange problem since using 3.16.x with captive portals. Downgrading to 3.15.


wyattbiker:
Using 3.15, I ve been connecting to captive portals / hotspots (such Starbucks ) and then I activate nordvpn (e.g. **nordvpn c us**), without having to reboot my Kubuntu. e.g. when I visit Starbucks the next day. 

But when I installed 3.16.x I can only connect to the captive portal the first time or after a reboot. The next day, I cannot connect to the captive portal, unless I reboot my machine. I've tried restarting the service including the Network service. Nothing seems to work. I must reboot. This happens not just at Starbuckses but also other captive portals.

The steps of connecting to a portal are as follows:

1) Select wifi (E.g. Starbucks), deactivate nordvpn.
2) Click on browser (e.g. 8.8.8.8) and accept the Starbucks login agreement.
3) Connect to nordvpn c us.

What I don't understand, what does nordvpn modify that will not allow to reconnect to an captive portal unless I reboot. Any suggestions to try? Is iptables modified in some manner? Anything I should look for?

EmilijusS:
Maybe related -  https://github.com/systemd/systemd/issues/27485

bartoszWojciechO:
nordvpn adds rules to the filter table, you'd have to try to remove the rules manually in order to debug. The app also configures it's own DNS servers, you can try to configure a custom server with `nordvpn set dns <server>`. If this setting is present no other DNS servers will be configured. You can reset this setting later on with `nordvpn set dns off`.

PR #914, opened 2025-06-10
Use pipeline tag

keliramu:


DoYouEvenCpp (approved):

ggediminass (approved):

.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
```
mariusSincovici, line 20: this must be in sync with the bellow value

keliramu: Not sure, in this line is it source ref or target ref? and what is the meaning of this condition?

keliramu: Clarified: on line #20 it means to cancel pipelines from the same github source, except main.

PR #915, opened 2025-06-11
LVPN-8165: Add proper automation test reports

pvraketa:
Added new test report solution. Now reports will be stored in **nordvpn/test/qa/artifactory/report.html**. Example of a report:
image

### DIFF
--- a/ci/test_deb.sh
+++ b/ci/test_deb.sh
@@ -71,7 +71,7 @@ if [[ -n ${LATTE:-} ]]; then
     fi
 fi
 
-python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true --html=reports/report.html "${args[@]}"
+python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true --html=artifacts/report.html "${args[@]}"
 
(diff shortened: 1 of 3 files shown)


ci/test_snap.sh
```diff
@@ -50,7 +50,7 @@ esac
 #     fi
 # fi
 
-python3 -m pytest -v --disable-pytest-warnings --timeout 180 -x -rsx --timeout-method=signal -o log_cli=true "${args[@]}"
+python3 -m pytest -v--disable-pytest-warnings --timeout 180 -x -rsx --timeout-method=signal -o log_cli=true --html=artifacts/report.html "${args[@]}"
```
DoYouEvenCpp: The removed space between `-v` and `--disable-pytest-warnings`, is it OK?

pvraketa: Nope, just typo, will remove that

ci/test_deb.sh
```diff
@@ -71,7 +71,8 @@ if [[ -n ${LATTE:-} ]]; then
     fi
 fi
 
-python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true "${args[@]}"
+python3 -m pip install -r /opt/ci/docker/tester/requirements.txt
```
ggediminass: Perhaps we should just update [Tester](https://github.com/-/nordsecurity/packages/container/package/nordvpn-linux%2Ftester) image itself, to contain these dependencies, so we do not need to install them each time?

bbrutall (approved):

mariusSincovici (approved):

PR #916, opened 2025-06-11
LVPN-6124: Have data on desktop environments used in linux distributions

thegravi:
Changes:
- Integrated metrics gathering and reporting mechanism:
    - `Desktop Environment` - e.g., gnome, kde, mate, etc.
    - `Display Protocol` - x11, wayland
    - `Custom` - ad hoc `key: value` reporting
- Introduced a new dedicated daemon RPC service `telemetry`. It is responsible for providing a way for child processes to submit various metrics for further processing in the moose library. Currently, this applies to `norduser` and `GUI`.
- The `norduser` process now has a full set of user session environment variables.
- Small refactor.

### DIFF
--- a/norduser/service/fork.go
+++ b/norduser/service/fork.go
@@ -291,7 +291,7 @@ func (s *systemGIDProvider) GetNordvpnGid() (uint32, error) {
 		return 0, errors.New("negative gid cannot be converted to uint32")
 	}
 
-	if gid > math.MaxUint32 {
+	if uint64(gid) > uint64(math.MaxUint32) {
 		return 0, errors.New("gid exceeds uint32 maximum value")
 	}
 


sysinfo/desktop_env_test.go
```diff
@@ -0,0 +1,52 @@
...
+func Test_getDesktopEnvironment(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	tests := []struct {
+		name     string
+		envValue string
+		want     string
+	}{
+		{"check for valid entry", "Unity", "unity"},
+		{"check for valid entry", "ubuntu:GNOME", "gnome"},
```
DoYouEvenCpp: nit: can you provide distincti name for the test, so it won't look like `check_for_valid_entry` and `check_for_valid_entry#01`?

sysinfo/desktop_env_test.go
```diff
@@ -0,0 +1,52 @@
...
+		{"check for valid entry with trailing spaces", "kde ", "kde"},
+		{"check for valid entry with random capital letters", "xFce", "xfce"},
+		{"check for valid entry with preceding spaces", " mate", "mate"},
+		{"check for empty entry", "", "none"},
+		{"check for with spaces", "    ", "none"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
DoYouEvenCpp: nit: since this subtest suite and the below one are almost akin, what about differentiating them a bit?
```suggestion
		t.Run("XDG_CURRENT_DESKTOP_"+tt.name, func(t *testing.T) {
```

sysinfo/desktop_env_test.go
```diff
@@ -0,0 +1,52 @@
...
+				return ""
+			}
+			if got := getDesktopEnvironment(mockEnv); got != tt.want {
+				t.Errorf("getDesktopEnvironment() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
DoYouEvenCpp: Similarly here, what about:
```suggestion
		t.Run("DESKTOP_SESSION_"+tt.name, func(t *testing.T) {
```

?

cmd/norduser/telemetry.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// ReportTelemetry reports underlying telemetry asynchronously.
+// It reports metrics once.
+func ReportTelemetry(conn *grpc.ClientConn, evt ReportEvent, wait bool) {
+	client := telemetrypb.NewTelemetryServiceClient(conn)
+	submitEmptyMetrics := evt == ReportOnExit // skip gathering actual system info on exit
+
+	var wg *sync.WaitGroup
+	if wait {
```
bartoszWojciechO, line 40: In `wait == true` case, maybe it would be simpler to just have the two calls in 42/43 run synchronously and block, rather than using a wait group? Unless there is a big benefit in running them concurrently.

cmd/norduser/telemetry.go
```diff
@@ -0,0 +1,110 @@
...
+	"sync"
+	"time"
+
+	telemetrypb "github.com/NordSecurity/nordvpn-linux/daemon/pb/telemetry/v1"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/sysinfo"
+	"google.golang.org/grpc"
+)
+
+const telemetryTimeout = 5 * time.Second
```
DoYouEvenCpp: nit: why not group all of the constants together?

cmd/daemon/events_linux.go
```diff
@@ -43,6 +44,7 @@ func (*dummyAnalytics) NotifyLANDiscovery(bool) error                 { return n
 func (*dummyAnalytics) NotifyVirtualLocation(bool) error              { return nil }
 func (*dummyAnalytics) NotifyPostquantumVpn(bool) error               { return nil }
 func (*dummyAnalytics) Stop() error                                   { return nil }
+func (*dummyAnalytics) OnTelemetry(telemetry.Metric, any) error       { return nil }
```
DoYouEvenCpp, line 47: Wouldn't `NotifyOnTelemetry` (?) be more coherent with the current method naming pattern here?

cmd/norduser/telemetry.go
thegravi: The reason I chose this approach is to eliminate the possibility that these two (low-prio) RPC calls would block the startup flow (for any reason, e.g., on the receiver's side), causing the systray to be delayed in starting up until the RPC calls time out. Your suggestion would work flawlessly on `exit()` case.

norduser/service/fork.go
```diff
@@ -237,3 +244,37 @@ func (c *ChildProcessNorduser) Restart(uid uint32) error {
 
 	return nil
 }
+
+// mergeUserSessionEnv tries to gather all environment variables from user session and merge with
+// the provided environment
+func mergeUserSessionEnv(uid, gid uint32, currentEnv *[]string) error {
```
DoYouEvenCpp: Is this covered by (any) unit test(s) maybe?

events/moose/moose.go
```diff
@@ -579,6 +580,50 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+		case telemetrypb.DisplayProtocol_DISPLAY_PROTOCOL_WAYLAND:
+			// set 'wayland' metric
+		case telemetrypb.DisplayProtocol_DISPLAY_PROTOCOL_X11:
+			// set 'x11' metric
+		case telemetrypb.DisplayProtocol_DISPLAY_PROTOCOL_UNKNOWN:
+		default:
+			// set 'unknown' metric (e.g. NordvpnappUnsetContextDeviceDisplayProtocol)
+		}
+
+	case telemetry.MetricCustom:
```
bartoszWojciechO: Is this something currently needed, or something supported by moose but not needed in our case?

cmd/daemon/events_linux.go
thegravi: The struct is currently named `Subscriber`, and it contains multiple methods prefixed with `Notify...`. However, this struct's core responsibility is to observe changes and react accordingly. It does not perform any actual publishing or notifying itself.
From my perspective, this naming pattern has originated in other parts of the codebase where publishing was implemented first. Those components introduced notify methods, and later, the same naming convention was reused across all subscribers, even in cases where the logic didn’t quite fit. As a result, the naming here feels misleading.
In this case, the analytics struct functions entirely as a receiving entity. It does not publish or notify others, so I wanted the naming to reflect that more accurately. One potential change could be to rename the remaining `Notify...` methods to describe their purpose better and align with the actual responsibilities of this struct.

@DoYouEvenCpp, what do you think?

norduser/service/fork.go
thegravi: Nope, gonna rewrite the function to make it testable.

cmd/daemon/events_linux.go
DoYouEvenCpp: Yes, agree completely!
Was personally under the same feeling when commented in the first place, but was also leaning towards methods name following the same pattern name.

However, since the discussion is in here - shall we then keep `OnTelemetry` and create a ticket to follow-up other methods' names in the future?

events/moose/moose.go
```diff
@@ -22,7 +22,6 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-	"os/exec"
```
devzbysiu: Isn't this breaking compilation? Looks like pipelines are failing to build the app

thegravi: Nothing in this file used `os`, therefore, vscode go's plugin removed it :shrug: . Gonna take a look at it.

thegravi: vscode go's plugin removes it, even though there are cases where `exec` is used :shrug: . Very odd..

events/moose/moose.go
```diff
@@ -579,6 +580,50 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+				return fmt.Errorf("unsetting desktop-environment: %w", err)
+			}
+		} else {
+			if err := s.response(moose.NordvpnappSetContextDeviceDesktopEnvironment(value.(string))); err != nil {
+				return fmt.Errorf("setting desktop-environment: %w", err)
+			}
+		}
+
+	case telemetry.MetricDisplayProtocol:
+		// TODO: missing moose metric support (e.g. NordvpnappSetContextDeviceDisplayProtocol)
```
bartoszWojciechO, line 641: Is there a task for this? Just so that this TODO won't get forgotten.

bartoszWojciechO (approved):

events/moose/moose.go
thegravi: Not particularly. I added it just for convenience to quickly test something for systray or GUI (in the future). I might remove it as well. :) No objections.

sysinfo/desktop_env_test.go
thegravi: done

sysinfo/desktop_env_test.go
thegravi: done

sysinfo/desktop_env_test.go
thegravi: done

cmd/norduser/telemetry.go
thegravi: done

norduser/service/fork.go
thegravi: done

cmd/daemon/events_linux.go
thegravi: Yes, sure. There will be incompatible interfaces after the change; therefore, more work will be needed after renaming these functions.

events/moose/moose.go
thegravi: fixed

cmd/norduser/telemetry.go
```diff
@@ -0,0 +1,113 @@
...
+		defer wg.Done()
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), telemetryTimeout)
+	defer cancel()
+
+	protocol := telemetrypb.DisplayProtocol_DISPLAY_PROTOCOL_UNSPECIFIED
+	if !submitEmpty {
+		switch sysinfo.GetDisplayProtocol() {
+		case x11DesktopEnv:
```
Savolro, line 100: This is probably very unlikely, but as we are getting any string here, we will ignore everything which is not x11 nor wayland. Maybe it would make sense to store it as a string?

thegravi: That would make the logic simpler and straightforward. But since it's just a loosely controlled string, there's no guarantee other clients using this RPC would stick to the same rules. Using strict predefined values not only improves consistency, but it also makes the interface more self-explanatory for anyone implementing it.
Maybe you see other viable use cases to keep using string as an argument?

events/moose/moose.go
devzbysiu: Yeah, I would suggest to remove it unless we know a clear use-case that will be needed soon. Otherwise it's safer without trying to predict the future. If the use-case will appear, we'll implement this and we'll have clear view of what's exactly needed

thegravi: Done

PR #918, opened 2025-06-13
LVPN-4994: Analytics appear to be sent from server device with snap package

thegravi:
Changes:
- Fixed a mechanism used to identify device type in the system
- Aggregated related system information components into a single package
- Some small refactoring

Depends on #916


events/moose/moose.go
```diff
@@ -159,26 +153,19 @@ func (s *Subscriber) Init(httpClient http.Client) error {
 		return fmt.Errorf("setting moose time zone: %w", err)
 	}
 
-	distroVersion, err := distro.ReleasePrettyName()
-	if err != nil {
-		return fmt.Errorf("determining device os: %w", err)
+	distroVersion := sysinfo.HostOSPrettyName()
```
devzbysiu: Looks like this breaks compilation :thinking: , `sysinfo.HostOSPrettyName` returns two values.

Then one line below, it's better to use returned `error` to check if there is something wrong. I think if `distroVersion` **is** empty then we have a problem, right? Not, when it's non-empty? Anyway, using `error` is most likely the way to go

sysinfo/device_type.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	SystemDeviceTypeUnknown SystemDeviceType = "unknown"
+	SystemDeviceTypeDesktop SystemDeviceType = "desktop"
+	SystemDeviceTypeServer  SystemDeviceType = "server"
+)
+
+// detectBySystemDefaultTarget determines the device type based on the default systemd target.
+func detectBySystemDefaultTarget() SystemDeviceType {
+	_, err := exec.LookPath("systemctl")
```
devzbysiu: What about containers? Are we OK with reporting containers as `SystemDeviceTypeUnknown` or it will be caught by other sources?

sysinfo/device_type.go
```diff
@@ -0,0 +1,100 @@
...
+	out, err := exec.Command("systemctl", "get-default").Output()
+	if err != nil {
+		return SystemDeviceTypeUnknown
+	}
+
+	switch strings.TrimSpace(string(out)) {
+	case "graphical.target":
+		return SystemDeviceTypeDesktop
+	case "multi-user.target":
+		return SystemDeviceTypeServer
```
devzbysiu: The default target can be modified, but I guess this is good enough for our case?

sysinfo/device_type_test.go
devzbysiu, line 1: The heuristics in `device_type.go` are hard to test, because of direct dependencies on the OS like `exec.LookPath`, `os.Getenv`. As a result, those tests are pretty basic. Additionally, those tests could be failing for external contributors if they will run them on some custom configuration. Also it looks like there are some failures on the CI :thinking:  (didn't check why exactly)
I think it would be worth it to abstract those OS dependencies so we have full control over the test execution. It will also help with writing corner case tests for some weird setups so the tests will also act as a documentation

sysinfo/host_info.go
```diff
@@ -0,0 +1,7 @@
+package sysinfo
+
+// HostInfo retrieves the complete system information, including kernel name, version,
+// architecture, and additional details about the environment.
+func HostInfo() string {
```
devzbysiu: Is this used somewhere?

sysinfo/host_name_test.go
```diff
@@ -0,0 +1,44 @@
...
+		result, err := readTagFromOSRelease(reader, tt.tag)
+
+		if (err != nil) != tt.wantErr || result != tt.expected {
+			t.Errorf("readTagFromOSRelease(%q) = %q, err: %v; want %q, err: %v",
+				tt.tag, result, err, tt.expected, tt.wantErr)
+		}
+	}
+
+	out, err := readTagFromOSRelease(strings.NewReader(""), "NAME")
+	if out != "" {
```
devzbysiu: [small] You can use `assert.Empty()`

sysinfo/host_name_test.go
```diff
@@ -0,0 +1,44 @@
...
+		t.Errorf("expected empty output, got: %q", out)
+	}
+
+	if err == nil {
```
devzbysiu: [small] you can use `assert.Nil()`

sysinfo/host_name_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"NAME", "Ubuntu", false},
+		{"PRETTY_NAME", "Ubuntu 22.04 LTS", false},
+		{"VERSION", "", true},
+	}
+
+	for _, tt := range tests {
+		reader := strings.NewReader(mockData) // Create a new reader per test iteration
+		result, err := readTagFromOSRelease(reader, tt.tag)
+
+		if (err != nil) != tt.wantErr || result != tt.expected {
```
devzbysiu: [small] Maybe using `assert.Equal()` would be easier to read?

sysinfo/host_name.go
```diff
@@ -0,0 +1,52 @@
...
+
+const etcOSReleaseFile = "/etc/os-release"
+
+// HostName retrieves the standard name of the currently running operating system.
+func HostOSName() (string, error) {
+	file, err := os.Open(etcOSReleaseFile)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
```
devzbysiu: Those are the same as in `HostOSPrettyName`. Maybe it would be worth introducing separate `func` that would open the file and read a specified tag from it? Like
```
func getOSReleaseField(tag string) (string, error) {
	file, err := os.Open(etcOSReleaseFile)
	if err != nil {
		return "", err
	}
	defer file.Close()
	return readTagFromOSRelease(file, tag)
}
```
, but then `HostOSName` would be just:

```
func HostOSName() (string, error) {
	return getOSReleaseField("NAME")
}
```

so maybe it's too much of tiny functions :smile: , I'll leave you the decision :smile:

sysinfo/kernel_name.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
+
+// uname retrieves OS information using the provided command runner.
+func uname(runner cmdRunner, flags string) string {
+	out, err := runner("uname", flags)
+	if err != nil || out == "" {
```
devzbysiu: Would be nice to log this case so it's easier to figure out what happened

sysinfo/kernel_name.go
```diff
@@ -0,0 +1,37 @@
+package sysinfo
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// kernelName defines default kernel name
+const kernelName = "Linux"
```
devzbysiu: [small] `defaultKernelName`?

sysinfo/display_desktop_env.go
```diff
@@ -21,3 +24,9 @@ func getDesktopEnvironment(readEnv envReader) string {
 
 	return strings.ToLower(de)
 }
+
+// DisplayDesktopEnvironment retrieves the current desktop environment.
+// This function only works in user sessions where the environment is populated with XDG info.
+func DisplayDesktopEnvironment() string {
```
devzbysiu: Slightly misleading naming, suggests like it's displaying something :smile:. Maybe `CurrentDesktopEnvironment`? Or just `DesktopEnvironment`?

sysinfo/kernel_name_test.go
```diff
@@ -0,0 +1,97 @@
+/*
+Package distro provides information about the current Linux distribution.
+*/
```
devzbysiu: [small] Maybe would be better to move those docs somewhere else?

sysinfo/kernel_name_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := defaultCmdRunner(tt.args.name, tt.args.args...)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("defaultCmdRunner() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if got != tt.want {
```
devzbysiu: [small] you can use `assert.Equal()`

sysinfo/kernel_name_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{"Valid Command - Uname Operating System", args{"uname", []string{"-o"}}, "GNU/Linux", false},
+		{"Invalid Command", args{"fakecmd", []string{}}, "", true},
+		{"Invalid Flag", args{"uname", []string{"-x"}}, "", true},
+		{"Empty Command", args{"", []string{}}, "", true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := defaultCmdRunner(tt.args.name, tt.args.args...)
+			if (err != nil) != tt.wantErr {
```
devzbysiu: [small] here also `assert.Equal()` may be easier to read

events/moose/moose.go
thegravi: Agree, forgot to change the usage after function refactor.

sysinfo/device_type.go
thegravi: Many targets can be swapped here, but this is just the straightforward/default usage. Did not want to overcomplicate things.

sysinfo/device_type_test.go
thegravi: Well noted, gonna change things a bit

sysinfo/host_info.go
thegravi: Currently not. I just moved from an old location without checking whether there was any usage.

sysinfo/host_name.go
thegravi: Well, at least it's DRY after this :))

sysinfo/display_desktop_env.go
thegravi: Initially, I wanted to group related information under common tags, such as:
`display`, `host`, `device`, kernel`

So, in this PR's world, it's a noun :D, but I totally agree with the misleading naming case. Gonna think of a workaround for this.

sysinfo/kernel_name_test.go
thegravi: Garbage leftover

events/moose/moose.go
thegravi: done

sysinfo/device_type_test.go
thegravi: done

sysinfo/display_desktop_env.go
thegravi: done

sysinfo/kernel_name.go
```diff
@@ -0,0 +1,48 @@
...
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
+
+// uname retrieves OS information using the provided command runner.
+func uname(runner cmdRunner, flags string) string {
+	out, err := runner("uname", flags)
+	if err != nil {
+		log.Printf("%s failed to execute 'uname %s': %v. Falling back to default: %s",
```
DoYouEvenCpp: nit: missing newline character at the end of the log

sysinfo/kernel_name.go
```diff
@@ -0,0 +1,48 @@
...
+			internal.ErrorPrefix, flags, err, defaultKernelName)
+		return defaultKernelName
+	}
+
+	if out == "" {
+		log.Printf("%s 'uname %s' returned empty output. Falling back to default: %s",
```
DoYouEvenCpp: nit: also here, missing newline at the end of log

sysinfo/kernel_name.go
thegravi: fixed

sysinfo/kernel_name.go
thegravi: fixed

sysinfo/device_type.go
thegravi: There are 3 sources, but they won't detect containers. Also, moose events currently support predefined values for server and desktop only.

devzbysiu (approved):

sysinfo/host_info.go
```diff
@@ -0,0 +1,7 @@
+package sysinfo
+
+// GetHostInfo retrieves the complete system information, including kernel name, version,
+// architecture, and additional details about the environment.
+func GetHostInfo() string {
```
keliramu, line 5: This fn is not used (except test), should it be removed? or there are future use plans?

keliramu (approved):

sysinfo/host_info.go
thegravi: https://github.com/NordSecurity/nordvpn-linux/pull/918#discussion_r2165915502

PR #919, opened 2025-06-13
LVPN-8136: Execute consent flow in cli on login and register

devzbysiu:
Changes:
- CLI reacts to `LoginStatus_CONSENT_MISSING` and triggers consent flow, then continues with login procedure

### DIFF
--- a/cli/cli_login.go
+++ b/cli/cli_login.go
@@ -24,6 +24,22 @@ const (
 
 func (c *cmd) Login(ctx *cli.Context) error {
 	resp, err := c.client.IsLoggedIn(context.Background(), &pb.Empty{})
+	if err != nil {
+		return formatError(err)
+	}
+	if resp.Status == pb.LoginStatus_CONSENT_MISSING {
+		// ask user for consent
+		if err := c.setAnalyticsFlow(); err != nil {
+			return formatError(err)
+		}
+	}
+
+	// continue with login
+	return c.loginCmd(ctx)
+}
+
+func (c *cmd) loginCmd(ctx *cli.Context) error {
+	resp, err := c.client.IsLoggedIn(context.Background(), &pb.Empty{})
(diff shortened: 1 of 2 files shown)


cli/cli_login.go
```diff
@@ -23,6 +23,22 @@ const (
...
+			return formatError(err)
+		}
+	}
+
+	// continue with login
+	return c.loginCmd(ctx)
+}
+
+func (c *cmd) loginCmd(ctx *cli.Context) error {
 	resp, err := c.client.IsLoggedIn(context.Background(), &pb.Empty{})
```
mariusSincovici: Isn't this duplicating the `func (c *cmd) Login(ctx *cli.Context) error {` start?

cli/cli_login.go
```diff
@@ -23,6 +23,22 @@ const (
 )
 
 func (c *cmd) Login(ctx *cli.Context) error {
+	resp, err := c.client.IsLoggedIn(context.Background(), &pb.Empty{})
+	if err != nil {
+		return formatError(err)
+	}
+	if resp.Status == pb.LoginStatus_CONSENT_MISSING {
```
mariusSincovici, line 35: Shouldn't this go into the `loginCmd`?

cli/cli_login.go
devzbysiu: True. It's slightly different, because first `IsLoggedIn` is checking for the consent and I'm not handling the case of already logged-in user, but this can be improved. I moved the check for `IsLoggedIn` from `loginCmd` to main `Login` func. So `Login` is more like precondition check, then `loginCmd` is executing the login.

cli/cli_login.go
devzbysiu: Now after I moved the `IsLoggedIn` all to `Login`, `Login` acts as a precondition check and `loginCmd` does the login. Do you think it would be better to move this to `loginCmd`?

PR #920, opened 2025-06-16
LVPN-8137: Execute consent flow in tray

devzbysiu:
Changes:
- `cli_click.go` has new action "consent", this way tray triggers consent flow on login
- this PR enables analytics consent feature by setting `AnalyticsConsent` field to `ConsentUndefined`
- login command in tray triggers the consent by opening `nordvpn://consent` with D-Bus

### DIFF
--- a/cli/cli_click.go
+++ b/cli/cli_click.go
@@ -85,13 +85,8 @@ func (c *cmd) Click(ctx *cli.Context) (err error) {
 				return nil
 
 			case "consent":
-				if err := c.setAnalyticsFlow(); err != nil {
-					return formatError(err)
-				}
-				// consent is triggered here only during login,
-				// so after consent flow is finished, continue with login
-				c.Login(ctx)
-				return nil
+				// login takes care of the analytics consent flow and continues with login
+				return c.Login(ctx)
 			}
 		}
 	}
--- a/cli/cli_click.go
+++ b/cli/cli_click.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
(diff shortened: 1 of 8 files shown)


tray/actions.go
```diff
@@ -8,20 +8,33 @@ import (
...
 	resp, err := ti.client.IsLoggedIn(context.Background(), &pb.Empty{})
-	if err != nil || resp.GetIsLoggedIn() {
+	if err != nil {
+		ti.notify("Login failed")
+		log.Println(internal.ErrorPrefix, "Failed to login:", err)
+		return
+	}
+	if resp.Status == pb.LoginStatus_CONSENT_MISSING {
+		// ask user for consent by opening terminal with consent flow,
+		openURI("nordvpn://consent")
```
mariusSincovici: [small] shouldn't this go somewhere as a const string?
Maybe a file with the schema and the commands, to see which are used, and then combined with the accepted ones from the click flow?

cli/cli_click.go
```diff
@@ -83,6 +83,15 @@ func (c *cmd) Click(ctx *cli.Context) (err error) {
 					return formatError(err)
 				}
 				return nil
+
+			case "consent":
+				if err := c.setAnalyticsFlow(); err != nil {
```
mariusSincovici: [small] why not call directly the c.Login? Isn't login already taking care of this flow.

mariusSincovici (approved):

cli/cli_click.go
devzbysiu: True. It's slightly less explicit, but I think it will be better to not have the consent everywhere

tray/actions.go
devzbysiu: Done

PR #922, opened 2025-06-16
LVPN-7413: Change moose DB permissions

mariusSincovici:

* Allow access to the moose DB only for root and users part of nordvpn group.
* Update FileWrite to always set file permissions, because os.WriteFile is not updating the permissions if the file already exists.

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -279,33 +279,8 @@ func main() {
 	dnsSetter := dns.NewSetter(infoSubject)
 	dnsHostSetter := dns.NewHostsFileSetter(dns.HostsFilePath)
 
-	eventsDbPath := filepath.Join(internal.DatFilesPath, "moose.db")
-	// TODO: remove once this is fixed: https://github.com/ziglang/zig/issues/11878
-	// P.S. this issue does not happen with Zig 0.10.0, but it requires Go 1.19+
-	if !internal.FileExists(eventsDbPath) {
-		_, err := internal.FileCreate(eventsDbPath, internal.PermUserRWGroupRWOthersR)
-		if err != nil {
-			log.Fatalln(err)
-		}
-	} else {
-		// Previously we created this file only with R permission for group, but fileshare daemon
-		// which runs with user permissions also needs to write to it. Need to always rewrite permission
-		// because of users updating from older version.
-		err = os.Chmod(eventsDbPath, internal.PermUserRWGroupRWOthersR)
-		if err != nil {
-			log.Println(err)
-		}
-
-		gid, err := internal.GetNordvpnGid()
-		if err != nil {
-			log.Println(err)
-		}
-
-		err = os.Chown(eventsDbPath, os.Getuid(), gid)
-		if err != nil {
-			log.Println(err)
-		}
-	}
+	eventsDbPath := filepath.Join(internal.DatFilesPathCommon, "moose.db")
+	createMooseDB(eventsDbPath)
 
 	machineID := machineIdGenerator.GetMachineID()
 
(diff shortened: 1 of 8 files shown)


cmd/daemon/main.go
```diff
@@ -656,3 +631,30 @@ func main() {
...
+func createMooseDB(eventsDbPath string) {
+	var permissions os.FileMode = internal.PermUserRWGroupRW
+
+	if !internal.FileExists(eventsDbPath) {
+		file, err := internal.FileCreate(eventsDbPath, permissions)
+		if err != nil {
+			log.Fatalln(err)
+		}
+		defer file.Close()
+	} else {
```
DoYouEvenCpp: I'm aware you mostly extracted bits into that function, but do we need else branch at all?
I guess we can create file if it doesn't exist, and then proceed with permission bits.

cmd/daemon/main.go
```diff
@@ -656,3 +631,30 @@ func main() {
...
+// If the file already exists then its permissions and user group are updated
+func createMooseDB(eventsDbPath string) {
+	var permissions os.FileMode = internal.PermUserRWGroupRW
+
+	if !internal.FileExists(eventsDbPath) {
+		file, err := internal.FileCreate(eventsDbPath, permissions)
+		if err != nil {
+			log.Fatalln(err)
+		}
+		defer file.Close()
```
DoYouEvenCpp: This `defer` is redundant here, isn't it?

cmd/daemon/main.go
```diff
@@ -656,3 +631,30 @@ func main() {
...
+
+// createMooseDB creates the moose DB with the desired permissions.
+// If the file already exists then its permissions and user group are updated
+func createMooseDB(eventsDbPath string) {
+	var permissions os.FileMode = internal.PermUserRWGroupRW
+
+	if !internal.FileExists(eventsDbPath) {
+		file, err := internal.FileCreate(eventsDbPath, permissions)
+		if err != nil {
+			log.Fatalln(err)
```
DoYouEvenCpp: Does this has to be `Fatalln`, eg. do we have to exit if moose.db can't be created? I see a big value in an early exit, however is this vital part of the application?

cmd/daemon/main.go
```diff
@@ -656,3 +631,30 @@ func main() {
 		log.Println(internal.ErrorPrefix, "stopping analytics:", err)
 	}
 }
+
+// createMooseDB creates the moose DB with the desired permissions.
+// If the file already exists then its permissions and user group are updated
+func createMooseDB(eventsDbPath string) {
```
DoYouEvenCpp: nit: `create` in the function name suggests if _always_ creates DB, while it does it only if it's not in the place. Essential bit of this function, is fixing the permissions, correct?
Then what about a slightly different name, like `assignMooseDBPermissions` and a mention in the comment, that the file is also created if not present?

cmd/daemon/main.go
```diff
@@ -656,3 +631,30 @@ func main() {
...
+	var permissions os.FileMode = internal.PermUserRWGroupRW
```
DoYouEvenCpp: nit: can't that be a `const` here?

internal/filesystem.go
```diff
@@ -240,7 +240,13 @@ func FileWrite(path string, contents []byte, permissions os.FileMode) error {
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(path, contents, permissions)
+	err = os.WriteFile(path, contents, permissions)
+	if err != nil {
+		return err
```
bartoszWojciechO: Can it be return `fmt.Errof("writing file: %s", err)`? Error would be easier to trace when it happens with such approach.

dfetti (approved):

internal/filesystem.go
mariusSincovici: Update all the places in the function to give better indication which step fails

cmd/daemon/main.go
mariusSincovici: Change the function and if the function doesn't exits it will create it and return the error from that operation, because if the file is newly created it will have the needed group and permissions.

cmd/daemon/main.go
mariusSincovici: Will be addressed in a separate task.

cmd/daemon/main.go
mariusSincovici: Changed

DoYouEvenCpp (approved):

PR #924, opened 2025-06-16
LVPN-8288: Add pytest.mark.xfail for failing test cases:

pvraketa:
LVPN-8288 Add pytest.mark.xfail for failing test cases:
  Added:
  * pytest xfail mark with related ticket for test cases


test/qa/test_connect.py
```diff
@@ -387,6 +387,7 @@ def test_connect_to_unavailable_servers(tech, proto, obfuscated):
         assert lib.is_connect_unsuccessful(ex)
 
 
+@pytest.mark.xfail(reason="LVPN-8096")
```
devzbysiu: Would it be possible to mark it as expected to fail based on the failure? Like: we know how the test fails when it's expected to fail, so maybe we could make the failure as expected only when it fails in such way?
Probably not with mark, but maybe manually inside the test? Or it's not worth it?

pvraketa: I would say that such a complicated decision is not worth implementation. In my PR is just one line, which will be removed after fixing this issue

PR #925, opened 2025-06-16
Hide IPv6 functionality from the app

dfetti:


cli/cli.go
```diff
@@ -179,19 +179,20 @@ func NewApp(version, environment, hash, salt string,
...
-				ArgsUsage: MsgSetBoolArgsUsage,
-				Description: fmt.Sprintf(
-					MsgSetBoolDescription,
-					SetIpv6UsageText,
-					"ipv6",
-					"ipv6",
-				),
-				BashComplete: cmd.SetBoolAutocomplete,
-			},
+			// LVPN-6928 (dfe): Hiding IPv6 functionality for now
```
Savolro: Why not just remove the code? It will be accessible in the git history anyways.

dfetti: Removed

cli/cli.go
Savolro, line 1: What about set dns and allowlist subnets functionality? They parse any IP address. Should we keep it that way?

dfetti: I checked and the app allows to set ipv6 subnets and dns servers even if it is disabled.. Should we stop it?

dfetti: Now allowlist subnets and dns accepts only IPv4.

bartoszWojciechO: If we are dropping IPv6 for allowlist, should it also be removed from routing and firewall internals? `daemon/routes/rule.go`, `firewall/iptables/iptables.go`, etc. But since we are pressed for time and those are sensitive places, maybe we could just add a tech debt task to clean it up in the next release.

keliramu (approved):

Savolro (approved):

cli/cli_set_dns.go
```diff
@@ -28,6 +28,7 @@ Example: nordvpn set dns 1.1.1.1 1.0.0.1
 
 Limits:
   Can set up to 3 DNS servers
+  Can set only IPv4 servers
```
Savolro: Server itself is not IPv4 or IPv6. Address to a server is

Savolro (changes requested):

cli/cli_set_dns.go
dfetti: Fair enough. `Can set only IPv4 server addresses` is ok?

Savolro (approved):

PR #926, opened 2025-06-17
LVPN-6206: Incorrect OS version in Snap

thegravi:
Changes:
- Introduced OS name retrieval via DBus


events/moose/moose.go
```diff
@@ -579,6 +568,50 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
 	))
 }
 
+func (s *Subscriber) OnTelemetry(metric telemetry.Metric, value any) error {
```
keliramu: Naming: imo, it should be something like 'NotifyTelemetry...' in the same style as other existing subscriber functions

thegravi: I understand your point, but I see this differently. The current structure follows an "observer" pattern, which primarily listens for events from the publishing node, justifying the name `Subscriber`. It doesn't notify itself, only consumes data. Because of this, using method names like `notify` can be confusing. That’s why I’ve begun to suggest clearer names and even opened a tech-debt ticket to help refactor this setup for better clarity.

daemon/telemetry/telemetry.go
```diff
@@ -0,0 +1,36 @@
...
+
+type MetricsListener func(metric Metric, value any) error
+
+type CustomField struct {
+	Label string
+	Value string
+}
+
+func New(listener MetricsListener) *Telemetry {
+	return &Telemetry{listen: listener}
```
keliramu: Naming: by how it is used here, to me it looks like `reporter` and not `listener` - someone would use this fn to report new value was set e.g. SetDesktopEnvironment or SetDisplayProtocol

thegravi: I've slightly changed the follow-up usage of this `listen` variable to reflect the more meaningful purpose.

daemon/telemetry/telemetry.go
```diff
@@ -2,31 +2,61 @@ package telemetry
...
 
 	pb "github.com/NordSecurity/nordvpn-linux/daemon/pb/telemetry/v1"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
-type MetricsListener func(metric Metric, value any) error
+const logTag = "[telemery]"
```
DoYouEvenCpp: nit: spelling, should be `telemetry`

sysinfo/host_name_dbus_src_test.go
```diff
@@ -0,0 +1,117 @@
...
+				t.Errorf("Expected result: %q, got: %q", tt.want, got)
+			}
+		})
+	}
+}
+
+func Test_NewHostname1DBusPropertyClient(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	// client := NewHostname1DBusPropertyClient(&dbus.Conn{})
```
DoYouEvenCpp: Is this still needed (also lines 115, 116)?

sysinfo/host_name_dbus_src.go
```diff
@@ -0,0 +1,58 @@
...
+	dbusHostname1PropOperatingSystemPrettyName string = "OperatingSystemPrettyName"
+	dbusHostname1PropOperatingSystemCPEName    string = "OperatingSystemCPEName"
+	dbusHostname1PropHomeURL                   string = "HomeURL"
+	dbusHostname1PropHardwareVendor            string = "HardwareVendor"
+	dbusHostname1PropHardwareModel             string = "HardwareModel"
+	dbusHostname1PropFirmwareVersion           string = "FirmwareVersion"
+	dbusHostname1PropFirmwareVendor            string = "FirmwareVendor"
+)
+
+// HostOSPrettyName retrieves user-friendly OS name using D-Bus communication with the system
```
DoYouEvenCpp: nit: outdated method name used in comment

sysinfo/host_name_dbus_src.go
```diff
@@ -0,0 +1,58 @@
...
+// hostname service.
+func GetHostOSPrettyName() (string, error) {
```
DoYouEvenCpp, line 37: Doesn't this implementation now makes `GetHostOSPrettyNameFallback` dead code?
Maybe we should still use it here as, well..., a _fallback_ mechanism when DBus fails?

sysinfo/host_name_dbus_src_test.go
thegravi: removed

sysinfo/host_name_dbus_src.go
thegravi: fixed

sysinfo/host_name_dbus_src.go
thegravi: I considered it. Since we're supporting deb and snap, the `HostOSPrettyName` implementation that uses a `/etc/os-release` file will not suffice for both cases. Technically, yes, it's dead code, at least for now. Therefore, I agree with removing it entirely, and if a need arises in the future, we can reintroduce it.

daemon/telemetry/telemetry.go
```diff
@@ -2,31 +2,61 @@ package telemetry
...
 
-type MetricsListener func(metric Metric, value any) error
+const logTag = "[telemetry]"
 
-type CustomField struct {
-	Label string
-	Value string
-}
+// MetricsListener defines a callback function used to report collected metrics.
+type MetricsListener func(metric Metric, value any) error
```
bartoszWojciechO: I think in other places we name function type `<Name>Func`. So here it would probably be something like `MetricsFunc` or `ListenFunc`. We don't have set conventions and I don't know if we follow this standard so though so it's a minor comment imo.

bartoszWojciechO (approved):

daemon/telemetry/telemetry.go
thegravi: done

daemon/telemetry/telemetry.go
thegravi: Updated. Closing.

PR #927, opened 2025-06-17
LVPN-8308: fully read SSH response before checking for exit status

bartoszWojciechO:
In case when SSH response is bigger than SSH window_size, reading recv_exit_status before fully reading the response could lead to a deadlock. recv_exit_status would block because no exit status could be received untill full response from the server is read, but client would not accept any more data because the buffer is full. It is a good practice to first read the response, then check for exit status.

### DIFF
--- a/test/qa/lib/ssh.py
+++ b/test/qa/lib/ssh.py
@@ -1,7 +1,6 @@
 import contextlib
 import json
 import time
-import socket
 from collections import namedtuple
 
 import paramiko
--- a/test/qa/lib/ssh.py
+++ b/test/qa/lib/ssh.py
@@ -28,14 +28,14 @@ class Ssh:
         self.client.connect(self.hostname, 22, username=self.username, password=self.password)
 
     def exec_command(self, command: str) -> str:
-        _, stdout, stderr = self.client.exec_command(command, timeout=30)
+        _, stdout, stderr = self.client.exec_command(command, timeout=10)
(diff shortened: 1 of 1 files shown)


test/qa/lib/ssh.py
```diff
@@ -27,11 +28,19 @@ def connect(self):
         self.client.connect(self.hostname, 22, username=self.username, password=self.password)
 
     def exec_command(self, command: str) -> str:
-        _, stdout, stderr = self.client.exec_command(command, timeout=10)
+        _, stdout, stderr = self.client.exec_command(command, timeout=30)
```
DoYouEvenCpp: Is timeout change  also a vital part of the fix? (the PR description doesn't mention it :wink: )

bartoszWojciechO: Debugging leftover. :smile:

test/qa/lib/ssh.py
```diff
@@ -27,11 +28,19 @@ def connect(self):
...
+        try:
+            output = stdout.read().decode()
+            error = stderr.read().decode()
+        except socket.TimeoutError as err:
+            stdout.close()
+            stderr.close()
+            raise RuntimeError("Socked timed out.") from err
```
mariusSincovici: [spelling] Socked

mariusSincovici (approved):

PR #930, opened 2025-06-17
LVPN-8101: Add support for moose debugger events

DoYouEvenCpp:
to be provided


events/events.go
```diff
@@ -101,6 +101,60 @@ type DataConnectChangeNotif struct {
...
+
+	// KeyBasedContextPaths allows defining custom context paths with associated values.
+	// Unlike GeneralContextPaths, this field enables creating new, event-specific context
+	// paths with custom values rather than reusing existing ones.
+	KeyBasedContextPaths []ContextValue
+}
+
+// WithJsonData adds JSON payload to the event
+func (e *MooseDebuggerEvent) WithJsonData(json string) *MooseDebuggerEvent {
+	e.JsonData = json
```
keliramu, line 117: Would be good to have basic json validation, but probably it is not here the best location as here chosen pattern does not return an error. Ideas?

cmd/daemon/main.go
```diff
@@ -608,6 +609,8 @@ func main() {
 		}
 	}()
 	rpc.StartJobs(statePublisher, heartBeatSubject)
+
+	daemonEvents.MooseDeveloper.DeveloperEvents.Publish(events.MooseDebuggerEvent{})
```
keliramu: This is for test only? should be removed?

events/moose/moose.go
```diff
@@ -579,6 +578,43 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+// It allows providing a custom JSON payload and context paths for the event.
+// For custom context paths, corresponding values must be of any of the following types: bool, float32, int32, int64, string.
+// Unsupported types are discarded.
+//
+// Parameters:
+//   - e: The MooseDebuggerEvent containing JSON data and context paths to process
+func (s *Subscriber) NotifyDebuggerEvent(e events.MooseDebuggerEvent) error {
+	combinedPaths := append([]string{}, e.GeneralContextPaths...)
+
+	key := moose.MooseNordvpnappGetDeveloperContextKey()
```
keliramu, line 624: Maybe json validation can be done here? as soon as possible i.e. validate what you have before you do anything more, and provide early feedback to calling party.

cmd/daemon/events_linux.go
```diff
@@ -13,36 +13,37 @@ import (
...
+func (*dummyAnalytics) NotifyMeshnet(bool) error                            { return nil }
+func (*dummyAnalytics) NotifyIpv6(bool) error                               { return nil }
+func (*dummyAnalytics) NotifyDefaults(any) error                            { return nil }
+func (*dummyAnalytics) NotifyConnect(events.DataConnect) error              { return nil }
+func (*dummyAnalytics) NotifyDisconnect(events.DataDisconnect) error        { return nil }
+func (*dummyAnalytics) NotifyLogin(events.DataAuthorization) error          { return nil }
+func (*dummyAnalytics) NotifyLogout(events.DataAuthorization) error         { return nil }
+func (*dummyAnalytics) NotifyMFA(bool) error                                { return nil }
+func (*dummyAnalytics) NotifyAccountCheck(any) error                        { return nil }
+func (*dummyAnalytics) NotifyDebuggerEvent(events.MooseDebuggerEvent) error { return nil }
```
devzbysiu, line 39: I see there is a `DebuggerEvent` vs `DeveloperEvent` in this PR, I think it would be better to have it consistent.
If I remember correctly, debugger events is the old name, now it's called developer events (double check :smile: )

events/events.go
DoYouEvenCpp: Yes, my idea was to make those building functions easy to use in a chained calls, thus no error reporting here.
I can however, provide basic json validation in the moose implementation, while preparing overall event to be sent out.

events/moose/moose.go
```diff
@@ -622,6 +623,43 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+// For custom context paths, corresponding values must be of any of the following types: bool, float32, int32, int64, string.
+// Unsupported types are discarded.
+//
+// Parameters:
+//   - e: The MooseDebuggerEvent containing JSON data and context paths to process
+func (s *Subscriber) NotifyDebuggerEvent(e events.MooseDebuggerEvent) error {
+	combinedPaths := append([]string{}, e.GeneralContextPaths...)
+	key := moose.MooseNordvpnappGetDeveloperContextKey()
+	for _, ctx := range e.KeyBasedContextPaths {
+		path := fmt.Sprintf("%s.%s", key, ctx.Path)
```
devzbysiu, line 626: Can't we reuse this `path` in `append`s below?

events/moose/moose.go
```diff
@@ -622,6 +623,43 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+			if val > math.MaxInt32 {
+				moose.MooseNordvpnappSetDeveloperEventContextLong(path, int64(val))
+			} else {
+				moose.MooseNordvpnappSetDeveloperEventContextInt(path, int32(val))
+			}
+		case string:
+			moose.MooseNordvpnappSetDeveloperEventContextString(path, v)
+			combinedPaths = append(combinedPaths, fmt.Sprintf("%s.%s", key, ctx.Path))
+		default:
+			log.Printf("%s Discarding unssupported type (%T) on path: %s", internal.WarningPrefix, ctx.Value, path)
```
devzbysiu: [typo] `unssupported`

events/moose/moose.go
```diff
@@ -622,6 +623,43 @@ func (s *Subscriber) NotifyRequestAPI(data events.DataRequestAPI) error {
...
+		path := fmt.Sprintf("%s.%s", key, ctx.Path)
+		switch v := ctx.Value.(type) {
+		case bool:
+			moose.MooseNordvpnappSetDeveloperEventContextBool(path, v)
+			combinedPaths = append(combinedPaths, fmt.Sprintf("%s.%s", key, ctx.Path))
+		case float32:
+			moose.MooseNordvpnappSetDeveloperEventContextFloat(path, v)
+			combinedPaths = append(combinedPaths, fmt.Sprintf("%s.%s", key, ctx.Path))
+		//deliberately omitted uint64
+		case int, int8, int16, int32, int64, uint, uint8, uint16, uint32:
```
devzbysiu, line 635: Do we need to keep the `path` for this case in `combinedPaths`? Other cases are appending the `path` to `combinedPaths` :thinking:

events/moose/moose.go
devzbysiu: [small] missing `\n` in log message

events/moose/moose.go
DoYouEvenCpp: Good catch! You're right. Missed that one.

PR #931, opened 2025-06-17
LVPN-8118: Replace setup/teardown functions with pytest fixtures

bbrutall:
There are duplicated setup/teardown functions for daemon start/login in test modules. To get rid of duplicated code, it was  decided to replace it with pytest fixtures.

### DIFF
--- a/test/qa/conftest.py
+++ b/test/qa/conftest.py
@@ -1,9 +1,11 @@
 import datetime
 import io
+import lib
 import subprocess
 import signal
 import threading
 import time
+import sh
 
 import dns.resolver
 import pytest
@@ -11,7 +13,7 @@ import pytest
 import sys
 import os
 
-from lib import logging, network
+from lib import logging, network, daemon, login, info, firewall
 
 pytest_plugins = ("lib.pytest_timeouts.pytest_timeouts")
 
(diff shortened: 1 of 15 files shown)


test/qa/conftest.py
```diff
@@ -1,17 +1,19 @@
...
+import sh
 
 import dns.resolver
 import pytest
 
 import sys
 import os
 
-from lib import logging, network
+from lib import logging, network, daemon, login, info, firewall
```
pvraketa, line 16: Please sort lines alphabetically. Also, how about adding a rule if more than 3 imports from the library, then move each imported module to a new line like this:
```suggestion
from lib import (
    daemon,
    firewall,
    info,
    logging,
    login,
    network,
)
```

test/qa/conftest.py
```diff
@@ -120,3 +122,59 @@ def _capture_traffic(stop_event):
...
+def disable_dns_ipv6_and_threat_protection():
+    """Disable DNS over IPv6 and threat protection settings."""
+    lib.set_dns("off")
+    lib.set_ipv6("off")
+    lib.set_threat_protection_lite("off")
+
+
+@pytest.fixture
+def disable_notifications():
+    """Disable notifications."""
```
pvraketa, line 187: What kind of notification are we talking about? Could you specify?

test/qa/conftest.py
```diff
@@ -120,3 +122,59 @@ def _capture_traffic(stop_event):
         process.kill()
     print(f"tshark out {process.stdout.read().strip()[-10:]} - {process.stderr.read().strip()[-10:]}")
     time.sleep(1)
+
+
+@pytest.fixture
+def collect_logs():
+    """Collect logs."""
```
pvraketa, line 129: What kind of logs are we collecting? Deamon? Journal? Execution log? Could you specify?

pvraketa:
What about test_update.py, test_settings.py, test_misc.py, test_login_nordaccount.py, test_fileshare.py, test_connect6.py? In these test files are also use of setup/teardown function of the daemon. Also, I see that test_meshnet has a couple of setup/teardown functions. How do you think, can we move them also as fixtures to conftest.py? Or will we do it in the scope of a different task?

test/qa/conftest.py
```diff
@@ -120,3 +122,59 @@ def _capture_traffic(stop_event):
...
+def add_and_delete_random_route():
+    """Add and delete a random network route."""
+    firewall.add_and_delete_random_route()
+
+
+@pytest.fixture
+def disable_dns_ipv6_and_threat_protection():
+    """Disable DNS over IPv6 and threat protection settings."""
+    lib.set_dns("off")
+    lib.set_ipv6("off")
```
devzbysiu: Is this still needed after https://github.com/NordSecurity/nordvpn-linux/pull/925/files

bbrutall: No, removed after rebase.

bbrutall:
> What about test_update.py, test_settings.py, test_misc.py, test_login_nordaccount.py, test_fileshare.py, test_connect6.py? In these test files are also use of setup/teardown function of the daemon. Also, I see that test_meshnet has a couple of setup/teardown functions. How do you think, can we move them also as fixtures to conftest.py? Or will we do it in the scope of a different task?



> What about test_update.py, test_settings.py, test_misc.py, test_login_nordaccount.py, test_fileshare.py, test_connect6.py? In these test files are also use of setup/teardown function of the daemon. Also, I see that test_meshnet has a couple of setup/teardown functions. How do you think, can we move them also as fixtures to conftest.py? Or will we do it in the scope of a different task?

It requires more effort, so it should be a different task.

dfetti (approved):

test/qa/conftest.py
```diff
@@ -120,3 +122,67 @@ def _capture_traffic(stop_event):
...
+    login.login_as("default")
+
+    yield
+
+    sh.nordvpn.logout("--persist-token")
+    daemon.stop()
+
+
+@pytest.fixture(scope="module")
+def unblock_network():
```
pvraketa: You should place this fixture in test_misc.py file or change it to this:
`def unblock_network(nordvpnd_scope_module)`
In this case, it will work and keep the right ordering

PR #932, opened 2025-06-18
LVPN-8307: Update tests for consent

devzbysiu:
Changes:
- adds `pexpect` as a dependency to tester and bumps tester image
- updates existing tests to take into account analytics consent
- adds new tests for analytics consent:
  - analytics consent displayed on login
  - invalid user input repeats the prompt
  - ctrl+c causes the prompt to appear again on next login
  - after confirming consent it does not appear again
  - `nordvpn set analytics` also triggers prompt
- helpers

### DIFF
--- a/ci/docker/tester/requirements.txt
+++ b/ci/docker/tester/requirements.txt
@@ -11,3 +11,4 @@ protobuf==5.28.3
 selenium==4.26.0
 grpcio==1.68.1
 pytest-html==4.1.1
+pexpect==4.8.0
--- a/test/qa/lib/__init__.py
+++ b/test/qa/lib/__init__.py
@@ -135,6 +135,18 @@ IPV6_SERVERS = [
 ]
 
 
+EXPECTED_CONSENT_MESSAGE = (
+    "We value your privacy.\n\n"
+    "That's why we want to be transparent about what data you agree to give us. "
+    "We only collect the bare minimum of information required to offer a smooth and stable VPN experience.\n\n"
+    'By pressing "y" (yes), you allow us to collect and use limited app performance data. '
+    "This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy. \n"
+    "https://my.nordaccount.com/legal/privacy-policy/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=settings_account-privacy_policy&nm=app&ns=nordvpn-linux-cli&nc=settings-privacy_policy\n\n"
+    'Press "n" (no) to send only the essential data our app needs to work.\n\n'
+    "Your browsing activities remain private, regardless of your choice.\n\n"
+)
+
+
 class Protocol(Enum):
     UDP = "UDP"
     TCP = "TCP"
@@ -376,3 +388,8 @@ def technology_to_upper_camel_case(tech: str) -> str:
             return "OpenVPN"
         case "NORDWHISPER":
             return "NordWhisper"
+
+
+def squash_whitespace(text: str) -> str:
+    """Normalize whitespace by collapsing all sequences of whitespace into single spaces."""
+    return re.sub(r'\s+', ' ', text.strip())
--- a/test/qa/lib/__init__.py
+++ b/test/qa/lib/__init__.py
@@ -148,7 +148,7 @@ Press "n" (no) to send only the essential data our app needs to work.
 Your browsing activities remain private, regardless of your choice.
 """
 WE_VALUE_YOUR_PRIVACY_MSG = "We value your privacy"
-USER_CONSENT_POMPT = "Do you allow us to collect and use limited app performance data\? \(y/n\)"
+USER_CONSENT_PROMPT = "Do you allow us to collect and use limited app performance data\? \(y/n\)"
 
 
 class UserConsentMode(str, Enum):
--- a/test/qa/lib/__init__.py
+++ b/test/qa/lib/__init__.py
@@ -392,4 +392,4 @@ def technology_to_upper_camel_case(tech: str) -> str:
 
 def squash_whitespace(text: str) -> str:
     """Normalize whitespace by collapsing all sequences of whitespace into single spaces."""
-    return re.sub(r'\s+', ' ', text.strip())
+    return ' '.join(text.split())
--- a/test/qa/lib/__init__.py
+++ b/test/qa/lib/__init__.py
@@ -135,18 +135,21 @@ IPV6_SERVERS = [
 ]
 
 
-EXPECTED_CONSENT_MESSAGE = (
-    "We value your privacy.\n\n"
-    "That's why we want to be transparent about what data you agree to give us. "
-    "We only collect the bare minimum of information required to offer a smooth and stable VPN experience.\n\n"
-    'By pressing "y" (yes), you allow us to collect and use limited app performance data. '
-    "This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy. \n"
-    "https://my.nordaccount.com/legal/privacy-policy/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=settings_account-privacy_policy&nm=app&ns=nordvpn-linux-cli&nc=settings-privacy_policy\n\n"
-    'Press "n" (no) to send only the essential data our app needs to work.\n\n'
-    "Your browsing activities remain private, regardless of your choice.\n\n"
-)
+EXPECTED_CONSENT_MESSAGE = """
+We value your privacy.
+
+That's why we want to be transparent about what data you agree to give us. We only collect the bare minimum of information required to offer a smooth and stable VPN experience.
+
+By pressing "y" (yes), you allow us to collect and use limited app performance data. This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy.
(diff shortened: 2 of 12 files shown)


test/qa/lib/__init__.py
```diff
@@ -376,3 +388,8 @@ def technology_to_upper_camel_case(tech: str) -> str:
             return "OpenVPN"
         case "NORDWHISPER":
             return "NordWhisper"
+
+
+def squash_whitespace(text: str) -> str:
+    """Normalize whitespace by collapsing all sequences of whitespace into single spaces."""
+    return re.sub(r'\s+', ' ', text.strip())
```
DoYouEvenCpp: nit: because of the fact I'm not the biggest fan in the universe of regexes, I can't help but propose a slightly different solution in here: 
```suggestion
    return '  '.join(text.split())
```

Which does the same, I guess. :wink:

test/qa/test_login.py
```diff
@@ -28,6 +31,98 @@ def teardown_function(function):  # noqa: ARG001
     logging.log()
 
 
+def test_analytics_consent_is_displayed_on_login():
+    cli = pexpect.spawn("nordvpn", args=["login"], encoding='utf-8', timeout=10)
```
DoYouEvenCpp: Care to extract common parts of the tests into a fixture? Possibly spawning the application, some of the `cli.expect` calls and possibly buffers could be unified.

Shouldn't be a lot of changes I guess (if it happens to be other way round, please disregard my comment then!).

test/qa/test_settings.py
```diff
@@ -229,13 +230,41 @@ def test_is_custom_dns_removed_after_setting_defaults(tech, proto, obfuscated, n
     assert not dns.is_set_for(nameserver)
 
 
+def test_set_analytics_starts_prompt_even_if_completed_before():
+    # first run: see prompt and respond
+    cli1 = pexpect.spawn("nordvpn", args=["set", "analytics"], encoding='utf-8', timeout=10)
+    cli1.expect(r"Do you allow us to collect and use limited app performance data\? \(y/n\)")
```
bbrutall, line 236: Propose to define a separate str variable instead of having this message duplicated.

test/qa/lib/login.py
```diff
@@ -33,12 +40,32 @@ def get_credentials(key) -> Credentials:
...
+
+def _analytics_value(mode: ConsentMode) -> str:
+    if mode == ConsentMode.Undefined:
+        raise Exception("can't set analytics with undefined consent")
+
+    if mode == ConsentMode.Granted:
+        return "on"
+
+    if mode == ConsentMode.Denied:
+        return "off"
```
bbrutall: Probably it makes sense to create separate class inherited from (str, Enum) and define these 'on' and 'off' values to return?

test/qa/test_login.py
```diff
@@ -28,6 +31,98 @@ def teardown_function(function):  # noqa: ARG001
...
+    # wait until the final prompt appears, then capture everything before it
+    cli.expect(r"Do you allow us to collect and use limited app performance data\? \(y/n\)")
```
bbrutall: Propose to define a separate str variable instead of having this message duplicated.

test/qa/lib/settings.py
```diff
@@ -109,9 +109,25 @@ def is_dns_disabled():
...
+    Returns True, if Analytics Consent is granted, False if it's denied.
+
+    If the consent was not declared. It raises an exception.
+    """
+    analytics_consent = Settings().get("analytics consent")
+    if analytics_consent == "granted":
+        return True
+
+    if analytics_consent == "denied":
+        return False
```
bbrutall: Probably it makes sense to create separate class inherited from (str, Enum) and define these 'granted' / 'denied'  / 'undefined' states?

test/qa/test_login.py
```diff
@@ -28,6 +31,98 @@ def teardown_function(function):  # noqa: ARG001
...
+    cli.sendline("blah")
+
+    # wait for error + prompt again
+    cli.expect(r"Do you allow us to collect and use limited app performance data\? \(y/n\)")
+    second_output = output_buffer.getvalue()[len(first_output):]  # take just the new part
+
+    assert "We value your privacy" in lib.squash_whitespace(first_output)
+    assert "We value your privacy" not in lib.squash_whitespace(second_output)
+    assert "Invalid response" in lib.squash_whitespace(second_output)
+    assert "(y/n)" in lib.squash_whitespace(second_output)
```
bbrutall: if the first assert raises, the next will be skipped? Do we expect to have all the asserts executed in such case?

test/qa/test_login.py
```diff
@@ -28,6 +31,98 @@ def teardown_function(function):  # noqa: ARG001
...
+
+    # wait for the consent prompt
+    cli1.expect(r"Do you allow us to collect and use limited app performance data\? \(y/n\)")
+
+    # simulate user hitting Ctrl+C
+    cli1.sendintr()  # this sends SIGINT like Ctrl+C
+    cli1.expect(pexpect.EOF)
+    first_output = buffer1.getvalue()
+
+    assert "We value your privacy" in lib.squash_whitespace(first_output), \
```
bbrutall: Message to verify can be defined as a separate variable to avoid duplicates.

test/qa/lib/login.py
devzbysiu: If we want to do that then it should be done for the whole codebase. Currently, we are using string literals everywhere. I'll create a task for this, otherwise we'll have a lot of changes only for this which would obscure more important changes

test/qa/test_login.py
devzbysiu: Done

test/qa/lib/settings.py
devzbysiu: Done

test/qa/test_login.py
devzbysiu: Done

test/qa/test_login.py
devzbysiu: Whatever is easier for debugging. I didn't put this in one big assert because the text we are asserting on is long on it's own, having this in one assert would be even worse to debug. Unless you have something else in mind?

test/qa/lib/__init__.py
devzbysiu: Too much spaces in your version :smile:
Updated

test/qa/test_login.py
devzbysiu: Done

ci/docker/tester/requirements.txt
```diff
@@ -11,3 +11,4 @@ protobuf==5.28.3
 selenium==4.26.0
 grpcio==1.68.1
 pytest-html==4.1.1
+pexpect==4.8.0
```
edvyno1, line 14: this is also needed for the tester-selenium test as well

devzbysiu: Yeah, see it now. Didn't know we have separate image for selenium. Updating

test/qa/lib/__init__.py
```diff
@@ -135,6 +135,25 @@
...
+    "That's why we want to be transparent about what data you agree to give us. "
+    "We only collect the bare minimum of information required to offer a smooth and stable VPN experience.\n\n"
+    'By pressing "y" (yes), you allow us to collect and use limited app performance data. '
+    "This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy. \n"
+    "https://my.nordaccount.com/legal/privacy-policy/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=settings_account-privacy_policy&nm=app&ns=nordvpn-linux-cli&nc=settings-privacy_policy\n\n"
+    'Press "n" (no) to send only the essential data our app needs to work.\n\n'
+    "Your browsing activities remain private, regardless of your choice.\n\n"
+)
+WE_VALUE_YOUR_PRIVACY_MSG = "We value your privacy"
+USER_CONSENT_POMPT = "Do you allow us to collect and use limited app performance data\? \(y/n\)"
```
mariusSincovici: [spelling] POMPT

ci/docker/tester/requirements.txt
edvyno1: don't forget to add requirements.txt change in `ci/docker/tester-selenium`

devzbysiu: Yep, yep. Done.

test/qa/lib/settings.py
```diff
@@ -137,7 +155,8 @@ def app_has_defaults_settings():
         "Firewall: enabled" in settings and
         "Firewall Mark: 0xe1f1" in settings and
         "Routing: enabled" in settings and
-        "Analytics: enabled" in settings and
+        # Analytics Consent is not restored to default on reset
+        ("Analytics Consent: granted" in settings or "Analytics Consent: denied" in settings) and
```
edvyno1: `granted/denied` changed in https://github.com/NordSecurity/nordvpn-linux/pull/934

test/qa/lib/__init__.py
devzbysiu: Done

test/qa/lib/settings.py
devzbysiu: Fixed

PR #934, opened 2025-06-18
LVPN-8136: Change setting labels

devzbysiu:
Changes:
- `AnalyticsConsent: granted/denied/undefined` changed to `User Consent: enabled/disabled/undefined`
- tests

### DIFF
--- a/nstrings/consent_mode.go
+++ b/nstrings/consent_mode.go
@@ -1,0 +1,17 @@
+package nstrings
+
+import "github.com/NordSecurity/nordvpn-linux/config/consent"
+
+// UserConsent returns user-friendly representation of `ConsentMode` mode
+func UserConsent(mode consent.ConsentMode) string {
+	switch mode {
+	case consent.ConsentMode_DENIED:
+		return disabled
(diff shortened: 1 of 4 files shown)


nstrings/consent_mode.go
```diff
@@ -0,0 +1,17 @@
...
+func UserConsent(mode consent.ConsentMode) string {
+	switch mode {
+	case consent.ConsentMode_DENIED:
+		return disabled
+	case consent.ConsentMode_GRANTED:
+		return enabled
+	case consent.ConsentMode_UNDEFINED:
+		return undefined
+	default:
+		return undefined
```
thegravi, line 15: Why skip `not set`? Just for full usability, it does not mean it must be used somewhere.

thegravi (approved):

nstrings/consent_mode.go
devzbysiu: `not set` is not used. It's was a leftover from the initial implementation. Removed

edvyno1 (approved):

PR #937, opened 2025-06-19
Remove ipv6 test leftovers

dfetti:


test/qa/test_connect6.py
```diff
@@ -18,14 +18,12 @@ def setup_function(function):  # noqa: ARG001
     daemon.start()
     login.login_as("default")
     logging.log()
-    print(sh.nordvpn.set.ipv6.on())
```
edvyno1, line 21: this removal makes the `connect6` tests into duplicated connect tests, and we are not running these in CI either. imo either remove the whole files or leave them be as is to be decided what we do with them later

test/qa/test_firewall6.py
```diff
@@ -39,7 +39,6 @@ def test_connected_firewall_disable(tech, proto, obfuscated):
     with lib.Defer(sh.nordvpn.disconnect):
         lib.set_technology_and_protocol(tech, proto, obfuscated)
 
-        lib.set_ipv6("on")
```
edvyno1, line 42: same idea as for connect tests, this makes it quite pointless too, either remove file or leave as is until later

dfetti: Ok, I will remove them

Savolro (approved):

edvyno1 (approved):

PR #943, opened 2025-06-24
LVPN-8006: add off-killswitch flag to set defaults

bartoszWojciechO:


config/manager.go
```diff
@@ -166,7 +166,7 @@ func (f *FilesystemConfigManager) save(c Config) error {
 // Reset config values to defaults.
 //
 // Thread-safe.
-func (f *FilesystemConfigManager) Reset(preserveLoginData bool) (retErr error) {
+func (f *FilesystemConfigManager) Reset(preserveLoginData bool, disableKillswitch bool) (retErr error) {
```
DoYouEvenCpp, line 169: Personally, I found boolean parameters passed as method parameters to be some sort of a code smell.  

Readability and usability of `Reset(true, true)` or `Reset (true, false)` isn't best in the world, I guess. It imposes on developers to reserve some extra capacity to comprehend what specific true/false value stands for.

What do you think about:
- extending the `Manager` interface a bit by making explicit `Reset` methods (eg., `ResetWithLoggingPreserved`, `ResetWithKillSwitchDisabled`)? biggest disadvantage is there gonna be multiple methods,  yet very clear/strict about the internal behaviour
- extend the passed parameters, to wrap them under some sort of a structure? (eg., this way this will be more extensible and explicit while running actual call to `Reset`)

DoYouEvenCpp: Or maybe at least, shall we define constants in the places where the method is actually done, so there's no `Reset(false, false` in the code :wink:

test/qa/test_allowlist_subnet.py
```diff
@@ -33,8 +33,7 @@ def teardown_function(function):  # noqa: ARG001
     logging.log(data=info.collect())
     logging.log()
 
-    sh.nordvpn.logout("--persist-token")
-    sh.nordvpn.set.defaults("--logout")
+    sh.nordvpn.set.defaults("--logout", "--off-killswitch")
```
DoYouEvenCpp: `--persist-token` isn't needed here, and in the below tests, any more?

config/manager.go
bartoszWojciechO: I don't necessarily disagree. Imo it would be better to have a techdebt task to discuss this as a code convention and refactor the code where necessary first. Otherwise we'd just have this method that's out of style with the others.

test/qa/test_allowlist_subnet.py
bartoszWojciechO: Yeah, `set defaults` always persists tokens when logging out.

cli/messages.go
```diff
@@ -371,14 +371,15 @@ Provide a [transfer_id] argument to list files in the specified transfer.`
...
+	SetDefaultsLogoutFlagText        = "Log out after restoring settings to their default values. Example: nordvpn set defaults --logout"
+	SetDefaultsOffKillswitchFlagText = "Turn off Kill Switch when restoring settings to their default values. Example: nordvpn set defaults --off-killswitch"
 
 	AnalyticsPolicyLink = "https://my.nordaccount.com/legal/privacy-policy/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=settings_account-privacy_policy&nm=app&ns=nordvpn-linux-cli&nc=settings-privacy_policy"
 	MsgConsentAgreement = `We value your privacy.
 
 That's why we want to be transparent about what data you agree to give us. We only collect the bare minimum of information required to offer a smooth and stable VPN experience.
 
-By pressing "y" (yes), you allow us to collect and use limited app performance data. This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy. 
+By pressing "y" (yes), you allow us to collect and use limited app performance data. This helps us keep our features relevant to your needs and fix issues faster, as explained in our Privacy Policy.
```
mariusSincovici: I think the space should not be removed in this case because it will remove the space from the final message

PR #944, opened 2025-06-24
LVPN-5971: unify UA header value sent out for app and moose

DoYouEvenCpp:
Changes under this PR:
- unify data being used for the UA (currently it's based on distro name in both cases)
- make the UA RFC9110 complianat
- small changes required to make user_agent testable

### DIFF
--- a/request/user_agent.go
+++ b/request/user_agent.go
@@ -1,0 +1,28 @@
+package request
+
+import (
+	"fmt"
+
+	"github.com/NordSecurity/nordvpn-linux/distro"
+)
+
+const appName = "NordApp Linux"
+
+// GetUserAgentValue generates a User-Agent header value compliant with RFC 9110 section 10.1.5.
+// It formats the user agent as "<application-name>/<version> (<platform-details>)" where:
+// - application-name is application identifier
+// - version is the application version
+// - platform-details contains the distro name of the currently running kernel
+//
+// Returns:
+//   - string: The formatted User-Agent string
+//   - error: An error if distribution information cannot be retrieved
+//
+// See: https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.5
+func GetUserAgentValue(version string, d distro.Distro) (string, error) {
+	distro_name, err := d.ReleasePrettyName()
+	if err != nil {
+		return "", fmt.Errorf("determining device os: %w", err)
+	}
+	return fmt.Sprintf("%s/%s (%s)", appName, version, distro_name), nil
+}
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -206,8 +206,10 @@ func main() {
 			log.Fatalln("Error on creating validator:", err)
 		}
 	}
-
-	userAgent := fmt.Sprintf("NordApp Linux %s %s", Version, distro.KernelName())
+	userAgent, err := request.GetUserAgentValue(Version, distro.NewDistro())
+	if err != nil {
+		log.Fatalln("Error on getting user agent value:", err)
+	}
 
 	httpGlobalCtx, httpCancel := context.WithCancel(context.Background())
 
--- a/distro/distro.go
+++ b/distro/distro.go
@@ -23,7 +23,56 @@ type osRelease struct {
 	PrettyName string
 }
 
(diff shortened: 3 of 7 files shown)


bartoszWojciechO (approved):

dfetti (approved):

sysinfo/host_name_dbus_src.go
```diff
@@ -46,8 +46,9 @@ func NewHostOsPrettyName() HostOsPrettyName {
 func (HostOsPrettyNameImpl) GetHostOSPrettyName() (string, error) {
 	conn, err := dbus.SystemBus()
 	if err != nil {
-		log.Printf("%s %s connecting to system dbus: %s", logTag, internal.ErrorPrefix, err)
-		return "", err
+		log.Printf("%s %s connecting to system dbus: %s\n", logTag, internal.ErrorPrefix, err)
+		log.Println(logTag, internal.WarningPrefix, "Attempting to read os-release file instead")
```
thegravi: I suggest not exposing the content of the following function in the log message, because it can change over time, and this message would become misleading.

sysinfo/host_name_dbus_src.go
```diff
@@ -60,7 +61,8 @@ func (HostOsPrettyNameImpl) GetHostOSPrettyName() (string, error) {
 	name, err := dbusutil.GetStringProperty(client, dbusHostname1PropOperatingSystemPrettyName)
 	if err != nil {
 		log.Printf("%s %s retrieving OS pretty name: %s", logTag, internal.WarningPrefix, err)
-		return "", err
+		log.Println(logTag, internal.WarningPrefix, "Attempting to read os-release file instead")
```
thegravi: Same here

thegravi (approved):

request/user_agent.go
```diff
@@ -0,0 +1,28 @@
...
+//   - error: An error if distribution information cannot be retrieved
+//
+// See: https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.5
+func GetUserAgentValue(version string, nameGetter sysinfo.HostOsPrettyName) (string, error) {
+	distro_name, err := nameGetter.GetHostOSPrettyName()
+	if err != nil {
+		return "", fmt.Errorf("determining device os: %w", err)
+	}
+	return fmt.Sprintf("%s/%s (%s)", appName, version, distro_name), nil
+}
```
thegravi: Why not just go with this approach:
```go
// osPrettyNameGetter defines the signature of a function returning OS pretty name
type osPrettyNameGetter func() (string, error)

func GetUserAgentValue(version string, getOsPrettyName osPrettyNameGetter) (string, error) {}
```

And eliminate usage of `NewHostOsPrettyName()`

You can still inject this getter function:
```go
userAgent, err := request.GetUserAgentValue(version, sysinfo.HostOsPrettyName)
```

DoYouEvenCpp: Good point!
Fixed.

sysinfo/host_name_dbus_src.go
DoYouEvenCpp: Fixed

sysinfo/host_name_dbus_src.go
DoYouEvenCpp: Fixed

mariusSincovici (approved):

daemon/api_repo.go
```diff
@@ -52,14 +55,12 @@ func (api *RepoAPI) DebianFileList() ([]byte, error) {
 
 	resp, err := api.request(fmt.Sprintf(core.DebFileinfoURLFormat, repoType, api.arch))
 	if err != nil {
-		//log.Printf("DebianFileList failed to fetch fileinfo. Error: %v.\n", err)
```
bartoszWojciechO, line 55: Why the comments? If there would be no logs, would it be possible to wrap the error so that it would be easier to trace when it's eventually logged? Something like `return nil, fmt.Errorf("Failed to fetch fileinfo: %w", err"`)

DoYouEvenCpp: Note, it used to be commented out in the first place :wink:
But sure, can restore and reuse them. Good idea!

PR #945, opened 2025-06-25
LVPN-6897: Add remote config fetcher

keliramu:


cmd/daemon/main.go
```diff
@@ -305,6 +304,11 @@ func main() {
 	daemonEvents.Service.Connect.Subscribe(loggerSubscriber.NotifyConnect)
 	daemonEvents.Settings.Publish(cfg)
 
+	rcConfig := getRemoteConfigGetter(Version, Environment, RemotePath, cdnAPI)
+	if err := rcConfig.LoadConfig(); err != nil {
+		log.Fatalln(err)
```
mariusSincovici: Why should the daemon get killed when RC loading fails. How can the users fix this? Wouldn't this kill the daemon at the OS startup when there is no internet connection?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+			return fmt.Errorf("writing/renaming files: %w", err)
+		}
+
+		// after all is done, set new hash
+		f.Hash = mainJsonHash.Hash
+	}
+	return nil
+}
+
+func walkFiles(targetPath, fileExt string, actionFunc func(string)) error {
```
mariusSincovici: [suggestion] there are some functions, like this, that are not specific to this feature and maybe they can be moved into FileSystem class?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return info.IsDir(), nil
+}
+
+func isHashValid(targetHash string, data ...[]byte) bool {
```
mariusSincovici: Isn't this function checking if hashes are equal to the provided data? Maybe a more suitable name would be `isHashEqual`?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+	bytesForHash := []byte{}
+	for _, b := range data {
```
mariusSincovici: [small] maybe you can use `bytes.Join(data, nil)`?
if not it would be good to have a check that len(b) + len(bytesForHash) is not allocating infinite data.

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+// cleanupFiles on transaction failure - remove tmp files
+func cleanupFiles(targetPath, fileExt string) error {
+	return walkFiles(targetPath, fileExt, func(path string) {
+		if err := os.Remove(path); err != nil {
+			log.Printf(internal.ErrorPrefix+" removing %s: %s\n", path, err)
+		}
+	})
+}
+
+func isValidExistingDir(path string) (bool, error) {
```
mariusSincovici: [small] `DirExists`, then is in sync with FileExists from FileSystem?

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if !result.Valid() {
+		var errStr string
+		for idx, desc := range result.Errors() {
+			if errStr == "" {
+				errStr = fmt.Sprintf("[%d] %s", idx, desc)
+			} else {
+				errStr = fmt.Sprintf("%s;; [%d] %s", errStr, idx, desc)
+			}
+			return fmt.Errorf(errStr)
```
mariusSincovici: Should this be here or outside of `for`?

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	schemaLoader := gojsonschema.NewStringLoader(string(schema))
+
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+	if err != nil {
+		return err
+	}
+	if !result.Valid() {
+		var errStr string
+		for idx, desc := range result.Errors() {
+			if errStr == "" {
```
mariusSincovici: Isn't this always true?

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	schemaLoader := gojsonschema.NewStringLoader(string(schema))
+
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+	if err != nil {
+		return err
+	}
+	if !result.Valid() {
+		var errStr string
+		for idx, desc := range result.Errors() {
```
mariusSincovici, line 41: same here.
It seams duplicated code, can't it be extracted into a separate function?

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	_ "embed"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/xeipuuv/gojsonschema"
+)
+
+// Validator generic interface for config validator
+type Validator interface {
```
mariusSincovici: At the moment I think `JsonValidator` is always used. Are there plans to change to use the interface?
Because for now JsonValidator doesn't seam to have any logic just to implement the interface.

config/remote/data.go
```diff
@@ -0,0 +1,88 @@
...
+	Version    int               `json:"version"` // JSON schema version
+	Configs    []Param           `json:"configs"`
+	Schema     string            `json:"schema"`
+	Name       string            // file name (main part)
+	Hash       string            // JSON file hash (from last download)
+	Params     map[string]*Param // parsed params
+	NeedUpdate bool
+}
+
+func (f Feature) FileName(basePath string) string {
```
mariusSincovici: Maybe change to `FilePath` instead, because there is already a Name member variable.
Same for `HashFilePath`.

config/remote/data.go
```diff
@@ -0,0 +1,88 @@
...
+	Type     string       `json:"value_type"`
+	Settings []ParamValue `json:"settings"`
+}
+
+// Feature is set of configs from one JSON file
+type Feature struct {
+	Version    int               `json:"version"` // JSON schema version
+	Configs    []Param           `json:"configs"`
+	Schema     string            `json:"schema"`
+	Name       string            // file name (main part)
```
mariusSincovici: Is this the feature name or the file name, or both? :smile:

config/remote/data.go
```diff
@@ -0,0 +1,88 @@
...
+	Hash       string            // JSON file hash (from last download)
```
mariusSincovici: Won't some members be in JSON with upper case while some with lower case because of missing `json:"`?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	// validate json by predefined schema
+	if err := NewJsonValidator().ValidateString(mainJsonStr); err != nil {
+		return fmt.Errorf("validating json: %w", err)
+	}
+
+	// load json into structures
+	var temp Feature
+	if err := json.Unmarshal(mainJsonStr, &temp); err != nil {
```
mariusSincovici: [suggestion] move this Unmarshal after checking if include files exist, because if a file is missing then is not needed.

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+		return err
+	}
+
+	validateIncludeFilesFunc := func(incFileName string) ([]byte, error) {
```
mariusSincovici: It might be good to move this outside so that the function can be tested?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+		return incJsonStr, nil
+	} //func()
+
+	incFiles, err := f.walkIncludeFiles(mainJsonStr, validateIncludeFilesFunc)
+	if err != nil {
+		return fmt.Errorf("loading include files: %w", err)
+	}
+
+	// verify content integrity
+	// if main json has include failes - hash should cover whole content
```
mariusSincovici: 'failes` -> files?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+			return nil, fmt.Errorf("loading include file hash: %w", err)
+		}
+		var incJsonHash jsonHash
+		if err = json.Unmarshal(incHashStr, &incJsonHash); err != nil {
+			return nil, fmt.Errorf("parsing include file hash: %w", err)
+		}
+		if !isHashValid(incJsonHash.Hash, incJsonStr) {
+			return nil, fmt.Errorf("include file integrity problem")
+		}
+		return incJsonStr, nil
```
mariusSincovici: shouldn't file schema also be checked here?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+	if !isHashValid(mainJsonHash.Hash, mainJsonStr, incFiles) {
+		return fmt.Errorf("main file integrity problem")
+	}
+
+	f.Params = make(map[string]*Param)
+	for _, cfgItem := range temp.Configs {
+		f.Params[cfgItem.Name] = &Param{Name: cfgItem.Name, Type: cfgItem.Type, Settings: []ParamValue{}}
+		for _, prm := range cfgItem.Settings {
+			var incVal string
+			switch cfgItem.Type {
```
mariusSincovici: This should go in a separate function that can be tested.
Maybe also have a `default` case that logs the skipped info in debug mode?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+					return fmt.Errorf("loading json object [%s]: %w", val, err)
+				}
+			case "file": //nolint
+				// primary field value is an include file name
+				val, err := prm.AsString()
+				if err != nil {
+					return fmt.Errorf("loading include file name as string value [%s]: %w", prm.Value, err)
+				}
+				// include file expect to be located in the same directory as main json file
+				dir, _ := filepath.Split(mainJsonFileName)
```
mariusSincovici: isn't dir == sourcePath?

config/remote/data.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	featureMain     = "nordvpn"
+	featureLibtelio = "libtelio"
+)
+
+type ParamValue struct {
+	Value      any    `json:"value"`
+	IncValue   string // include file content
```
mariusSincovici: won't this also be upper case in json?

Can `IncValue` contain something else than the other file content?

cmd/daemon/main.go
```diff
@@ -305,6 +304,11 @@ func main() {
 	daemonEvents.Service.Connect.Subscribe(loggerSubscriber.NotifyConnect)
 	daemonEvents.Settings.Publish(cfg)
 
+	rcConfig := getRemoteConfigGetter(Version, Environment, RemotePath, cdnAPI)
```
devzbysiu: After some PRs were merged, there is now `config.BuildTarget` struct which contains `Version` and `Environment`, so just FYI, maybe we could use it here

cmd/daemon/main.go
devzbysiu: Maybe we could add some message here to make it  easier to find the source of an error?

cmd/daemon/no_cdnrc_cfg.go
```diff
@@ -12,6 +12,13 @@ func (r RemoteConfigGetterStub) GetTelioConfig() (string, error) {
 	return "", fmt.Errorf("no remote config getter was compiled into the app")
 }
 
-func getRemoteConfigGetter(string) RemoteConfigGetterStub {
+type RemoteStorage interface {
```
devzbysiu: Maybe this could be moved to some other place and shared to not duplicate it in `cdnrc_cfg.go`?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+		return fmt.Errorf("downloading main hash file: %w", err)
+	}
+
+	var mainJsonHash jsonHash
+	if err = json.Unmarshal(mainJsonHashStr, &mainJsonHash); err != nil {
+		return fmt.Errorf("parsing main hash file: %w", err)
+	}
+
+	// main hash covers the include files as well
+	if f.Hash != mainJsonHash.Hash {
```
devzbysiu: [small] maybe we can invert this and have precondition with early exit, so it will be less nesting 
```go
if f.Hash == mainJsonHash.Hash {
  return nil
}
```

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+
+// walkIncludeFiles
+func (f *Feature) walkIncludeFiles(mainJason []byte, fileActionFunc func(string) ([]byte, error)) ([]byte, error) {
+	var temp Feature
+	if err := json.Unmarshal(mainJason, &temp); err != nil {
+		return nil, err
+	}
+	incFilesJson := []byte{}
+	for _, cfgItem := range temp.Configs {
+		for _, prm := range cfgItem.Settings {
```
devzbysiu: `prm` stands for `primary`?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+				incVal = string(jsn)
+			}
+			f.Params[cfgItem.Name].Settings = append(f.Params[cfgItem.Name].Settings,
+				ParamValue{Value: prm.Value, IncValue: incVal, AppVersion: prm.AppVersion, Weight: prm.Weight})
+		}
+	}
+	return nil
+}
+
+// walkIncludeFiles
```
devzbysiu: TODO? :smile:

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+
+	// main hash covers the include files as well
+	if f.Hash != mainJsonHash.Hash {
+		mainJsonStr, err := cdn.GetRemoteFile(f.FileName(cdnBasePath))
+		if err != nil {
+			return fmt.Errorf("downloading main file: %w", err)
+		}
+
+		// validate json against predefined schema
+		if err = NewJsonValidator().ValidateString(mainJsonStr); err != nil {
```
devzbysiu: [small] maybe this could also be a dependency so make testing easier? 
On the other hand we can manipulate main JSON, so we can testing error conditions here also

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	constraint, err := semver.NewConstraint(verConstraint)
+	if err != nil {
+		return false, fmt.Errorf("invalid version constraint: %w", err)
+	}
+	return constraint.Check(v), nil
+}
+
+// download main json file and check if include files should be downloaded
+func (f *Feature) download(cdn RemoteStorage, cdnBasePath, targetPath string) (err error) {
```
devzbysiu: Maybe we could split the responsibility of this so the testing would be easier? Right now `RemoteStorage` is responsible for downloading, but then this func does schema validation, include resolution, hash validation, File IO

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+type FeatureConfig interface {
+	IsFeatureEnabled(featureName string) bool
+	GetFeatureParam(featureName, paramName string) (string, error)
+}
+
+type ConfigLoader interface {
+	LoadConfig() error
+}
+
+var (
```
mariusSincovici: [small] `const`?

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+		Features:       make(FeatureMap),
+	}
+	rc.Features.Add(featureMain)
+	rc.Features.Add(featureLibtelio)
+	return rc
+}
+
+// LoadConfig download from remote or load from disk
+func (c *CdnRemoteConfig) LoadConfig() error {
+	useOnlyLocalConfig := c.appEnvironment == "dev" && os.Getenv(envUseLocalConfig) != "" // forced load from disk?
```
mariusSincovici: Maybe use `internal.IsDevEnv(Environment)` instead of string comparison?

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+			if err := f.download(c.cdn, filepath.Join(c.remotePath, c.appEnvironment), c.localCachePath); err != nil {
+				log.Println(internal.ErrorPrefix, "failed downloading config for [", f.Name, "]:", err)
+				continue
+			}
+			log.Println(internal.DebugPrefix, "Feature [", f.Name, "] config downloaded.")
+		}
+	}
+
+	// lock only when loading from local files
+	c.mu.Lock()
```
mariusSincovici, line 144: why needed to lock only when loading from file? There are also changes to the internal feature struct when downloading the file.

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+	envUseLocalConfig = "USE_LOCAL_CONFIG"
+)
+
+type CdnRemoteConfig struct {
+	appVersion     string
+	appEnvironment string
+	localCachePath string
+	remotePath     string
+	cdn            RemoteStorage
+	Features       FeatureMap
```
mariusSincovici: Does this need to be accessible from outside?

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+		sort.Slice(matches, func(i, j int) bool {
+			return matches[i].Weight > matches[j].Weight
+		})
+		return &matches[0]
+	}
+	return nil
+}
+
+// TODO/FIXME: add `rollout` support
+func (c *CdnRemoteConfig) GetTelioConfig() (string, error) {
```
mariusSincovici, line 179: Couldn't this reuse `GetFeatureParam`, or maybe both use some common function?
In this way the `item` type can also be verified before returning `item.IncValue`.

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return data.Version, nil
+}
+func extractVersionJsonStr(jsonStr []byte) (int, error) {
+	var data struct {
+		Version int `json:"version"`
+	}
+	if err := json.Unmarshal(jsonStr, &data); err != nil {
+		return 0, fmt.Errorf("parsing json file: %w", err)
```
mariusSincovici: [small] `json file`, the message might not be fully correct. The parameter in this case is a []byte and not a file.

core/cdn.go
```diff
@@ -124,7 +124,7 @@ func (api *CDNAPI) ConfigTemplate(isObfuscated bool, method string) (http.Header
 	if err != nil {
 		return nil, nil, err
 	}
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
```
mariusSincovici: Maybe this should be changed to `MaxBytesReadAll` instead?

config/remote/validator_test.go
```diff
@@ -0,0 +1,42 @@
+package remote
```
devzbysiu, line 1: Would be nice to add negative tests

core/cdn_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/NordSecurity/nordvpn-linux/daemon/response"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/events/subs"
+	"github.com/NordSecurity/nordvpn-linux/request"
+)
+
+func TestCDN(t *testing.T) {
+	var err error
+	var validator response.Validator
+	if true {
```
devzbysiu: TODO?

config/remote/remote_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/daemon/response"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/events/subs"
+	"github.com/NordSecurity/nordvpn-linux/request"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
```
devzbysiu: [small] `const`?

config/remote/config.go
```diff
@@ -0,0 +1,384 @@
...
+		if err != nil {
+			return fmt.Errorf("downloading main file: %w", err)
+		}
+
+		// validate json against predefined schema
+		if err = NewJsonValidator().ValidateString(mainJsonStr); err != nil {
+			return fmt.Errorf("validating main: %w", err)
+		}
+
+		downloadIncludeFilesFunc := func(incFileName string) ([]byte, error) {
```
devzbysiu: This is very similar to `validateIncludeFilesFunc`, maybe this could be extracted and reused at least partially?

config/remote/validator.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return data.Version, nil
+}
+
+func (v *JsonValidator) ValidateFile(name string) error {
```
devzbysiu: Is this needed? I think it's not used anywhere :thinking:

config/remote/remote.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+type CdnRemoteConfig struct {
+	appVersion     string
+	appEnvironment string
+	localCachePath string
+	remotePath     string
+	cdn            RemoteStorage
+	Features       FeatureMap
+	mu             sync.Mutex
```
devzbysiu: `LoadConfig` is called as a job, may this introduce some random latency during `GetFeatureParam`/`IsFeatureEnabled`/`GetTelioConfig`? Maybe `RWMutex` would be safer?

config/remote/data.go
```diff
@@ -0,0 +1,88 @@
...
+	v, ok := pv.Value.(bool)
+	if !ok {
+		return false, fmt.Errorf("not a bool")
+	}
+	return v, nil
+}
+func (pv ParamValue) AsStringArray() ([]string, error) {
+	v, ok := pv.Value.([]any)
+	if !ok {
+		return nil, fmt.Errorf("not an array")
```
devzbysiu: [small] Maybe would be nice to also add `%T` to have an info what type we got

cmd/daemon/main.go
keliramu: ok

cmd/daemon/no_cdnrc_cfg.go
keliramu: In `go` it is not like in `C` to be it like `include` and used in all places, in `go` interface is defined on consumer side, usually

config/remote/config.go
keliramu: Include file while still `json` but can be of arbitrary content e.g. `libtelio` config and sometimes it is not possible to have schema or it would be restrictive to force for it to have schema defined.

config/remote/config.go
keliramu: yes, it is, changed

config/remote/config.go
keliramu: it's `parameter`, changed var name to `param`

config/remote/data.go
keliramu: `IncValue` does not have `json`, so, it can be private (lower case). Its value is set to include file content.

config/remote/data.go
keliramu: yes, sure, thx

config/remote/data.go
keliramu: It is `feature` name which is used to construct a config file name

config/remote/data.go
keliramu: some member variables does not have `json` and can be private /  lover case, changed

config/remote/data.go
keliramu: make sense, changed, thx

config/remote/remote.go
keliramu: ok

config/remote/remote.go
keliramu: good point, only interface is used outside, thx

config/remote/remote.go
keliramu: yes, changed

config/remote/remote.go
keliramu: Changed `download` part not to modify `feature` (in download part new hash value was set - eliminated that)

keliramu: `download` part is communicating to the network, so, might be variable in time - would be good to avoid locking there.

config/remote/remote.go
keliramu: yes, good catch, thx

config/remote/remote_test.go
keliramu: yes, thx

config/remote/config.go
keliramu: yes, good catch, thx

config/remote/config.go
keliramu: ok

config/remote/remote.go
keliramu: no, actually need this type to be returned in `main` package so, it has to be public

config/remote/validator.go
keliramu: it's true only on 1st iteration

config/remote/validator.go
keliramu: yes, outside, thx

config/remote/validator.go
keliramu: removed `ValidateFile` as it is not used now, no more duplicate code

config/remote/validator.go
keliramu: not needed for now, removed

config/remote/validator.go
keliramu: fixed

config/remote/validator.go
keliramu: Ok, removed interface - not needed for now

core/cdn.go
keliramu: yes, good catch, thx

config/remote/config.go
keliramu: moved

config/remote/config.go
keliramu: added `bytes.Join` also added file size check in other place - now validate json file size being downloaded and loaded from the disk - prevention against too big files.

cmd/daemon/main.go
keliramu: right, no need to kill daemon in case of RC load failure.

keliramu: Added.

cmd/daemon/no_cdnrc_cfg.go
devzbysiu: Right. Yeah and the duplication is only because of build directives for `cdnrc`. Ok :+1:

config/remote/remote.go
keliramu: `LoadConfig` involves communication to the network but it has timeouts of 15sec, but locking is done only after files are finished downloading from server and when info is being loaded from the file on disk which is more predictable in how much time it can take (it should be fast).

core/cdn_test.go
keliramu: File removed, `cdn` is being unit-tested in `remote_test.go`

config/remote/validator_test.go
keliramu: Added.

config/remote/config.go
keliramu: added comment

config/remote/config.go
keliramu: ok, moved generic functions

config/remote/config.go
keliramu: yes, moved to FileSystem

config/remote/config.go
keliramu: yes, done

config/remote/config.go
keliramu: yes, done

config/remote/config.go
keliramu: yes, done

config/remote/config.go
keliramu: yes, done

config/remote/remote.go
```diff
@@ -0,0 +1,224 @@
...
+	IsFeatureEnabled(featureName string) bool
+	GetFeatureParam(featureName, paramName string) (string, error)
+}
+
+type ConfigLoader interface {
+	LoadConfig() error
+}
+
+const (
+	envUseLocalConfig = "USE_LOCAL_CONFIG"
```
devzbysiu: [small] Maybe would be good to have `NORDVPN_` prefix here like `NORDVPN_USE_LOCAL_CONFIG` to minimize the risk of conflicts if something else has such env var

devzbysiu, line 36: Maybe it would be safer to have `NORDVPN_` prefix here to avoid conflicts if something else uses such env var

config/remote/remote.go
```diff
@@ -0,0 +1,224 @@
...
+			log.Println(internal.ErrorPrefix, "failed loading feature [", f.name, "] config from the disk:", err)
+			continue
+		}
+		log.Println(internal.InfoPrefix, "feature [", f.name, "] config loaded from:", c.localCachePath)
+	}
+
+	return nil
+}
+
+func findMatchingRecord(ss []ParamValue, ver string) *ParamValue {
```
devzbysiu: We probably don't need to store the `matches` and sort them if we take only one with highest  `Weight`, we could just keep one and update it if matching record appears with higher `Weight`

cmd/daemon/cdnrc_cfg.go
```diff
@@ -0,0 +1,17 @@
...
+
+package main
+
+import (
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+type RemoteStorage interface {
```
mariusSincovici: This is defined in 3 places. Isn't it possible to reuse the one from `remote`?

cmd/daemon/main.go
```diff
@@ -326,6 +325,11 @@ func main() {
 	daemonEvents.Service.Connect.Subscribe(loggerSubscriber.NotifyConnect)
 	daemonEvents.Settings.Publish(cfg)
 
+	rcConfig := getRemoteConfigGetter(buildTarget, RemotePath, cdnAPI)
+	if err := rcConfig.LoadConfig(); err != nil {
```
mariusSincovici: This will make lots of operations including download from the internet, wouldn't this slowdown the daemon startup?
Also the same will be one more time executed when the Jobs are starting in `StartJobs`

config/remote/config.go
```diff
@@ -0,0 +1,334 @@
...
+			}
+			// validate field
+			incVal, err := validateField(cfgItem, param, fileReadFunc)
+			if err != nil {
+				return err
+			}
+			// store valid values in the map
+			f.params[cfgItem.Name].Settings = append(f.params[cfgItem.Name].Settings,
+				ParamValue{Value: param.Value, incValue: string(incVal), AppVersion: param.AppVersion, Weight: param.Weight})
+			// after all is done, set new hash //TODO/FIXME: test this!
```
devzbysiu: Do we have a test for this already?

devzbysiu (approved):

daemon/jobs.go
```diff
@@ -37,6 +39,15 @@ func (r *RPC) StartJobs(
 		log.Println(internal.WarningPrefix, "job countries schedule error:", err)
 	}
 
+	_, err := r.scheduler.NewJob(gocron.DurationJob(3*time.Minute), gocron.NewTask(func() { // TODO/FIXME: change time/frequency
```
mariusSincovici: [small] 
* isn't 3 Minutes too often? Also maybe move the duration in a const like `heartBeatPeriod`.
* And if you want to test that the test that the job works move it in a separate function with a duration param.

daemon/jobs.go
```diff
@@ -37,6 +39,15 @@ func (r *RPC) StartJobs(
...
+		if err := remoteConfigLoader.LoadConfig(); err != nil {
+			log.Printf("JobConfigLoader: %s", err)
```
mariusSincovici: `internal.ErrorString`?

config/remote/remote.go
mariusSincovici: Since it is not a bool, maybe RC_LOCAL_CONFIG_PATH, in case we'll have other configs in the future?

config/remote/config.go
```diff
@@ -0,0 +1,334 @@
...
+		return fmt.Errorf("main file integrity problem")
+	}
+
+	// load json into structures
+	var temp Feature
+	if err := json.Unmarshal(mainJsonStr, &temp); err != nil {
+		return err
+	}
+
+	f.params = make(map[string]*Param)
```
mariusSincovici: if there is an error after this `f.params` would remain with partial or incorrect data. is it ok?
same with `f.hash`

config/remote/config.go
```diff
@@ -0,0 +1,334 @@
...
+			// validate field
+			incVal, err := validateField(cfgItem, param, fileReadFunc)
+			if err != nil {
+				return err
+			}
+			// store valid values in the map
+			f.params[cfgItem.Name].Settings = append(f.params[cfgItem.Name].Settings,
+				ParamValue{Value: param.Value, incValue: string(incVal), AppVersion: param.AppVersion, Weight: param.Weight})
+			// after all is done, set new hash //TODO/FIXME: test this!
+			f.hash = hash(bytes.Join([][]byte{mainJsonStr, incFiles}, nil))
```
mariusSincovici: If `download` and `load` are called in the same time there can be a race on `f.hash` or incorrect value is used.
Also does it need to be inside `for`? because it seams to not be related to the `for`? :thinking:

config/remote/data.go
```diff
@@ -0,0 +1,98 @@
...
+	fieldTypeNumber = "number"
+	fieldTypeBool   = "boolean"
+	fieldTypeArray  = "array"
+	fieldTypeObject = "object"
+	fieldTypeFile   = "file"
+)
+
+type ParamValue struct {
+	Value      any    `json:"value"`
+	incValue   string // include file content
```
mariusSincovici: shouldn't this be ignored by json?

config/remote/data.go
```diff
@@ -0,0 +1,98 @@
...
+	Type     string       `json:"value_type"`
+	Settings []ParamValue `json:"settings"`
+}
+
+// Feature is set of configs from one JSON file
+type Feature struct {
+	Version int               `json:"version"` // JSON schema version
+	Configs []Param           `json:"configs"`
+	Schema  string            `json:"schema"`
+	name    string            // file name (main part)
```
mariusSincovici, line 80: same for this with this params, shouldn't they be ignored by unmarshal?

config/remote/remote.go
```diff
@@ -0,0 +1,224 @@
...
+	matches := []ParamValue{}
+	for _, s := range ss {
+		// find all my version matching records
+		ok, err := isVersionMatching(ver, s.AppVersion)
```
mariusSincovici, line 160: we should add more selectors in here, otherwise older version might not be able to correctly select.

config/remote/data.go
```diff
@@ -0,0 +1,98 @@
...
+	hash    string            // JSON file hash (from last download)
+	params  map[string]*Param // parsed params
+}
+
+func (f Feature) FilePath(basePath string) string {
+	return filepath.Join(basePath, f.name) + ".json"
```
mariusSincovici, line 86: [small] maybe add a test for this and next to check if the names + extensions are correct.

config/remote/remote.go
keliramu: Changed env var name to `RC_USE_LOCAL_CONFIG` and it is just a flag. Avoiding possibility to specify just another path for local configs ending with multiple configs and possibly having confusion which is which.

config/remote/remote.go
keliramu: created task for it.

config/remote/remote.go
keliramu: yes, thx

config/remote/data.go
keliramu: unexported fields are ignored by json.unmarshal

config/remote/data.go
keliramu: unexported fields are ignored by json.unmarshal

config/remote/config.go
keliramu: good catch, changed to using intermediate variable, and hash to be set after the loop

cmd/daemon/cdnrc_cfg.go
keliramu: made only single place

cmd/daemon/no_cdnrc_cfg.go
keliramu: made interface definition in single place

config/remote/config.go
keliramu: good catch, added RWMutex

config/remote/data.go
keliramu: added unit test

config/remote/config.go
keliramu: done

daemon/jobs.go
keliramu: added error prefix

daemon/jobs.go
keliramu: added possibility to specify env variable with number of minutes between job runs

cmd/daemon/main.go
keliramu: implemented RC non-blocking load on start with multiple attempts and with exponential backoff on failure

config/remote/config.go
```diff
@@ -0,0 +1,335 @@
...
+	readFile(string) ([]byte, error)
+}
+
+type validator interface {
+	validate([]byte) error
+}
+
+// download main json file and check if include files should be downloaded,
+// return `true` if remote config was really downloaded.
+func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBasePath, targetPath string) (success bool, err error) {
```
mariusSincovici, line 62: [very minor] `success bool` at the moment is only used for logging, maybe it is not needed?

config/remote/config.go
```diff
@@ -0,0 +1,335 @@
...
+	if err = json.Unmarshal(incHashStr, &incJsonHash); err != nil {
+		return nil, fmt.Errorf("parsing include file hash: %w", err)
+	}
+	if !isHashEqual(incJsonHash.Hash, incJsonStr) {
+		return nil, fmt.Errorf("include file integrity problem, expected hash[%s] got hash[%s]", incJsonHash.Hash, hash(incJsonStr))
+	}
+	// write an include json to file
+	incFileTargetPath := filepath.Join(trgBasePath, incFileName) + tmpExt
+	if err = fw.writeFile(incFileTargetPath, incJsonStr, internal.PermUserRW); err != nil {
+		return nil, fmt.Errorf("wrting include file: %w", err)
```
mariusSincovici: [spelling] `wrting` also bellow

config/remote/config.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		fileVal, err = fileReader(val)
+		if err != nil {
+			return nil, fmt.Errorf("loading include file name as string value [%s]: %w", paramVal.Value, err)
+		}
+	}
+	return fileVal, nil
+}
+
+// walkIncludeFiles iterate through include files and do action on them
```
mariusSincovici: ` and do action on them` is this leftover from when the function was generic?

config/remote/config.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	var incJsonHash jsonHash
+	if err = json.Unmarshal(incHashStr, &incJsonHash); err != nil {
+		return nil, fmt.Errorf("parsing include file hash: %w", err)
+	}
+	if !isHashEqual(incJsonHash.Hash, incJsonStr) {
+		return nil, fmt.Errorf("include file integrity problem, expected hash[%s] got hash[%s]", incJsonHash.Hash, hash(incJsonStr))
+	}
+	// write an include json to file
+	incFileTargetPath := filepath.Join(trgBasePath, incFileName) + tmpExt
```
mariusSincovici: [suggestion] Would it be ok to write only when `trgBasePath != ""`?

daemon/jobs.go
```diff
@@ -98,6 +103,46 @@ func (r *RPC) StartJobs(
 	}()
 }
 
+func (r *RPC) StartRemoteConfigLoaderJob(
+	remoteConfigLoader remote.ConfigLoader,
+) {
+	// on first try - load remote config in non-blocking goroutine
+	go func(rcl remote.ConfigLoader) {
+		// try to load remote config 5 times with exponential backoff
+		for i := 0; i < 5; i++ {
```
mariusSincovici, line 112: [suggestion from IDE :smile: ] `for i := range 5 {`

daemon/jobs.go
```diff
@@ -98,6 +103,46 @@ func (r *RPC) StartJobs(
...
+	if internal.IsDevEnv(string(r.environment)) && os.Getenv(envRcLoadTime) != "" {
+		tm, err := strconv.Atoi(os.Getenv(envRcLoadTime))
+		if err != nil {
+			log.Println(internal.WarningPrefix, "converting rc load time:", err)
+		} else {
+			if tm > 3 && tm < 100 {
+				rcLoadTime = time.Duration(tm) * time.Minute
+			}
+		}
+	}
```
devzbysiu, line 135: [small] maybe a log msg would be useful here to log what's the final `rcLoadTime` to be sure that override worked when used

config/remote/remote.go
```diff
@@ -0,0 +1,242 @@
...
+func (c *CdnRemoteConfig) LoadConfig() error {
+	useOnlyLocalConfig := internal.IsDevEnv(c.appEnvironment) && os.Getenv(envUseLocalConfig) != "" // forced load from disk?
+	if !useOnlyLocalConfig {
+		c.mu.RLock()
+		for _, f := range c.features {
+			dnld, err := f.download(cdnFileGetter{cdn: c.cdn}, jsonFileReaderWriter{}, jsonValidator{}, filepath.Join(c.remotePath, c.appEnvironment), c.localCachePath)
+			if err != nil {
+				log.Println(internal.ErrorPrefix, "failed downloading feature [", f.name, "] remote config:", err)
+				if isNetworkRetryable(err) {
+					return err
```
mariusSincovici: missing `RUnlock`.
Might be easier to move this entire block including `useOnlyLocalConfig := internal.IsDev` to a separate function `downloadFiles` and then you might be able to use defer unlock?

daemon/job_templates.go
```diff
@@ -24,16 +25,19 @@ func JobTemplates(cdn core.CDN) func() {
 
 			headers, _, err := cdn.ConfigTemplate(isObfuscated, http.MethodHead)
 			if err != nil {
+				log.Println(internal.WarningPrefix, "doanloding (MethodHead) config template:", err)
```
mariusSincovici: [spelling] `doanloding` also for others

cmd/daemon/main.go
```diff
@@ -326,6 +325,8 @@ func main() {
 	daemonEvents.Service.Connect.Subscribe(loggerSubscriber.NotifyConnect)
 	daemonEvents.Settings.Publish(cfg)
 
+	rcConfig := getRemoteConfigGetter(buildTarget, RemotePath, cdnAPI)
```
mariusSincovici, line 328: I think the download will fail when KS is enabled and not connected to VPN, shouldn't it work? :thinking:

keliramu: there is separate task for fwmark

config/remote/config.go
keliramu: fixed

config/remote/config.go
keliramu: It's needed later to correctly write to the log if feature config file was really downloaded, or it is not if there is no changes on the remote.

daemon/jobs.go
keliramu: Right, I had this suggestion and implementation, but CI linter was against it, so, had to get back to this implementation.

config/remote/remote.go
keliramu: Split into multiple functions to be able to leverage `defer`

cmd/daemon/main.go
keliramu: for `fwmark` have created separate task

daemon/jobs.go
keliramu: ok, added log, it's useful

config/remote/config.go
keliramu: `trgBasePath=""` is valid case

config/remote/config.go
keliramu: yes: TestGetUpdatedTelioConfig

PR #950, opened 2025-06-26
LVPN-7695 use native sqlite

Savolro:


.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,5 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.1.1
+      # TODO: Change this to a tagged version
```
Savolro: Please resolve this TODO

ci/docker/builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/nordsecurity/nordvpn-linux/golang:ubuntu-20.04-20241203
```
edvyno1, line 1: why swap from ubuntu to debian?

edvyno1: explained in later docker image, EOL and glibc versions

ci/docker/golang/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
+# TODO: Change base image to ubuntu:22.04 on 2026-06-30 (glibc 2.35)
 # Then to debian:bookworm-slim on 2027-06-01 (glibc 2.36)
 # Then to ubuntu:24:04 on 2028-06-30 (glibc 2.39)
 
 LABEL org.opencontainers.image.source=https://github.com/NordSecurity/nordvpn-linux
 
 # Based on: https://github.com/docker-library/golang/blob/master/1.22/bookworm/Dockerfile
-ENV GOLANG_VERSION 1.23.3
-ENV GOLANG_SHA256 a0afb9744c00648bafb1b90b4aba5bdb86f424f02f9275399ce0c20b93a2c3a8
+ENV GOLANG_VERSION 1.22.12
```
mariusSincovici, line 10: why is the version decreased?

ci/docker/golang/Dockerfile
```diff
@@ -1,15 +1,14 @@
-FROM ubuntu:20.04
-# (glibc 2.29)
-# TODO: Change base image to to debian:bullseye-slim on 2025-05-29 (glibc 2.31)
-# Then to ubuntu:22.04 on 2026-06-30 (glibc 2.35)
+FROM debian:bullseye-slim
+# (glibc 2.31)
```
mariusSincovici, line 2: Was the comment from before wrong regarding the glibc version?
As far as I've found `ubuntu:20.04` has also version 2.31:thinking:

contrib/scriptlets/deb/postinst
```diff
@@ -33,6 +33,11 @@ echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
 # literally nothing) fixes the issue.
 ldconfig -v > /dev/null 2>&1
 
+ln -sf $(ldconfig -p | awk '/libsqlite3.so.0/ && /=>/ {print $NF; exit}') /usr/lib/nordvpn/libsqlite3.so
```
mariusSincovici, line 36: why is this needed?

magefiles/mage.go
```diff
@@ -20,15 +20,15 @@ import (
 
 const (
 	registryPrefix         = "ghcr.io/nordsecurity/nordvpn-linux/"
-	imageBuilder           = registryPrefix + "builder:1.3.4"
-	imagePackager          = registryPrefix + "packager:1.3.1"
-	imageDepender          = registryPrefix + "depender:1.3.1"
-	imageSnapPackager      = registryPrefix + "snaper:1.0.0"
-	imageProtobufGenerator = registryPrefix + "generator:1.4.1"
+	imageBuilder           = registryPrefix + "builder:1.4.0"
+	imagePackager          = registryPrefix + "packager:1.3.2"
```
mariusSincovici, line 24: [small] since the base OS  for the new docker images is changed to a different distro and version, shouldn't the minor version be increased?

ci/docker/golang/Dockerfile
Savolro: go-licenses (depender image) was misbehaving when there is a a mismatch between toolchain version defined in the `go.mod` file and the actual one.

ci/docker/golang/Dockerfile
Savolro: Seems like so

contrib/scriptlets/deb/postinst
Savolro: Explained in the commit message. As this is done in multiple places for all of the packages, I decided not to add it separately. Do you think it would make sense to do so?

magefiles/mage.go
Savolro: I treated this as internal changes that did not change functionality of the images, that would usually happen with our images by adding or removing packages. On the other hand, built-in os functionality changed. The difference is that we generally do not care about those.

mariusSincovici (approved):

edvyno1 (approved):

ci/docker/ruster/Dockerfile
```diff
@@ -53,17 +53,6 @@ RUN set -eux; \
     /usr/lib/arm-linux-gnueabi \
     /usr/lib/aarch64-linux-gnu
 
-RUN --mount=type=secret,id=gl_access_token export GL_ACCESS_TOKEN="$(cat /run/secrets/gl_access_token)"; \
```
devzbysiu, line 56: There is also readme for  this image that suggests way to build it with build arg and secret mount which we don't need anymore if this is removed, so probably whole readme can go

devzbysiu (approved):

ci/docker/ruster/Dockerfile
```diff
@@ -75,29 +63,29 @@ RUN set -eux; \
...
+    ldconfig; \
+    rm -rf /var/lib/apt/lists/*
 
-RUN set -eux; \
-  dpkg --add-architecture i386;\
-  dpkg --add-architecture armhf;  \
-  dpkg --add-architecture armel; \
-  dpkg --add-architecture arm64; \
-  ldconfig
+# This is not a dangerous hack. `libsqlite3-dev:armel` package is broken on Debian 11. Therefore,
```
DoYouEvenCpp, line 82: Is there any bug-tracker issue on Debian11 for broken lissqlite3-dev? (might be nice to link it in here)

PR #951, opened 2025-06-30
LVPN-8062: Provide rollout group value

DoYouEvenCpp:
Changes under this PR:
- add a method calculating rollout group value
- value is based on machineID, SHA256 hash of it is calculated and then used to provide a value between range of [1, max] (where max must not exceed `DefaultMaxGroup`)
- tests


config/remote/utils.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+const DefaultMaxGroup = 100
+
+type Group struct {
+	value int
+}
```
thegravi: How would you access this value from the client side?

config/remote/utils.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+const DefaultMaxGroup = 100
```
thegravi: Shall we explain the purpose of this constant? :)

config/remote/utils.go
```diff
@@ -0,0 +1,43 @@
...
+type Group struct {
+	value int
+}
+
+// NewGroup creates a new Group instance based on a provided UUID and maximum group size.
+// It computes a group value by hashing the UUID and deriving a number between 1 and provided max (inclusive).
+//
+// Parameters:
+//   - uuid: The UUID used as the basis for group assignment
+//   - max: The maximum possible group value (must be positive and not exceed DefaultMaxGroup)
```
thegravi: I think it would be more reasonable to add information about the purpose of this argument, so that it is used to determine the position in the rollout group. What do you think?

DoYouEvenCpp: It got reworked, the `max` parameter is not valid anymore.

config/remote/utils.go
DoYouEvenCpp: Fixed!

config/remote/utils.go
DoYouEvenCpp: Fixed as well.

config/remote/utils.go
```diff
@@ -0,0 +1,27 @@
...
+// It computes a group value by hashing the UUID and deriving a number between 1 and defaultMaxGroup (inclusive).
+//
+// Parameters:
+//   - uuid: The UUID used as the basis for group assignment
+//
+// Returns:
+//   - RolloutGroup: A new RolloutGroup instance with the computed value
+func GenerateRolloutGroup(uuid uuid.UUID) int {
+	hash := sha256.Sum256(uuid[:])
+	num := binary.BigEndian.Uint32(hash[:])
```
devzbysiu, line 24: Aren't we loosing sha256 uniform distribution with this? Looks like `Uint32` is taking only first 4 bytes out of 32 bytes of sha256
I'm not sure if we can count on uniform distribution of just part of the sha 🤔

config/remote/utils_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// calculate specific brackets counts
+	brackets := make([]int, 5)
+	for _, g := range groups {
+		bracketIndex := (g - 1) / bracketSize
+		brackets[bracketIndex]++
+	}
+
+	// some expected statistics per bracket
+	expectedCount := iterations / 5
```
devzbysiu: [small] maybe number of brackets could also be put into a const and reused

devzbysiu (approved):

config/remote/utils_test.go
DoYouEvenCpp: Fixed.

PR #952, opened 2025-06-30
LVPN-6747: API response "invalid_creds" is not handled correctly

thegravi:
Changes
- Wrap `DefaultAPI` into `SmartClientAPI`, which internally manages token renewal and delegates error handling for custom business logic.
- Introduce `LoginTokenManager` to handle the lifecycle of the `login token` (get, set, renew, invalidate).
- Introduce `LoginTokenValidator`, responsible for verifying the existence and expiration of the `login token` based on its creation origin.
- Add specific error messages when the `login token` is revoked using CLI commands.
- Fix the issue where, if the `login token` is revoked, the user can still connect to any VPN server until they log out.


auth/auth.go
```diff
@@ -337,68 +324,44 @@ func (r *RenewingChecker) fetchSaveServices(userId int64, data *config.TokenData
...
+	services, err := r.creds.Services()
 	if err != nil {
 		return nil, fmt.Errorf("fetching available services: %w", err)
 	}
-
 	return services, nil
 }
 
-// saveLoginToken persists only token related data,
+// saveLoginCreds persists only token related data,
```
mariusSincovici: Is the comment still correct?

core/simple_client_api.go
```diff
@@ -0,0 +1,569 @@
...
+	// Payments returns a list of payments done by the user
+	Payments(token string) ([]PaymentResponse, error)
+}
+
+type SimpleClientAPI struct {
+	agent     string
+	baseURL   string
+	client    *http.Client
+	validator response.Validator
+	mu        sync.Mutex
```
mariusSincovici, line 62: What should this protect? Because I think is used only in Base(), but all the other functions use directly the member variables.

core/constants.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+var (
+	// ManualLoginTokenExpiryDate represents the expiration date for manually created tokens
+	// Setting a very big expiration date here as real expiration date is unknown just from the
+	// token, and there is no way to check for it. In case token is used but expired, automatic
+	// logout will happen. See: core/login_token_manager.go
+	// Note: bigger year cannot be used as time.Parse cannot parse year longer than 4 digits as
+	// of Go 1.21
+	ManualLoginTokenExpiryDate = time.Date(9999, time.December, 31, 0, 0, 0, 0, time.UTC).Format(internal.ServerDateFormat)
```
mariusSincovici, line 16: [minor] why not to go into the constants from `internal`?

core/login_token_validator.go
```diff
@@ -0,0 +1,60 @@
...
+	case ManualLoginTokenExpiryDate:
+		return l.validateCredibility(token)
+
+	default:
+		return l.validateExpiryDate(expiryDate)
+	}
+}
+
+func (l LoginTokenValidator) validateCredibility(token string) error {
+	var isFormatValid = regexp.MustCompile(`^[a-f0-9]*$`).MatchString
```
mariusSincovici: is it sure that the format will not change to include other chars?

core/login_token_validator.go
```diff
@@ -0,0 +1,60 @@
...
+	if !isFormatValid(token) {
+		return ErrLoginTokenExpired
```
mariusSincovici: It would be good to add the reason it was expired. 
Because in this case it is not expired, but the format is not correct. Maybe something with fmt.Errors?

core/login_token_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	if !ok {
+		return "", fmt.Errorf("there is no token")
+	}
+
+	return data.Token, nil
+}
+
+// Renew iterates over stored user tokens and attempts to renew each one that has expired.
+// Returns a combined error if one or more renewals fail.
+func (l *LoginTokenManager) Renew() error {
```
mariusSincovici, line 54: Wouldn't it make sense to have only one renew executed at a time?

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+		input.Events.User.Logout.Publish(events.DataAuthorization{
+			DurationMs:   max(int(time.Since(start).Milliseconds()), 1),
+			EventTrigger: events.TriggerApp,
+			EventStatus:  status,
+		})
+	}(logoutStartTime)
+
+	var cfg config.Config
+	if err := input.ConfigManager.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
```
mariusSincovici: Maybe add some more info to know where it failed actually. The same for other logging places

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+		return LogoutResult{Status: internal.CodeFailure, Err: nil}
```
mariusSincovici: Why not pass `err` in `Err` outside?
Also in other places.

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+// ForceLogoutWithoutToken performs user logout operation without using login toking
+func ForceLogoutWithoutToken(input ForceLogoutWithoutTokenInput) (logoutResult LogoutResult) {
+	logoutStartTime := time.Now()
+	input.Events.User.Logout.Publish(events.DataAuthorization{
+		DurationMs:   -1,
+		EventTrigger: events.TriggerApp,
+		EventStatus:  events.StatusAttempt,
+	})
+
+	defer func(start time.Time) {
```
mariusSincovici: why not just use `logoutStartTime` in here without passing it as a param, or maybe move the entire defer to a new function? 
Because `start` time is the only param passed to this, while other are directly used and I think it complicates a bit when reading.

It also seams to be duplicated with the one from `Logout`, except the `EventTrigger` type?

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+		status := events.StatusSuccess
+		if logoutResult.Err != nil &&
+			logoutResult.Status != 0 &&
+			logoutResult.Status != internal.CodeSuccess &&
```
mariusSincovici, line 154: Why not only different from `internal.CodeSuccess` and Error != nil?

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+		log.Println(internal.ErrorPrefix, err)
+		return LogoutResult{Status: internal.CodeFailure, Err: nil}
+	}
+
+	if err := input.NcClient.Stop(); err != nil {
+		log.Println(internal.WarningPrefix, err)
+	}
+
+	if err := input.ConfigManager.SaveWith(func(c config.Config) config.Config {
+		delete(c.TokensData, cfg.AutoConnectData.ID)
```
mariusSincovici: This seams to be duplicated with `Logout`. Would it be possible to have it extracted into a separate function? This would also make its testing easier.

daemon/access/access_logout.go
```diff
@@ -0,0 +1,205 @@
...
+		c.AutoConnectData.ID = 0
+		c.Mesh = false
+		c.MeshPrivateKey = ""
+		return c
+	}); err != nil {
+		return LogoutResult{Status: 0, Err: err}
+	}
+
+	input.Publisher.Publish("user logged out")
```
mariusSincovici: [minor] in some places `input.Publisher.Publish` is used for logging while in others Print, would i be possible to use only one?

daemon/rpc_logout.go
```diff
@@ -2,119 +2,27 @@ package daemon
...
+		Events:         r.events,
+		Publisher:      r.publisher,
+		PersistToken:   in.GetPersistToken(),
+		DisconnectFunc: r.DoDisconnect,
+	})
 
-	// RenewToken being empty means user logged in using Access Token
-	if !in.GetPersistToken() && tokenData.RenewToken == "" {
-		return &pb.Payload{Type: internal.CodeTokenInvalidated}, nil
+	if result.Status == 0 {
```
mariusSincovici, line 23: maybe return always when `result.Err != nil`.

daemon/access/access_logout.go
mariusSincovici, line 1: There are some comments to the file, but majority of code seams to be just moved from other places.

cmd/daemon/main.go
```diff
@@ -294,7 +280,36 @@ func main() {
...
-	deviceID := fmt.Sprintf("%x_%d", sha256.Sum256([]byte(machineID.String()+Salt)), machineIdGenerator.GetUsedInformationMask())
+	deviceID := fmt.Sprintf(
+		"%x_%d",
+		sha256.Sum256([]byte(machineID.String()+Salt)),
+		machineIdGenerator.GetUsedInformationMask(),
+	)
+
+	loginTokenErrHandlingReg := core.NewErrorHandlingRegistry[func(uid int64)]()
+	// encapsulating initialization logic
+	clientAPI, loginTokenManager := func() (core.ClientAPI, core.TokenManager) {
```
mariusSincovici, line 290: wouldn't it make sense to have this in a separate function? Then is also more clear what are the actual parameters needed to create the objects and can also be tested if needed.

cmd/daemon/main.go
```diff
@@ -459,19 +481,51 @@ func main() {
...
 		nc.MqttClientBuilder{},
 		infoSubject,
 		errSubject,
 		meshnetEvents.PeerUpdate,
-		nc.NewCredsFetcher(defaultAPI, fsystem))
+		nc.NewCredsFetcher(clientAPI, fsystem))
+
+	// on token invalidation (unauthorized access, missing server resources, invalid request)
+	// perform user log-out action
+	loginTokenErrHandlingReg.AddMulti(
```
mariusSincovici: same with this, separate function to be able to test that it works?

cmd/daemon/main.go
```diff
@@ -459,19 +481,51 @@ func main() {
...
+		[]error{core.ErrUnauthorized, core.ErrNotFound, core.ErrBadRequest},
```
mariusSincovici: Wouldn't the app disconnect the user whenever is a bad request? Wouldn't this be risky, if the app creates a bad request for something, then the user is logged out :thinking:

daemon/rpc_account.go
```diff
@@ -115,9 +122,8 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
 	accountInfo.ExpiresAt = tokenData.ServiceExpiry
 	accountInfo.LastDedicatedIpExpiresAt = dedicatedIPExpirationDate
 
-	currentUser, err := r.credentialsAPI.CurrentUser(tokenData.Token)
+	currentUser, err := r.credentialsAPI.CurrentUser()
 	if err != nil {
-		log.Println(internal.ErrorPrefix, "retrieving user:", err)
```
mariusSincovici, line 120: Why remove the log?

daemon/access/access_disconnect.go
```diff
@@ -0,0 +1,59 @@
...
+			Duration:             time.Since(start),
+			Error:                err,
+		})
+	}(startTime)
+
+	if err = input.Networker.Stop(); err != nil {
+		err = fmt.Errorf("stopping networker: %w", err)
+		return true, err
+	}
+	if err = input.ConfigManager.Load(&cfg); err != nil {
```
mariusSincovici: Maybe move the entire loading from file before the defer. Because if this or the above fails, `cfg` will have invalid/incorrect information.

auth/auth_test.go
```diff
@@ -124,6 +124,10 @@ func (p *mockErrPublisher) Publish(e error) {
 	p.err = e
 }
 
+type mockTokenManager struct {
```
devzbysiu: Maybe this could be moved to `test/mock`?

daemon/data_models.go
```diff
@@ -158,44 +160,43 @@ func (data *VersionData) save() error {
...
+	a.cache.Invalidate()
 }
 
-func (a *AccountData) get(respectValidityPeriod bool) (*pb.AccountResponse, bool) {
-	if !a.isSet {
-		return nil, false
+// return true if cached data was used, otherwise false
+func (a *AccountData) get(requestFreshFetch bool) (*pb.AccountResponse, bool) {
+	if requestFreshFetch {
+		data, err := a.cache.Fetch()
```
mariusSincovici: Isn't `Fetch` function nil?

daemon/data_models.go
```diff
@@ -158,44 +160,43 @@ func (data *VersionData) save() error {
...
-}
-
-type cacheValidityFunc func(time.Time) bool
-
 type AccountData struct {
-	accountData            *pb.AccountResponse
-	isSet                  bool
-	updatedAt              time.Time
-	checkCacheValidityFunc cacheValidityFunc
+	cache *caching.Cache[*pb.AccountResponse]
```
mariusSincovici, line 164: Since this has its own mutex inside, does it make sense to also lock the entire DataManager when accessing it? :thinking:

daemon/data_models.go
```diff
@@ -158,44 +160,43 @@ func (data *VersionData) save() error {
...
-	a.isSet = false
-	a.accountData = &pb.AccountResponse{}
+	a.cache.Invalidate()
 }
 
-func (a *AccountData) get(respectValidityPeriod bool) (*pb.AccountResponse, bool) {
-	if !a.isSet {
-		return nil, false
+// return true if cached data was used, otherwise false
+func (a *AccountData) get(requestFreshFetch bool) (*pb.AccountResponse, bool) {
```
mariusSincovici: Isn't `requestFreshFetch` changing the behavior? 
Because now the account is always fetched and the availability period is ignored and `nordvpn account`.

auth/auth_test.go
devzbysiu, line 1: `IsLoggedIn` has now different function signature, maybe it would be good to cover error cases with tests?

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
+package core
+
+import (
+	"reflect"
+	"sync"
+)
+
+type ErrorHandlingRegistry[ErrHandler any] struct {
+	m    sync.Mutex
+	pool map[error][]ErrHandler
```
devzbysiu: Wouldn't be safer to use `string` a key? If you'll use it with different `error` instances but having the same error messages, it will be interpreted as different key even when the meaning of the error is the same

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// GetHandlers returns a deep copy list of the provided error handlers.
+func (e *ErrorHandlingRegistry[ErrHandler]) GetHandlers(err error) []ErrHandler {
+	e.m.Lock()
+	defer e.m.Unlock()
+
+	original := e.pool[err]
+	copied := make([]ErrHandler, len(original))
+	copy(copied, e.pool[err])
```
devzbysiu: You can copy `original` here, no need for lookup by `err` again

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
...
+		panic("ErrHandler must be a function")
+	}
+
+	return &ErrorHandlingRegistry[T]{
+		pool: make(map[error][]T),
+	}
+}
+
+// Add registers an ErrHandler for a specific error.
+func (e *ErrorHandlingRegistry[ErrHandler]) Add(err error, handler ErrHandler) {
```
devzbysiu: Is this needed? I see it used only in tests. Maybe `AddMulti` is enough?

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
...
+	e.pool[err] = append(e.pool[err], handler)
+}
+
+// AddMulti registers an ErrHandler for a list of errors.
+func (e *ErrorHandlingRegistry[ErrHandler]) AddMulti(errs []error, handler ErrHandler) {
+	e.m.Lock()
+	defer e.m.Unlock()
+
+	for _, err := range errs {
+		e.pool[err] = append(e.pool[err], handler)
```
devzbysiu: Should we check for handler already existing in the `pool[err]`? Now if it's there, it will be duplicated. Or maybe just document this behavior if we don't need the check

core/err_handler_registry_test.go
```diff
@@ -0,0 +1,159 @@
+package core_test
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"gotest.tools/v3/assert"
```
devzbysiu: We are using mostly `github.com/stretchr/testify/assert` also keep in mind that there is a difference in how failed assertions are processed. With `testify` failed assertion is not stopping the test (unless you use `require`) but will report the error and mark the test as failed. Gotest stops the test execution, so if you have multiple assertions, you won't see the errors in next assertions in the test

core/err_handler_registry_test.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+func assertPanic(t *testing.T, fn func(), msg string) {
```
devzbysiu: There is a way to assert on panic in `testify/assert` without need for custom handler and having to play with `recover()`

core/helper_test.go
```diff
@@ -140,8 +140,8 @@ func CheckServer(listener net.Listener, attempts int) {
 
 // testNewDefaultAPI returns a pointer to initialized and
```
devzbysiu: This needs to be `testNewSimpleAPI` now

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
+package core
+
+import (
+	"reflect"
+	"sync"
+)
+
+type ErrorHandlingRegistry[ErrHandler any] struct {
```
devzbysiu: Do we need to have this generic? Maybe we could have handler defined as a func type and just use this type inline in the `pool` field?
Would be nice to avoid having runtime check if possible

core/constants.go
thegravi: I thought that this is kind of a "backend" thing; therefore, why should it be treated as an internal for all components (cli, tray) of our app? Do you see any other usages apart the daemon?

core/login_token_validator.go
thegravi: I took it just like it was used elsewhere. We can put it into the constants section to be publicly accessible and therefore reusable and easier to maintain. What do you think?

core/login_token_validator.go
thegravi: Good point, this is too generic for this specific case. Probably, I will try to add an additional error and later handle it upstream. I would like these specific errors to be checkable in the upper layers.

core/login_token_manager.go
thegravi: Here, I ported the underlying checks from the previous use. What about checking only for the current user? I mean use `autoconnect.id` for a single check.
Why do we want to check other users, since the request indirectly comes (via tray or CLI) only from the currently logged-in user?

daemon/access/access_logout.go
thegravi: I don't understand the issue. Here, `input.Publisher.Publish` is used, what about other places where `Print`?? is used? Can you explain more?

Some cases just logs the message, other actually published and event.

daemon/access/access_logout.go
thegravi: Possibly, there are places left where I've forgotten to clean up 👀

cmd/daemon/main.go
thegravi: Yes, I wanted to do that, but unfortunately, there would be a function with too many arguments 😞

cmd/daemon/main.go
thegravi: This was the initial logic behind. I was not sure about `BadRequest` and `NotFound` cases, and when they could occur; therefore, I just reused them. If you know for sure that those are not suitable for this logic, I will gladly remove/move elsewhere.

core/err_handler_registry.go
thegravi: Sure, missed that :)

core/simple_client_api.go
thegravi: Mesh interfaces use it also.

core/err_handler_registry.go
```diff
@@ -0,0 +1,54 @@
...
+	m    sync.Mutex
+	pool map[error][]ErrHandler
+}
+
+// NewErrorHandlingRegistry creates a new ErrorHandlingRegistry with initialized pool.
+// It is thread safe
+func NewErrorHandlingRegistry[T any]() *ErrorHandlingRegistry[T] {
+	var zero T
+	if reflect.TypeOf(zero).Kind() != reflect.Func {
```
mariusSincovici: Maybe if it needs to be a function, wouldn't work to accept only the parameter and then `Add` to accept something like func(T)? In this way the panic is not needed anymore. 
And if more than 1 param is needed, structs can be used instead?

auth/auth.go
thegravi: fixed

auth/auth.go
```diff
@@ -91,29 +92,33 @@ func NewRenewingChecker(cm config.Manager,
...
 // IsLoggedIn reports user login status.
 //
 // Thread safe.
-func (r *RenewingChecker) IsLoggedIn() bool {
+func (r *RenewingChecker) IsLoggedIn() (bool, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
+	if err := r.loginTokenManager.Renew(); err != nil {
+		return false, err
```
bartoszWojciechO: Can you wrap the errors so that they would be easier to trace from the logs? For example in this line, `fmt.Errof("renewing token: %w", err")`

daemon/rpc_account.go
thegravi: accidentally, reverting back

core/err_handler_registry.go
mariusSincovici: [small] you can use `slices.Clone` so it will make the allocation an copy

daemon/access/access_disconnect.go
thegravi: good point

daemon/data_models.go
thegravi: This is a public method that is always implemented. You may be asking about the provided one. If so, that case is handled internally. During fetching (without the provided function), if there was no cached data before fetching, it would simply return an error, and if data already existed, stale data would be retrieved.

daemon/data_models.go
thegravi: Yes! Now I remember what I wanted to change here but forgot 🙃

daemon/data_models.go
thegravi: I might interpreted the logic incorrectly. The initial one was very confusing. RPC would set this "Full" flag, whose value would then be converted to `respectValidityPeriod`, which is contradictory in my opinion. I understand that if full request is provided, meaning it wants to retrieve fresh data and not cached, then we do not respect that validity period. I'm I wrong? Could you enlighten me regarding this logic?

daemon/access/access_logout.go
thegravi: There is this custom case below which has status 0 as default invalid status 🤷🏻‍♂️

daemon/access/access_logout.go
thegravi: fixed defer

daemon/access/access_logout.go
thegravi: In this PR, I tried not to touch other code except the one I needed to change; therefore, many cases are left unchanged, even this one :)

daemon/access/access_logout.go
thegravi: Sure, I've extracted it, but logic differs from `Logout`. It also removes autoconnect.id, mesh cfg and logout event is published with different DataAuthorization values.

cmd/daemon/main.go
thegravi: same with that :) There would be a function with way too many arguments, o a struct with a great number of fields. The question is, are we up to?

core/login_token_validator.go
mariusSincovici: Yes, I've just saw is the same in another part.
Then would it work to move this regex check to a function and reuse the same errors? Because in `LoginWithToken` a different error is returned, `internal.CodeTokenInvalid`

daemon/access/access_logout.go
mariusSincovici: Yes, but in that case `error` is set. I would recommend to treat `0` as an error case and only `internal.CodeSuccess` is considered successful.

daemon/access/access_disconnect.go
```diff
@@ -0,0 +1,58 @@
...
+		err = fmt.Errorf("loading config: %w", err)
+		return true, err
+	}
+
+	defer func() {
```
mariusSincovici, line 32: I think this will emit events only on success. 
What I was suggesting was to move 
```
if err = input.ConfigManager.Load(&cfg); err != nil {
		err = fmt.Errorf("loading config: %w", err)
		return true, err
	}
````
before defer block and keep defer before stop :smile: . In this way the members from publish have correct value
```
	if err = input.ConfigManager.Load(&cfg); err != nil {
		err = fmt.Errorf("loading config: %w", err)
		return true, err
	}
 defer ...
if err = input.Networker.Stop(); err != nil {
```

File: core/simple_client_api.go (hunk 2 above)
Comment (mariusSincovici) [reply core/simple_client_api.go:62]: ooo, then ignore

File: core/login_token_manager.go [hunk 34]
```diff
@@ -0,0 +1,203 @@
...
+	resp, err := l.tokenRenewAPICall(data.RenewToken, *data.IdempotencyKey)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+// tokenDataRemover returns a function that removes the token data associated with the user ID
+// from the config object.
+func (l *LoginTokenManager) tokenDataRemover(uid int64) config.SaveFunc {
```
bartoszWojciechO, line 185: If this function is only called once, maybe it would be better to inline it?

core/login_token_validator.go
thegravi: Those 2 errors are different, one is using the error interface, while another is a pure integer.

auth/auth_test.go
thegravi: done

core/err_handler_registry.go
thegravi: done

core/login_token_validator.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (l LoginTokenValidator) validateCredibility(token string) error {
+	var isFormatValid = internal.AccessTokenFormatValidator
+	if !isFormatValid(token) {
+		return fmt.Errorf("invalid access token format: %w", ErrLoginTokenExpired)
+	}
+
+	// this is the least expensive api call that needs authentication
+	_, err := l.credsAPI.CurrentUser(token)
```
bartoszWojciechO, line 42: Is this idempotent per token? If it's not if users running some commands in a loop could unknowingly spam the request.

core/err_handler_registry.go
thegravi: I've simplified the struct to accept 1 or more error for the same handler

core/err_handler_registry.go
thegravi: Thought about it, but wanted to make the client responsible for this. Perhaps, I will make a not in the description

core/err_handler_registry_test.go
thegravi: fixed

core/err_handler_registry_test.go
thegravi: done

core/helper_test.go
thegravi: done

core/err_handler_registry.go
thegravi: Each different error handling registry could support different handling mechanism; therefore, I can only suggest removing `any` and use `func(T)`. However, with this approach we can only have a single argument and if more would be needed then a struct should be used.

core/login_token_validator.go
thegravi: The content should not be changed as long as those tokens are associated with the same user.

core/login_token_manager.go
thegravi: What advantage would we gain at the expense of readability and testability?

auth/auth.go
thegravi: sure

core/err_handler_registry.go
thegravi: Of course, it would, but I do not want to ditch elegance and simplicity for these misuses. With strings, you can also make such mistakes when registering one error string and later using another. In this way, you would also need to synchronize those meanings somehow.
My original idea is to use predefined errors, just like we have them in `errors.go`

core/login_token_manager.go
bartoszWojciechO: Imo it could be more readable to have this call inlined(it would require less of a mental map when reading the code). Is this function used in testing?

thegravi: It is since now

bartoszWojciechO: Fine by me in this case.

bartoszWojciechO (approved):

daemon/rpc_test.go
```diff
@@ -69,12 +69,20 @@ var (
 )
 
 func testRPC() *RPC {
-	api := core.NewDefaultAPI(
-		"1.0.0",
-		"",
-		http.DefaultClient,
-		response.NoopValidator{},
+	type mockLoginTokenManager struct {
```
devzbysiu, line 72: This is also done in `auth/token_test.go`, maybe it could be moved to somewhere in `test/mock` and reused?

internal/caching/expiring_cache.go
```diff
@@ -0,0 +1,149 @@
...
+// NewCacheWithTTL creates a new cache with a custom validity period and fetch function
+func NewCacheWithTTL[T any](validity time.Duration, fetchFunc func() (T, error)) *Cache[T] {
+	return &Cache[T]{
+		ttl:       validity,
+		fetchFunc: fetchFunc,
+	}
+}
+
+// Get returns the data either from cache or fetches from origin depending on validity period
+func (c *Cache[T]) Get() (T, error) {
```
devzbysiu, line 38: Do we have performance issues when using just write lock for the whole `Get` or there is other reason to have double locking here?

internal/caching/expiring_cache.go
```diff
@@ -0,0 +1,149 @@
...
+	// check again under the write lock to avoid double fetching
+	if c.isValid() {
+		return c.entry.data, nil
+	}
+
+	return c.fetchLocked()
+}
+
+// GetEvenIfStale returns data regardless of staleness but with a bool indicating freshness
+func (c *Cache[T]) GetEvenIfStale() (data T, fresh bool, err error) {
```
devzbysiu, line 59: Is this used? I see it only in tests :thinking:

daemon/data_models.go
```diff
@@ -158,44 +160,54 @@ func (data *VersionData) save() error {
...
-	if respectValidityPeriod && !a.checkCacheValidityFunc(a.updatedAt) {
-		a.isSet = false
+	if respectDataExpiry && err != nil {
 		return nil, false
 	}
 
-	return proto.Clone(a.accountData).(*pb.AccountResponse), a.isSet
+	// when not respecting expiry, we can use stale data
+	if !respectDataExpiry && err != nil {
+		if isStaleDataError := errors.Is(err, caching.ErrStaleData); !isStaleDataError {
```
devzbysiu: [small] I think `if !error.Is(..) {` is clear enough, no need for `isStaleDataError` var IMO

devzbysiu (approved):

internal/caching/expiring_cache.go
thegravi: The first `read` lock attempts to access data if it's valid, allowing other threads to read the data simultaneously without blocking. But if the data is invalid, it then tries to acquire a `write` lock, which blocks all other access to the underlying data until the lock is released. This situation would prevent other threads from accessing the data until we update it.
When I designed this caching module, I kept concurrency in mind; therefore, this double locking is implemented to prevent unnecessary blocking.

internal/caching/expiring_cache.go
thegravi: Not anymore, gonna remove it.

nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=952) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=952) before we can accept your contribution.<br/>**1** out of **5** committers have signed the CLA.<br/><br/>:white_check_mark: dfetti<br/>:x: bbrutall<br/>:x: edvyno1<br/>:x: keliramu<br/>:x: thegravi<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.nordsec.com/check/NordSecurity/nordvpn-linux?pullRequest=952) it.</sub>

PR #953, opened 2025-07-02
LVPN-7891: fix grpc tests

bartoszWojciechO:
Added synchronization to grpc tests, added a wait for grpc server to become available and switched to using a single socket in case of multiple subscribers test.

### DIFF
--- a/test/qa/test_grpc.py
+++ b/test/qa/test_grpc.py
@@ -60,11 +60,11 @@ def test_tunnel_update_notifications_before_and_after_connect():
                b in zip(result, expected_states, strict=True))
 
 
-def collect_state_changes_guard(stop_at: int, tracked_states: Sequence[str], subscribtion_barrier: Barrier, timeout: int = 10) -> Sequence[state_pb2.AppState]:
+def collect_state_changes_guard(stop_at: int, tracked_states: Sequence[str], subscribtion_barrier: Barrier, timeout: int = 30) -> Sequence[state_pb2.AppState]:
     with grpc.insecure_channel(NORDVPND_SOCKET) as channel:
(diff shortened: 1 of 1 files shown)


DoYouEvenCpp (approved):

test/qa/test_grpc.py
```diff
@@ -43,20 +47,27 @@ def test_tunnel_update_notifications_before_and_after_connect():
...
-def collect_state_changes(stop_at: int, tracked_states: Sequence[str], timeout: int = 10) -> Sequence[state_pb2.AppState]:
+def collect_state_changes_guard(stop_at: int, tracked_states: Sequence[str], subscribtion_barrier: Barrier, timeout: int = 10) -> Sequence[state_pb2.AppState]:
     with grpc.insecure_channel(NORDVPND_SOCKET) as channel:
+        return collect_state_changes(channel, stop_at, tracked_states, subscribtion_barrier, timeout)
+
+def collect_state_changes(channel: grpc.Channel, stop_at: int, tracked_states: Sequence[str], subscribtion_barrier: Barrier, timeout: int = 10) -> Sequence[state_pb2.AppState]:
+        grpc.channel_ready_future(channel).result(timeout=timeout)
         stub = service_pb2_grpc.DaemonStub(channel)
+        subscribtion_barrier.wait()
         response_stream = stub.SubscribeToStateChanges(
```
mariusSincovici, line 71: why is `timeout` needed?
based on what I've found and tested related to the timeout param, the gRPC will close the connection after 10 seconds, no matter if there are still messages received. So this behaves also like a "max open duration".

bartoszWojciechO: Probably to put a constraint on the app to provide all of the expected state changes within some set time.

PR #956, opened 2025-07-03
Remove completely IPv6 from linux app

dfetti:


cmd/daemon/main.go
```diff
@@ -363,14 +363,18 @@ func main() {
 	daemonEvents.User.Subscribe(statePublisher)
 	configEvents.Subscribe(statePublisher)
 
+	// Always disable IPv6 with sysctl in the system
+	// TODO (dfe): Should I leave it into networker and do Block every time a new adapter is added?
```
mariusSincovici: In the past there was an issue with the IPv6 disable. Even if all the sysctl values were set to disabled IPv6 when a new adapter was inserted into the system would have IPv6 enabled. Because of this IPv6 is disabled every time a new interface is detected.

daemon/device/device.go
```diff
@@ -55,26 +54,13 @@ func ListPhysical() ([]net.Interface, error) {
 	}
 
 	if len(interfaces) == len(vInterfaces) {
-		gateway4, err4 := DefaultGateway(false)
-		gateway6, err6 := DefaultGateway(true)
+		gateway4, err4 := DefaultGateway()
```
mariusSincovici: [small] maybe remove the `4` from the names?

protobuf/daemon/settings.proto
```diff
@@ -27,20 +27,19 @@ message Settings {
...
-  repeated string dns = 10;
-  bool threat_protection_lite = 11;
-  config.Protocol protocol = 12;
-  bool lan_discovery = 13;
-  Allowlist allowlist = 14;
-  bool obfuscate = 15;
-  bool virtualLocation = 16;
-  bool postquantum_vpn = 17;
-  UserSpecificSettings user_settings = 18;
+  bool meshnet = 7;
```
mariusSincovici: Why is it needed to change the values for the other members?

cmd/daemon/main.go
dfetti: Now it's done once at the start and after that every time there are changes in network interfaces

networker/listener.go
```diff
@@ -12,11 +12,11 @@ func (c *Combined) Reconnect(stateIsUp bool) {
...
 	defer c.mu.Unlock()
 
-	if c.isVpnSet {
-		// disable IPv6 as soon as possible to prevent leaks when a new Ethernet is inserted
-		if err := c.disableIPv6IfNeeded(); err != nil {
-			log.Println(internal.ErrorPrefix, "refreshing network", err)
-		}
+	// Disable IPv6 as soon as possible to prevent leaks when a new adapter is inserted
+	// TODO (dfe): Should condition it if c.isVpnSet or something similar?
+	blockError := c.ipv6Blocker.Block()
```
mariusSincovici: it should be set only when `if c.isVpnSet {` otherwise IPv6 would be disabled also when user is not connected to VPN.

networker/networker.go
```diff
@@ -208,6 +204,20 @@ func (netw *Combined) Start(
 ) (err error) {
 	netw.mu.Lock()
 	defer netw.mu.Unlock()
+
+	// Always disable IPv6 with sysctl in the system
+	// We also do that when there is a change in network interfaces
+	blockError := netw.ipv6Blocker.Block()
+	if blockError != nil {
+		log.Println(internal.ErrorPrefix, "Failed to block ipv6 using sysctl", blockError)
```
mariusSincovici: I think before failing to block IPv6 was fatal and now is not anymore. If the IPv6 existed but is not blocked then user will have leaks.

networker/networker.go
```diff
@@ -426,8 +420,17 @@ func (netw *Combined) restart(
 		return err
 	}
 
-	if err := netw.disableIPv6IfNeeded(); err != nil {
-		log.Println(internal.ErrorPrefix, "failed to disable ipv6", err)
+	// Always disable IPv6 with sysctl in the system
+	// We also do that when there is a change in network interfaces
+	blockError := netw.ipv6Blocker.Block()
```
mariusSincovici: Maybe move this and the next `if` in a separate function, to know that they always need to be executed together?

networker/networker.go
```diff
@@ -594,56 +583,128 @@ func (netw *Combined) unblockTraffic() error {
...
+func (netw *Combined) blockIPv6Traffic() error {
 	ifaces, err := netw.devices()
 	if err != nil {
 		return err
 	}
 
 	err = netw.fw.Add([]firewall.Rule{
 		{
-			Name:        "vpn_allowlist_icmp6_errors",
+			Name:        "vpn_block_icmp6_errors",
```
mariusSincovici: Cannot comment on line 710, shouldn't `Allow:       true,` be false? The same for all the others in the function?
Or wouldn't be sufficient to just block all the IPv6 traffic without specific restriction?

dfetti: Yes, it should be false. And for the other question, I don't know..

daemon/device/device.go
```diff
@@ -55,26 +54,13 @@ func ListPhysical() ([]net.Interface, error) {
...
 		}
 
-		if err4 != nil {
-			return nil, err4
-		}
-		if err6 != nil {
-			return nil, err6
-		}
-		return nil, errors.New("unable to retrieve default gateway")
+		return nil, err
```
mariusSincovici: [suggestion] if you could combine the error with the `unable to retrieve default gateway` previous message would be better to understand from where it came.

networker/networker.go
mariusSincovici: I think it can fully be delete this. and then there are no special cases for IPv6 and everything is just blocked

protobuf/daemon/settings.proto
```diff
@@ -27,7 +27,6 @@ message Settings {
   bool firewall = 2;
   bool kill_switch = 3;
   AutoconnectData auto_connect_data = 4;
-  bool ipv6 = 5;
```
devzbysiu, line 30: This requires bump of the version here https://github.com/NordSecurity/nordvpn-linux/blob/main/protobuf/daemon/common.proto#L9 and also updates on GUI side

devzbysiu: Nope, bump not needed. My bad. But we have ipv6 settings reference in daemon mock

cmd/daemon/main.go
```diff
@@ -142,6 +142,12 @@ func main() {
 		config.StdFilesystemHandle{},
 		configEvents.Config,
 	)
+
+	// Remove any remains of IPv6 settings
+	if err := fsystem.SaveWith(sanitizeConfig); err != nil {
+		log.Println(internal.ErrorPrefix, "failed to remove IPv6 entries from settings")
```
devzbysiu: Maybe add `err` to log output so it's easier to see what's the issue?

cmd/daemon/main.go
```diff
@@ -188,7 +194,7 @@ func main() {
 		stateModule,
 		stateFlag,
 		chainPrefix,
-		iptables.FilterSupportedIPTables(internal.GetSupportedIPTables()),
```
devzbysiu, line 202: Can this func be removed now when it's not used here?

cmd/daemon/main.go
```diff
@@ -695,3 +701,36 @@ func assignMooseDBPermissions(eventsDbPath string) error {
 	}
 	return nil
 }
+
+func sanitizeConfig(c config.Config) config.Config {
```
devzbysiu: Maybe it would be better to name this in a more specific way? Like `removeIPv6Remains` or something like that? So next time when we'll be cleaning something else it will be less tempting to add it to this func :smile:

daemon/device/device.go
```diff
@@ -100,11 +86,8 @@ func ifaceListContains(list []net.Interface, device net.Interface) bool {
 // Linux generally has only a single default gateway. Although it can
 // have more than one default gateway by using routing tables, only one
 // is allowed per routing table.
-func DefaultGateway(ipv6 bool) (net.Interface, error) {
+func DefaultGateway() (net.Interface, error) {
 	version := "-4"
```
devzbysiu: [small] This can be inlined now

daemon/dns/nameservers.go
```diff
@@ -32,13 +28,9 @@ func NewNameServers(servers []string) *NameServers {
 }
 
 // Get nameservers selected by the given criteria.
-func (n *NameServers) Get(isThreatProtectionLite bool, ipv6 bool) []string {
+func (n *NameServers) Get(isThreatProtectionLite bool) []string {
 	if isThreatProtectionLite {
 		var nameservers []string
```
devzbysiu: This and the `append` below is probably not needed now right? :thinking: we can just use `v4Nameservers`

daemon/dns/nameservers.go
```diff
@@ -49,10 +41,6 @@ func (n *NameServers) Get(isThreatProtectionLite bool, ipv6 bool) []string {
 	}
 
 	var nameservers []string
-	if ipv6 {
-		nameservers = append(nameservers, primaryNameserver6, secondaryNameserver6)
-	}
-
 	nameservers = append(nameservers, primaryNameserver4, secondaryNameserver4)
```
devzbysiu: Here we probably also don't need need `append`, maybe we can even just have one const with primary and secondary nameservers combined

daemon/firewall/iptables_manager/iptables_manager.go
```diff
@@ -9,17 +9,14 @@ import (
...
-	ip6tablesCommand = "ip6tables"
+	iptablesCommand = "iptables"
 )
 
-// IpVersion determines which version of iptables command should be used, i.e iptables, ip6tables or both
+// IpVersion determines which version of iptables command should be used, i.e iptables
 type IpVersion int
 
 const (
 	IPv4 IpVersion = iota
```
devzbysiu, line 19: Do we need all of this now? :thinking:

daemon/firewall/firewall_manager.go
devzbysiu, line 1: Hmm, is this whole thing even used? :thinking:

daemon/firewall/firewall_manager_test.go
devzbysiu, line 1: This probably can also go if `firewall_manager` is not used

daemon/firewall/iptables_manager/iptables_manager.go
devzbysiu, line 1: This is used only in `firewall_manager.go` which is not used (or I'm blind :smile: ), so maybe this also can be deleted? Then its tests as well?

daemon/firewall/firewall_manager.go
bartoszWojciechO: Not used as of yet. This is an abandoned task to refactor firewall management. There are tasks to come back to this in the backlog, but no time yet.

daemon/firewall/iptables/iptables.go
```diff
@@ -216,32 +216,26 @@ func ruleToIPTables(rule firewall.Rule, module string, stateFlag string, chainPr
 												newRule := generateIPTablesRule(
 													chain, target, iface, remoteNetwork, localNetwork, protocol, pRange,
 													module, stateFlag, rule.ConnectionStates, chainPrefix, portFlag,
-													icmpv6Type, rule.HopLimit, nil, nil,
-													rule.Comment, mark,
+													icmpv6Type, rule.HopLimit, nil, nil, rule.Comment, mark,
 												)
-												if rule.Ipv6Only || remoteNetwork.Addr().Is6() || localNetwork.Addr().Is6() {
+												if rule.Ipv6Only {
+													// We should have just our block rules here
```
bartoszWojciechO: Do we need any IPv6 rulse if the protocol is blocked in the kernel? Maybe it's indeed better to keep them, for security(in case blocking in kernel fails).

dfetti: Discussed with Marius about this and decided to keep a generic IPv6 block rule just for safety. Working on it.

cmd/daemon/main.go
dfetti: I'm using it. It actually checks if the command is available on the system.

daemon/firewall/firewall_manager.go
dfetti: I'm leaving it as it is..

daemon/firewall/iptables_manager/iptables_manager.go
dfetti: As Bartosz said it's a work in progress and I'm leaving it as it is

daemon/dns/nameservers.go
```diff
@@ -32,29 +28,18 @@ func NewNameServers(servers []string) *NameServers {
...
 	if isThreatProtectionLite {
 		var nameservers []string
-		if ipv6 {
-			// @TODO remove once config is updated
-			nameservers = append(nameservers, threatProtectionLitePrimaryNameserver6, threatProtectionLiteSecondaryNameserver6)
-		}
-		v4Nameservers := n.servers
-		if len(v4Nameservers) == 0 {
-			v4Nameservers = []string{threatProtectionLitePrimaryNameserver4, threatProtectionLiteSecondaryNameserver4}
+		nameservers = append(nameservers, n.servers...)
```
mariusSincovici: [suggestion]
```var nameservers []string
		if len(nameservers) == 0 {
			nameservers = []string{threatProtectionLitePrimaryNameserver4, threatProtectionLiteSecondaryNameserver4}
		} else {
			nameservers = slices.Clone(n.servers)
		}
```

networker/networker.go
```diff
@@ -267,6 +261,8 @@ func (netw *Combined) start(
 
 	netw.publisher.Publish("starting vpn")
 
+	netw.blockIPv6()
```
mariusSincovici: If blocking IPv6 fails then the connection should stop. Because it can result in leaks on IPv6.

events/logger/logger.go
```diff
@@ -207,39 +207,29 @@ func maskIPRouteOutput(output string) string {
...
-	for _, arg := range []string{"4", "6"} {
-		// #nosec G204 -- arg values are known before even running the program
-		out, err := exec.Command("ip", "-"+arg, "route", "show", "table", "all").CombinedOutput()
-		if err != nil {
-			continue
-		}
-		maskedOutput := maskIPRouteOutput(string(out))
-		builder.WriteString("Routes for ipv" + arg + ":\n")
-		builder.WriteString(maskedOutput)
+	arg := "4"
```
devzbysiu: [small] This probably can be inlined

network/connection_test.go
```diff
@@ -19,20 +19,14 @@ func TestDefaultEndpoint(t *testing.T) {
...
 	serverIps := []netip.Addr{
 		netip.MustParseAddr("1.1.1.1"),
-		netip.MustParseAddr("2606:4700:4700::1111"),
 	}
 	endpoint := DefaultEndpoint(&resolver, serverIps)
 
 	ip, err := endpoint.ip()
 	assert.NoError(t, err)
-	assert.Equal(t, serverIps[1], ip)
+	assert.Equal(t, serverIps[0], ip)
```
devzbysiu, line 27: Isn't it testing the same thing? `ip()` now just delegates to `ip4()` so this is effectively the same block as below

networker/networker.go
```diff
@@ -267,6 +261,13 @@ func (netw *Combined) start(
 
 	netw.publisher.Publish("starting vpn")
 
+	// Always disable IPv6 with sysctl in the system
+	// We also do that when there is a change in network interfaces
+	if err = netw.ipv6Blocker.Block(); err != nil {
+		log.Println(internal.ErrorPrefix, "Failed to block ipv6 using sysctl ", err)
```
devzbysiu: Right now there are three blocks of code with the same log message if blocking fails. I think adding more context to the log message would help with debugging in the future

daemon/firewall/iptables/iptables_test.go
```diff
@@ -874,13 +806,8 @@ func TestRuleToIPTables(t *testing.T) {
 				countingOnly = true
 				assert.Equal(t, tt.ipv4Count, len(allRules[ipv4Table]))
 			}
-			if tt.ipv6Count > 0 {
-				countingOnly = true
-				assert.Equal(t, tt.ipv6Count, len(allRules[ipv6Table]))
-			}
 			if countingOnly == false {
```
devzbysiu: This could probably be simplified by having condition only on `tt.ipv4Count` now when we don't have `ipv6Count`

devzbysiu (approved):

PR #964, opened 2025-07-10
LVPN-5030: Add snap info into app version output

mariusSincovici:

Change the version output to add [snap] at the end of the version when the application is running in snap.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -111,7 +111,7 @@ func NewApp(version, environment, hash, salt string,
 	}
 
 	cli.VersionPrinter = func(c *cli.Context) {
-		fmt.Printf("NordVPN Version %s\n", c.App.Version)
+		cmd.Version(c)
 	}
 	cli.BashCompletionFlag = &cli.BoolFlag{
 		Name:   "complete",
@@ -350,10 +350,8 @@ func NewApp(version, environment, hash, salt string,
 		return nil
 	}
 
-	app.Version = version
-	if internal.IsDevEnv(environment) {
-		app.Version = fmt.Sprintf("%s - %s (%s)", version, internal.Environment(environment), hash)
-	}
+	app.Version = composeAppVersion(version, environment, snapconf.IsUnderSnap())
+
 	app.Commands = []*cli.Command{
 		{
 			Name:               "account",
@@ -1397,4 +1395,21 @@ func readForConfirmationBlockUntilValid(r io.Reader, prompt string) bool {
 		}
 		fmt.Println(InputParsingError)
 	}
+}
+
+// composeAppVersion concatenates the provided information to produce a string
+// representing the application version in the format: 1.2.3 [snap] - dev
+// Later it is used by the version command to display to the user
+func composeAppVersion(buildVersion string, environment string, isSnap bool) string {
+	env := ""
+	if internal.IsDevEnv(environment) {
+		env = fmt.Sprintf(" - %s", environment)
+	}
+
+	snap := ""
+	if isSnap {
+		snap = " [snap]"
+	}
+
+	return fmt.Sprintf("%s%s%s", buildVersion, snap, env)
 }
--- a/cli/cli_version.go
+++ b/cli/cli_version.go
@@ -3,46 +3,20 @@ package cli
 import (
 	"context"
 	"fmt"
-	"strings"
-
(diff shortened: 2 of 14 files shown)


magefiles/scripts.go
```diff
@@ -59,24 +59,6 @@ func getVersions() (map[string]string, error) {
...
 }
 
-// TODO: replace with information coming from the Go toolchain
-func getGitInfo() (*gitInfo, error) {
-	hash, err := sh.Output("git", "rev-parse", "--short", "HEAD")
-	if err != nil {
-		return nil, err
-	}
-
-	version, err := getLatestVersion()
```
DoYouEvenCpp, line 69: With that change `getLatestVersion()` is now a dead code (and subsequently `getLatestVersionChangelog`)

magefiles/mage.go
```diff
@@ -305,16 +305,9 @@ func buildPackageDocker(ctx context.Context, packageType string, buildFlags stri
 		mg.Deps(Build.OpenvpnDocker)
 	}
 
-	git, err := getGitInfo()
```
DoYouEvenCpp, line 308: Can you please check if now `gitInfo` (see: https://github.com/NordSecurity/nordvpn-linux/blob/main/magefiles/scripts.go#L19) is not used at all with your changes?

cli/cli_version.go
```diff
@@ -39,6 +40,9 @@ func (c *cmd) Version(ctx *cli.Context) error {
 	if resp.Metadata != "" {
 		fmt.Printf("+%s", resp.Metadata)
 	}
+	if snapconf.IsUnderSnap() {
+		fmt.Printf(" [snap]")
```
devzbysiu: [small] Maybe we could put this logic in some func and reuse? It's kind of awkward, because here we are using `Major`, `Minor`, `Patch` to build version, but in `cli.go` we just get a string, so it won't be super nice anyway. 
If not then maybe we could just put this `" [snap]"` tag into `const` and reuse?

magefiles/scripts.go
mariusSincovici: Removed all the unneeded functions.

magefiles/mage.go
mariusSincovici: deleted it.

cli/cli_version.go
mariusSincovici: Change the logic, now version and -v will return the same information

cli/cli.go
```diff
@@ -1398,3 +1396,17 @@ func readForConfirmationBlockUntilValid(r io.Reader, prompt string) bool {
 		fmt.Println(InputParsingError)
 	}
 }
+
+func composeAppVersion(buildVersion string, environment string, isSnap bool) string {
```
devzbysiu, line 1403: We have `config.BuildTarget` that contains version and environment. Not sure if it makes sense to reuse here and also it does not have information if this is snap build currently, so maybe doesn't make sense to use it, but leaving a comment here just FYI

cli/cli_version_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli/v2"
+)
+
+// TestVersion checks that the cli.Version command adds when needed the outdated information to the version
+func TestVersion(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	mockClient := mockDaemonClient{}
```
devzbysiu: Does it make sense to move `mockDaemonClient` and `captureOutput` to somewhere under `test/` as testing utilities? :thinking:  It's defined in a another test file now

cli/cli.go
mariusSincovici: `config.BuildTarget` is on the deamon and cannot be used on the client..
I've added some comments to the function.

test/helpers/capture_output.go
```diff
@@ -0,0 +1,32 @@
+package helpers
```
mariusSincovici, line 1: didn't had other name ideas. if you have other preferences let me know :smile: .

cli/cli_version_test.go
mariusSincovici: I've moved them to the test

PR #966, opened 2025-07-11
LVPN-8097: Add remote config analytics

DoYouEvenCpp:
Changes:
- added functions to generate RC analytics, using debugger events
- details of each analytics event based on the feature design
- fix for redundant application path while generating dev + context based paths
- UT to cover generic implementation


config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+	if e.Error != "" {
+		eventToMarshal.Result = rcFailure
+	} else {
+		eventToMarshal.Result = rcSuccess
+	}
+
+	jsonData, err := json.Marshal(eventToMarshal)
+	if err != nil {
+		log.Printf("%s%s Failed to marshal RCEvent to JSON %s. Defaulting to empty\n", internal.WarningPrefix, logPrefix, err)
+		jsonData = []byte("{}")
```
bartoszWojciechO: Would we get empty events in case of serialization error? Is this a desired result?

config/remote/utils.go
```diff
@@ -3,13 +3,28 @@ package remote
...
-// defaultMaxGroup represents the maximum value for a rollout group,
-// effectively making the value to be in range of 1-100 (inclusive) to reflect percentage-based groups.
-const defaultMaxGroup uint32 = 100
+const (
+	baseKey = "remote-config"
+
+	// defaultMaxGroup represents the maximum value for a rollout group,
+	// effectively making the value to be in range of 1-100 (inclusive) to reflect percentage-based groups.
+	defaultMaxGroup  uint32 = 100
+	logPrefix               = "[Remote Confg]"
```
devzbysiu: [typo] `Confg`

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+const (
+	RCDownloadErrorRemoteHashNotFound RCDownloadErrorKind = "remote_hash_not_found"
+	RCDownloadErrorRemoteFileNotFound RCDownloadErrorKind = "remote_file_not_found"
+	RCDownloadErrorIntegrity          RCDownloadErrorKind = "integrity_error"
+	RCDownloadErrorFileDownload       RCDownloadErrorKind = "file_download_error"
+	RCDownloadErrorNetwork            RCDownloadErrorKind = "network_error"
+	RCDownloadErrorOther              RCDownloadErrorKind = "other_error"
+)
+
+type Context struct {
```
devzbysiu: Would it be possible to have a different name for this? We have go `context.Context`, then there is `cli.Context` also we are sometimes talking about moose context so maybe would be better to have something else here to not increase the chances for confusion during discussions :smile:

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+	Timestamp        time.Time   `json:"timestamp"`
+}
+
+func NewRCDownloadSuccess(ctx Context, client, featureName string) RCEvent {
+	return RCEvent{
+		Context:     ctx,
+		Client:      client,
+		FeatureName: featureName,
+		Type:        RCDownloadSuccess,
+		Timestamp:   time.Now(),
```
devzbysiu: Won't this be problematic from testing perspective that we have inline dependency on time? :thinking:

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+	MessageNamespace string `json:"namespace"`
+	Result           string `json:"result"`
+	Subscope         string `json:"subscope"`
+	Client           string `json:"client,omitempty"`
+	FeatureName      string
+	Type             RCEventType `json:"type"`
+	Timestamp        time.Time   `json:"timestamp"`
+}
+
+func NewRCDownloadSuccess(ctx Context, client, featureName string) RCEvent {
```
devzbysiu: Maybe we could have enum for name of the feature and use it instead of `string`?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+		FeatureRollout: featureRollout,
+	}
+
+	messageBytes, err := json.Marshal(payload)
+	if err != nil {
+		log.Printf("%s%s Failed to marshal partial rollout message: %s. Defaulting to an empty\n", internal.WarningPrefix, logPrefix, err)
+		messageBytes = []byte{}
+	}
+
+	detail := RCEventDetails{
```
devzbysiu: [typo?] `detail` vs `details`?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+		Client:         client,
+		FeatureName:    featureName,
+		Type:           eventType,
+		Timestamp:      time.Now(),
+		RCEventDetails: detail,
+	}
+}
+
+func NewRCPartialRollout(ctx Context, client, featureName, errorKind string, featureRollout int) RCEvent {
+	//messagae format is slightly different for partial rollout events
```
devzbysiu: [typo] `messagae`

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
 	value := int(num%defaultMaxGroup) + 1
 	return value
 }
+
```
mariusSincovici: Wouldn't make sense to move this entire new code in a separate file?
For example analytics.go, because this is kind of very specific topic and not some utility function like the above.

config/remote/utils.go
```diff
@@ -3,13 +3,28 @@ package remote
...
 	"github.com/google/uuid"
+
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
-// defaultMaxGroup represents the maximum value for a rollout group,
-// effectively making the value to be in range of 1-100 (inclusive) to reflect percentage-based groups.
-const defaultMaxGroup uint32 = 100
+const (
```
mariusSincovici: If more an more things are defined into the const block, maybe it would be time to have a constants.go file for remote?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+
+// RCEvent is the main analytics event structure.
+type RCEvent struct {
+	Context
+	RCEventDetails
+	MessageNamespace string `json:"namespace"`
+	Result           string `json:"result"`
+	Subscope         string `json:"subscope"`
+	Client           string `json:"client,omitempty"`
+	FeatureName      string
```
mariusSincovici: Shouldn't this also have another json name, for example starting with lower case?

config/remote/utils.go
mariusSincovici: This is part of a struct that has custom json keys, shouldn't also this have them?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+	return RCEvent{
+		Context:     ctx,
+		Client:      client,
+		FeatureName: featureName,
+		Type:        RCDownloadSuccess,
+		Timestamp:   time.Now(),
+	}
+}
+
+func NewRCDownloadFailure(ctx Context, client, featureName string, errorKind RCDownloadErrorKind, errorMessage string) RCEvent {
```
mariusSincovici: [minor] Wouldn't be more clear that the functions `NewRC...` create an event if their names jave event instead of `RC`. For example `NewDownloadFailureEvent`? Because the functions are already from the `remote` package so it will be clear is related to RC?

config/remote/utils.go
mariusSincovici: Will the info from here be updated? For example what if the app changes the internet connection and has different ISP.
For name some suggestions `ClientContext`, `UserInfo`, `AppEnvironment`...?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
 	value := int(num%defaultMaxGroup) + 1
 	return value
 }
+
+// RCEventType defines the type of remote config analytics event.
+type RCEventType string
```
mariusSincovici: Not sure how it will be used, but why are RCEventType and RCDownloadErrorKind string, why not ints?

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+		FeatureName:    featureName,
+		Type:           eventType,
+		Timestamp:      time.Now(),
+		RCEventDetails: detail,
+	}
+}
+
+func NewRCPartialRollout(ctx Context, client, featureName, errorKind string, featureRollout int) RCEvent {
+	//messagae format is slightly different for partial rollout events
+	payload := struct {
```
mariusSincovici: Maybe move the struct outside of the function, but not expose it.
Otherwise it would also not be available for tests.

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+		FeatureName    string `json:"feature.name"`
+		RolloutGroup   int    `json:"rollout_group"`
+		FeatureRollout int    `json:"feature.rollout"`
```
mariusSincovici: why in json have `.`? :thinking:

config/remote/utils.go
```diff
@@ -25,3 +40,173 @@ func GenerateRolloutGroup(uuid uuid.UUID) int {
...
+		Context:        ctx,
+		Client:         client,
+		FeatureName:    featureName,
+		Type:           eventType,
+		Timestamp:      time.Now(),
+		RCEventDetails: detail,
+	}
+}
+
+func NewRCPartialRollout(ctx Context, client, featureName, errorKind string, featureRollout int) RCEvent {
```
mariusSincovici: Shouldn't `errorKind` be an `error` and then internally is converted to string? 
Isn't it possible to pass wrapped errors?

config/remote/utils.go
DoYouEvenCpp: Agree. Will remove also the `RC` prefix from other types/names as this indeed seems redundant in the context of `remote` package.

config/remote/utils.go
DoYouEvenCpp: Actually, wanted to encapsulate it and not expose event outside of the function, however I do see your point and am happy to extract it to a local struct, to make it more testable.

Currently, this might be tested by running a generated JSON comparison (e.g.., here: https://github.com/NordSecurity/nordvpn-linux/pull/966/files#diff-dd4f0f78cd79464e18eb12150b4d4292ae0b8dccf3c9e73f06680004dbb25761R105)

config/remote/utils.go
DoYouEvenCpp: Changes to int based.

config/remote/utils.go
DoYouEvenCpp: For the name bit: switched to `UserInfo`
And for the other bit: note the events are created ad-hoc, thus it wasn't designed to create them and store for a long(er) lifetime purposes. If however that occurs to be a case here, then yes - we'll have to update some of the fields (like the ISP, for example).

config/remote/utils.go
DoYouEvenCpp: Good question. As of now, it shouldn't be a problem (as the timestamp doesn't look crucial in here and also we haven't yet clarified out specific use cases).
Anyway, in order to not neglect testability, I've added a possibility to inject a custom time getter method.
If that become too tedious, we can always deliver a factory method, that can nicely hide all the details (eg. passing time.Now to event creation methods).

config/remote/utils.go
DoYouEvenCpp: Will pull in, into created constants.go, https://github.com/NordSecurity/nordvpn-linux/blob/main/config/remote/data.go#L8 and adapt usage of them in analytics as well.

config/remote/utils.go
DoYouEvenCpp: I guess you have a fair point in here.

I'll change it a bit, so in case of an json parsing error we still send out at least all the non-custom fields as a fallback mechanism.

config/remote/analytics.go
```diff
@@ -0,0 +1,171 @@
...
+type Event struct {
+	UserInfo
+	EventDetails
+	MessageNamespace string    `json:"namespace"`
+	Result           string    `json:"result"`
+	Subscope         string    `json:"subscope"`
+	Client           string    `json:"client,omitempty"`
+	FeatureName      string    `json:"feature_name"`
+	Type             EventType `json:"type"`
+	Timestamp        time.Time `json:"timestamp"`
```
keliramu: What is the purpose of adding `Timestamp`? Server is adding record timestamp at the time of receive and store in db already.

config/remote/remote.go
```diff
@@ -54,9 +54,9 @@ func NewCdnRemoteConfig(buildTarget config.BuildTarget, remotePath, localPath st
 		cdn:            cdn,
 		features:       make(FeatureMap),
 	}
-	rc.features.Add(FeatureMain)
-	rc.features.Add(FeatureLibtelio)
-	rc.features.Add(FeatureMeshnet)
+	rc.features.Add(FeatureMain.String())
```
mariusSincovici, line 59: [small] Wouldn't it make sense to change `rc.features` time from string?

config/remote/analytics.go
```diff
@@ -0,0 +1,171 @@
...
+		eventToMarshal.Result = rcFailure
+	} else {
+		eventToMarshal.Result = rcSuccess
+	}
+
+	jsonData, err := json.Marshal(eventToMarshal)
+	if err != nil {
+		log.Printf("%s%s Failed to marshal Event to JSON %s. Fallback to a limited set of data\n", internal.WarningPrefix, logPrefix, err)
+		jsonData = []byte(
+			`{"namespace":"` + messageNamespace + `",` +
```
mariusSincovici: Wouldn't be safer to use an `Event` constructed here and Marshal it?

bartoszWojciechO (approved):

config/remote/analytics.go
DoYouEvenCpp: Removed.

devzbysiu (approved):

config/remote/remote.go
DoYouEvenCpp: Am afraid this might just not fit into scope of this PR, as it requires far more advanced changes than just code reorganisation.

PR #969, opened 2025-07-16
Hide meshnet commands from cli

dfetti:


cli/cli.go
```diff
@@ -1413,3 +1425,20 @@ func composeAppVersion(buildVersion string, environment string, isSnap bool) str
 
 	return fmt.Sprintf("%s%s%s", buildVersion, snap, env)
 }
+
+func shouldHideMeshnet() bool {
+	value := os.Getenv("HIDE_MESHNET")
+
+	if value != "" {
+		parsed, err := strconv.ParseBool(value)
+		if err != nil {
```
bartoszWojciechO, line 1438: Will this be different when we have remote config? Shouldn't app default to hiding a feature on any failures?

cli/cli.go
```diff
@@ -121,222 +122,231 @@ func NewApp(version, environment, hash, salt string,
...
-				BashComplete: cmd.SetAutoConnectAutoComplete,
-				ArgsUsage:    SetAutoConnectArgsUsageText,
-				Description:  SetAutoConnectDescription,
-				Flags: []cli.Flag{
-					&cli.StringFlag{
-						Name:    "group",
-						Aliases: []string{"g"},
-						Usage:   ConnectFlagGroupUsageText,
-					},
+	setSubcommands := []*cli.Command{
```
mariusSincovici: [suggestion] move this and the logic to show/hide meshnet in a separate function. then is also testable.

cli/cli.go
```diff
@@ -121,222 +122,231 @@ func NewApp(version, environment, hash, salt string,
...
+		ArgsUsage:    MsgSetMeshnetArgsUsage,
+		Description:  MsgSetMeshnetDescription,
+		Action:       cmd.MeshSet,
+		BashComplete: cmd.SetBoolAutocomplete,
+	}
+
+	shouldHideMeshnet := shouldHideMeshnet()
+
+	if !shouldHideMeshnet {
+		setSubcommands = append(setSubcommands, &setMeshCommand)
```
mariusSincovici: Should a message be displayed instead of removing the command?

cli/cli.go
dfetti: Yes, it will be different. I just wanted to be careful to not break the tests.

bartoszWojciechO (approved):

cli/cli.go
dfetti: Let's just merge it as it is for now and deal with UX things later when we have the answers.

PR #970, opened 2025-07-17
LVPN-8543: Save rollout group in local config

bartoszWojciechO:


config/static_config.go
```diff
@@ -0,0 +1,129 @@
...
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+// var staticConfigFilename = internal.DatFilesPathCommon + "install_static.dat"
```
devzbysiu: Leftover?

config/static_config.go
```diff
@@ -0,0 +1,129 @@
...
+var staticConfigFilename = filepath.Join(internal.DatFilesPathCommon, "install_static.dat")
+var (
+	ErrStaticValueAlreadySet    = errors.New("static value already configured")
+	ErrStaticValueNotConfigured = errors.New("static value was not configured")
+	ErrFailedToReadConfigFile   = errors.New("failed to read static config file")
+)
+
+type ConfigState int
```
devzbysiu: Does it need to be public? It's returned only by `tryInitStaticConfig` which is private and it's stored in the `FilesystemStaticConfigManager` but also as a private field

config/static_config.go
```diff
@@ -0,0 +1,129 @@
...
+
+const (
+	NoFile ConfigState = iota
+	FailedToInitialize
+	Initialized
+)
```
devzbysiu: Those also could be private maybe?

config/static_config.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if cfg.RolloutGroup == 0 {
+		return 0, ErrStaticValueNotConfigured
+	}
+
+	return cfg.RolloutGroup, nil
+}
+
+func (s *FilesystemStaticConfigManager) SetRolloutGroup(rolloutGroup int) error {
```
devzbysiu, line 113: If not the custom domain type with checks for its value, then we should have preconditions for `rolloutGroup` here to check it fits in `[1; defaultMaxGroup]`

config/static_config.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type StaticConfig struct {
+	RolloutGroup int `json:"rollout_group,omitempty"`
+}
+
+// StaticConfigManager stores values which remain constant thoroughout app's lifetime
+type StaticConfigManager interface {
+	GetRolloutGroup() (int, error)
+	SetRolloutGroup(int) error
```
devzbysiu, line 40: Do we want to go with custom domain-specific type here instead of `int`, which would not allow to be created with incorrect values? :thinking: 
Then we don't need to do any validation in `SetRolloutGroup` etc

bartoszWojciechO: `int` would be better imo, but do you have any idea in mind on how to make it work with serialization/deserialization?

devzbysiu: Yeah, we would probably have to implement marshaling interface for this new type :thinking:. We could do `type RolloutGroup int` but then we are loosing the advantage of having constructor with validation, so it doesn't give us much :/

protobuf/daemon/telemetry/v1/service.proto
```diff
@@ -10,9 +10,6 @@ import "google/protobuf/empty.proto";
 import "fields.proto";
 
 service TelemetryService {
-  // Reports a generic entry for flexible extension
-  rpc SetCustomEntry(CustomEntry) returns (google.protobuf.Empty);
```
mariusSincovici: is by mistake?

config/static_config.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// FilesystemStaticConfigManager saves and reads values to a config file
+type FilesystemStaticConfigManager struct {
```
mariusSincovici, line 44: It would be good to have a mutex to protect access to members

protobuf/daemon/telemetry/v1/service.proto
bartoszWojciechO: This is a leftover that was supposed to be cleaned up from an unrelated PR. It breaks protobuf compilation, so I just removed it here. I can move it somewhere else if we want a separate PR for it.

config/static_config.go
```diff
@@ -0,0 +1,141 @@
...
+	ErrFailedToReadConfigFile   = errors.New("failed to read static config file")
+	ErrRolloutGroupOutOfBounds  = errors.New("rollout group out of bounds")
+)
+
+type configState int
+
+const (
+	noFile configState = iota
+	failedToInitialize
+	initialized
```
thegravi: If these are shared among whole config package, then it could mislead of be shadowed elsewhere without any prefix of something. Especially `initialized`.

config/static_config.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (s *FilesystemStaticConfigManager) GetRolloutGroup() (int, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	cfg, err := s.getConfig()
+	if err != nil {
+		return 0, err
+	}
```
thegravi: Later on you have a wrapped error in the same position, but not here.

```suggestion
	if err != nil {
		return 0, fmt.Errorf("failed to read config: %w", err)
	}
```

config/static_config.go
```diff
@@ -0,0 +1,141 @@
...
+	s.cfg = cfg
+	s.state = state
+
+	return s.cfg, nil
+}
+
+func (s *FilesystemStaticConfigManager) GetRolloutGroup() (int, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
thegravi: We can use here `read-mutex` instead of rw

config/static_config.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return cfg.RolloutGroup, nil
+}
+
+func (s *FilesystemStaticConfigManager) SetRolloutGroup(rolloutGroup int) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if rolloutGroup < 1 || rolloutGroup > 100 {
```
thegravi: 🪄 numbers

```suggestion
	if rolloutGroup < minRolloutGroupValue || rolloutGroup > maxRolloutGrouptValue {
```

config/static_config.go
```diff
@@ -0,0 +1,141 @@
...
+		return ErrRolloutGroupOutOfBounds
+	}
+
+	cfg, err := s.getConfig()
+	if err != nil {
+		return fmt.Errorf("failed to read config: %w", err)
+	}
+
+	if cfg.RolloutGroup != 0 {
```
thegravi: Why not have a dedicated constant for all these cases regarding comparison/returning `0`? To make every usage consistent.

```suggestion
	if cfg.RolloutGroup != unsetRolloutGroupValue {
```

cmd/daemon/main.go
```diff
@@ -292,6 +293,13 @@ func main() {
 
 	machineID := machineIdGenerator.GetMachineID()
 
+	staticCfgManager := config.NewFilesystemStaticConfigManager()
+	if err := staticCfgManager.SetRolloutGroup(remote.GenerateRolloutGroup(machineID)); err != nil {
+		if !errors.Is(err, config.ErrStaticValueAlreadySet) {
+			log.Println(internal.ErrorPrefix, "failed to configure rollout group:", err)
+		}
+	}
+
```
thegravi: Since this performs only on disk changes, can we wrap this execution in a function and move it outside?

PR #973, opened 2025-07-21
Lvpn 5404 tests execution for snap

mariusSincovici:


ci/snap/vagrant/Vagrantfile
```diff
@@ -0,0 +1,19 @@
...
+
+  config.vm.network "private_network", type: "dhcp"
+  # mount the project folder. WORKDIR must be set
+  config.vm.synced_folder ENV["WORKDIR"], "/vagrant"
+
+  # install dependencies
+  config.vm.provision "shell", path: "provision_snap.sh"
+
+  config.vm.define "ubuntu2204", autostart: false do |ubuntu|
+    ubuntu.vm.box = "generic/ubuntu2204"
```
thegravi, line 21: Shall we also add build with GUI option? Since we now have one. Maybe with different box name alias. This could be handy when performing manual testing.

ci/qa_tests_env.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euxo pipefail
+
+# Defines env variables common when running pytests
+
+# Disable the TUI loader indicator to prevent interference during automated tests
+export DISABLE_TUI_LOADER=1
+
+# go code cover dir
+export COVERDIR="covdatafiles"
+export GOCOVERDIR="${WORKDIR}/$COVERDIR"
```
thegravi, line 12: [nit] Check whether `WORKDIR` is defined.

ci/qa_run_tests.sh
```diff
@@ -0,0 +1,65 @@
...
+
+categories="${1}"
+pattern="${2:-}"
+
+# check that some env variables are set before running the tests
+: "${DISABLE_TUI_LOADER:?Environment variable DISABLE_TUI_LOADER must be set}"
+: "${COVERDIR:?Environment variable COVERDIR must be set}"
+: "${GOCOVERDIR:?Environment variable GOCOVERDIR must be set}"
+
+cd "${WORKDIR}"/test/qa || exit
```
thegravi, line 24: [nit] `WORKDIR` check also

ci/test_deb.sh
```diff
@@ -1,10 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euxo pipefail
 
-categories="${1}"
-pattern="${2:-}"
 
-export COVERDIR="covdatafiles"
+# load the env vars needed for tests
+source "${WORKDIR}/ci/qa_tests_env.sh"
```
thegravi, line 6: [nit] also here?

ci/test_snap.sh
```diff
@@ -1,65 +1,18 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euxo pipefail
 
-category="${1}"
-pattern="${2}"
-
-export COVERDIR="covdatafiles"
+source "${WORKDIR}/ci/qa_tests_env.sh"
```
thegravi, line 4: [nit] And here `WORKDIR`

ci/snap/vagrant/Vagrantfile
```diff
@@ -0,0 +1,19 @@
+Vagrant.configure("2") do |config|
+  config.vm.provider :libvirt do |libvirt|
+  end
+
+  config.vm.network "private_network", type: "dhcp"
+  # mount the project folder. WORKDIR must be set
+  config.vm.synced_folder ENV["WORKDIR"], "/vagrant"
```
thegravi: [nit] Check for existence of `WORKDIR`

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+
+func vagrantUp(vagrantEnv VagrantEnv) error {
+	return runVagrantCmd(vagrantEnv, "up", vagrantEnv.BoxName)
+}
+
+func vagrantStop(vagrantEnv VagrantEnv) {
+	runVagrantCmd(vagrantEnv, "halt", "-f", vagrantEnv.BoxName)
+	if vagrantEnv.ShouldDestroyVM {
+		runVagrantCmd(vagrantEnv, "destroy", "-f", vagrantEnv.BoxName)
+	}
```
thegravi: Silent error handling

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return cmd.Wait()
+}
+
+func streamLines(r io.Reader, out io.Writer) {
+	sc := bufio.NewScanner(r)
+	for sc.Scan() {
+		fmt.Fprintln(out, sc.Text())
+	}
```
thegravi, line 86: [small] Why not also check scanner errors?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+	vagrantFileDir := path.Join(cwd, "ci", "snap", "vagrant")
+
+	fmt.Println("WORKDIR=", cwd)
+	fmt.Println("Vagrantfile folder", vagrantFileDir)
+
+	shouldDestroyVM := false
+	if val, ok := env["SNAP_TEST_DESTROY_VM_ON_EXIT"]; ok {
+		shouldDestroyVM, err = nstrings.BoolFromString(val)
+		if err != nil {
+			shouldDestroyVM = false
```
thegravi: [small] Log message regarding conversion error?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return err
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return err
+	}
+
+	go streamLines(stdout, os.Stdout)
+	go streamLines(stderr, os.Stderr)
```
thegravi, line 72: Who is responsible for waiting for these and making sure that they are stopped at some point?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+		runVagrantCmd(vagrantEnv, "destroy", "-f", vagrantEnv.BoxName)
+	}
+}
+
+// buildVagrantEnvEnv will construct a VagrantEnv storing all the env variables needed
+// Used env variables:
+// * SNAP_TEST_BOX_NAME specifies what box to use, for example generic/ubuntu2204
+// * SNAP_TEST_DESTROY_VM_ON_EXIT - optional, when set to 1|true it will delete the VM after running the tests
+// * NA_TESTS_CREDENTIALS - test credentials
+func buildVagrantEnvEnv() (VagrantEnv, error) {
```
thegravi: Quirky naming :)

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return VagrantEnv{}, err
+	}
+
+	env, err := getEnv()
+	if err != nil {
+		return VagrantEnv{}, err
+	}
+
+	boxName := env["SNAP_TEST_BOX_NAME"]
```
thegravi, line 143: Shouldn't we also check the name format? So, nothing unsupported would be provided

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+
+	vagrantEnv := VagrantEnv{
+		Cwd:             cwd,
+		VagrantFileDir:  vagrantFileDir,
+		BoxName:         boxName,
+		ShouldDestroyVM: shouldDestroyVM,
+		RemoteCwd:       "/vagrant",
+		TestCredentials: newSecret(env["NA_TESTS_CREDENTIALS"]),
```
thegravi, line 167: Is there a check for `NA_TESTS_CREDENTIALS` content?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+
+	go streamLines(stdout, os.Stdout)
+	go streamLines(stderr, os.Stderr)
+
+	// Start the command
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	return cmd.Wait()
```
thegravi, line 79: What about providing a timeout if things get stuck?

Maybe also add error logging here before returning `cmd.Wait()` response?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
+//go:build mage
```
thegravi, line 1: Is there a check anywhere to validate whether `vagrant` is installed on the system?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+		vagrantEnv.RemoteCwd,
+		strings.Join(env, " "),
+		strings.Join(args, " "),
+	)
+	fmt.Println("executing remote command:", fullRemoteCmd)
+	return runVagrantCmd(vagrantEnv, "ssh", vagrantEnv.BoxName, "-c", fullRemoteCmd)
+}
+
+func vagrantUp(vagrantEnv VagrantEnv) error {
+	return runVagrantCmd(vagrantEnv, "up", vagrantEnv.BoxName)
```
thegravi, line 113: I think it would be wise to check whether the VM is already running before starting a new one.

TESTING.md
```diff
@@ -77,6 +77,10 @@ To run tests without rebuilding everything each time use `test:qaDockerFast` ins
 ### Test credentials for QA tests
 `NA_TESTS_CREDENTIALS` environment variable is used to configure test credentials. It is a JSON
 object containing a key(string):value(credentials) map.
+For running the tests using Vagrant for the snap build the following environment variable are required:
+* `SNAP_TEST_BOX_NAME` - represents the configuration name used by vagrant from [Vagrantfile](ci/snap/vagrant/Vagrantfile).
+* `SNAP_TEST_DESTROY_VM_ON_EXIT` - [optional] a boolean value that controls if vagrant should destroy the created virtual machine after running the tests. By default is is `false`.
```
DoYouEvenCpp: [nit] A typo in `By default is is false.` which should be `By default it is false`. 
However I propose to formulate it using passive-voice, eg. `By default set to false`

ci/qa_run_tests.sh
```diff
@@ -0,0 +1,65 @@
...
+LOGS_FOLDER="${WORKDIR}"/dist/logs
+
+mkdir -p "${LOGS_FOLDER}"
+mkdir -p "${ARTIFACTS_FOLDER}"
+mkdir -p "${GOCOVERDIR}"
+
+python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true \
+--html="${WORKDIR}"/dist/test_artifacts/report.html --self-contained-html  --junitxml="${WORKDIR}"/dist/test_artifacts/report.xml "${args[@]}"
+
+# # To print goroutine profile when debugging:
```
DoYouEvenCpp, line 58: Do we need that code, or simply a leftover?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/NordSecurity/nordvpn-linux/nstrings"
+)
+
+// Define a new type to hide it's content from prints
```
DoYouEvenCpp: [nit] it's -> its

ci/snap/vagrant/provision_snap.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
DoYouEvenCpp: In the other place(s) it's `#!/usr/bin/env bash` - care to unify it here as well?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+func newSecret(val string) secret {
+	return secret{value: val}
+}
+func (s secret) GoString() string {
+	return s.String()
+}
+
+func (s secret) String() string {
+	const display = 10
+	str := string(s.value)
```
DoYouEvenCpp: s.value is already a string, no?

magefiles/vagrant.go
```diff
@@ -0,0 +1,183 @@
...
+
+// runCommandInVM - runs a command using vagrant ssh inside the VM
+func runCommandInVM(vagrantEnv VagrantEnv, args []string) error {
+	env := []string{
+		fmt.Sprintf("WORKDIR=%s", vagrantEnv.RemoteCwd),
+		fmt.Sprintf("NA_TESTS_CREDENTIALS='%s'", vagrantEnv.TestCredentials),
+	}
+
+	// compose the remote command to be on single string:
+	// cd /vagrant + env + <cmd + params>
```
DoYouEvenCpp: I guess we could be more specific here, and instead of `+` in the comment, make it almost akin to the actual command to be applied (eg.: `cd <RemoteCwd> && ...` )

magefiles/vagrant.go
DoYouEvenCpp: Also, assignment to `false` seems redundant here. `BoolFromString` returns `false` in case of any error.

ci/qa_run_tests.sh
mariusSincovici: The problem with `WORKDIR` is that it is used everywhere in the code, in go, python, bash . So the tests will most likely not even get here if it is not set. I also think is not bringing too much value.

The other variables are checked because they are kind of optional, specially for `DISABLE_TUI_LOADER`.
I'll probably remove the checks also for the others, except `DISABLE_TUI_LOADER` because it doesn't bring anything :smile:.

ci/qa_run_tests.sh
mariusSincovici: This is moved from other place.
 I'm not sure if anyone uses this in tests. In debug builds the pprof is running, so I was thinking to let it here in case someone checks something to see that it is available :thinking:. 
 But if it doesn't make sense, it can be deleted.

ci/snap/vagrant/Vagrantfile
mariusSincovici: This file will be updated in the future, but not now. The new configs will be added as start to be used. For this was also introduced the new env var `SNAP_TEST_BOX_NAME` to be able to support future configs.

ci/snap/vagrant/provision_snap.sh
mariusSincovici: yes, for sure. missed it.

magefiles/vagrant.go
mariusSincovici: If vagrant is not installed in the system then the commands will fail to execute, with command not found.
Because this is not using go module to talk with vagrant, just starts vagrant commands.
I've thought it doesn't bring too much if scripts print `vagrant not installed` vs `command not found vagrant` or similar :laughing:. 
But I can add if you think is useful.

magefiles/vagrant.go
mariusSincovici: right, actually I've changed to have the `secret` just as a string alias.

magefiles/vagrant.go
mariusSincovici: Nobody, they will be stopped in 2 ways, the command ends or the mage command finished execution.

magefiles/vagrant.go
mariusSincovici: Who is calling this will need to log the error, otherwise it will be duplicated.

Adding timeout is complicated because it is hard to find a good value, maybe the command execution takes lot of time. For example to setup the vagrant box maybe takes 1-2-3 minutes. Same with running the tests, can be 10 minutes, depending on the tests.

magefiles/vagrant.go
mariusSincovici: Not really needed, because if the name is not found by vagrant into Vagrantfile configs then it will just fail to start the VM. Also this will require to know now if in the future there will be custom VMs with custom name format created.

magefiles/vagrant.go
mariusSincovici: I've changed to return the error so the caller will decide what to to,

`BoolFromString` is not guaranteed to return false when there is an error. It is recommended that if a function returns an error its, then result must not be used.

magefiles/vagrant.go
mariusSincovici: No validation is made for this, because it is not used here but in tests. If this is incorrect then the tests will fail with the exact error.

TESTING.md
mariusSincovici: done

magefiles/vagrant.go
mariusSincovici: done

magefiles/vagrant.go
mariusSincovici: done

magefiles/vagrant.go
mariusSincovici: done

magefiles/vagrant.go
mariusSincovici: Depends what should be checked. To run the tests it is not important if it is running, because vagrant will consider it ok.

But what should it happen if it is running, stop it/error/give instructions how to stop?

magefiles/vagrant.go
mariusSincovici: added logs

magefiles/vagrant.go
mariusSincovici: done

magefiles/vagrant.go
DoYouEvenCpp: For the record - this is the implementation of `BoolFromString` we talk about here, right? - https://github.com/NordSecurity/nordvpn-linux/blob/main/nstrings/bool.go#L34

Yes, the false when an error is emitted has a special meaning - ideally the method should return a triboolean (or any other tri-state boolean),

ci/qa_run_tests.sh
thegravi: fair enough

magefiles/vagrant.go
thegravi: Why not just put something as a safeguard, like 1h?

magefiles/vagrant.go
thegravi: But in a log message you can say, please do sudo apt-get install etc... :D But this would require to detect os/distro. So not worth it.

magefiles/vagrant.go
thegravi: Isn't the VM stopped or paused after tests are completed? Or is it just left "dangling"?

mariusSincovici: The VM is stopped in the `defer` from [here](https://github.com/NordSecurity/nordvpn-linux/pull/973/files#diff-2c74c6c8984c3cc8e2a2f46fa1a2a2e3aacbe4a12a602457fc5525f488bc78ceR182) if start was successful. It might also be destroyed, depending on the env variable.
The only way when might not be stopped is if ctrl+c is pressed and then `mage` command is killed.

thegravi (approved):

DoYouEvenCpp (approved):

devzbysiu (approved):

PR #974, opened 2025-07-23
LVPN-6747: Access and TP token integration into session store

thegravi:
Changes:
- Reworked `TokenManager` into `AccessTokenSessionStore` for a more generic approach to session management
- Implemented `TrustedPassSessionStore` for managing trusted-pass token


auth/auth.go
```diff
@@ -212,33 +215,6 @@ func (r *RenewingChecker) renew(uid int64, data config.TokenData) error {
...
-			if errors.Is(err, core.ErrUnauthorized) ||
-				errors.Is(err, core.ErrNotFound) ||
-				errors.Is(err, core.ErrBadRequest) {
-				return r.cm.SaveWith(Logout(uid, r.logoutPub))
-			}
-		}
-
 		if err := r.cm.SaveWith(saveLoginCreds(uid, data)); err != nil {
 			return err
 		}
```
mariusSincovici, line 220: isn't [line 225](https://github.com/NordSecurity/nordvpn-linux/pull/974/files#diff-f5c9a39fac1719f2a73593994b931d8a874449548e8a3753e1be46e761e503eaR225) calling the same thing as it was in `if err := store.Renew(); err != nil {`?

cmd/daemon/init.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/NordSecurity/nordvpn-linux/session"
+	"github.com/google/uuid"
+)
+
+func buildAccessTokenSessionStore(
+	confman config.Manager,
+	errRegistry *internal.ErrorHandlingRegistry[int64],
+	clientAPI core.RawClientAPI,
+) session.SessionStore {
+	return session.NewAccessTokenSessionStore(
```
mariusSincovici, line 18: would it be possible to move in a separate functions the parameters because is quite hard to follow where one object creation is finished?

session/validators.go
```diff
@@ -0,0 +1,82 @@
...
+	return v.ValidateFunc(carrier.GetUsername(), carrier.GetPassword())
+}
+
+// NewCredentialValidator creates a validator that checks session credentials.
+func NewCredentialValidator(fn func(username, password string) error) SessionStoreValidator {
+	return &CredentialValidator{ValidateFunc: fn}
+}
+
+type CompositeValidator struct {
+	Validators []SessionStoreValidator
```
mariusSincovici: do they need to be public?

session/access_token_store_validators.go
```diff
@@ -0,0 +1,54 @@
...
+
+	return v.ValidateFunc(tokenProvider.GetToken(), expiryProvider.GetExpiry())
+}
+
+// NewManualAccessTokenValidator creates a validator that verifies manually issued access tokens.
+func NewManualAccessTokenValidator(api AnyIdempotentAPICallWithToken) SessionStoreValidator {
+	return &manualAccessTokenValidator{
+		ValidateFunc: func(token string, expiryDate time.Time) error {
+			if expiryDate.Equal(ManualAccessTokenExpiryDate) {
+				// here we have a manually issuesed access token
```
mariusSincovici: [spelling] issuesed

session/access_token_store_validators.go
```diff
@@ -0,0 +1,54 @@
...
+					return fmt.Errorf("invalid access token format: %w", ErrAccessTokenExpired)
+				}
+
+				err := api(token)
+				// the only interest is in whether we receive "unauthorized access" error
+				if errors.Is(err, ErrUnauthorized) {
+					return ErrAccessTokenRevoked
+				}
+
+				return nil
```
mariusSincovici: [small] this return might not be needed

session/access_token_store_validators.go
```diff
@@ -0,0 +1,54 @@
...
+		return errors.New("unsupported store for expiry validation")
+	}
+
+	return v.ValidateFunc(tokenProvider.GetToken(), expiryProvider.GetExpiry())
+}
+
+// NewManualAccessTokenValidator creates a validator that verifies manually issued access tokens.
+func NewManualAccessTokenValidator(api AnyIdempotentAPICallWithToken) SessionStoreValidator {
+	return &manualAccessTokenValidator{
+		ValidateFunc: func(token string, expiryDate time.Time) error {
```
mariusSincovici: would it be possible to move the func in a separate function?

cmd/daemon/init.go
```diff
@@ -0,0 +1,90 @@
...
+		confman,
+		session.NewCompositeValidator(
+			session.NewExpiryValidator(),
+			session.NewManualAccessTokenValidator(func(token string) error {
+				// this is the least expensive api call that needs authentication
+				_, err := clientAPI.CurrentUser(token)
```
mariusSincovici: how often will this be executed?
Maybe have a time and not to be executed more often than x hour.

session/access_token_store.go
```diff
@@ -0,0 +1,125 @@
...
+		return c
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *AccessTokenSessionStore) SetToken(value string) error {
```
mariusSincovici: Why not define the functions in the same file like the other class implementations `AccessTokenSessionStore`?

session/trusted_pass_store.go
```diff
@@ -0,0 +1,125 @@
...
+		return c
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *TrustedPassSessionStore) SetToken(value string) error {
```
mariusSincovici: Same with `TrustedPassSessionStore`, why have it in multiple files?

session/access_token_session_store.go
```diff
@@ -0,0 +1,155 @@
...
+		validator:          validator,
+		renewAPICall:       renewAPICall,
+		session:            newAccessTokenSession(cfgManager),
+	}
+}
+
+// Renew iterates over stored user tokens and attempts to renew each one that has expired.
+// Returns a combined error if one or more renewals fail.
+func (s *AccessTokenSessionStore) Renew() error {
+	var cfg config.Config
```
mariusSincovici: Is this needed here? Could the loading be moved after validating?

session/access_token_store.go
```diff
@@ -0,0 +1,125 @@
...
+		c.TokensData[c.AutoConnectData.ID] = data
+		return c
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
mariusSincovici: I didn't saw for this functions to be executed somewhere, except tests. Are they or will they be called later?

session/trusted_pass_session_store.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func (m *TrustedPassSessionStore) renewToken(data *config.TokenData) error {
+	resp, err := m.renewAPICall(data.Token)
+	if err != nil {
+		return fmt.Errorf("getting trusted pass token data: %w", err)
+	}
+
+	if err := m.SetToken(resp.Token); err != nil {
```
mariusSincovici: Wouldn't it make sense to have a single call for all the setters? 
Because now it results in writing the file 3 times in a row. Also if there are parallel calls will use partial info, some updated and some not.

session/trusted_pass_store.go
```diff
@@ -0,0 +1,125 @@
...
+	return cfg.ExpiresAt
+}
+
+// implements ExpirableSession
+func (s *TrustedPassSessionStore) IsExpired() bool {
+	cfg, err := s.session.get()
+	if err != nil {
+		return true
+	}
+	return time.Now().After(cfg.ExpiresAt)
```
mariusSincovici: Maybe not so problematic, but this will renew all the tokens or update.

session/trusted_pass_store.go
```diff
@@ -0,0 +1,125 @@
...
+		return trustedPassConfig{}, err
+	}
+
+	data, ok := cfg.TokensData[cfg.AutoConnectData.ID]
+	if !ok {
+		return trustedPassConfig{}, errors.New("non existing data")
+	}
+
+	// must contain valid data
+	expiryTime, _ := time.Parse(internal.ServerDateFormat, data.TrustedPassTokenExpiry)
```
mariusSincovici: When the `data.TrustedPassTokenExpiry` is empty then `expiryTime` is still used. It would be safer to set to something before using it.

session/accessors.go
mariusSincovici, line 1: Are this functions used somewhere?

session/accessors.go
```diff
@@ -0,0 +1,38 @@
+package session
+
+import (
+	"errors"
+	"time"
+)
+
+// GetToken retrieves a token from a SessionTokenProvider
+func GetToken(store interface{}) (string, error) {
```
mariusSincovici: Why not use the `SessionTokenProvider` for param instead of dynamically checking the type and return an error?
Same for the others.

session/trusted_pass_session_store.go
```diff
@@ -0,0 +1,146 @@
...
+	if isNotValid {
+		if err := m.renewIfOAuth(&data, cfg.AutoConnectData.ID); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (m *TrustedPassSessionStore) Invalidate(reason error) error {
```
mariusSincovici: Shouldn't this delete the stored token from the config?

session/trusted_pass_store_validators.go
```diff
@@ -0,0 +1,32 @@
...
+type OwnerIDValidator struct{}
+
+// Validate
+func (v *OwnerIDValidator) Validate(session any) error {
+	carrier, ok := session.(SessionOwnerProvider)
+	if !ok {
+		return errors.New("unsupported store for owner id validation")
+	}
+
+	if carrier.GetOwnerID() != TrustedPassOwnerID {
```
mariusSincovici: What would happen if there is no trusted pass for manually generated tokens?

session/validators.go
```diff
@@ -0,0 +1,82 @@
+package session
+
+import "errors"
+
+type TokenValidator struct {
+	ValidateFunc func(token string) error
+}
+
+// Validate checks if the provided session has a valid token.
+func (v *TokenValidator) Validate(session any) error {
```
mariusSincovici: same here, could it session of type `SessionTokenProvider`?

session/access_token_session_store.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return nil
+}
+
+// Invalidate triggers error handlers for all stored user tokens using the provided error.
+// It does not modify or remove any tokens from storage and leaves this responsibility to the
+// client.
+func (s *AccessTokenSessionStore) Invalidate(reason error) error {
+	var cfg config.Config
+	if err := s.cfgManager.Load(&cfg); err != nil {
```
mariusSincovici: Shouldn't the tokens be deleted from the settings? or is made from somewhere else?

session/trusted_pass_session_store.go
```diff
@@ -0,0 +1,146 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+const (
+	// it's predefined value, but not retrievable from any API
```
DoYouEvenCpp, line 14: nit: then maybe we could have a set of common constants, predefining different time-based values (a.k.a. expiration) somewhere accessible for all the packages in the project? (but this looks like a minor, future development idea)

session/trusted_pass_session_store.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (m *TrustedPassSessionStore) isLogoutNeeded(err error) bool {
+	return errors.Is(err, ErrUnauthorized) ||
+		errors.Is(err, ErrNotFound) ||
+		errors.Is(err, ErrBadRequest)
+}
+
+func (m *TrustedPassSessionStore) doLogout(reason error, uid int64) {
+	// err  handler will take care of event publishing (which will need to be catched for logout)
```
DoYouEvenCpp: ultra nit: there's an extra space between `err` and `handler` there ;)

session/trusted_pass_store_validators.go
```diff
@@ -0,0 +1,32 @@
...
+		return ErrInvalidOwnerId
+	}
+
+	return nil
+}
+
+// NewOwnerIDValidator
```
devzbysiu: Missing doc comments?

session/trusted_pass_store_validators.go
```diff
@@ -0,0 +1,32 @@
...
+	ErrInvalidOwnerId = errors.New("invalid owner id")
+)
+
+const (
+	TrustedPassOwnerID = "nordvpn"
+)
+
+type OwnerIDValidator struct{}
+
+// Validate
```
devzbysiu: Missing doc comments?

session/validators_test.go
devzbysiu, line 1: Why not table driven tests? At least for those tests that are the very similar

session/access_token_store_validators.go
thegravi: This is the only thing that this function does - it encapsulates another one. Is there a reason why you think that this is not a suitable location for it?

session/access_token_store.go
thegravi: The rationale was that since the publicly exposed functions are `Renew` and `Invalidate` and those getter/setters are used indirectly via accessors, that is why I decided to group them closely with the `store` object. Since it's the same package and the coupling is intentional, I decided it's a suitable location.

session/access_token_store.go
thegravi: Which one? `SetToken()`?

session/trusted_pass_session_store.go
thegravi: Here, I depend on the bright future where we have a configuration caching mechanism already implemented 🙏🏻 . This would justify extra calls without loading every time, hence the comment at the beginning of the file.

session/trusted_pass_store.go
thegravi: If it is empty, then the time is the 1st year of our Common Era. Is that safe enough to be used as expired? :)

session/trusted_pass_session_store.go
thegravi: We can discuss. The initial idea was to allow the client to decide how to "invalidate", hence the error handling registry module.

session/accessors.go
thegravi: Because we are dealing with generic `SessionStore` interfaces, which by default do not implement any of these session providers.

session/validators.go
```diff
@@ -0,0 +1,82 @@
...
+	carrier, ok := session.(SessionTokenProvider)
+	if !ok {
+		return errors.New("unsupported store for token validation")
+	}
+
+	return v.ValidateFunc(carrier.GetToken())
+}
+
+// NewTokenValidator creates a validator that checks session tokens.
+func NewTokenValidator(fn func(token string) error) SessionStoreValidator {
```
devzbysiu: `fn` should be checked, otherwise I can pass `nil` and it will blew up with panic when you try to call it

session/validators.go
```diff
@@ -0,0 +1,82 @@
...
+	return &TokenValidator{ValidateFunc: fn}
+}
+
+type ExpiryValidator struct{}
+
+// Validate checks if the provided session has expired.
+func (v *ExpiryValidator) Validate(session any) error {
+	expirable, ok := session.(ExpirableSession)
+	if !ok {
+		return errors.New("unsupported store for expiration validation")
```
devzbysiu: Would be nice to have `session` type logged here and in other such places to make debugging easier

session/access_token_session_store.go
thegravi: In the daemon main, client-side business logic is integrated into the error handling registry, which is invoked from any session store. Inside a logout is called, where those parameters are deleted. It is done on specific errors.

session/trusted_pass_store_validators.go
thegravi: Manually or not, TP needs an access token to retrieve its token. The source how the initial access token was obtained, I think, is irrelevant.

session/trusted_pass_store.go
```diff
@@ -0,0 +1,128 @@
...
+func (s *TrustedPassSessionStore) SetToken(value string) error {
+	cfg, err := s.session.get()
+	if err != nil {
+		return err
+	}
+	cfg.Token = value
+	return s.session.set(cfg)
+}
+
+func (s *TrustedPassSessionStore) SetOwnerID(value string) error {
```
devzbysiu: All the setters here are called in one place and every setter is reading and writing to file, why not to have one call that sets everything that's needed all at once? Maybe even something similar as in `config/manager.go` - `SaveWith` where you get the config, do mutation on it and save

session/trusted_pass_store.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return trustedPassConfig{
+		Token:     data.TrustedPassToken,
+		OwnerID:   data.TrustedPassOwnerID,
+		ExpiresAt: expiryTime,
+	}, nil
+}
+
+// TODO: needs global caching
```
devzbysiu: Why? What problem do we need to solve with caching? :thinking: Similarly [here](https://github.com/NordSecurity/nordvpn-linux/pull/974/files#diff-4a216653629fec33ef9c5a8b6af7c88ea519d2b7c21ba67571b32fb04f3e62d2R21)

session/validators.go
```diff
@@ -0,0 +1,82 @@
...
+
+// NewTokenValidator creates a validator that checks session tokens.
+func NewTokenValidator(fn func(token string) error) SessionStoreValidator {
+	return &TokenValidator{ValidateFunc: fn}
+}
+
+type ExpiryValidator struct{}
+
+// Validate checks if the provided session has expired.
+func (v *ExpiryValidator) Validate(session any) error {
```
devzbysiu: Is there a reason you are using `any` in those calls? It's sacrificing the use of type system so usually there should be a good reason for this. Is it only because of `CompositeValidator` or there is something else here?

session/access_token_store_validators.go
devzbysiu: I second what @mariusSincovici says. It worsens readability and I don't know any benefit of this

cmd/daemon/init.go
```diff
@@ -0,0 +1,90 @@
...
+				// map to internal errors
+				if errors.Is(err, core.ErrUnauthorized) {
+					return session.ErrUnauthorized
+				}
+				return nil
+			}),
+		),
+		errRegistry,
+		func(token string, idempotencyKey uuid.UUID) (*session.AccessTokenResponse, error) {
```
devzbysiu: Also here with inline func, why not to move it to a separate, named func?

cmd/daemon/init.go
```diff
@@ -0,0 +1,90 @@
...
+
+			// map to internal errors
+			switch {
+			case errors.Is(err, core.ErrBadRequest):
+				err = session.ErrBadRequest
+			case errors.Is(err, core.ErrUnauthorized):
+				err = session.ErrUnauthorized
+			case errors.Is(err, core.ErrNotFound):
+				err = session.ErrNotFound
+			}
```
devzbysiu: Duplicate with the mapping below. Maybe would be better to have it in a single place? Unless we are sure that this will be often changed and for a different reasons

auth/token_expiration_test.go
```diff
@@ -0,0 +1,304 @@
...
+		StatusCode: http.StatusUnauthorized,
+		Header: http.Header{
+			"Content-Type": []string{"application/json"},
+		},
+		Body:          io.NopCloser(bytes.NewReader([]byte(errorBody))),
+		ContentLength: int64(len(errorBody)),
+	}, nil
+}
+
+func createDefaultRoundTripper(rt *mockRoundTripperWithExpiration) func(req *http.Request) (*http.Response, error) {
```
devzbysiu, line 106: This could be done with `t.Helper()`, then you can fail with `t.Error()` or similar instead of panic and it will show the line in the test that failed

auth/token_test.go
```diff
@@ -108,11 +107,12 @@ func TestTokenRenewWithBadConnection(t *testing.T) {
 		expectedIdempotencyKey: idempotencyKey,
 	}
 
+	uid := int64(1)
 	cm := memoryConfigManager{
 		c: config.Config{
-			AutoConnectData: config.AutoConnectData{ID: 1},
+			AutoConnectData: config.AutoConnectData{ID: uid},
```
devzbysiu, line 114: Here `ID` was set to `1` and then `TokensData` below had token under the key `0`, now it's both `1`. Was it wrong before? :thinking: 
Similar [here](https://github.com/NordSecurity/nordvpn-linux/pull/974/files/80e40dc610ae3915ba196101995b2904af41b96d#diff-64505921a130d998dc87eefe666aabc3bced755658b9313148319b5e1dd53325R242)

session/validators_test.go
thegravi: Decided to do it this way :)

core/login_token_manager_test.go
```diff
@@ -1,773 +1,775 @@
...
-		},
-		&mockErrRegsitry,
-		validator,
-	)
-
-	err := tokenman.Renew()
-	assert.NoError(t, err)
-	assert.Empty(t, mockCfg.c.TokensData, "TokensData must be empty")
-}
+// WILL BE REUSED FOR SESSIONSTORE testing
```
devzbysiu: Is this leftover?

daemon/rpc_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"testing"
 	"time"
 
+	testsession "github.com/NordSecurity/nordvpn-linux/test/mock/session"
```
devzbysiu, line 17: [small] in other places this is called `mocksession`, would be nice to be consistent maybe, but not a big deal

session/validators.go
thegravi: Wanted to do that, but forgot. Thanks.

session/trusted_pass_store.go
thegravi: Constant data loading from the disk. With in-memory chaching we would make a quick copy without need to access the disk. When saving changes we would update not only the cache, but also the disk data.

session/validators.go
thegravi: It's used for all SessionStore interfaces, not only for this. And this lookup is not pricey at all.

cmd/daemon/init.go
thegravi: It is already changed.

cmd/daemon/main.go
```diff
@@ -498,9 +491,9 @@ func main() {
 		meshnetEvents.PeerUpdate,
 		nc.NewCredsFetcher(clientAPI, fsystem))
 
-	// on token invalidation (unauthorized access, missing server resources, invalid request)
+	// on session invalidation (unauthorized access, missing server resources, invalid request)
 	// perform user log-out action
-	loginTokenErrHandlingReg.Add(
+	invalidSessionErrHandlingReg.Add(
 		func(uid int64) {
```
mariusSincovici: Now this parameter is never used, but it requires all the sessions to load the settings file and pass it.
Wouldn't it make sense to remove it?

session/access_token_session_store.go
```diff
@@ -0,0 +1,155 @@
...
+			td.RenewToken = data.RenewToken
+			td.TokenExpiry = data.TokenExpiry
+			c.TokensData[uid] = td
+			return c
+		})
+	}
+
+	if errors.Is(err, ErrUnauthorized) || errors.Is(err, ErrNotFound) || errors.Is(err, ErrBadRequest) {
+		defer s.invokeClientErrorHandlers(uid, err)
+		if err := s.cfgManager.SaveWith(func(c config.Config) config.Config {
```
mariusSincovici: here the data is deleted from the settings, why not let also here the error handle do the cleanup?

session/access_token_session_store.go
```diff
@@ -0,0 +1,155 @@
...
+			delete(c.TokensData, uid)
+			return c
+		}); err != nil {
+			return fmt.Errorf("removing token data: %w", err)
+		}
+	}
+
+	return nil
```
mariusSincovici, line 160: If there is an error, function will return no error.

session/access_token_session_store.go
```diff
@@ -0,0 +1,155 @@
...
+			td := c.TokensData[uid]
+			td.Token = data.Token
+			td.RenewToken = data.RenewToken
+			td.TokenExpiry = data.TokenExpiry
+			c.TokensData[uid] = td
+			return c
+		})
+	}
+
+	if errors.Is(err, ErrUnauthorized) || errors.Is(err, ErrNotFound) || errors.Is(err, ErrBadRequest) {
```
mariusSincovici: Wouldn't this duplicate what the error handlers are doing? Because now there will be 2 lists with errors defined

session/errors.go
```diff
@@ -0,0 +1,32 @@
+package session
+
+import (
+	"errors"
+)
+
+// generic errors
+var (
```
mariusSincovici, line 7: Now there are multiple errors created in different places for the same things. Is this really needed?

Because of this when the API returns `Unauthorized` error it will not do automatic logout, because this `ErrUnauthorized` != from core.ErrUnauthorized. One starts with capital letter.

auth/token_expiration_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+const (
+	defaultTokenValue     = "valid-token"
+	renewedTokenValue     = "new-token-value"
+	testUsername          = "testuser"
+	testEmail             = "test@example.com"
+	defaultExpirationDays = 24 * time.Hour
+)
+
+type mockConfigManager struct {
```
mariusSincovici, line 34: Would it be possible to have this in some mock class in one place, because now it is duplicated 4 times?
And the same implementation exists in `mock.ConfigManager`

cmd/daemon/init.go
mariusSincovici: Otherwise this makes a call to the API every few seconds for the manual tokens :smile:

cmd/daemon/main.go
thegravi: Yes, it's residue. Gonna remove.

core/login_token_manager_test.go
thegravi: Yes, gonna remove.

auth/token_test.go
thegravi: Before we had the implementation where we loop over all map entries (hence UID), that's why it passed. But I do not see any value in doing so since we are only using 1 UID per machine.

auth/token_expiration_test.go
```diff
@@ -0,0 +1,304 @@
...
+	idempotencyKey := uuid.New()
+	futureExpiry := time.Now().Add(defaultExpirationDays).Format(internal.ServerDateFormat)
+
+	api, mockRT, _ := setupTestEnvironment(idempotencyKey, futureExpiry, nil)
+
+	mockRT.respError = errors.New("network error")
+	mockRT.ForceExpiration()
+
+	resp, err := api.CurrentUser()
+	assert.Error(t, err)
```
DoYouEvenCpp, line 318: Here we could also assert on error type, like:
`assert.ErrorIs(t, err, core.ErrUnauthorized)`, correct?

auth/token_test.go
```diff
@@ -244,27 +257,39 @@ func Test_TokenRenewForcesUserLogout(t *testing.T) {
...
+			func(token string, idempotencyKey uuid.UUID) (*session.AccessTokenResponse, error) {
+				resp, err := api.TokenRenew(token, idempotencyKey)
+				if err != nil {
+					return nil, err
+				}
+				return &session.AccessTokenResponse{
+					Token:      resp.Token,
+					RenewToken: resp.RenewToken,
+					ExpiresAt:  resp.ExpiresAt,
+				}, err
```
DoYouEvenCpp: [ultra-nit] instead of returning `err` here explicitly, `nil` might be used instead.
Reason: nil makes one immediately sure it's a non-error case and requires really no time spent on reading the code and understanding the logic in here.

cmd/daemon/init.go
DoYouEvenCpp: Or at least some comment describing what validators are used for the logic.

cmd/daemon/init.go
```diff
@@ -0,0 +1,101 @@
...
+			if errors.Is(err, core.ErrUnauthorized) {
+				return session.ErrUnauthorized
+			}
+			return nil
+		}),
+	)
+}
+
+func convertCoreToSessionError(err error) error {
+	switch {
```
DoYouEvenCpp: `ErrForbidden` skipped intentionally?

cmd/daemon/init.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func buildTrustedPassSessionStoreValidators() session.SessionStoreValidator {
+	return session.NewCompositeValidator(session.NewExpiryValidator(), session.NewOwnerIDValidator())
+}
+
+func buildTrustedPassSessionStoreAPIRenewalCall(clientAPI core.ClientAPI) session.TrustedPassRenewalAPICall {
+	return func(token string) (*session.TrustedPassAccessTokenResponse, error) {
+		resp, err := clientAPI.TrustedPassToken()
+		if err != nil {
```
DoYouEvenCpp, line 57: [nit] the logic of checking if err happened isn't coherent. Sometimes you test at first against `err == nil` while in other places it's the opposite (eg. ,`err != nil`).
What about aligning it to a common approach? (my choice would be to always at first test if an error happened, and if so bail out early - but I'm fine with any approach, as long as it's unified :wink: )

config/token.go
```diff
@@ -3,16 +3,17 @@ package config
...
+	IsOAuth                bool       `json:"is_oauth,omitempty"`
+	TrustedPassToken       string     `json:"trusted_pass_token,omitempty"`
+	TrustedPassOwnerID     string     `json:"trusted_pass_owner_id,omitempty"`
+	ServiceExpiry          string     `json:"service_expiry,omitempty"`
+	NordLynxPrivateKey     string     `json:"nordlynx_private_key"`
+	OpenVPNUsername        string     `json:"openvpn_username"`
+	OpenVPNPassword        string     `json:"openvpn_password"`
+	NCData                 NCData     `json:"nc_data,omitempty"`
+	IdempotencyKey         *uuid.UUID `json:"idempotency_key,omitempty"`
+	TrustedPassTokenExpiry string     `json:"trusted_pass_token_expiry,omitempty"`
```
DoYouEvenCpp: [nit] Care to movie it a few lines above, just to group that fellow with others starting with `TrustedPass...` ?

session/validators.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("unsupported store for token validation: got type %T", session)
+	}
+
+	return v.ValidateFunc(carrier.GetToken())
+}
+
+// NewTokenValidator creates a validator that checks session tokens.
+func NewTokenValidator(fn func(token string) error) SessionStoreValidator {
+	if fn == nil {
+		return &InvalidSessionValidator{}
```
DoYouEvenCpp: What about a login a warning here, that an invalid validator is about to get created?

session/validators.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("unsupported store for credentials validation: got type %T", session)
+	}
+
+	return v.ValidateFunc(carrier.GetUsername(), carrier.GetPassword())
+}
+
+// NewCredentialValidator creates a validator that checks session credentials.
+func NewCredentialValidator(fn func(username, password string) error) SessionStoreValidator {
+	if fn == nil {
+		return &InvalidSessionValidator{}
```
DoYouEvenCpp: I am all for running input validation and producing an invalid validator here, however the `InvalidSessionValidator` isn't quite generic. At least its name is not.

So my suggestion is to rename it to something less specific. `InvalidValidator` maybe?

And I guess getting rid of an _invalid_ validator in favour of running input validation directly in the constructors, and returning nil in a NOK case doesn't seem like an trivial change to introduce, correct?

session/access_token_session_store.go
```diff
@@ -0,0 +1,157 @@
...
+		cfgManager:         cfgManager,
+		errHandlerRegistry: errorHandlingRegistry,
+		validator:          validator,
+		renewAPICall:       renewAPICall,
+		session:            newAccessTokenSession(cfgManager),
+	}
+}
+
+// Renew iterates over stored user tokens and attempts to renew each one that has expired.
+// Returns a combined error if one or more renewals fail.
```
DoYouEvenCpp: Is the error returned here actually a combined one?
Also, I guess it tries to renew only for a token that matches `cfg.AutoConnectData.ID`, correct?

session/access_token_session_store.go
```diff
@@ -0,0 +1,157 @@
...
+	key := uuid.New()
+	data.IdempotencyKey = &key
+	err := s.cfgManager.SaveWith(func(c config.Config) config.Config {
+		user := c.TokensData[uid]
+		user.IdempotencyKey = data.IdempotencyKey
+		c.TokensData[uid] = user
+		return c
+	})
+
+	if err != nil {
```
DoYouEvenCpp, line 177: Does it make sense to fallback to previous parameters (eg., the idempotency key) in case of an error?
In `tryUpdate...`  the `try` bit suggests that in case of a failed attempt to udpate the data, the key remains unchanged.
Or is it only me here? :smile: 

Anyway, what about either keeping previous data and in case of an error restoring it back or slightly different method name (or a word or two of explanation in the missing comment section)?

DoYouEvenCpp (commented):
Two general remarks:
- if that's not a big issue, please plan some time to provide missing comments (even most brief ones) to newly provided code
- it would be really nice to keep track of all the TODOs with internal tickets (ideally, snap IDs along respective todos)

cmd/daemon/init.go
thegravi: done

session/validators.go
thegravi: answered before

cmd/daemon/init.go
thegravi: done

daemon/rpc_test.go
thegravi: I tried to go along with the current alias naming in this file, which uses `test<package>`. Therefore, I moved from the previous way I called my alias.

cmd/daemon/init.go
thegravi: yes, API does not return this one

session/validators.go
```diff
@@ -0,0 +1,111 @@
...
+			"TokenValidator creation failed due to an invalid function argument.")
+		return &InvalidSessionValidator{}
+	}
+	return &TokenValidator{ValidateFunc: fn}
+}
+
+type ExpiryValidator struct{}
+
+// Validate checks if the provided session has expired.
+func (v *ExpiryValidator) Validate(session interface{}) error {
```
devzbysiu: Why change from `any` to `interface{}`? 🤔
(edit: sorry - forgot we have a meeting today)

thegravi: Even though it's an alias, I thought it would be more understandable not to provide built-in types, but now I don't think that was a good idea.

session/trusted_pass_session_store.go
```diff
@@ -0,0 +1,203 @@
...
+// TrustedPassRenewalAPICall renews TrustedPass tokens
+type TrustedPassRenewalAPICall func(token string) (*TrustedPassAccessTokenResponse, error)
+
+// TrustedPassSessionStore manages TrustedPass-based sessions
+type TrustedPassSessionStore struct {
+	cfgManager         config.Manager
+	errHandlerRegistry *internal.ErrorHandlingRegistry[error]
+	renewAPICall       TrustedPassRenewalAPICall
+
+	// optional external validator with
```
mariusSincovici: [small] is `with` extra? :thinking:

session/access_token_session_store.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return nil
+}
+
+// HandleError processes errors that occur during session operations.
+// It returns nil if the error was not handled, or the error itself if it was.
+func (s *AccessTokenSessionStore) HandleError(reason error) error {
+	handlers := s.errHandlerRegistry.GetHandlers(reason)
+	if len(handlers) == 0 {
+		return nil
```
mariusSincovici, line 114: Wouldn't be good to log that nobody is registered?

session/access_token_session_store.go
```diff
@@ -0,0 +1,213 @@
...
+	err = s.cfgManager.SaveWith(func(c config.Config) config.Config {
+		data := c.TokensData[uid]
+		data.Token = resp.Token
+		data.RenewToken = resp.RenewToken
+		data.TokenExpiry = expTime.Format(internal.ServerDateFormat)
+		c.TokensData[uid] = data
+		return c
+	})
+
+	if err != nil {
```
mariusSincovici, line 156: [small] wouldn't work `return err` or add context to the error?

test/mock/config_manager.go
```diff
@@ -20,33 +23,94 @@ func NewMockConfigManager() *ConfigManager {
 	return &m
 }
 
+// deepCopyConfig creates a complete deep copy of the config to prevent race conditions
+// when accessing maps and pointers.
+func deepCopyConfig(src config.Config) config.Config {
```
mariusSincovici: Wouldn't it be easier to marshal and unmarshal the struct? Then future members will be supported automatically?

mariusSincovici (approved):

DoYouEvenCpp (approved):

test/mock/config_manager.go
devzbysiu: Yeah, marshaling and unmarshaling is probably the best way, it should cover adding new members. 

I would also suggest to move this to be a method of `config.Config` so it's closer to the source and harder to forget in case new fields are added (but the marshaling and unmarshaling should cover that well enough) or if this will be needed in other places then it's easier to discover having it as a method.

Also having tests for this would be nice so we can detect regression in case new member is added but somehow breaks this

thegravi: @devzbysiu, what tests are you talking about? This code is used as a mock.

devzbysiu: Aaah, yeah. WTF :joy: , my bad :smile:

session/trusted_pass_session_store.go
thegravi: unfinished thought, gonna remove it

daemon/session_error_handler.go
```diff
@@ -0,0 +1,100 @@
...
+		state.logoutInProgress = true
+		state.mu.Unlock()
+
+		defer func() {
+			state.mu.Lock()
+			state.logoutInProgress = false
+			state.mu.Unlock()
+		}()
+
+		log.Printf("%s Session error detected: %v. Forcing logout.", internal.DebugPrefix, reason)
```
devzbysiu: [small] add `\n` to log msg here and other `Printf` places without it

session/validation_helpers_test.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+func TestValidateExpiry_TimeBoundary(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	// Test with time very close to now
+	almostNow := time.Now().Add(1 * time.Millisecond)
+	err := ValidateExpiry(almostNow)
+	// This might pass or fail depending on timing, but should not panic
+	if err != nil {
```
devzbysiu: I can break the implementation of `ValidateExpiry` to return always `nil` and this test will pass. Similarly I can break it to always return `ErrSessionExpired` and it will pass.
This test will fail only when error was returned, but it's different than `ErrSessionExpired`. Not sure if we have meaningful value from it.

If you want to test this deterministically, then the time should be abstracted away and you'll have full control over the flow of time. There is also `testing/synctest` which has a `Clock` that allows controlling time with `.Advance` method, but I think it's not fully available for our go version :thinking:

session/validation_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+// ValidateExpiry checks if the provided expiry time has not passed
+func ValidateExpiry(expiry time.Time) error {
+	if time.Now().After(expiry) {
+		return ErrSessionExpired
+	}
+	return nil
+}
+
+// ValidateToken checks if the token is not empty
+func ValidateToken(token string) error {
```
devzbysiu: Does it need to be public?

session/validation_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	if token == "" {
+		return ErrInvalidToken
+	}
+	return nil
+}
+
+// ValidateAccessTokenFormat checks if the access token has valid format (hexadecimal characters only)
+func ValidateAccessTokenFormat(token string) error {
+	if err := ValidateToken(token); err != nil {
```
devzbysiu: Maybe we don't need this `ValidateToken`? It checks if the token is empty and if it is you return `ErrInvalidToken`, then regex validation is done and if it fails it also returns `ErrInvalidToken`, but the regex allows for 0 or more repetitions, so maybe we can make the regex as `^[a-f0-9]+$` and then you don't need `ValidateToken`?

internal/constants.go
```diff
@@ -144,7 +144,9 @@ var (
 )
 
 var (
-	AccessTokenFormatValidator = regexp.MustCompile(`^[a-f0-9]*$`).MatchString
+	AccessTokenFormatValidator      = regexp.MustCompile(`^[a-f0-9]*$`).MatchString
```
devzbysiu: Is it OK for it to allow empty strings? It's used in `rpc_login.go` directly

session/validation_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if !internal.AccessTokenFormatValidator(token) {
+		return ErrInvalidToken
+	}
+	return nil
+}
+
+// ValidateTrustedPassTokenFormat checks if the TrustedPass token is not empty and has valid format
+func ValidateTrustedPassTokenFormat(token string) error {
+	if err := ValidateToken(token); err != nil {
```
devzbysiu: This is probably not needed, because `internal.TrustedPassTokenFormatValidator` does not allow for empty strings

internal/constants.go
```diff
@@ -144,7 +144,9 @@ var (
...
+	RenewalTokenFormatValidator     = regexp.MustCompile(`^[a-f0-9]+$`).MatchString
+	TrustedPassTokenFormatValidator = regexp.MustCompile(`^[A-Za-z0-9_-]+$`).MatchString
```
devzbysiu: [small] shouldn't those validators be named like `ValidateXYZ` since those are methods and not objects? :thinking:

session/validation_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+	if !internal.TrustedPassTokenFormatValidator(token) {
+		return ErrInvalidToken
+	}
+	return nil
+}
+
+// ValidateOwnerID checks if the owner ID is not empty
+func ValidateOwnerID(ownerID string) error {
```
devzbysiu: Maybe this could be inlined similar to how `ValidateRenewToken` has empty check inlined? 

Or if it's needed like that then maybe it should be named something like `CheckNotEmpty` or something similar. `ValidateOwnerID` seems to vague for what it's doing.
And then it can be private maybe?

session/access_token_session_store.go
```diff
@@ -0,0 +1,214 @@
...
+
+	if err := ValidateExpiry(cfg.ExpiresAt); err != nil {
+		return fmt.Errorf("validating access token: %w", err)
+	}
+
+	// Run external validation if available
+	if s.externalValidator != nil {
+		if err := s.externalValidator(cfg.Token); err != nil {
+			return err
+		}
```
devzbysiu: [small] can it be just `return s.externalValidator(cfg.Token)`?

devzbysiu (approved):

nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=974) <br/>All committers have signed the CLA.

PR #975, opened 2025-07-23
Add GetFeatureToggles gRPC

dfetti:
We need that to know which features are enabled and available for the cli/gui.


cli/cli_features_toggles.go
```diff
@@ -0,0 +1,22 @@
...
+type FeatureToggles struct {
+	meshnetEnabled bool
+}
+
+func (c *cmd) GetFeatureToggles() FeatureToggles {
+	featureToggles, err := c.client.GetFeatureToggles(context.Background(), &pb.Empty{})
+
+	if err != nil {
+		// TODO: Log an error here. We need ui/ux for that I think
+		return FeatureToggles{meshnetEnabled: false}
```
mariusSincovici: Maybe make a function in which the "default" values are defined. That function can also be used into the daemon when there are no RC available.
Also you could log the error. I don't think special message is needed since is not something the user can see and the app should work with some default values.

mariusSincovici (approved):

daemon/rpc_feature_toggles.go
```diff
@@ -0,0 +1,13 @@
+package daemon
+
+import (
+	"context"
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+)
+
+// Returns the list of feature toggles fetched from the remote config
+// TODO (dfe): Use the actual info from the remote config
```
devzbysiu, line 10: Do we have a task for this or is it being worked on in a separate PR? Just to not forget it

protobuf/daemon/features.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package pb;
+
+option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
+
+message FeatureToggles {
+  bool meshnetEnabled = 1;
```
devzbysiu: Shouldn't it be in `snake_case`?

devzbysiu (approved):

protobuf/daemon/features.proto
dfetti: You're right, it should be snake case.

daemon/rpc_feature_toggles.go
dfetti: There is a task for that. It's waiting for the merge of remote config PR

PR #976, opened 2025-07-24
LVPN-8650: track client usage

bartoszWojciechO:
Added gRPC interceptors to track which client is used to call desired gRPC methods based on the metadata and method name.


daemon/clientid/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+
+	clientString = clientIDToClientString(pb.ClientID(clientID))
+	if clientString == "" {
+		return
+	}
+
+	itemName := fullMethodNameToItemName(fullMethod)
+	if itemName == "" {
+		return
+	}
```
devzbysiu: Maybe this should be first precondition, so if we have a call which does not support this reporting then we exit early and don't need to play with getting client id from the metadata?

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+
+	clientID, err := strconv.Atoi(metadata[0])
+	if err != nil {
+		return
+	}
+
+	clientString = clientIDToClientString(pb.ClientID(clientID))
+	if clientString == "" {
+		return
+	}
```
devzbysiu: Maybe this chunk could be moved to separate `func` that returns also error? So it would be like:
- if we pass the precondition about supported method name, from comment on lines below
- then we have to have valid client id
- so we return error if it's invalid

And then in this func here, we can log this error? Now if something is wrong we would fail silently

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+func (c *CliendIDMiddleware) UnaryMiddleware(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo) (interface{}, error) {
+	c.notifyAboutClickEvent(ctx, info.FullMethod)
+
+	return nil, nil
+}
+
+func (c *CliendIDMiddleware) StreamMiddleware(srv interface{},
```
devzbysiu, line 122: Do we need stream middleware also? :thinking: If I'm not mistaken, we are streaming only for the status updates which is not supported by this reporting anyway :thinking: , or is there anything else streaming?

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+
+func clientIDToClientString(clientID pb.ClientID) string {
+	switch clientID {
+	case pb.ClientID_CLI:
+		return cliClientString
+	case pb.ClientID_GUI:
+		return guiClientString
+	case pb.ClientID_TRAY:
+		return trayClientString
+	default:
```
mariusSincovici: [suggestion] not to add default, if possible, so it gets an error when there a new values added?

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+	if md, ok = metadata.FromIncomingContext(ctx); !ok {
+		return
+	}
+
+	var metadata []string
+	if metadata = md.Get(clientIDMetadataKey); len(metadata) == 0 {
+		return
+	}
+
+	clientID, err := strconv.Atoi(metadata[0])
```
mariusSincovici: [minor] use the parseInt with max size 32bit?

daemon/clientid/middleware.go
mariusSincovici: Connect command is also a stream.

protobuf/daemon/telemetry/v1/service.proto
```diff
@@ -10,9 +10,6 @@ import "google/protobuf/empty.proto";
 import "fields.proto";
 
 service TelemetryService {
-  // Reports a generic entry for flexible extension
-  rpc SetCustomEntry(CustomEntry) returns (google.protobuf.Empty);
```
mariusSincovici: is it intentional the removal?

mariusSincovici (approved):

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,128 @@
...
+	case pb.Daemon_Disconnect_FullMethodName:
+		return disconnectItemName
+	case pb.Daemon_LoginOAuth2_FullMethodName:
+		return loginItemName
+	case pb.Daemon_Logout_FullMethodName:
+		return logoutItemName
+	default:
+		return ""
+	}
+}
```
thegravi, line 66: [small] If there were more items, I suggest using hashing, like a map for example

protobuf/daemon/telemetry/v1/service.proto
thegravi: It was a leftover from my implementation because we decided not to have custom logic. So it's alright to remove.

daemon/clientid/middleware.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "parsing client ID:", err)
+		return
+	}
+
+	c.publisher.Publish(events.UiItemsAction{
+		ItemName:      itemName,
+		ItemType:      "button",
+		FormReference: clientString,
+	})
```
thegravi, line 110: [small] It should not be forgotten that there might be other triggers, not only the `button` when the GUI gets more attention to this (soon). And that would need to be changed from the source side, because here we would not know the correct trigger type.

thegravi (approved):

daemon/clientid/middleware.go
bartoszWojciechO: For now I'd keep it as a switch, because it's easier to read. Let's switch to hasihing when we start tracking more RPCs.

PR #977, opened 2025-07-24
LVPN-6747: VPN credentials integration into session store

thegravi:


session/validation_helpers.go
```diff
@@ -48,6 +48,22 @@ func ValidateRenewToken(renewToken string) error {
 	return nil
 }
 
-// Type-safe external validators for each session type
+// ValidateOpenVPNCredentials checks if OpenVPN credentials are valid
+func ValidateOpenVPNCredentials(username, password string) error {
```
devzbysiu: IMO `Validate` is too vague here. Also doc comment is imprecise. You can still have non empty credentials which are wrong, but you'll call `Validate...` and you won't get an error which is a bit misleading. How about `RequireOpenVPNCredentials` or `CheckOpenVPNCredentialsPresent` or something like that?

Similarly below in `ValidateNordLynxPrivateKey`

session/vpn_creds_sessions_store.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+// Renew VPN credentials if they have expired.
+func (s *VPNCredentialsSessionStore) Renew() error {
+	if err := s.validate(); err == nil {
+		return nil
+	}
+
+	if s.renewAPICall == nil {
```
devzbysiu, line 47: Ideally, the error should be returned as close as possible to the source of an issue because it's easier to debug. Maybe this could be done at the time of construction?

session/vpn_creds_sessions_store.go
```diff
@@ -0,0 +1,142 @@
...
+		return errors.New("renewal API call not configured")
+	}
+
+	resp, err := s.renewAPICall()
+	if err != nil {
+		return s.HandleError(err)
```
devzbysiu, line 53: Hmm, so this will silently return `nil` if there is an error, but there are no handlers for it in the `errHandlerRegistry`? Or is this somehow validated that registry cannot be empty?

thegravi: This will return nil only for those errors that we are interested in -> registered via error registry, otherwise use will be logged out on any error

cmd/daemon/init.go
```diff
@@ -64,3 +65,47 @@ func buildTrustedPassSessionStoreAPIRenewalCall(clientAPI core.ClientAPI) sessio
...
+func buildVPNCredentialsSessionStore(
+	confman config.Manager,
+	errRegistry *internal.ErrorHandlingRegistry[error],
+	clientAPI core.ClientAPI,
+) session.SessionStore {
+	return session.NewVPNCredentialsSessionStore(
+		confman,
+		errRegistry,
+		buildVPNCredentialsSessionStoreAPIRenewalCall(confman, clientAPI),
+		nil, // No external validator for now
```
devzbysiu, line 78: Is external validator used at all in any of the session stores? :smile:

cmd/daemon/init.go
```diff
@@ -64,3 +65,47 @@ func buildTrustedPassSessionStoreAPIRenewalCall(clientAPI core.ClientAPI) sessio
...
+			return nil, err
+		}
+
+		data, ok := cfg.TokensData[cfg.AutoConnectData.ID]
+		if !ok {
+			return nil, errors.New("there is not data")
+		}
+
+		// actual API call passed into Session Store object
+		return func(token string) (*session.VPNCredentialsResponse, error) {
```
devzbysiu: Can this be just a separate function or there is a benefit of this inline call?

session/validation_helpers.go
```diff
@@ -48,6 +48,23 @@ func ValidateRenewToken(renewToken string) error {
 	return nil
 }
 
+// ValidateOpenVPNCredentialsPresence checks if OpenVPN credentials are valid
```
devzbysiu: Probably this could be updates as well to tell it's checking if creds are empty. Same for doc comment of `ValidateNordLynxPrivateKeyPresence`

cmd/daemon/init.go
```diff
@@ -64,3 +65,47 @@ func buildTrustedPassSessionStoreAPIRenewalCall(clientAPI core.ClientAPI) sessio
 		}, nil
 	}
 }
+
+func buildVPNCredentialsSessionStore(
```
devzbysiu, line 69: Maybe this could be inlined? It just delegates to the constructor

devzbysiu (approved):

cmd/daemon/init.go
thegravi: Nope, left it for flexibility if we would need to validate something additionally in the future.

bartoszWojciechO (approved):

session/vpn_creds_sessions_store.go
```diff
@@ -0,0 +1,142 @@
...
+
+type VPNCredentialsResponse struct {
+	Username           string
+	Password           string
+	NordLynxPrivateKey string
+}
+
+type VPNCredentialsRenewalAPICall func() (*VPNCredentialsResponse, error)
+
+type vpnCredentialsConfig struct {
```
mariusSincovici: Why is this needed? Isn't the same as `VPNCredentialsResponse`?

mariusSincovici: Why is this needed? Isn't the same as `VPNCredentialsResponse`?

cmd/daemon/init.go
thegravi: One time it's hard to read, so need to extract to separate function, next time "can we inline it" :D What are the criteria for these requests?

session/vpn_creds_sessions_store.go
thegravi: It's the same, but the naming for struct tells different usage

cmd/daemon/init.go
devzbysiu: I think hard to read comment was about inline anonymous function which you extracted to `buildVPNCredentialsSessionStoreAPIRenewalCall` below

This comment is about having a function which just delegates to constructor and does nothing else

thegravi: There were comments for the access token session store to extract these lambdas to separate functions for readability, that is why `build...` functions exist.

PR #980, opened 2025-07-24
Use the meshnet feature toggle from the remote config

dfetti:


daemon/rpc_feature_toggles.go
```diff
@@ -3,11 +3,12 @@ package daemon
...
 
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 )
 
 // Returns the list of feature toggles fetched from the remote config
-// TODO (dfe): Use the actual info from the remote config
 func (r *RPC) GetFeatureToggles(ctx context.Context, in *pb.Empty) (*pb.FeatureToggles, error) {
-	return &pb.FeatureToggles{MeshnetEnabled: true}, nil
+	meshnetEnabled := r.remoteConfigGetter.IsFeatureEnabled(remote.FeatureMeshnet)
```
keliramu: no need for intermediate variable

keliramu (approved):

devzbysiu (approved):

daemon/rpc_feature_toggles.go
dfetti: The formatter will break it in more than 2 lines otherwise.. And it will also look cleaner when we will have more feature toggles.

DoYouEvenCpp (approved):

thegravi (approved):

PR #981, opened 2025-07-25
Fix API usage in tests

dfetti:


test/qa/lib/server.py
```diff
@@ -52,11 +52,23 @@ def get_hostname_by(technology="", protocol="", obfuscated="", group_id=""):
     }
 
     if group_id != "":
-        group_id = group_ids[group_id]
+        try:
+            group_id = int(group_id)
+        except ValueError:
+            group_id = group_ids[group_id]
+    elif tech_id == 15 or tech_id == 17:    # Obfuscated
```
bartoszWojciechO: Maybe it would be good to use common constants here and in the map in line 41?

test/qa/lib/server.py
```diff
@@ -94,5 +106,5 @@ def get_dedicated_ip() -> None | str:
 
     headers = {'Accept': 'application/json'}
     response = requests.get(f'https://api.nordvpn.com/v1/servers?&filters[servers.id]={dip_server_id}', headers=headers, timeout=5)
-
-    return str(response.json()[0]['hostname'].split(".")[0])
+    server_info = response.json()[0]
```
DoYouEvenCpp, line 118: nit: maybe we could add here a check if response actually contains what we expect? In case it doesn't we can produce here a meaningful error message (which eventually, can ease a bit debugging)

test/qa/lib/server.py
dfetti: Yes, I will refine the code after I will make everything work. If you have any other suggestions, drop them here.

test/qa/lib/server.py
```diff
@@ -53,12 +53,24 @@ def get_hostname_by(technology="", protocol="", obfuscated="", group_id=""):
...
     # api limits
     time.sleep(2)
-    url = f"https://api.nordvpn.com/v1/servers?limit=10&filters[servers.status]=online&filters[servers_technologies]={tech_id}&filters[servers_groups]={group_id}"
+    # recommendations endpoint does not return DIP servers
+    if group_id == 9:
+        endpoint = "v1/servers"
+    else:
+        endpoint = "v1/servers/recommendations"
+
+    url = f"https://api.nordvpn.com/{endpoint}?limit=10&filters[servers.status]=online&filters[servers_technologies][id]={tech_id}&filters[servers_groups][id]={group_id}"
```
DoYouEvenCpp: nit: what about:
`url = f"https://api.nordvpn.com/{'v1/servers' if group_id == 9 else 'v1/servers/recommendations'}?limit=10&filters[servers.status]=online&filters[servers_technologies][id]={tech_id}&filters[servers_groups][id]={group_id}"`

Then we can get rid of a few lines of the code just above

dfetti: Don't really like this approach and anyway I slightly changed the code a bit.

test/qa/lib/server.py
```diff
@@ -14,52 +65,23 @@ def __init__(self, server_info):
...
-        "Obfuscated_Servers": "17",
-        "Europe": "19",
-        "The_Americas": "21",
-        "Asia_Pacific": "23",
-        "Africa_The_Middle_East_And_India": "25",
-    }
-
-    if group_id != "":
-        group_id = group_ids[group_id]
+    (endpoint, tech_id, group_id) = get_request_parameters(technology, protocol, obfuscated, group_name)
```
thegravi: Why don't we use try/catch for a missing keyword inside the function?

Or something like this inside to avoid an exception, but receive `None` in those cases:
```python
tech_id = TECH_IDS.get(technology, {}).get(protocol, {}).get(f"{obfuscated}", None)
```

dfetti: Good idea. I added that.

thegravi (approved):

PR #983, opened 2025-07-29
LVPN-6753: run openvpn with --auth-nocache

bartoszWojciechO:


DoYouEvenCpp (approved):
nit: care to create an e2e test that checks if there's no warning about caching password while using openVPN?

PR #986, opened 2025-07-30
Fix settings output parsing in tests

thegravi:


bartoszWojciechO (approved):

test/qa/lib/settings.py
```diff
@@ -11,18 +11,26 @@
 
 class Settings:
     def __init__(self):
-        output = sh.nordvpn("settings").strip(" \r-\n")
+        output = sh.nordvpn("settings").replace("\r", "")
```
mariusSincovici: This needs to be replaced with `sh_no_tty`

thegravi: Does it change the behavior for /r/n?

mariusSincovici: yes, the application will not sent the loading indicator in that case

thegravi: Now it's irrelevant, closing

PR #998, opened 2025-08-05
LVPN-8576: Integrate remote config analytics

DoYouEvenCpp:
Changes under this PR:
- renamed `config/remote/analytics.go` to `config/remote/analytics_builder.go` (a respective rename was also done for tests eg.:  `analytics_test.go`)
- added `config/remote/analytics_.go` which wraps analytics events functionality
- extended `DownloadErrorKind` enum to provide greater granularity for possible erroneous cases while downloading remote config files
- renamed `MooseDebuggerEvent` to `DebuggerEvent`
- extracted RC related errors into a separate `errors.go` package
- removed API, AppVersion and Country from `UserInfo` 
  - also removed the structure type completely, now only user's rollout group is  stored in RC's `Analytics`
- tests

### DIFF
--- a/test/qa/test_autoconnect.py
+++ b/test/qa/test_autoconnect.py
@@ -32,12 +32,16 @@ def autoconnect_base_test(group):
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_default(tech, proto, obfuscated):
+    """Manual TC: LVPN-6779"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test("")
 
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_not_autoconnect(tech, proto, obfuscated):
+    """Manual TC: LVPN-6780"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     output = sh.nordvpn.set.autoconnect.off()
@@ -50,19 +54,25 @@ def test_not_autoconnect(tech, proto, obfuscated):
 @pytest.mark.parametrize("group", lib.COUNTRIES + lib.COUNTRY_CODES)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_to_country(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-6781"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
 
 @pytest.mark.parametrize("group", lib.CITIES)
-@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_to_city(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-6784"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_to_random_server_by_name(tech, proto, obfuscated):
+    """Manual TC: LVPN-6782"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     server_info = server.get_hostname_by(tech, proto, obfuscated)
@@ -74,6 +84,8 @@ def test_autoconnect_to_random_server_by_name(tech, proto, obfuscated):
 @pytest.mark.parametrize("group", lib.STANDARD_GROUPS)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_to_standard_group(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-8424"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
@@ -81,6 +93,8 @@ def test_autoconnect_to_standard_group(tech, proto, obfuscated, group):
 @pytest.mark.parametrize("group", lib.ADDITIONAL_GROUPS)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER)
 def test_autoconnect_to_additional_group(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-6786"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
@@ -88,6 +102,8 @@ def test_autoconnect_to_additional_group(tech, proto, obfuscated, group):
 @pytest.mark.parametrize("group", lib.ADDITIONAL_GROUPS_NORDWHISPER)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.NORDWHISPER_TECHNOLOGY)
 def test_nordwhisper_autoconnect_to_additional_group(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-6786"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
@@ -95,6 +111,8 @@ def test_nordwhisper_autoconnect_to_additional_group(tech, proto, obfuscated, gr
 @pytest.mark.parametrize("group", lib.DEDICATED_IP_GROUPS)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER)
 def test_autoconnect_to_ovpn_group(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-563"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
@@ -102,12 +120,16 @@ def test_autoconnect_to_ovpn_group(tech, proto, obfuscated, group):
 @pytest.mark.parametrize("group", lib.OVPN_OBFUSCATED_GROUPS)
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.OBFUSCATED_TECHNOLOGIES)
 def test_autoconnect_to_obfuscated_group(tech, proto, obfuscated, group):
+    """Manual TC: LVPN-410"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
 
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
 def test_autoconnect_virtual_country(tech, proto, obfuscated):
+    """Manual TC: LVPN-8549"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     sh.nordvpn.set("virtual-location", "on")
 
@@ -120,6 +142,8 @@ def test_autoconnect_virtual_country(tech, proto, obfuscated):
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
 def test_autoconnect_virtual_country_disabled(tech, proto, obfuscated):
+    """Manual TC: LVPN-8548"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     # fix in LVPN-8449
@@ -139,6 +163,8 @@ def test_autoconnect_virtual_country_disabled(tech, proto, obfuscated):
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
 def test_autoconnect_to_unavailable_groups(tech, proto, obfuscated):
+    """Manual TC: LVPN-8431"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     unavailable_groups = daemon.get_unavailable_groups()
@@ -153,6 +179,8 @@ def test_autoconnect_to_unavailable_groups(tech, proto, obfuscated):
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.OBFUSCATED_TECHNOLOGIES)
 def test_prevent_autoconnect_enable_to_non_obfuscated_servers_when_obfuscation_is_on(tech, proto, obfuscated):
+    """Manual TC: LVPN-8581"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     unavailable_groups = daemon.get_unavailable_groups()
@@ -172,6 +200,8 @@ def test_prevent_autoconnect_enable_to_non_obfuscated_servers_when_obfuscation_i
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.OBFUSCATED_TECHNOLOGIES)
 def test_prevent_obfuscate_disable_with_autoconnect_enabled_to_obfuscated_server(tech, proto, obfuscated):
+    """Manual TC: LVPN-5847"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     available_groups = str(sh.nordvpn.groups(_tty_out=False)).strip().split()
@@ -191,6 +221,8 @@ def test_prevent_obfuscate_disable_with_autoconnect_enabled_to_obfuscated_server
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
 def test_prevent_autoconnect_enable_to_obfuscated_servers_when_obfuscation_is_off(tech, proto, obfuscated):
+    """Manual TC: LVPN-8591"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     with pytest.raises(sh.ErrorReturnCode_1) as ex:
@@ -207,6 +239,8 @@ def test_prevent_autoconnect_enable_to_obfuscated_servers_when_obfuscation_is_of
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.OVPN_STANDARD_TECHNOLOGIES)
 def test_prevent_obfuscate_enable_with_autoconnect_set_to_nonobfuscated(tech, proto, obfuscated):
+    """Manual TC: LVPN-5848"""
+
     lib.set_technology_and_protocol(tech, proto, obfuscated)
 
     available_groups = str(sh.nordvpn.groups(_tty_out=False)).strip().split()


config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
-			Message: errorMessage,
+// Analytics defines an interface for reporting various analytics events related to downloads,
+// local usage, JSON parsing, and partial rollouts. Implementations of this interface are
+// responsible for handling the notification logic for each event type.
+type Analytics interface {
+	NotifyDownload(string, string)
+	NotifyDownloadFailure(string, string, DownloadError)
+	NotifyLocalUse(string, string, error)
+	NotifyJsonParse(string, string, error)
+	NotifyPartialRollout(string, string, int, bool)
```
thegravi: [small] Why not add variable names to have the whole picture in one place without searching for a specific implementation?

config/remote/config.go
```diff
@@ -97,41 +107,42 @@ func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBaseP
...
 	// validate json against predefined schema
 	if err = jv.validate(mainJsonStr); err != nil {
-		return false, fmt.Errorf("validating main: %w", err)
+		return false, &DownloadError{Kind: DownloadErrorParsing, Cause: fmt.Errorf("validating main: %w", err)}
 	}
 
 	incFiles, err := walkIncludeFiles(mainJsonStr, cdnBasePath, targetPath, cdn, fw)
 	if err != nil {
-		return false, fmt.Errorf("downloading include files: %w", err)
+		//TODO: fixme!
```
devzbysiu: Leftover?

config/remote/config.go
```diff
@@ -22,6 +22,16 @@ type jsonHash struct {
 	Hash string `json:"hash"`
 }
 
+// DownloadError holds information about a specific download faillure that is then used while reporting analytics about RemoteConfig.
```
devzbysiu: [typo] `faillure`

config/remote/remote_test.go
```diff
@@ -426,19 +427,24 @@ func (e *RemoteConfigEventHandler) RemoteConfigUpdate(c RemoteConfigEvent) error
 }
 
 func newTestRemoteConfig(ver, env string, cdn core.RemoteStorage) *CdnRemoteConfig {
+	mle := subs.Subject[events.DebuggerEvent]{}
```
devzbysiu: `mle`? :thinking:

config/remote/remote.go
```diff
@@ -182,12 +192,15 @@ func (c *CdnRemoteConfig) load() {
...
+	for _, f := range c.features.keys() {
+		feature := c.features.get(f)
+		if err := feature.load(c.localCachePath, jsonFileReaderWriter{}, jsonValidator{}); err != nil {
+			log.Println(internal.ErrorPrefix, "failed loading feature [", feature.name, "] config from the disk:", err)
+			c.analytics.NotifyLocalUse("cli", feature.name, err)
 			continue
 		}
-		log.Println(internal.InfoPrefix, "feature [", f.name, "] config loaded from:", c.localCachePath)
+		log.Println(internal.InfoPrefix, "feature [", feature.name, "] config loaded from:", c.localCachePath)
+		c.analytics.NotifyLocalUse("cli", feature.name, nil)
```
devzbysiu: Maybe this `"cli"` should be a `const`? Or maybe even we could have an enum for clients

config/remote/data.go
```diff
@@ -83,10 +83,37 @@ func (f Feature) HashFilePath(basePath string) string {
...
+}
 
-func (m *FeatureMap) Add(name string) {
-	(*m)[name] = &Feature{
+func (m *FeatureMap) features() map[string]*Feature {
+	return m.featureMap
+}
+
+func (m *FeatureMap) add(name string) {
+	m.featureKeys = append(m.featureKeys, name)
```
devzbysiu: Won't this duplicate keys in `featureKeys` if the key was already added?

config/remote/data.go
```diff
@@ -83,10 +83,37 @@ func (f Feature) HashFilePath(basePath string) string {
...
+	}
+}
+
+func (m *FeatureMap) keys() []string {
+	return m.featureKeys
+}
 
-func (m *FeatureMap) Add(name string) {
-	(*m)[name] = &Feature{
+func (m *FeatureMap) features() map[string]*Feature {
```
devzbysiu: Is this only for testing?

config/remote/constants.go
```diff
@@ -55,51 +55,59 @@ func (e EventType) String() string {
...
+	case DownloadErrorParsing:
+		return "parsing_error"
+	case DownloadErrorRemoteFileNotFound:
+		return "remote_file_not_found"
+	case DownloadErrorRemoteHashNotFound:
+		return "remote_hash_not_found"
+	case DownloadErrorWriteHash:
+		return "write_hash_error"
+	case DownloadErrorWriteJson:
+		return "write_json_error"
```
devzbysiu: Wouldn't be easier to put those in a map and then get the string from the map based on the kind?

config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
+}
+type MooseAnalytics struct {
```
devzbysiu: Shouldn't this be maybe something like `RCAnalytics`? Constructor is named with `..RemoteConfigAnalytics` but returns `MooseAnalytics` and this name is big vague I think, because we already have moose analytics :smile:

config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
-func NewDownloadFailureEvent(info UserInfo, client string, featureName FeatureName, errorKind DownloadErrorKind, errorMessage string) Event {
-	return Event{
-		UserInfo:    info,
-		Client:      client,
-		FeatureName: featureName.String(),
-		Event:       DownloadFailure.String(),
-		EventDetails: EventDetails{
-			Error:   errorKind.String(),
-			Message: errorMessage,
+// Analytics defines an interface for reporting various analytics events related to downloads,
```
devzbysiu: Maybe add also something that it's about remote config analytics

config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
+type Analytics interface {
+	NotifyDownload(string, string)
+	NotifyDownloadFailure(string, string, DownloadError)
+	NotifyLocalUse(string, string, error)
+	NotifyJsonParse(string, string, error)
+	NotifyPartialRollout(string, string, int, bool)
+}
+type MooseAnalytics struct {
+	publisher events.PublishSubcriber[events.DebuggerEvent]
+	ctx       Event
```
devzbysiu: This is kind of weird :thinking:. We are publishing events using this `MooseAnalytics` but it already holds an `Event` which is called `ctx` :thinking:

config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
+// Analytics defines an interface for reporting various analytics events related to downloads,
+// local usage, JSON parsing, and partial rollouts. Implementations of this interface are
+// responsible for handling the notification logic for each event type.
+type Analytics interface {
+	NotifyDownload(string, string)
+	NotifyDownloadFailure(string, string, DownloadError)
+	NotifyLocalUse(string, string, error)
+	NotifyJsonParse(string, string, error)
+	NotifyPartialRollout(string, string, int, bool)
+}
```
devzbysiu, line 15: The naming here is slightly confusing, because those `NotifyXYZ` methods we have so far in `main` are reacting to events send via `.Publish` so like:
- [rpc_connect.go](https://github.com/NordSecurity/nordvpn-linux/blob/lvpn-8576-integrate-analytics/daemon/rpc_connect.go#L218-L218) publishes
- and [NotifyConnect](https://github.com/NordSecurity/nordvpn-linux/blob/lvpn-8576-integrate-analytics/events/moose/moose.go#L455-L455) receives it and "notifies" moose

In case of this interface the `NotifyXYZ` methods are producers, not consumers, so maybe different name would be enough?

config/remote/analytics.go
```diff
@@ -1,168 +1,61 @@
...
-		Event:       DownloadFailure.String(),
-		EventDetails: EventDetails{
-			Error:   errorKind.String(),
-			Message: errorMessage,
+// Analytics defines an interface for reporting various analytics events related to downloads,
+// local usage, JSON parsing, and partial rollouts. Implementations of this interface are
+// responsible for handling the notification logic for each event type.
+type Analytics interface {
+	NotifyDownload(string, string)
+	NotifyDownloadFailure(string, string, DownloadError)
```
devzbysiu: Also, maybe we could accept whole structs here instead of separate args? With structs it would be harder to put args in wrong order if we have multiple args with the same type

The implementations also use factory methods to produce event and immediately transform it to `DebuggerEvent` like:

```
func (ma *MooseAnalytics) NotifyDownload(client, featureName string) {
	ma.publisher.Publish(*NewDownloadSuccessEvent(ma.ctx.UserInfo, client, featureName).ToDebuggerEvent())
}
```
So maybe we could get rid of one layer this way.

The downside is that you would have to construct the event at the moment of calling the methods, but that's what we have so far in `main` - we construct the event and call `.Publish(event)`.

Not a big deal, just something to take into account

config/remote/constants.go
DoYouEvenCpp: Main reason why I chose switch, is a potential extension to the DownloadErrorKind. With a leverage of static-code analysis tool we would be able to easily spot a not handled case. Which might not be the case in map-bases approach. Not a big advantage, but still a small step toward code hygiene. :wink: 

If however you still would like to switch into map-based approach, then I'm perfectly happy to change that.

config/remote/config.go
DoYouEvenCpp: Fixed

config/remote/remote_test.go
DoYouEvenCpp: Fixed

config/remote/data.go
DoYouEvenCpp: Good catch! Fixed.

config/remote/data.go
DoYouEvenCpp: Indeed, seems unused. Removed this method then.

config/remote/analytics.go
DoYouEvenCpp: Good point.Renamed to `RemoteConfigAnalytics`

config/remote/analytics.go
DoYouEvenCpp: Updated the comment

config/remote/analytics.go
DoYouEvenCpp: Fixed. Switched to `userInfo` (same goes about the type used)

config/remote/analytics.go
DoYouEvenCpp: Switched to `Emit...` regarding interface methods.

config/remote/analytics.go
DoYouEvenCpp: I guess this is a really good idea - will address it however under a separate PR, once all the RC-related work ins in place.

config/remote/config.go
```diff
@@ -73,21 +83,21 @@ func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBaseP
 	}()
 
 	if f.name == "" {
-		return false, fmt.Errorf("feature name is not set")
+		return false, &DownloadError{Kind: DownloadErrorOther, Cause: fmt.Errorf("feature name is not set")}
```
mariusSincovici: [small] Maybe to have a helper function `NewDownloadError` then if new members are added to the struct you can get easier all the places that need to be changed.

config/remote/config.go
```diff
@@ -22,6 +22,16 @@ type jsonHash struct {
 	Hash string `json:"hash"`
 }
 
+// DownloadError holds information about a specific download failure that is then used while reporting analytics about RemoteConfig.
+type DownloadError struct {
```
mariusSincovici: Maybe implement support for unwrap error?

config/remote/data.go
```diff
@@ -83,10 +83,35 @@ func (f Feature) HashFilePath(basePath string) string {
...
+
+func (m *FeatureMap) keys() []string {
+	return m.featureKeys
+}
 
-func (m *FeatureMap) Add(name string) {
-	(*m)[name] = &Feature{
+func (m *FeatureMap) add(name string) {
+	if _, ok := m.featureMap[name]; !ok {
+		m.featureKeys = append(m.featureKeys, name)
```
mariusSincovici, line 114: can `featureKeys` be `nil`?

DoYouEvenCpp: To my best knowledge there should be no case where those could be `nil`.

config/remote/config.go
DoYouEvenCpp: Added.

config/remote/config.go
DoYouEvenCpp: Added helper method.

config/remote/remote.go
```diff
@@ -212,23 +233,32 @@ func findMatchingRecord(ss []ParamValue, ver string, rollout int) (match *ParamV
...
+	if match != nil {
+		if match.Rollout > c.rolloutGroup {
+			c.analytics.EmitPartialRolloutEvent(ClientCli, featureName, match.Rollout, partialRolloutPerformedFailure)
+			match = nil
+		} else {
+			c.analytics.EmitPartialRolloutEvent(ClientCli, featureName, match.Rollout, partialRolloutPerformedSuccess)
+		}
+	} else {
+		//when there's no match (eg., due to a version value mismatch) emit the partial rollout event failure
+		//TODO: change the structure of this event to hold error anyway?
```
devzbysiu: Leftover or something for other PR?

devzbysiu (approved):

config/remote/remote.go
DoYouEvenCpp: Note for this branch, where the match is nil, we must emit _some_ value of feature rollout group (now made it zero, arbitrarily)  - this is why I added this `TODO` here. So we're either fine about it, or an future update might be needed in here.

config/remote/analytics_builder_test.go
```diff
@@ -0,0 +1,152 @@
...
+			event:           NewRolloutEvent(rolloutGroup, "test-client", FeatureMeshnet, 50, true),
+			expectedResult:  rolloutYes,
+			expectedError:   "meshnet 42 / 50",
+			expectedMessage: FeatureMeshnet,
+		},
+		{
+			name:            "Rollout Failure",
+			event:           NewRolloutEvent(rolloutGroup, "test-client", FeatureMeshnet, 50, false),
+			expectedResult:  rolloutNo,
+			expectedError:   "meshnet 42 / 50",
```
keliramu: minor: if it is rollout failure then expected error should be like this `meshnet 65 / 50` ie. user app rollout group should be bigger than feature's

config/remote/analytics_builder_test.go
```diff
@@ -0,0 +1,152 @@
...
+			event:           NewDownloadFailureEvent(rolloutGroup, "client", FeatureMain, DownloadErrorNetwork, "timeout"),
+			expectedResult:  rcFailure,
+			expectedError:   DownloadErrorNetwork.String(),
+			expectedMessage: "timeout",
+		},
+		{
+			name:            "JSON Parse Failure",
+			event:           NewJSONParseEventFailure(rolloutGroup, "client", FeatureLibtelio, LoadErrorMainHashJsonParsing, "bad hash"),
+			expectedResult:  rcFailure,
+			expectedError:   "main_hash_json_parsing_error",
```
keliramu: this should be `LoadErrorMainHashJsonParsing.String()`

config/remote/analytics_builder_test.go
keliramu: to be completely correct - need `rolloutGroup` be part of the test case not like now it's global for all test cases

config/remote/analytics_test.go
```diff
@@ -1,177 +1,290 @@
...
+	analytics := NewRemoteConfigAnalytics(publisher, rolloutGroup)
+
+	return &analyticsTestFixture{
+		publisher:  publisher,
+		subscriber: subscriber,
+		analytics:  analytics,
 	}
+}
+
+// The tests verify that MooseAnalytics correctly publishes analytics events
```
keliramu: Moose should not be mentioned here, or if it needs - it should under build tag protection

config/remote/analytics_test.go
```diff
@@ -1,177 +1,290 @@
...
-			name:            "Rollout Success",
-			event:           NewRolloutEvent(ctx, "test-client", FeatureMeshnet, 50, true),
-			expectedResult:  rolloutYes,
-			expectedError:   "meshnet 42 / 50",
-			expectedMessage: FeatureMeshnet.String(),
+			name: "EmitDownloadEvent success",
+			action: func(a Analytics) {
+				a.EmitDownloadEvent(client, feature)
+			},
+			expectedEventName: `"event":"rc_download_success"`,
```
keliramu: Predefined constants should be used here e.g. `DownloadSuccess.String()`

keliramu: And in all following test cases

config/remote/analytics_test.go
```diff
@@ -1,177 +1,290 @@
...
-			event:           NewLocalUseEvent(ctx, "client", FeatureLibtelio),
-			expectedResult:  rcSuccess,
-			expectedError:   "",
-			expectedMessage: "",
+			name: "EmitPartialRolloutEvent failure",
+			action: func(a Analytics) {
+				a.EmitPartialRolloutEvent(client, feature, 7, false)
+			},
+			expectedEventName: `"event":"rc_rollout"`,
+			expectedDetails:   `"error":"meshnet 42 / 7"`,
```
keliramu: To be correct, expected details should be where second number is bigger e.g. `meshnet 42 / 74`

config/remote/analytics_builder_test.go
DoYouEvenCpp: Not entirely sure what do you mean here, however the test suite of analytics uses `rolloutGroup` in a more modifiable fashion. Here, the tests are solely about building up analytics events.

config/remote/analytics_builder.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+// EventDetails holds optional details for an event.
+type EventDetails struct {
+	Error              string `json:"error"`
+	Message            string `json:"message"`
+	RolloutFeatureName string `json:"feature_name"`
```
keliramu: Looks like it's unnecessary duplication of field `RolloutFeatureName` - it's the same value as `FeatureName` - I'd suggest to remove one of them. Also, does additional structure `EventDetails` make any sense? I would leave just one `Event`.

DoYouEvenCpp: Fixed, removed the `EventDetails` completely, and moved unique fields into `Event` structure.

cmd/daemon/main.go
```diff
@@ -349,7 +349,8 @@ func main() {
 		log.Println(internal.ErrorPrefix, "getting rollout group:", err)
 		// in case of error, rollout group is `0`
 	}
-	rcConfig := getRemoteConfigGetter(buildTarget, RemotePath, cdnAPI, rolloutGroup)
+	rcAnalytics := remote.NewRemoteConfigAnalytics(daemonEvents.Debugger.DebuggerEvents, rolloutGroup)
+	rcConfig := getRemoteConfigGetter(buildTarget, RemotePath, cdnAPI, rcAnalytics, rolloutGroup)
```
thegravi: Can we eliminate rcAnalytics var and directly pass it to `getRemoveConfigGetter`?

config/remote/analytics.go
```diff
@@ -1,168 +1,54 @@
...
-	ISP          string
-	RolloutGroup int
+// Analytics defines an interface for reporting various analytics related to remote config
+// including: download, RC local usage, JSON parsing, and partial rollouts event.
+type Analytics interface {
+	EmitDownloadEvent(string, string)
+	EmitDownloadFailureEvent(string, string, DownloadError)
+	EmitLocalUseEvent(string, string, error)
+	EmitJsonParseFailureEvent(string, string, LoadError)
+	EmitPartialRolloutEvent(string, string, int, bool)
```
thegravi: Why not also provide variable names? What should be expected from the implementation structure point of view.

config/remote/analytics.go
```diff
@@ -1,168 +1,54 @@
...
-}
-
-func NewDownloadSuccessEvent(info UserInfo, client string, featureName FeatureName) Event {
-	return Event{
-		UserInfo:    info,
-		Client:      client,
-		FeatureName: featureName.String(),
-		Event:       DownloadSuccess.String(),
-	}
+func NewRemoteConfigAnalytics(publisher events.PublishSubcriber[events.DebuggerEvent], rg int) *RemoteConfigAnalytics {
```
thegravi: Can we use more descriptive names? rg - rolloutGroup?

config/remote/data.go
```diff
@@ -16,11 +16,21 @@ const (
 )
 
 type ParamValue struct {
-	Value      any    `json:"value"`
-	incValue   string // include file content, unexported, ingnored by json.unmarshal
-	AppVersion string `json:"app_version"`
-	Weight     int    `json:"weight"`
-	Rollout    int    `json:"rollout"`
+	Value         any    `json:"value"`
+	incValue      string // include file content, unexported, ingnored by json.unmarshal
```
thegravi: typo `ingnored`

config/remote/remote.go
```diff
@@ -187,19 +197,29 @@ func (c *CdnRemoteConfig) load() {
...
+				}
+			}
+			log.Println(internal.ErrorPrefix, "failed loading feature [", feature.name, "] config from the disk:", err)
+			c.analytics.EmitLocalUseEvent(ClientCli, feature.name, err)
 			continue
 		}
-		log.Println(internal.InfoPrefix, "feature [", f.name, "] config loaded from:", c.localCachePath)
+		log.Println(internal.InfoPrefix, "feature [", feature.name, "] config loaded from:", c.localCachePath)
+		c.analytics.EmitLocalUseEvent(ClientCli, feature.name, nil)
 	}
```
thegravi: I try to avoid deeper nesting than 3 levels. Also, you can use `switch` for error checking.

DoYouEvenCpp: Agree! However I guess `switch` will also increase cyclomatic complexity here, so maybe for the sake of code clarity I'll extract error handling to a function.

config/remote/remote.go
```diff
@@ -165,41 +168,78 @@ func (c *CdnRemoteConfig) download() (bool, error) {
...
 	newChangesDownloaded := false
 
-	for _, f := range c.features {
-		dnld, err := f.download(cdnFileGetter{cdn: c.cdn}, jsonFileReaderWriter{}, jsonValidator{}, filepath.Join(c.remotePath, c.appEnvironment), c.localCachePath)
+	for _, f := range c.features.keys() {
+		feature := c.features.get(f)
+		dnld, err := feature.download(cdnFileGetter{cdn: c.cdn}, jsonFileReaderWriter{}, jsonValidator{}, filepath.Join(c.remotePath, c.appEnvironment), c.localCachePath)
 		if err != nil {
-			log.Println(internal.ErrorPrefix, "failed downloading feature [", f.name, "] remote config:", err)
+			log.Println(internal.ErrorPrefix, "failed downloading feature [", feature, "] remote config:", err)
```
keliramu: here `feature` should be string to be printed but it's structure - need to use `feature.name`

config/remote/remote.go
```diff
@@ -165,41 +168,78 @@ func (c *CdnRemoteConfig) download() (bool, error) {
...
+			}
 			if isNetworkRetryable(err) {
 				return false, err
 			}
 			continue
 		}
 		if dnld {
 			// only if remote config was really downloaded
-			log.Println(internal.InfoPrefix, "feature [", f.name, "] remote config downloaded to:", c.localCachePath)
+			log.Println(internal.InfoPrefix, "feature [", feature, "] remote config downloaded to:", c.localCachePath)
```
keliramu: here feature should be string to be printed but it's structure - need to use feature.name

keliramu (approved):

PR #999, opened 2025-08-06
Implement prevention to sym/hard link attack

keliramu:


internal/filesystemsec.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	defer unix.Close(fd)
+
+	// get file info from already open file descriptor (second check)
+	var stat unix.Stat_t
+	if err := unix.Fstat(fd, &stat); err != nil {
+		return nil, fmt.Errorf("fstat failed: %w", err)
+	}
+
+	// validate file size again
```
bartoszWojciechO: Why do we need to validate the size two times?

internal/filesystemsec.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	fd, err := unix.Open(cleanPath, unix.O_RDONLY|unix.O_NOFOLLOW, 0)
+	if err != nil {
+		if !errors.Is(err, unix.EINVAL) {
+			return nil, fmt.Errorf("open failed: %w", err)
+		}
+		// fallback to regular open if O_NOFOLLOW is not supported
+		// but verify it's not a symlink first
+		if err := VerifyNotLink(cleanPath); err != nil {
```
bartoszWojciechO: Maybe https://pkg.go.dev/golang.org/x/sys/unix#Fstat can be used for checking?

internal/filesystemsec.go
keliramu: Optimized to have only single size check

internal/filesystemsec.go
keliramu: `Fstat` operates with already open file and `VerifyNotLink` uses `Lstat`. What is your point? plz clarify.

config/remote/config.go
```diff
@@ -68,7 +68,10 @@ func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBaseP
 
 	defer func() {
 		if err != nil {
-			internal.CleanupTmpFiles(targetPath, tmpExt)
+			validDir, _ := internal.IsValidExistingDir(targetPath)
```
mariusSincovici: would it be possible to check if there is an error returned and that case not to use the `validDir`?

config/remote/remote.go
```diff
@@ -129,8 +125,11 @@ func (c *CdnRemoteConfig) LoadConfig() error {
 	var err error
 	reloadDone := false
 	c.initOnce.Do(func() {
-		c.load() // on start init cache from disk
-		reloadDone = true
+		validDir, _ := internal.IsValidExistingDir(c.localCachePath)
```
mariusSincovici: same here, to check the error?

internal/filesystemsec.go
```diff
@@ -0,0 +1,241 @@
...
+// CheckPathForSymlinks checks if any component of the path is a symlink
+func CheckPathForSymlinks(path string) error {
+	var pathComponents []string
+	var checkPath string
+
+	// handle absolute and relative paths
+	cleanPath := filepath.Clean(path)
+
+	if filepath.IsAbs(cleanPath) {
+		pathComponents = strings.Split(cleanPath, string(os.PathSeparator))
```
mariusSincovici: is there a different for this between if/else? or it can be moved outside?

internal/filesystemsec.go
```diff
@@ -0,0 +1,241 @@
...
+		info, err := os.Lstat(cleanPath)
+		if err != nil {
+			return nil, fmt.Errorf("stat failed: %w", err)
+		}
+		if info.Size() > MaxBytesLimit {
+			return nil, fmt.Errorf("%w: %d bytes", ErrFileTooLarge, info.Size())
+		}
+
+		// use regular file read with size limit
+		file, err := os.Open(cleanPath)
```
mariusSincovici: why not reopen the file without the `O_NOFOLLOW` and then just run the same code a for when it exists?

internal/filesystemsec.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// remove existing file if it exists (prevents hardlink issues)
+	if FileExists(cleanPath) {
+		if err := os.Remove(cleanPath); err != nil {
+			return fmt.Errorf("removing existing file: %w", err)
+		}
+	}
+
+	// atomic rename
+	if err := os.Rename(tmpname, cleanPath); err != nil {
```
mariusSincovici, line 209: If rename fails tmp file is never deleted because tmpfile==nil

internal/filesystemsec.go
bartoszWojciechO: Isn't the file opened at this point(line 29)? Wouldn't it be possible to do the same validation using the file descriptor?

keliramu: Here we are in `err/fail` block when `O_NOFOLLOW` is not supported and `fd` is not valid here

bartoszWojciechO: Got it, thank you.

config/remote/config.go
keliramu: Don't want to trigger false negative signals here. If problem with `targetPath` would exist - it will be checked and reported much earlier in the process.

config/remote/remote.go
keliramu: Added error check

internal/filesystemsec.go
keliramu: Right, `Split` line can be moved outside if/else

internal/filesystemsec.go
keliramu: fixed

internal/filesystemsec.go
keliramu: in `defer` block modified tmp file cleanup to not depend on if tmp file was closed

PR #1001, opened 2025-08-07
LVPN 8626: Add firewall mark into CDN HTTP transport

DoYouEvenCpp:
Changes under this PR:
- CDN is now provided with a firewall mark based HTTP transport
- reconstructed the way how is `threatProtectionLiteServers` now built
   - it creates now a temporary CDN object just to fetch the list of servers
- e2e test to cover the change


cmd/daemon/main.go
```diff
@@ -266,6 +266,13 @@ func main() {
 		log.Println(internal.WarningPrefix, "failed to set buffer size for HTTP/3:", err)
 	}
 
+	//once resolver is created, let's recreate httpsClientSimple transport with the firewall mark
+	httpClientSimple.Transport = request.NewHTTPReTransport(
```
mariusSincovici: shouldn't like [239](https://github.com/NordSecurity/nordvpn-linux/pull/1001/files#diff-82554fd7bcb5b77d26245b5fc8d10f32b79340ede041fe0f672ae845cff4f08fR239) be removed?

DoYouEvenCpp: Yes, wanted to remove it but the HTTP transport _is_ required to create resolver which then is crucial to properly apply firewall mark. 

Which is a bit of an internal cross-dependencies (between HTTP transport and resolver).

Weighting it all up, I decided to recreate it instead of providing here a significant rework of HTTP transport and the resolver in order to decouple those two (am happy however to create a follow-up ticket, if that's something we want to handle in the future)

cmd/daemon/main.go
```diff
@@ -234,33 +237,34 @@ func main() {
 
 	httpGlobalCtx, httpCancel := context.WithCancel(context.Background())
 
+	var cdnAPI *core.CDNAPI
+
+	var threatProtectionLiteServers *dns.NameServers
+	resolver := func() network.DNSResolver {
```
keliramu: Why not use new type you just defined on line 130, as return type

nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=1001) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=1001) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.nordsec.com/check/NordSecurity/nordvpn-linux?pullRequest=1001) it.</sub>

mariusSincovici (approved):

PR #1003, opened 2025-08-07
LVPN-8129: Prepare core suite of automation tests

bbrutall:
- Implemented the initial test automation infrastructure for the RemoteConfig feature.
- Added a few automation tests from the core suite to get things started.

### DIFF
--- a/test/qa/conftest.py
+++ b/test/qa/conftest.py
@@ -1,5 +1,7 @@
 import datetime
 import io
+from urllib.parse import urlparse
+
 import lib
 import subprocess
 import signal
@@ -14,6 +16,10 @@ import sys
 import os
 
 from lib import logging, network, daemon, login, info, firewall
+
+from lib.remote_config_manager import RemoteConfigManager, LOCAL_CACHE_DIR, REMOTE_DIR
+from lib.logging import FILE
+from lib.log_reader import LogReader
 
 pytest_plugins = ("lib.pytest_timeouts.pytest_timeouts")
 
@@ -185,3 +191,107 @@ def disable_dns_and_threat_protection():
 def disable_notifications():
     """Disable notifications."""
     lib.set_notify("off")
+
+
+@pytest.fixture
+def default_config() -> dict:
+    """Fixture to provide a default config."""
+    return dict()
+
+
+@pytest.fixture
+def rc_config_manager(default_config: dict
+) -> RemoteConfigManager:
+    """
+    Fixture to create and provide an instance of "RemoteConfigManager".
+
+    :param default_config: A dictionary containing the default config values.
+
+    :return: An instance of the "RemoteConfigManager".
+    """
+    return RemoteConfigManager(
+        env="dev",
+        cache_dir=LOCAL_CACHE_DIR,
+        default_config=default_config
+    )
+
+
+@pytest.fixture
+def daemon_log_reader() -> LogReader:
+    """
+    Fixture to provide an instance of the "LogReader" class.
+
+    :return: an instance of the "LogReader" class with the default log file path.
+    """
+    return LogReader(FILE)
+
+
+@pytest.fixture
+def daemon_log_cursor(daemon_log_reader) -> int:
+    """
+    Fixture to get the current cursor for the log file.
+
+    Uses the "LogReader" class to determine the current EOF cursor.
+    If the log file is not found or empty, the cursor is set to 0.
+
+    :param daemon_log_reader: An instance of the "LogReader" class.
+
+    :return: An integer cursor representing the current end of the log file..
+    """
+    cursor = daemon_log_reader.get_cursor()
+
+    print(f"Cursor: {cursor}")
+    return cursor
+
+
+@pytest.fixture
+def clean_cache_files(rc_config_manager):
+    """
+    Fixture to clean up the files in "LOCAL_CACHE_DIR" directory before running tests.
+
+    :param rc_config_manager: An instance of the "RemoteConfigManager" class.
+    """
+    if rc_config_manager.set_permissions_cache_dir():
+        subprocess.run(["sudo", "rm", "-rf", LOCAL_CACHE_DIR], check=True)
+    else:
+        print(f"Unable to clean {LOCAL_CACHE_DIR}, as it does not exist.")
+
+    yield
+
+
+@pytest.fixture
+def disable_remote_endpoint():
+    """
+    Fixture to temporarily disable connections to "remote endpoint" by modifying "/etc/hosts".
+
+    This fixture performs the following steps:
+    1. Backs up the current "/etc/hosts" file to "/etc/hosts.bak".
+    2. Appends the following entry "0.0.0.0 %remote endpoint%" to "/etc/hosts" to block "remote endpoint".
+        This ensures the application cannot reach this endpoint (e.g., to download remote config files).
+    3. Restores the original "/etc/hosts" file after the test completes.
+
+    :raises ValueError: If REMOTE_DIR is not set.
+    """
+    if not REMOTE_DIR:
+        raise ValueError("REMOTE_DIR environment variable is not set.")
+    domain = urlparse(REMOTE_DIR).netloc
+
+    hosts_original = "/etc/hosts"
+    hosts_backup = "/etc/hosts.bak"
+
+    subprocess.run(["sudo", "cp", hosts_original, hosts_backup], check=True)
+    print(f"Backup created at {hosts_backup}")
+
+    print(f"Adding '0.0.0.0 {domain}' to {hosts_original}")
+    cmd = ["sudo", "bash", "-c", f'echo "0.0.0.0 {domain}" >> /etc/hosts']
+    subprocess.run(cmd, check=True)
+
+    yield
+
+    try:
+        subprocess.run(["sudo", "cp", hosts_backup, hosts_original], check=True)
+        subprocess.run(["sudo", "chmod", "644", hosts_original], check=True)
+        print(f"{hosts_original} restored from {hosts_backup}.")
+    except Exception as e: # noqa: BLE001
+        print(f"Got an error during restoring {hosts_original} from {hosts_backup}: {e}")
+        raise
--- a/test/qa/lib/log_reader.py
+++ b/test/qa/lib/log_reader.py
@@ -1,0 +1,143 @@
+import os
+import time
+
+
+class LogReader:
+    """A utility class for reading logs, partially or fully, and waiting for specific messages in the log file."""
+
+    def __init__(self, log_path: str):
+        """
+        Initializes the "LogReader" with the path to the log file.
+
+        :param log_path: The path to the log file.
+        """
+        self.log_path = log_path
+
+    def get_full_log(self) -> str:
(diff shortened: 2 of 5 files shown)


test/qa/conftest.py
```diff
@@ -185,3 +191,103 @@ def disable_dns_and_threat_protection():
 def disable_notifications():
     """Disable notifications."""
     lib.set_notify("off")
+
+
+@pytest.fixture
+def default_config() -> dict:
```
mariusSincovici, line 197: Is this the default config for RC only or something in general, that can also contain RC?

test/qa/conftest.py
```diff
@@ -185,3 +191,103 @@ def disable_dns_and_threat_protection():
...
+    """Fixture to provide a default config."""
+    return dict()
+
+
+@pytest.fixture
+def config_manager(default_config: dict
```
mariusSincovici: [small] wouldn't be more clear that this is the RC manager by adding `rc_`config_manager?

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+import uuid
+
+import pytest
+
+from lib import daemon
+from lib.log_reader import LogReader
+from lib.remote_config_manager import LOCAL_CACHE_DIR
+
+RC_REMOTE_MESSAGES = [
+    "[Info] feature [ meshnet ] remote config downloaded to: /var/lib/nordvpn/conf",
```
devzbysiu, line 10: Isn't the path different in snap?

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+
+    for hash_file in rc_hash_files:
+        config = config_manager.read_config(hash_file)
+        rc_hash_data_before.append((hash_file, config.get("hash")))
+
+        # Set new hash value
+        config.set("hash", hash_value)
+        config.save_to_file(hash_file)
+
+        hash_value_after = config_manager.read_config(hash_file).get("hash")
```
devzbysiu: Maybe this `"hash"` could be put in some var? It's used in few places

test/qa/conftest.py
```diff
@@ -185,3 +191,103 @@ def disable_dns_and_threat_protection():
...
+    subprocess.run(["sudo", "cp", hosts_original, hosts_backup], check=True)
+    print(f"Backup created at {hosts_backup}")
+
+    print(f"Adding '0.0.0.0 {domain}' to {hosts_original}")
+    cmd = ["sudo", "bash", "-c", f'echo "0.0.0.0 {domain}" >> /etc/hosts']
+    subprocess.run(cmd, check=True)
+
+    yield
+
+    subprocess.run(["sudo", "cp", hosts_backup, hosts_original], check=True)
```
mariusSincovici: Shouldn't this revert be put in some try/catch or equivalent? this specially for when the tests are executed locally.

test/qa/lib/log_reader.py
```diff
@@ -0,0 +1,145 @@
...
+                    print("All specified messages have been found in the log.")
+                    return True
+
+            except FileNotFoundError:
+                print(f"Log file '{self.log_path}' not found. Retrying...")
+
+            time.sleep(interval)
+
+        print(f"Messages not found within {timeout} seconds: {messages_to_find}")
+        return False if not return_not_found else (False, messages_to_find)
```
mariusSincovici, line 124: wouldn't this return in some cases bool and other tuple?

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+        subprocess.run(["sudo", "rm", hash_file], check=True)
+
+    daemon.start()
+
+    assert daemon_log_reader.wait_for_messages(
+        messages=RC_INITIAL_RUN_MESSAGES, cursor=cursor
+    )
+
+
+def test_disable_remote_config_download_and_verify_log_messages_not_found(
```
devzbysiu, line 138: Won't this test pass also when there is something wrong with RC? Maybe we should first check that it's working, then disable remote endpoint and then check that logs are not there?

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+    )
+
+
+def test_local_hash_files_removal_and_daemon_restart(
+    initialized_app_with_remote_config, # noqa: ARG001
+    daemon_log_reader,
+    config_manager,
+) -> None:
+    """
+    Test to verify the application's behavior when local hash configu files are deleted and the daemon is restarted.
```
devzbysiu: [typo] `configu`

test/qa/lib/remote_config_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class RemoteConfigManager:
+    """Manages remote, local, and backup config retrieval, hash checking, and caching."""
+
+    def __init__(
+        self, env: str, cache_dir: str, default_config: dict
+    ) -> None:
+        self.env = env
+        self.cache_dir = LOCAL_CACHE_DIR or cache_dir
```
mariusSincovici: shouldn't it be the other way `cach or LOCAL_CACHE_DIR`? Because `LOCAL_CACHE_DIR` is always set.

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+
+    rc_hash_files = config_manager.get_local_hash_files()
+
+    cursor = daemon_log_reader.get_cursor()
+
+    daemon.stop()
+
+    for hash_file in rc_hash_files:
+        print(f"Deleting hash file: {hash_file}")
+        subprocess.run(["sudo", "rm", hash_file], check=True)
```
devzbysiu: Maybe this could be part of the `config_manager`? Something like `config_manager.delete_hash_files()`

test/qa/test_remote_config.py
```diff
@@ -0,0 +1,245 @@
...
+    2. Identify and remove files that match the naming pattern "*-hash.json".
+    3. Stop the daemon after identifying and deleting the relevant files.
+    4. Restart the daemon after modifications to the local config files.
+    5. Verify that the application logs the appropriate messages
+       ("RC_INITIAL_RUN_MESSAGES") after the daemon startup.
+
+    :raises AssertionError: If the expected log messages are not found after deleting hash files.
+    """
+
+    rc_hash_files = config_manager.get_local_hash_files()
```
devzbysiu: What if this returns empty list? Maybe we should assert that some files are already there so RC worked, then we remove them, then restart the daemon and then we see that it  downloads the files again

test/qa/test_remote_config.py
mariusSincovici: [spelling] `configu`

test/qa/conftest.py
```diff
@@ -185,3 +191,103 @@ def disable_dns_and_threat_protection():
...
+
+
+@pytest.fixture
+def clean_cache_dir(config_manager):
+    """
+    Fixture to clean up the "LOCAL_CACHE_DIR" directory before running tests.
+
+    :param config_manager: An instance of the "RemoteConfigManager" class.
+    """
+    if config_manager.set_permissions_cache_dir():
```
mariusSincovici: This and the next operations might not be referring to the same folder in the future. They are now, but `config_manager` gets the cache as param and fallback to `LOCAL_CACHE_DIR`.
Also is it needed to change the permissions before deleting, because deletion is made with sudo?

ci/test_deb.sh
```diff
@@ -74,7 +74,7 @@ fi
 # Disable the TUI loader indicator to prevent interference during automated tests
 export DISABLE_TUI_LOADER=1
 
-python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true \
```
edvyno1: `-x` flag is for stopping on first failure, please keep this on

test/qa/test_remote_config.py
bbrutall: "hash" is used only 3 times within a small code block, so I suggest it's  fine to use "hash" directly each time. If it would be use  many times, or across multiple functions or files, - yes it makes sense to put it in some var.

ci/test_deb.sh
bbrutall: Why do we need to stop all the tests after the first test failure?

edvyno1: faster feedback loop when a test fails, for day to day development failing fast saves time and resources

test/qa/test_remote_config.py
bbrutall: Yes, it's different.

test/qa/lib/log_reader.py
bbrutall: yes, when we explicilty set `return_not_found=True`

test/qa/conftest.py
bbrutall: Yes.

test/qa/lib/remote_config_manager.py
bbrutall: you are right, need to swap it `cache_dir or LOCAL_CACHE_DIR`

test/qa/conftest.py
bbrutall: Yes, it makes sense.

test/qa/conftest.py
bbrutall: I don't mind :)

test/qa/conftest.py
bbrutall: 2nd one, but now it's just a placeholder.

test/qa/test_remote_config.py
bbrutall: This test is using `initialized_app_with_remote_config` as a core happy path verification. If there are no local RC files - test will fail on setup phase.

test/qa/test_remote_config.py
bbrutall: You are right, included `initialized_app_with_remote_config` fixture.

test/qa/lib/remote_config_manager.py
```diff
@@ -0,0 +1,151 @@
...
+            raise ValueError("REMOTE_FILES environment variable is not set.")
+
+        remote_files = REMOTE_FILES.split(",")
+        configs = {}
+
+        for rc_file in remote_files:
+            full_url = os.path.join(REMOTE_DIR, rc_file.strip())
+            try:
+                print(f"Fetching config from: {full_url}")
+                response = requests.get(full_url, timeout=5)
```
ggediminass, line 137: Why `timeout=5`? Maybe it should become a parameter in `get_remote_configs` function? :thinking:

test/qa/test_remote_config.py
bbrutall: Yes, makes sense.

test/qa/lib/remote_config_manager.py
bbrutall: It's best practice to make timeouts configurable,  but it’s not always necessary. If the server doesn’t respond within 5 seconds the request will raise RequestException exception. If such an issue occurs, we'll investigate it and handle it properly using the configurable increased timeout value.

PR #1005, opened 2025-08-08
Add docstrings to Settings tests

naujalyte:


test/qa/test_settings.py
```diff
@@ -60,6 +83,15 @@ def test_technology_set_options(tech, proto, obfuscated):
...
     lib.set_technology_and_protocol(tech, proto, obfuscated)
+    """
+    Manual TCs:
+        [openvpn-udp-on] - LVPN-8836
+        [openvpn-tcp-on] - LVPN-8737
+        [openvpn-udp-off] - LVPN-8834
+        [openvpn-tcp-off] - LVPN-8835
+        [nordlynx--] - LVPN-8837
+        [nordwhisper--] - LVPN-8833
+    """  # noqa: D205, D400
```
devzbysiu: Shouldn't those comments be directly under the function declaration? Similar for `test_set_defaults_when_logged_out_2nd_set` below

naujalyte: yes, did not notice that it was in the wrong place, thanks, will update :+1:

test/qa/test_settings.py
```diff
@@ -20,6 +20,8 @@ def teardown_function(function):  # noqa: ARG001
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES_BASIC1)
 def test_obfuscate_nonobfucated(tech, proto, obfuscated):
+    """Manual TC: LVPN-788"""  # noqa: D400
```
edvyno1: maybe do not enforce `D400`, remove it from `ruff.toml`, I don't see a lot of gain from just having or not having a dot at the end of the first line, even more so in this scenario where it simply does not make sense

test/qa/test_settings.py
```diff
@@ -30,6 +32,11 @@ def test_obfuscate_nonobfucated(tech, proto, obfuscated):
 
 @pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES_BASIC2 + lib.TECHNOLOGIES_BASIC1)
 def test_set_technology(tech, proto, obfuscated):  # noqa: ARG001
+    """
+    Manual TCs:
+        [openvpn-udp-off] - LVPN-615
+        [nordlynx--] - LVPN-601
+    """  # noqa: D205, D400
```
edvyno1: `D205` can be solved by just having a blank line at the end of between the `Manual TCs` header and the ticket IDs, they should all look something like this:
```
    """
    Manual TC(s):

        LVPN-whatever
    """

naujalyte: Fixed :heavy_check_mark:

test/qa/test_settings.py
naujalyte: Fixed ✔️

test/qa/test_settings.py
naujalyte: Fixed ✔️

PR #1021, opened 2025-08-14
LVPN-6704: Logout reasons reporting

thegravi:


cmd/daemon/main.go
```diff
@@ -304,19 +303,23 @@ func main() {
 		machineIdGenerator.GetUsedInformationMask(),
 	)
 
-	invalidSessionErrHandlingReg := internal.NewErrorHandlingRegistry[error]()
-
-	// encapsulating initialization logic
-	clientAPI, accessTokenSessionStore := func() (core.ClientAPI, session.SessionStore) {
-		api := core.NewSimpleAPI(
+	// Build client API and session stores
+	clientAPI, sessionBuilder := func() (core.ClientAPI, *SessionStoresBuilder) {
```
mariusSincovici: [suggestion] Wouldn't make sense to, extract this in a function and add a unit-test that uses reflection to check that the function will create all the struct members?

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+			}
+		}
+		return events.ReasonNone
+	})
+}
+
+// registerVPNCredsHandlers configures VPN credentials error handlers
+func (b *SessionStoresBuilder) registerVPNCredsHandlers(logoutHandler *daemon.LogoutHandler) {
+	errs := []error{
+		core.ErrUnauthorized,
```
mariusSincovici, line 144: Are `core.ErrUnauthorized` and `core.ErrBadRequest` needed, because in `Register` only the other errors are checked?

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+	errs := []error{
+		core.ErrBadRequest,
+		core.ErrUnauthorized,
+		core.ErrNotFound,
+		session.ErrInvalidToken,
+		session.ErrInvalidOwnerID,
+		session.ErrMissingTrustedPassResponse,
+	}
+
+	logoutHandler.Register(b.registries.trustedPass, errs, alwaysReasonNone)
```
mariusSincovici, line 193: Is this and the `logoutHandler.Register(b.registries.ncCreds...` needed since they don't do anything?

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+	)
+	return b.stores.accessToken
+}
+
+// BuildVPNCredsStore builds and returns the VPN credentials session store
+func (b *SessionStoresBuilder) BuildVPNCredsStore(clientAPI core.ClientAPI) session.SessionStore {
+	b.stores.vpnCreds = session.NewVPNCredentialsSessionStore(
+		b.confman,
+		b.registries.vpnCreds,
+		renewVPNCredentials(b.confman, clientAPI),
```
mariusSincovici, line 68: This function receives a `smartAPI` that has a reference to the AccessTokenSessionStore. But on the renew here it still uses config manager to get the token. 
Maybe the credentials interface needs to be changed not to request a token?

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+
+// renewVPNCredentials creates a VPN credentials renewal function
+func renewVPNCredentials(confman config.Manager, api core.ClientAPI) session.VPNCredentialsRenewalAPICall {
+	return func() (*session.VPNCredentialsResponse, error) {
+		data, err := getTokenData(confman)
+		if err != nil {
+			return nil, err
+		}
+
+		resp, err := api.ServiceCredentials(data.Token)
```
mariusSincovici, line 270: I was referring to this token

core/errors.go
```diff
@@ -61,7 +61,7 @@ func ExtractError(resp *http.Response) error {
 	}
 
 	var info apiError
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
```
mariusSincovici: This needs to be replaced with `MaxBytesReadAll`.

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+// errorRegistries holds all error handling registries
+type errorRegistries struct {
+	accessToken *internal.ErrorHandlingRegistry[error]
+	vpnCreds    *internal.ErrorHandlingRegistry[error]
+	trustedPass *internal.ErrorHandlingRegistry[error]
+	ncCreds     *internal.ErrorHandlingRegistry[error]
+}
+
+// NewSessionStoresBuilder creates a new builder for session stores
+func NewSessionStoresBuilder(confman config.Manager) *SessionStoresBuilder {
```
DoYouEvenCpp, line 41: Wont this (returning structure instead of an interface) cause any troubles testing-wise?

cmd/daemon/init_session_stores.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+}
+
+// BuildAccessTokenStore builds and returns the access token session store
+func (b *SessionStoresBuilder) BuildAccessTokenStore(rawClientAPI core.RawClientAPI) *session.AccessTokenSessionStore {
+	b.stores.accessToken = session.NewAccessTokenSessionStore(
+		b.confman,
+		b.registries.accessToken,
+		renewAccessToken(rawClientAPI),
+		nil,
```
DoYouEvenCpp: [further-development] I'm perfectly aware that this has been already discussed and reviewed (in other PRs), but now come to think about it, how do you fancy idea of passing a noop validator instead of a `nil`?

Just a `nil` is not the most verbose thing in the world, and I guess passing there a `noopValidator` (or something similar) would put one already into a correct context.

daemon/session_error_handler.go
```diff
@@ -25,76 +24,88 @@ type SessionErrorHandlerDependencies struct {
...
-		// Prevent concurrent logout attempts
-		state.mu.Lock()
-		if state.logoutInProgress {
-			state.mu.Unlock()
-			log.Printf(
-				"%s Session error detected but logout already in progress, ignoring: %v",
-				internal.DebugPrefix,
-				reason)
+		// Prevent concurrent logouts
+		h.state.Lock()
```
mariusSincovici: Wouldn't make sense to use atomic or a single mutex without extra bool?

For mutex the first to be a `tryLock` and if it fails someone else is executing the code, otherwise keep the lock until defer.
With atomic the CompareAndSwap, maybe?

cmd/daemon/init_session_stores.go
thegravi: These are the errors that are registered to be 'tracked/monitored' if they occur, which forces the session store `Renew` method to return an error because we **care** about it. Otherwise renewal process would ignore these errors as not significant to the renewal process and tell nothing bad happened and 'we are still logged in'. The following handling does not use these errors for forced logout because there is no relevant exception code to be used.

cmd/daemon/init_session_stores.go
thegravi: They log out the user. There is `alwaysReasonNone` meaning logout without exception code.

cmd/daemon/init_session_stores.go
thegravi: Two functions need an explicitly provided token because they are used during the login phase, when a token is not yet stored in storage.

daemon/session_error_handler.go
thegravi: Good point, will use try-lock approach.

dfetti (approved):

cmd/daemon/main_test.go
```diff
@@ -0,0 +1,29 @@
...
+	)
+
+	assert.NotNil(t, clientAPI)
+	assert.NotNil(t, sessionBuilder)
+
+	stores := sessionBuilder.GetStores()
+	assert.Len(t, stores, 4)
+
+	for i, store := range stores {
+		assert.NotNil(t, store, "Store at index %d should not be nil", i)
```
DoYouEvenCpp, line 27: I'm not sure about usefulness of this test case, particularly this check.
From what I understand, the session stores can't be really nil, correct?

We can test against the quantity of stores created, but this test doesn't say much about actual objects' type.
If the invariant is the sessions stores are always non nil, we can snap an assertion in the damon/main maybe, to return early in case they're not created due any reason (or handle this situation accordingly).

cmd/daemon/init_session_stores_test.go
```diff
@@ -0,0 +1,889 @@
...
+func setupTestConfig() *config.Config {
+	uid := int64(123)
+	idempotencyKey := uuid.New()
+	userID := uuid.New()
+
+	return &config.Config{
+		AutoConnectData: config.AutoConnectData{ID: uid},
+		TokensData: map[int64]config.TokenData{
+			uid: {
+				Token:          "ab78bb36299d442fa0715fb53b5e3e57",
```
DoYouEvenCpp: [nit] this is a bit of a petty remark, however  all of those magic-values here (tokens, dates, token's names, etc.) makes following tests logic a bit harder. Can we please extract those to a constants, here, for the sake of tests clarity?

cmd/daemon/init_session_stores_test.go
```diff
@@ -0,0 +1,889 @@
...
+			case "vpnCreds":
+				registry = builder.registries.vpnCreds
+			case "trustedPass":
+				registry = builder.registries.trustedPass
+			case "ncCreds":
+				registry = builder.registries.ncCreds
+			}
+
+			handlers := registry.GetHandlers(tt.apiError)
+			if len(handlers) > 0 {
```
DoYouEvenCpp, line 712: This check seems redundant here, right?

cmd/daemon/init_session_stores_test.go
```diff
@@ -0,0 +1,889 @@
...
+				AuthChecker:            &mockauth.AuthCheckerMock{},
+				PublishDisconnectFunc:  func(events.DataDisconnect) {},
+				DebugPublisherFunc:     func(string) {},
+			})
+
+			builder.registerAccessTokenHandlers(logoutHandler)
+			builder.registerVPNCredsHandlers(logoutHandler)
+
+			handlers := builder.registries.vpnCreds.GetHandlers(tt.vpnCredsError)
+			if len(handlers) > 0 {
```
DoYouEvenCpp, line 813: And again here, is that really needed?
The flow here is perfectly controlled (the handlers), so it should be stable all the time, right?
Thus I guess we can simply assert against handlers length, and then simply run it?

Or is there any deeper  logic, I simply miss to spot?

core/errors.go
```diff
@@ -75,42 +74,45 @@ func ExtractError(resp *http.Response) error {
...
-		nicknameHasDoubleHyphensCode = 101131 // nickname contains double hyphens
-		invalidCharsCode             = 101132 // nickname contains invalid characters
+		rateLimitReachCode                   = 101126 // rate limit reached (max allowed nickname changes per user per week)
+		nicknameTooLongCode                  = 101127 // nickname too long
+		duplicateNicknameCode                = 101128 // duplicate nickname (nickname already exist)
+		forbiddenWordCode                    = 101129 // nickname with forbidden word
+		invalidPrefixOrSuffixCode            = 101130 // nickname contains invalid prefix or suffix
+		nicknameHasDoubleHyphensCode         = 101131 // nickname contains double hyphens
+		invalidCharsCode                     = 101132 // nickname contains invalid characters
+		maxMachineCountReached               = 101120
```
DoYouEvenCpp: [nit] Care to provide comments for the new constants (even though are quite self-explanatory), to keep the code coherent in here?

internal/errors.go
```diff
@@ -2,8 +2,37 @@ package internal
 
 import (
 	"errors"
+	"fmt"
 )
 
+type CodedError struct {
```
DoYouEvenCpp, line 9: Can you please provide here a comment what's the purpose of this? After a while it might not be obvious what was the sole purpose of `coded errors` :wink:

session/access_token_session_store_silent_renewal_test.go
```diff
@@ -0,0 +1,455 @@
...
+	cfgManager.Cfg = &cfg
+
+	errorRegistry := internal.NewErrorHandlingRegistry[error]()
+
+	renewAPICall := func(token string, key uuid.UUID) (*AccessTokenResponse, error) {
+		assert.Equal(t, "old-token", token)
+		assert.Equal(t, idempotencyKey, key)
+		return &AccessTokenResponse{
+			Token:      "ab78bb36299d442fa0715fb53b5e3e58",
+			RenewToken: "deadbeef1234567890abcdef1234567890",
```
DoYouEvenCpp, line 83: > deadbeef

> Ah, I see you are a man of culture as well


:wink:

session/access_token_session_store_silent_renewal_test.go
```diff
@@ -0,0 +1,455 @@
...
+	cfgManager := mock.NewMockConfigManager()
+	cfgManager.Cfg = &cfg
+
+	errorRegistry := internal.NewErrorHandlingRegistry[error]()
+
+	renewAPICall := func(token string, key uuid.UUID) (*AccessTokenResponse, error) {
+		assert.Equal(t, "old-token", token)
+		assert.Equal(t, idempotencyKey, key)
+		return &AccessTokenResponse{
+			Token:      "ab78bb36299d442fa0715fb53b5e3e58",
```
DoYouEvenCpp, line 82: [nit] Same as in the other files, can you please make them a constants?

cmd/daemon/init_session_stores.go
devzbysiu: I would go even further and remove it entirely. Correct me if I'm wrong, but I think it's not used in any session store. If something like this will be needed it should be introduced then - precisely for the needed use-case. Now it's a dead code, including type definitions from `validation_helpers.go`

cmd/daemon/init_session_stores_test.go
```diff
@@ -0,0 +1,889 @@
...
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
+	mockauth "github.com/NordSecurity/nordvpn-linux/test/mock/auth"
+	mocknetworker "github.com/NordSecurity/nordvpn-linux/test/mock/networker"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type mockRawClientAPI struct {
```
devzbysiu, line 30: There is `mockRawClientAPI` in `session_error_handler_test.go` already. Could it be put in `test/mock` and reused or it's too different and not worth it to merge those two?

daemon/access/access_logout.go
```diff
@@ -136,15 +136,23 @@ type ForceLogoutWithoutTokenInput struct {
...
+	Reason                 events.ReasonCode
 }
 
 // ForceLogoutWithoutToken performs user logout operation without using login toking
 func ForceLogoutWithoutToken(input ForceLogoutWithoutTokenInput) (logoutResult LogoutResult) {
 	logoutStartTime := time.Now()
+
+	// Log the reason if provided
+	if input.Reason != events.ReasonNone {
+		log.Printf("%s Forcing logout due to: %v", internal.DebugPrefix, input.Reason)
```
devzbysiu: Missing `\n` in the `Printf` msg

session/trusted_pass_session_store.go
```diff
@@ -61,12 +68,12 @@ func (s *TrustedPassSessionStore) Renew() error {
...
 	data, ok := cfg.TokensData[uid]
 	if !ok {
-		return fmt.Errorf("there is no data")
+		return fmt.Errorf("there is no data during during trusted pass session renewal")
 	}
 
 	// check if everything is valid or data renewal is required
-	if err := s.validate(); err != nil {
-		if err = s.renewIfOAuth(uid, &data); err != nil {
+	if options.forceRenewal || s.validate(false) != nil {
```
devzbysiu: If possible, bool flags should be avoided, it's usually a code smell:
- it's harder to know immediately what this flag is doing without inspecting the function body (eventually docs if provided and up to date)
- it forces the function to "know" about different use cases

(there are other downsides also like increased cyclomatic complexity, but for small functions this is not a big deal)
Instead I would suggest to have separate functions that do their own thing with clear names so devs don't need to look inside if not needed.
I think this is easy to do in this case, so should be doable. If this would be problematic, then at least we can have an enum instead of bool maybe, so it would be obvious what it's doing instead of `validate(false)`

session/trusted_pass_session_store.go
```diff
@@ -100,18 +107,19 @@ func (s *TrustedPassSessionStore) HandleError(reason error) error {
...
+func (s *TrustedPassSessionStore) validate(skipExpiry bool) error {
 	cfg, err := s.getConfig()
 	if err != nil {
 		return err
 	}
 
-	if err := ValidateTrustedPassTokenFormat(cfg.Token); err != nil {
-		return err
+	if !skipExpiry {
+		if err := ValidateExpiry(cfg.ExpiresAt); err != nil {
```
devzbysiu: The order of the validation (expiry vs token format) was changed so we can potentially exit early or there was a different reason?

session/trusted_pass_session_store.go
```diff
@@ -125,26 +133,42 @@ func (s *TrustedPassSessionStore) validate() error {
 	return nil
 }
 
-func (s *TrustedPassSessionStore) renewToken(uid int64, data *config.TokenData) error {
+func (s *TrustedPassSessionStore) renewToken(uid int64, data *config.TokenData, skipErrorHandlers bool) error {
```
devzbysiu: Same here with the bool flag. And the problematic parts are even more apparent now - many new exit points introduced :smile: 

Eventually, if it would be hard to refactor, then you can introduce small helper to deal with repeated 
```
if skipErrorHandlers {
    return err
}
return s.HandleError(err)
```

session/trusted_pass_session_store.go
```diff
@@ -154,18 +178,21 @@ func (s *TrustedPassSessionStore) renewToken(uid int64, data *config.TokenData)
...
 	if err != nil {
-		return err
+		return fmt.Errorf("saving trusted pass session config: %w", err)
 	}
 
 	return nil
 }
 
-func (s *TrustedPassSessionStore) renewIfOAuth(uid int64, data *config.TokenData) error {
+func (s *TrustedPassSessionStore) renewIfOAuth(uid int64, data *config.TokenData, skipErrorHandlers bool) error {
```
devzbysiu: Same here with the flag

session/trusted_pass_session_store_silent_renewal_test.go
devzbysiu, line 1: Tests are kind of similar after quick look, maybe would be better to make them table driven?

session/trusted_pass_session_store.go
```diff
@@ -61,12 +68,12 @@ func (s *TrustedPassSessionStore) Renew() error {
 	uid := cfg.AutoConnectData.ID
 	data, ok := cfg.TokensData[uid]
 	if !ok {
-		return fmt.Errorf("there is no data")
+		return fmt.Errorf("there is no data during during trusted pass session renewal")
```
devzbysiu: [typo] `during during`

session/errors.go
```diff
@@ -26,6 +35,12 @@ var (
 	// ErrMissingNCCredentials indicates that NC credentials (username/password) are missing.
 	ErrMissingNCCredentials = errors.New("missing nc credentials")
 
-	// ErrInvalidEndpoint indicates that the endpoint is invalid or empty.
-	ErrInvalidEndpoint = errors.New("invalid endpoint")
+	// ErrMissingNCCredentials indicates that NC credentials credentials are missing
+	ErrMissingNCCredentialsResponse = errors.New("renewal API returned nil response")
+
+	// ErrMissingEndpoint indicates that the endpoint is empty.
+	ErrMissingEndpoint = errors.New("invalid endpoint")
```
devzbysiu: Should it be now `"missing endpoint"` or `"empty endpoint"`?

session/errors.go
```diff
@@ -26,6 +35,12 @@ var (
 	// ErrMissingNCCredentials indicates that NC credentials (username/password) are missing.
 	ErrMissingNCCredentials = errors.New("missing nc credentials")
 
-	// ErrInvalidEndpoint indicates that the endpoint is invalid or empty.
-	ErrInvalidEndpoint = errors.New("invalid endpoint")
+	// ErrMissingNCCredentials indicates that NC credentials credentials are missing
```
devzbysiu: Should be `ErrMissingNCCredentialsResponse` and probably the rest of the comment should also be updated

events/events.go
```diff
@@ -156,11 +156,27 @@ type DataDisconnect struct {
 	Error                 error
 }
 
+type ReasonCode int32
+
+const (
+	// no exceptions
+	ReasonNone                         ReasonCode = -1
```
mariusSincovici: [small] Wouldn't make sense to add a comment for this to say that the logout will still be made but without giving any reason?
Because at first read I had the impression `ReasonNone` means don't do the logout :laughing:.

cmd/daemon/init_session_stores.go
mariusSincovici: Yeah, maybe in future we can put the entire logic, also the one from the login RPC, into the AccessTokenSessionStore so there will be only one place to get the token thru the entire app.

mariusSincovici (approved):

cmd/daemon/init_session_stores.go
thegravi: Up until now, there was none. Do you see anything worth testing that can't be done right now?

cmd/daemon/main_test.go
thegravi: Since it is an interface, it can be nil. This test just guards from passing stores that are not properly created. It's important to identify whether "store builder" created all stores inside `buildClientAPIAndSessionStores` helper function or not.
Or am I missing your point?

cmd/daemon/init_session_stores_test.go
thegravi: The latter assertion is based on this code.

cmd/daemon/init_session_stores_test.go
thegravi: Nope, see other comment below.

cmd/daemon/init_session_stores_test.go
thegravi: I tried to make them reusable in both places, but it gets messy and requires extra configuration on both ends. Therefore, I left it as a separate mockup.

session/trusted_pass_session_store.go
thegravi: This one was left without latest update just like for other stores. Now this function does not have any arguments.

session/trusted_pass_session_store.go
thegravi: It's changes. Force renewal bypasses validation and tries to update.

session/trusted_pass_session_store_silent_renewal_test.go
thegravi: It gets too complex when placed in the table.

devzbysiu (approved):

cmd/daemon/main_test.go
DoYouEvenCpp: My gripe was about testing what we initialise in `buildClientAPIAndSessionStores` in the daemons' main. And since  the initialisation process runs always the same, I assumed we should have here all non nils - right? (otherwise we'd be in a bit of pickle, thus my other comment about asserting/raising an error directly in the production code)

cmd/daemon/init_session_stores.go
DoYouEvenCpp: I guess it's all good in the end. :)

PR #1043, opened 2025-08-22
LVPN-8953: Add early config load on start

keliramu:
Changes:
- try to load config early from disk, if available, for libtelio to have config on initialization;
- fix rollout apply condition: case then feature target rollout value is missing and is 0 - this means good for all;


edvyno1 (approved):

cmd/daemon/main.go
```diff
@@ -373,6 +373,9 @@ func main() {
 		rolloutGroup,
 	)
 
+	// try to load config from disk if it was previously downloaded
+	rcConfig.LoadConfigFromDisk()
```
thegravi: And if it fails, despite the log message, would the app be aware, or would it just use `default` one?

config/remote/remote.go
```diff
@@ -30,6 +30,7 @@ type FeatureConfig interface {
 
 type ConfigLoader interface {
 	LoadConfig() error
+	LoadConfigFromDisk()
```
thegravi: `LoadConfig` for remote (+fallback local) has this error information. What is the purpose of not having for local one?

cmd/daemon/main.go
keliramu: If it fails it should not break anything - that's the goal. There can be legit situation when there is nothing to load from disk - either not exists, or exists garbage. In following steps download from remote server will be attempted and valid config will be downloaded and saved for next time to load from disk.

config/remote/remote.go
keliramu: If it fails - it should not impact anything, no need to return error. Just needed new method for this interface (maybe naming could be improved). Alternative would be to add `bool` parameter into existing one - but that would create unnecessary tangling and not easy to read/comprehend, and would be prone to future human errors.

config/remote/remote.go
```diff
@@ -215,18 +220,20 @@ func (c *CdnRemoteConfig) reportLoadError(featureName string, err error) {
...
+				c.reportLoadError(feature.name, err)
+				log.Printf("%s failed loading feature [%s] config from the disk: %s\n", internal.ErrorPrefix, feature.name, err)
+			}
 			continue
 		}
-		log.Println(internal.InfoPrefix, "feature [", feature.name, "] config loaded from:", c.localCachePath)
+		log.Printf("%s feature [%s] config loaded from: %s\n", internal.InfoPrefix, feature.name, c.localCachePath)
 		c.analytics.EmitLocalUseEvent(ClientCli, feature.name, nil)
 	}
 }
```
thegravi:
```suggestion
func (c *CdnRemoteConfig) loadSilent() {
	c.mu.Lock()
	defer c.mu.Unlock()
	c.doLoad(false)
}

func (c *CdnRemoteConfig) loadStrict() { // or loadAndReport
	c.mu.Lock()
	defer c.mu.Unlock()
	c.doLoad(true)
}

func (c *CdnRemoteConfig) doLoad(reportErrors bool) {
	for _, f := range c.features.keys() {
		feature := c.features.get(f)
		if err := feature.load(c.localCachePath, jsonFileReaderWriter{}, jsonValidator{}); err != nil {
			if reportErrors {
				c.reportLoadError(feature.name, err)
				log.Printf("%s failed loading feature [%s] config from the disk: %s\n", internal.ErrorPrefix, feature.name, err)
			}
			continue
		}
		log.Printf("%s feature [%s] config loaded from: %s\n", internal.InfoPrefix, feature.name, c.localCachePath)
		c.analytics.EmitLocalUseEvent(ClientCli, feature.name, nil)
	}
}```

**Review by thegravi — APPROVED** [review]

File: config/remote/remote.go [hunk 4]
```diff
@@ -220,10 +220,21 @@ func (c *CdnRemoteConfig) reportLoadError(featureName string, err error) {
...
+}
+
+func (c *CdnRemoteConfig) load() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c._load(true)
+}
+
+func (c *CdnRemoteConfig) _load(reportErrors bool) {
```
thegravi: [nit] Do we want to prefix the function with an underscore? I saw some places where `do<sometihng>` pattern was used with encapsulating functions. Shall we use it here as well?

keliramu: For a moment thought it would be good idea, but changed it to your suggested name.

PR #1046, opened 2025-08-22
Lvpn 8508 recent connections sync support

thegravi:


daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,183 @@
...
+// New connections are placed at the beginning of the store
+func (r *RecentConnectionsStore) Add(conn VPNConnection) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if err := r.checkExistence(); err != nil {
+		return err
+	}
+
+	connections, err := r.loadLocked()
```
mariusSincovici: [small] In general external functions make the lock and private ones don't. Would it work also here the same logic? and the `loadLocked` can be changed to `load`? :smile:

daemon/recents/model.go
```diff
@@ -0,0 +1,12 @@
+package recents
+
+import "github.com/NordSecurity/nordvpn-linux/config"
+
+type Model struct {
+	Country            string
```
mariusSincovici: wouldn't make sense to have also the `json:`, like for other structs?

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,183 @@
...
+	"sync"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+const (
+	// maxRecentConnections defines the maximum number of recent connections to store
+	maxRecentConnections = 10
+	labelUnidentified    = "Unidentified"
```
mariusSincovici: Displaying this to the user would not bring too much gain. If the app is not able to construct maybe it should not display anything instead?

daemon/rpc_connect.go
```diff
@@ -275,6 +277,16 @@ func (r *RPC) connect(
 	event.EventStatus = events.StatusSuccess
 	event.DurationMs = getElapsedTime(connectingStartTime)
 	r.events.Service.Connect.Publish(event)
+	r.recentVPNConnStore.Add(recents.NewVPNConnection(
+		recents.Model{
+			Country:            country.Name,
```
mariusSincovici: This should store and/or the country code.

daemon/rpc_connect.go
```diff
@@ -275,6 +277,16 @@ func (r *RPC) connect(
...
+			City:               city,
+			SpecificServer:     strings.Split(server.Hostname, ".")[0],
+			SpecificServerName: server.Name,
+			Group:              event.TargetServerGroup,
```
mariusSincovici: For group there is an enum for groups that can be used instead when sent to the clients. Couldn't that be sent instead?
Or wouldn't make sense to store also the `parameters := GetServerParameters(` and then the exact same parameters would be given to reconnect?

daemon/rpc_get_recent_connections.go
```diff
@@ -0,0 +1,42 @@
...
+	for i, v := range values {
+		conn := &v.Connection
+		rcValues[i] = &pb.RecentConnection{
+			ConnectionModel: &pb.RecentConnectionModel{
+				Country:            conn.Country,
+				City:               conn.City,
+				SpecificServer:     conn.SpecificServer,
+				SpecificServerName: conn.SpecificServerName,
+				Group:              conn.Group,
+				ConnectionType:     pb.ServerSelectionRule(conn.ConnectionType),
```
mariusSincovici: this would force the 2 enums to have the same values otherwise the conversion will not work correctly.
Wouldn't make sense to only keep one enum, the one from proto definition?

internal/constants.go
```diff
@@ -120,9 +120,10 @@ var (
 
 	DatFilesPath = filepath.Join(AppDataPath, "data")
 
-	DatFilesPathCommon    = filepath.Join(AppDataPathCommon, "data")
-	ConfigFilesPathCommon = filepath.Join(AppDataPath, "conf")
-	StaticConfigFilename  = filepath.Join(DatFilesPathCommon, "install_static.dat")
+	DatFilesPathCommon           = filepath.Join(AppDataPathCommon, "data")
+	ConfigFilesPathCommon        = filepath.Join(DatFilesPath, "conf")
+	StaticConfigFilename         = filepath.Join(DatFilesPathCommon, "install_static.dat")
+	RecentVPNConnectionsFilename = filepath.Join(DatFilesPathCommon, "recent_conn.dat")
```
mariusSincovici: [small] why not entire word for connections, `recent_connections`? :smile:

tray/tray.go
```diff
@@ -21,6 +23,7 @@ const (
 	NotifierStartDelay        = 3 * time.Second
 	PollingUpdateInterval     = 5 * time.Second
 	PollingFullUpdateInterval = 60 * time.Second
+	CountryListUpdateInterval = 60 * time.Minute
```
mariusSincovici, line 27: this is not going to work if obfuscate option is enabled/disabled

tray/tray.go
```diff
@@ -62,6 +65,51 @@ func (ai *accountInfo) reset() {
...
+		cp.mu.Unlock()
+		return out, nil
+	}
+	cp.mu.Unlock()
+
+	resp, err := client.Countries(context.Background(), &pb.Empty{})
+	if err != nil {
+		return nil, err
+	}
+	result := sortedConnections(resp.Servers)
```
mariusSincovici, line 89: Aren't the countries already sorted by the daemon?

tray/tray.go
```diff
@@ -62,6 +65,51 @@ func (ai *accountInfo) reset() {
...
+
+	cp.mu.Lock()
+	cp.countries = result
+	cp.countriesUpdateTime = time.Now()
+	out := append([]string(nil), cp.countries...)
```
mariusSincovici: [small] `slices.Clone` might be a bit more clear?

tray/monitor.go
```diff
@@ -249,6 +272,7 @@ func (ti *Instance) pollingMonitor() {
 					ti.redraw(ti.updateAccountInfo())
 				}
 				ti.redraw(ti.updateVpnStatus())
+				ti.redraw(ti.updateCountryList())
```
mariusSincovici: wouldn't be sufficient to have a single `redraw(ti.updateVpnStatus() | ti.updateCountryList())`? So it prevents from coping the servers list and doing all the comparisons?

tray/menu.go
```diff
@@ -89,69 +142,233 @@ func addDaemonErrorSection(ti *Instance) {
...
+	countries []string,
+	addedConnections map[string]bool,
+) {
+	parent.AddSubMenuItem(labelCountries, tooltipCountries).Disable()
+	for _, country := range countries {
+		if addedConnections[country] {
+			continue
+		}
+
+		title := strings.ReplaceAll(country, "_", " ")
```
mariusSincovici, line 257: country code can be used instead

tray/menu.go
```diff
@@ -89,69 +142,233 @@ func addDaemonErrorSection(ti *Instance) {
...
+
+		if ti.connect(country, "") {
 			ti.updateChan <- true
-		}()
+		}
 	}
-	systray.AddSeparator()
+}
+
+func connectByConnectionModel(
```
mariusSincovici: This will not work for countries/cities that contain space in their names.

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,183 @@
...
+	connections = slices.Insert(connections, 0, conn)
+	if len(connections) > maxRecentConnections {
+		connections = connections[:maxRecentConnections]
+	}
+
+	return r.saveLocked(connections)
+}
+
+// Clean removes all stored connection information
+func (r *RecentConnectionsStore) Clean() error {
```
mariusSincovici, line 93: When are the recent connections deleted?

protobuf/daemon/recent_connections.proto
```diff
@@ -0,0 +1,54 @@
...
+  // List of recent connections information
+  repeated RecentConnection connections = 1;
+}
+
+message RecentConnection {
+  // Connection model
+  RecentConnectionModel connection_model = 1;
+
+  // Human readable connection label
+  string display_label = 2;
```
mariusSincovici: Maybe this should not be part of the response, because this format can only be used by the tray. If in the future other clients(GUI+CLI) will use this will not be usable, unless future formatting.

protobuf/daemon/service.proto
```diff
@@ -22,50 +22,79 @@ import "purchase.proto";
...
 import "servers.proto";
 import "defaults.proto";
+import "recent_connections.proto";
 
 service Daemon {
+  // ==================== Authentication ====================
+  rpc IsLoggedIn(Empty) returns (IsLoggedInResponse);
+  rpc LoginWithToken(LoginWithTokenRequest) returns (LoginResponse);
+  rpc LoginOAuth2(LoginOAuth2Request) returns (LoginOAuth2Response);
+  rpc LoginOAuth2Callback(LoginOAuth2CallbackRequest)
```
mariusSincovici, line 32: [small] any reason it is on a new line the return?
Same for other places

daemon/rpc_get_recent_connections.go
thegravi: This is the plan in the later PR, did not want to bloat it with unrelated changes to systray core implementation.

protobuf/daemon/service.proto
thegravi: Applied protobuf syntax formatter

tray/menu.go
thegravi: I agree, this would solve several problems at once.

daemon/recents/recent_connections.go
thegravi: Sure

daemon/rpc_connect.go
thegravi: ~~If we ditch specific servers, then I could tailor it~~
Couldn't due to `import cycle`

tray/tray.go
thegravi: Nope.

tray/monitor.go
thegravi: I went a little bit further and optimized all renderings.

daemon/recents/recent_connections.go
thegravi: I like to always provide a full lifecycle - if you can create, then you can delete. In this case, currently, there are no options unless the GUI implements them.

protobuf/daemon/recent_connections.proto
thegravi: totally agree, changed it

daemon/recents/recent_connections.go
thegravi: Gonna add during 'logout' and 'set defaults' operations.

tray/menu.go
thegravi: It's a list of country strings. We don't have codes here.

mariusSincovici: yeah, right missed that you actually remove the `_` to have space

cmd/daemon/main.go
```diff
@@ -373,6 +374,9 @@ func main() {
 		rolloutGroup,
 	)
 
+	// try to load config from disk if it was previously downloaded
+	rcConfig.TryPreload()
```
mariusSincovici: is this needed? or it was added after merging main to branch?

daemon/recents/model.go
```diff
@@ -0,0 +1,15 @@
+package recents
+
+import (
+	"github.com/NordSecurity/nordvpn-linux/config"
+)
+
+type Model struct {
+	Country            string                     `json:"country"`
```
mariusSincovici, line 8: Is it country or country code?

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+// Get retrieves all stored VPN connections from the store
+func (r *RecentConnectionsStore) Get() ([]Model, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if err := r.checkExistence(); err != nil {
+		return nil, fmt.Errorf("%s %s getting recent connections: %w\n", logTag, internal.ErrorPrefix, err)
```
mariusSincovici: The error format looks more like a print. Probably it should not include `ErrorPrefix` and \n inside. The build in linter is also marking this as warning :).

Shouldn't the the error level be something that the caller decides? Because maybe it is a recoverable error and it will be logged as warning or something else.

daemon/rpc_get_recent_connections.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// GetRecentConnections retrieves recent vpn connections from store
+func (r *RPC) GetRecentConnections(
+	ctx context.Context,
+	in *pb.RecentConnectionsRequest,
+) (*pb.RecentConnectionsResponse, error) {
+	values, err := r.recentVPNConnStore.Get()
+	if err != nil {
+		return nil, fmt.Errorf("%s getting recent vpn connections: %w", internal.ErrorPrefix, err)
```
mariusSincovici: :thinking:  this adds another set of [error] to the errors that already have [error]

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%s %s getting recent vpn connections: %w", logTag, internal.ErrorPrefix, err)
+	}
+
+	return conns, nil
+}
+
+func (r *RecentConnectionsStore) find(model Model, list []Model) int {
+	return slices.IndexFunc(list, func(m Model) bool {
+		return m.Country == model.Country &&
```
mariusSincovici: doesn't it work `m == model`?

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (r *RecentConnectionsStore) load() ([]Model, error) {
+	data, err := r.fsHandle.ReadFile(r.path)
+	if err != nil {
+		return nil, fmt.Errorf("reading recent connections store: %w", err)
+	}
+
+	var connections []Model
+	if err := json.Unmarshal(data, &connections); err != nil {
```
mariusSincovici, line 124: If Unmarshal fails, shouldn't the app just delete the file and start from scratch? :thinking: 

Because if it fails once to unmarshal it will probably never work so then it will never be able to add/read the recent connections ever, unless file is deleted manually.

daemon/recents/recent_connections.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *RecentConnectionsStore) save(values []Model) error {
+	data, err := json.Marshal(values)
+	if err != nil {
+		return fmt.Errorf("marshaling vpn connections store: %w", err)
+	}
+
+	if err := r.fsHandle.WriteFile(r.path, data, internal.PermUserRW); err != nil {
+		return fmt.Errorf("writing vpn connections store: %w", err)
+	}
+	defer log.Println("saved recent conns:", values)
```
mariusSincovici: why `defer` since is the last line in the file? Also could the log have the tag + prefix?

daemon/rpc_get_recent_connections.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	if in.Limit > 0 && int(in.Limit) < len(values) {
+		values = values[:in.Limit]
+	}
+
+	rcValues := make([]*pb.RecentConnectionModel, len(values))
+	for i, v := range values {
+		rcValues[i] = &pb.RecentConnectionModel{
+			Country:            v.Country,
```
mariusSincovici, line 31: The country code is not set?

protobuf/daemon/recent_connections.proto
```diff
@@ -0,0 +1,49 @@
...
+  // Specific server identifier
+  string specific_server = 6;
+
+  // Connection type enum
+  ServerSelectionRule connection_type = 7;
+}
+
+message RecentConnectionsRequest {
+  // Limit maximum entries to be returned, 0 - no limit
+  int64 limit = 1;
```
mariusSincovici: Since this is kind of optional, wouldn't be good to mark it as `optional`?

daemon/rpc_logout.go
```diff
@@ -109,6 +110,10 @@ func (r *RPC) Logout(ctx context.Context, in *pb.LogoutRequest) (payload *pb.Pay
 		return nil, err
 	}
 
+	if err := r.recentVPNConnStore.Clean(); err != nil {
+		return nil, fmt.Errorf("executing logout operation: %w\n", err)
```
mariusSincovici: Maybe this should only be a log message? 
Because here the user is already logged out. And if an error is displayed, then at the next attempt to logout will stop at the beginning of the function saying user is not logged in.

daemon/rpc_set_defaults.go
```diff
@@ -50,6 +51,12 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
 		}, nil
 	}
 
+	if err := r.recentVPNConnStore.Clean(); err != nil {
+		return &pb.Payload{
+			Type: internal.CodeFailure,
```
mariusSincovici: This should be a new error code without `fmt.Errorf("resetting to defaults: %w\n", err)`, because:
* on the CLI the error message will be displayed to the user directly
* on the GUI this results on an exception and it will be very hard to convert the dynamic error message to some predefined user message that needs to be displayed on the GUI.

tray/menu.go
```diff
@@ -89,69 +137,179 @@ func addDaemonErrorSection(ti *Instance) {
...
+		if !open {
+			return
 		}
+		if ti.connect("", "") {
+			ti.updateChan <- true
+		}
+	}
+}
+
+func addConnectionSelector(ti *Instance) {
```
mariusSincovici: [tiny :smile: ] Wouldn't be more appropriate something like `AddConnectToItem`?

tray/recents.go
```diff
@@ -0,0 +1,116 @@
...
+	switch model.ConnectionType {
+	case config.ServerSelectionRule_RECOMMENDED:
+		return ti.connect("", "")
+
+	case config.ServerSelectionRule_CITY:
+		city_str := strings.ReplaceAll(model.City, " ", "_")
+		return ti.connect(city_str, "")
+
+	case config.ServerSelectionRule_COUNTRY:
+		country_str := strings.ReplaceAll(model.Country, " ", "_")
```
mariusSincovici: Country code can be used instead.

tray/recents.go
```diff
@@ -0,0 +1,116 @@
...
+	case config.ServerSelectionRule_SPECIFIC_SERVER:
+		return ti.connect(model.SpecificServer, "")
+
+	case config.ServerSelectionRule_GROUP:
+		group_str := strings.ReplaceAll(model.Group.String(), "_", " ")
+		return ti.connect("", group_str)
+
+	case config.ServerSelectionRule_COUNTRY_WITH_GROUP:
+		group_str := strings.ReplaceAll(model.Group.String(), "_", " ")
+		country_str := strings.ReplaceAll(model.Country, " ", "_")
```
mariusSincovici: same here, country code?

daemon/rpc_connect.go
```diff
@@ -276,13 +278,39 @@ func (r *RPC) connect(
 	event.DurationMs = getElapsedTime(connectingStartTime)
 	r.events.Service.Connect.Publish(event)
 
+	if isRecentConnectionSupported(event.TargetServerSelection) {
+		recentModel := recents.Model{
+			CountryCode:        event.TargetServerCountryCode,
+			Country:            event.TargetServerCountry,
+			City:               event.TargetServerCity,
+			SpecificServer:     strings.Split(event.TargetServerDomain, ".")[0],
```
mariusSincovici: Normally this should never happen, but wouldn't be  be good to check if [0] exists, just in case something breaks into the servers list?

tray/recents.go
```diff
@@ -0,0 +1,116 @@
...
+		return ""
+
+	case config.ServerSelectionRule_COUNTRY:
+		return conn.Country
+
+	case config.ServerSelectionRule_SPECIFIC_SERVER:
+		return conn.SpecificServerName
+
+	case config.ServerSelectionRule_GROUP:
+		return strings.ReplaceAll(conn.Group.String(), "_", " ")
```
mariusSincovici, line 37: It might look nicer if there would be a map with the values defined for each group type for display. Because now they are all upper case.

tray/tray.go
mariusSincovici: later PR

tray/recents.go
thegravi: This is done in following PR where grpc state stream is used instead if polling.

daemon/recents/model.go
thegravi: country code is lower in the list

daemon/recents/recent_connections.go
thegravi: Correct, I went with an inertia when writing log /error messages :). Gonna change these.

daemon/recents/recent_connections.go
thegravi: It does. Previously, I had a non-built-in type, and it was unable to implicitly compare, but after removing it, I left this part unchanged.

daemon/recents/recent_connections.go
thegravi: It was a part of the testing code, removing it.

edvyno1 (approved):

PR #1048, opened 2025-08-25
Example mockcdn for local and CI tests

edvyno1:


ci/docker/mockcdn/dl_non_rc_files.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mariusSincovici: [small] it might be more portable `#!/usr/bin/env bash`

ci/docker/mockcdn/server.py
```diff
@@ -0,0 +1,10 @@
+import http.server
+import socketserver
+
+PORT = 8000
```
mariusSincovici: [very small] wouldn't work to set in docker file the port number as env variable?

ci/qa_run_tests.sh
```diff
@@ -52,6 +52,15 @@ mkdir -p "${LOGS_FOLDER}"
 mkdir -p "${ARTIFACTS_FOLDER}"
 mkdir -p "${GOCOVERDIR}"
 
+if [[ -n ${NORD_CDN_URL:-} ]]; then
```
mariusSincovici: This is deb specific change. Could you add it instead into `test_deb.sh` file?

ci/docker/mockcdn/dl_non_rc_files.sh
edvyno1: applied

ci/docker/mockcdn/server.py
edvyno1: changed and looks to be working, will not rebuild and repush the container as there is no need for changing ports as of now

ci/qa_run_tests.sh
edvyno1: moved

PR #1052, opened 2025-08-26
LVPN-8994: Improve static config

keliramu:
Issues addressed:
- do not do init twice;
- do not report error if on first start file is not present yet (which is normal case, not error);

### DIFF
--- a/config/static_config.go
+++ b/config/static_config.go
@@ -23,7 +23,7 @@ const (
 )
 
 type StaticConfig struct {
-	RolloutGroup int `json:"rollout_group,omitempty"`
+	RolloutGroup int `json:"rollout_group"`
 }
 
 // StaticConfigManager stores values which remain constant throughout app's lifetime
@@ -45,19 +45,29 @@ func NewFilesystemStaticConfigManager() *FilesystemStaticConfigManager {
 }
 
 // loadConfig reads the config from disk
-func (s *FilesystemStaticConfigManager) loadConfig() (StaticConfig, error) {
-	data, err := s.fs.ReadFile(internal.StaticConfigFilename)
+func (s *FilesystemStaticConfigManager) loadConfig() (sc StaticConfig, err error) {
+	defer func() {
+		if err != nil {
+			// in case read/load fails, try to write default/empty values
+			if data, err := json.Marshal(StaticConfig{}); err == nil {
+				s.fs.WriteFile(internal.StaticConfigFilename, data, internal.PermUserRW)
+			}
+		}
+	}()
+
+	var data []byte
+	data, err = s.fs.ReadFile(internal.StaticConfigFilename)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
 			// file doesn't exist yet, return empty config
 			return StaticConfig{}, nil
 		}
-		return StaticConfig{}, fmt.Errorf("reading config file: %w", err)
+		return StaticConfig{}, fmt.Errorf("reading static config file: %w", err)
 	}
 
 	var cfg StaticConfig
-	if err := json.Unmarshal(data, &cfg); err != nil {
-		return StaticConfig{}, fmt.Errorf("unmarshaling config: %w", err)
+	if err = json.Unmarshal(data, &cfg); err != nil {
+		return StaticConfig{}, fmt.Errorf("unmarshaling static config: %w", err)
 	}
 
 	return cfg, nil
@@ -67,11 +77,11 @@ func (s *FilesystemStaticConfigManager) loadConfig() (StaticConfig, error) {
 func (s *FilesystemStaticConfigManager) saveConfig(cfg StaticConfig) error {
 	data, err := json.Marshal(cfg)
 	if err != nil {
-		return fmt.Errorf("marshaling config: %w", err)
+		return fmt.Errorf("marshaling static config: %w", err)
 	}
 
 	if err := s.fs.WriteFile(internal.StaticConfigFilename, data, internal.PermUserRW); err != nil {
-		return fmt.Errorf("writing config file: %w", err)
+		return fmt.Errorf("writing static config file: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


config/static_config.go
```diff
@@ -51,10 +51,11 @@ type FilesystemStaticConfigManager struct {
 func tryInitStaticConfig(fs FilesystemHandle) (StaticConfig, configState) {
```
thegravi: [small] What if we accumulate logs and return them? Later, we can entirely ditch them - not logging. And we can also wrap the underlying error when reading a file, since it can be a permission issue, non-existence, or an IO error.

thegravi (approved):

config/static_config.go
keliramu: plz check whole logic, there is a state management logic implemented, and on first init only `file not-exists` error is suppressed and other errors are reported

thegravi: That is why I left that comment. What I mean is that we could refine the error message even more by adding the sequence where this try-init was invoked: during ctor or later when reading data. It is your call to choose.

mariusSincovici: The init logic should be simplified and when the file doesn't exists or there is an error to read/unmarshal it, then the file should be recreated with the default options.

keliramu: Simplified whole logic, plz recheck.

keliramu: Simplified whole logic, plz recheck.

config/static_config.go
```diff
@@ -4,103 +4,86 @@ import (
...
+	data, err := s.fs.ReadFile(internal.StaticConfigFilename)
 	if err != nil {
-		log.Println(internal.ErrorPrefix, "failed to load static config:", err)
 		if errors.Is(err, os.ErrNotExist) {
-			return StaticConfig{}, staticConfigState_noFile
+			// file doesn't exist yet, return empty config
+			return StaticConfig{}, nil
 		}
-		return StaticConfig{}, staticConfigState_failedToInitialize
+		return StaticConfig{}, fmt.Errorf("reading config file: %w", err)
```
thegravi: [small] Shall we add `static config`?

config/static_config.go
```diff
@@ -4,103 +4,86 @@ import (
...
 	}
 
 	var cfg StaticConfig
-	err = json.Unmarshal(cfgFile, &cfg)
-	if err != nil {
-		log.Println(internal.ErrorPrefix, "failed to unmarshal static config:", err)
-		return cfg, staticConfigState_failedToInitialize
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return StaticConfig{}, fmt.Errorf("unmarshaling config: %w", err)
```
thegravi: [small] Here also - `static config`

I see there are more locations where this could be applied for more detailed logging, which one `config` is used.

config/static_config.go
```diff
@@ -4,103 +4,86 @@ import (
...
 
 func (s *FilesystemStaticConfigManager) GetRolloutGroup() (int, error) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	cfg, err := s.getConfig()
+	cfg, err := s.loadConfig()
 	if err != nil {
-		return 0, fmt.Errorf("failed to read config: %w", err)
+		return 0, err
```
thegravi: [small] I'd like to have a more detailed error message - where the load config error originated from.

config/static_config.go
```diff
@@ -114,33 +97,19 @@ func (s *FilesystemStaticConfigManager) SetRolloutGroup(rolloutGroup int) error
...
-
 	if rolloutGroup < rolloutGroupMin || rolloutGroup > rolloutGroupMax {
 		return ErrRolloutGroupOutOfBounds
 	}
 
-	cfg, err := s.getConfig()
+	cfg, err := s.loadConfig()
 	if err != nil {
-		return fmt.Errorf("failed to read config: %w", err)
+		return err
```
thegravi: Also here, more detailed.

config/static_config.go
```diff
@@ -114,33 +97,19 @@ func (s *FilesystemStaticConfigManager) SetRolloutGroup(rolloutGroup int) error
...
-	}
-
-	err = s.fs.WriteFile(staticConfigFilename, json, internal.PermUserRW)
-	if err != nil {
-		return fmt.Errorf("failed to save static config file: %w", err)
-	}
-
-	s.cfg = cfg
-	return nil
+	return s.saveConfig(cfg)
```
thegravi: [small] Here also.

config/static_config.go
```diff
@@ -4,103 +4,86 @@ import (
...
+func NewFilesystemStaticConfigManager() *FilesystemStaticConfigManager {
+	return &FilesystemStaticConfigManager{
+		fs: StdFilesystemHandle{},
+	}
 }
 
-func tryInitStaticConfig(fs FilesystemHandle) (StaticConfig, configState) {
-	cfgFile, err := fs.ReadFile(staticConfigFilename)
+// loadConfig reads the config from disk
+func (s *FilesystemStaticConfigManager) loadConfig() (StaticConfig, error) {
```
mariusSincovici: Maybe if the file loading fails, for any reason it should be recreated empty. 
Because if for some reason Unmarshal/read fails, then the app will never be able to save anything, until it is manually deleted.

keliramu: ok, make sense, added recovery logic on load error

config/static_config.go
keliramu: done

config/static_config.go
keliramu: done

config/static_config.go
keliramu: done

config/static_config.go
keliramu: done

config/static_config.go
keliramu: done

mariusSincovici (approved):

PR #1059, opened 2025-08-29
LVPN-8653: Add client id metadata on the client side

bartoszWojciechO:


clientid/client.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+// SetMetadataUnaryInterceptor inserts metadata into unary gRPC
+func (i *SetClientIDInterceptor) SetMetadataUnaryInterceptor(ctx context.Context,
+	method string,
+	req, reply any, cc *grpc.ClientConn,
+	invoker grpc.UnaryInvoker,
+	opts ...grpc.CallOption) error {
+	metadataCtx := metadata.AppendToOutgoingContext(ctx, "client-id", strconv.Itoa(int(i.clientID.Number())))
```
devzbysiu: Maybe you can reuse `clientIDMetadataKey` const for those `client-id`s here and below?

cmd/cli/main.go
```diff
@@ -88,14 +90,18 @@ func main() {
 	}
 	log.SetOutput(fileLogger)
 
+	clientIDMetadataInterceptor := clientid.NewInsertClientIDInterceptor(pb.ClientID_CLI)
+
 	loaderInterceptor := cli.LoaderInterceptor{}
 	conn, err := grpc.Dial(
```
DoYouEvenCpp, line 96: [further development] The `grpc.Dial` is marked as deprecated - do we have a plan to start using suggested `NewClient`?

bartoszWojciechO: Yeah, we have a tech debt task for that.

PR #1062, opened 2025-08-29
Move denyDNS rules to filter table

edvyno1:


networker/networker.go
```diff
@@ -752,6 +752,14 @@ func (netw *Combined) setAllowlist(allowlist config.Allowlist) error {
 			}
 		}
 	}
+	// if port 53 is whitelisted - do not add drop-dns rules
+	if !allowlist.Ports.TCP[53] && !allowlist.Ports.UDP[53] {
+		// disable DNS traffic to private LAN ranges - to prevent DNS leaks
+		// when /etc/resolv.conf has nameserver default gateway
+		if err := netw.denyDNS(); err != nil {
+			return err
```
bartoszWojciechO, line 760: Can you wrap err in `fmt.Errof("denying DNS: %w", err)`

bartoszWojciechO (approved):

networker/networker.go
edvyno1: it may be a bit redundant since denyDNS itself is already returning a formatted error:
`return fmt.Errorf("adding deny-private-dns rule to firewall: %w", err)`

mariusSincovici (approved):

PR #1063, opened 2025-08-29
LVPN-8938: introduce automatic security assesments over the repo

DoYouEvenCpp:
As part of the hardening actions on the NordVPN repository, the cyclic security assessments to be introduced via the OpenSSF github actions.

Changes under this PR:
- added scorecards workflow file, that is to get trigger each Monday early in the night hours
- updated the README file with a badge containing results of latest scan results (pointing to a more detailed information)

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -7,6 +7,7 @@ on:
     - cron: '0 3 * * 1'
   push:
     branches: [ main ]
+  workflow_dispatch:
 
 # Declare default permissions as read only.
 permissions: read-all


.github/workflows/scorecards.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-22.04
+    permissions:
+      # Needed if using Code scanning alerts
+      security-events: write
+      # Needed for GitHub OIDC token if publish_results is true
+      id-token: write
+
+    steps:
+    - name: "Checkout code"
+      uses: actions/checkout@v5
```
devzbysiu: I probably forgot to do that, but it would be nice to pin the versions to specific commit instead of `@vX`. Similarly for the rest of third party actions

DoYouEvenCpp: Ok, went back to commit sha based versions then.

mariusSincovici (approved):

PR #1064, opened 2025-08-29
Add missing automated tests

naujalyte:


test/qa/test_connect.py
```diff
@@ -572,3 +572,29 @@ def test_connect_to_dedicated_ip(tech, proto, obfuscated):
...
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES)
+def test_connect_fails_virtual_location_disabled(tech, proto, obfuscated):
+    """Manual TC: LVPN-8533"""
+
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+
+    sh.nordvpn.set("virtual-location", "off")
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        sh.nordvpn(get_alias(), "India")
```
thegravi: [small] If this is very specific to these tests, shall we extract it as a constant for other related tests to be a reference?

mariusSincovici: This should not be hardcoded. Instead there is a helper function to get the list of countries that have only virtual servers and connect to one. 
Something like this is used in other tests:
```
   virtual_countries = lib.get_virtual_countries()
    assert len(virtual_countries) > 0
    country = random.choice(virtual_countries)
```
Maybe it can also be moved into a function that returns a random country, not to copy/paste same 3 lines of code. :)

test/qa/test_connect.py
```diff
@@ -572,3 +572,29 @@ def test_connect_to_dedicated_ip(tech, proto, obfuscated):
...
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.OBFUSCATED_TECHNOLOGIES)
+def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscated):
+    """Manual TC: LVPN-5771"""
+
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+
+    with pytest.raises(sh.ErrorReturnCode_1) as ex:
+        sh.nordvpn(get_alias(), "India")
+
+    assert "The specified server is not available at the moment or does not support your connection settings." in str(ex.value)
```
mariusSincovici: `str(ex.value)` can be problematic if the string is long. It will not find the message because when it is converted to `str` it adds `...` into the middle.
It should be replaced with something else.

test/qa/test_settings.py
```diff
@@ -266,11 +266,21 @@ def test_set_analytics_starts_prompt_even_if_completed_before():
...
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
+def test_set_defaults_no_logout(tech, proto, obfuscated):
+    """Manual TC: LVPN-9029"""
+
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+
+    sh.nordvpn.set("virtual-location", "off")
+    sh.nordvpn.set("lan-discovery", "on")
+
+    assert not settings.is_virtual_location_enabled()
```
mariusSincovici, line 278: each of this lines will do an `nordvpn settings`. Is it intended? Or can it be just a `Settings()` object created and then reused in the entire function.

test/qa/lib/settings.py
```diff
@@ -127,6 +127,11 @@ def is_dns_disabled():
     return Settings().get("DNS") == "disabled"
 
 
+def is_tray_enabled():
```
mariusSincovici, line 130: [small] This functions are useful, but they might be a bit misused, because each of them will do a `nordvpn settings`. And in some tests this functions are called one after the other resulting in lots of commands.

test/qa/test_connect.py
naujalyte: Updated

test/qa/test_connect.py
naujalyte: Updated

test/qa/test_settings.py
naujalyte: Will be refactored in separate PR

test/qa/lib/settings.py
naujalyte: Will be refactored in separate PR

mariusSincovici (approved):

test/qa/lib/settings.py
```diff
@@ -127,6 +127,11 @@ def is_dns_disabled():
...
+    """Returns True, if Tray is enabled in application settings."""
```
DoYouEvenCpp, line 131:
```suggestion
    """Returns True, if Tray is enabled in application settings. False otherwise."""
```

PR #1065, opened 2025-08-29
LVPN-8889 implement automation tests for rc feature part 2

pvraketa:
Added:
 * new test cases
 * new fixtures 
 * new constants for test cases
Changed:
* Increased timeout of execution
* stop/start/restart daemon methods

### DIFF
--- a/daemon/rpc.go
+++ b/daemon/rpc.go
@@ -61,6 +61,7 @@ type RPC struct {
 	consentChecker      ConsentChecker
 	recentVPNConnStore  *recents.RecentConnectionsStore
 	dataUpdateEvents    *daemonevents.DataUpdateEvents
+	initialLoginType    pb.LoginType // memorize what action started: Login or Signup (Register)
 	pb.UnimplementedDaemonServer
 }
 
--- a/daemon/rpc_cities_test.go
+++ b/daemon/rpc_cities_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/events/subs"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
+	coremock "github.com/NordSecurity/nordvpn-linux/test/mock/core"
 	"github.com/NordSecurity/nordvpn-linux/test/mock/networker"
 	testnorduser "github.com/NordSecurity/nordvpn-linux/test/mock/norduser/service"
 	"github.com/stretchr/testify/assert"
@@ -106,9 +108,9 @@ func TestRPCCities(t *testing.T) {
 				dm:        dm,
 				norduser:  &testnorduser.MockNorduserCombinedService{},
 				netw:      &networker.Mock{},
-				ncClient:  mockNC{},
+				ncClient:  &mock.NotificationClientMock{},
 				publisher: &subs.Subject[string]{},
-				api:       mockApi{},
+				api:       &coremock.CredentialsAPIMock{},
 			}
 
 			payload, _ := rpc.Cities(context.Background(), &pb.CitiesRequest{Country: test.countryName})
--- a/daemon/rpc_connect_test.go
+++ b/daemon/rpc_connect_test.go
@@ -30,16 +30,6 @@ type mockRPCServer struct {
 
 func (m *mockRPCServer) Send(p *pb.Payload) error { m.msg = p; return nil }
 
-type mockAuthenticationAPI struct{}
-
-func (mockAuthenticationAPI) Login(bool) (string, error) {
-	return "", nil
-}
-
-func (mockAuthenticationAPI) Token(string) (*core.LoginResponse, error) {
-	return nil, nil
-}
-
 // deterministicServersAPI provides deterministic server selection for connect tests
 // while reusing the existing serversList() and countriesList() functions
 type deterministicServersAPI struct {
--- a/daemon/rpc_countries_test.go
+++ b/daemon/rpc_countries_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/events/subs"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
+	coremock "github.com/NordSecurity/nordvpn-linux/test/mock/core"
 	"github.com/NordSecurity/nordvpn-linux/test/mock/networker"
 	testnorduser "github.com/NordSecurity/nordvpn-linux/test/mock/norduser/service"
 	"github.com/stretchr/testify/assert"
@@ -81,9 +83,9 @@ func TestRPCCountries(t *testing.T) {
 				dm:        dm,
 				norduser:  &testnorduser.MockNorduserCombinedService{},
 				netw:      &networker.Mock{},
-				ncClient:  mockNC{},
+				ncClient:  &mock.NotificationClientMock{},
 				publisher: &subs.Subject[string]{},
-				api:       mockApi{},
+				api:       &coremock.CredentialsAPIMock{},
 			}
 			payload, _ := rpc.Countries(context.Background(), &pb.Empty{})
 
--- a/daemon/rpc_groups_test.go
+++ b/daemon/rpc_groups_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/events/subs"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
+	coremock "github.com/NordSecurity/nordvpn-linux/test/mock/core"
 	"github.com/NordSecurity/nordvpn-linux/test/mock/networker"
 	testnorduser "github.com/NordSecurity/nordvpn-linux/test/mock/norduser/service"
 	"github.com/stretchr/testify/assert"
@@ -47,9 +49,9 @@ func TestRPCGroups(t *testing.T) {
 				dm:        test.dm,
 				norduser:  &testnorduser.MockNorduserCombinedService{},
 				netw:      &networker.Mock{},
-				ncClient:  mockNC{},
+				ncClient:  &mock.NotificationClientMock{},
 				publisher: &subs.Subject[string]{},
-				api:       mockApi{},
+				api:       &coremock.CredentialsAPIMock{},
 			}
 			payload, _ := rpc.Groups(context.Background(), &pb.Empty{})
 
@@ -125,9 +127,9 @@ func TestRPCGroups_Successful(t *testing.T) {
 				dm:        dm,
 				norduser:  &testnorduser.MockNorduserCombinedService{},
 				netw:      &networker.Mock{},
-				ncClient:  mockNC{},
+				ncClient:  &mock.NotificationClientMock{},
 				publisher: &subs.Subject[string]{},
-				api:       mockApi{},
+				api:       &coremock.CredentialsAPIMock{},
 			}
 			payload, _ := rpc.Groups(context.Background(), &pb.Empty{})
 
--- a/daemon/rpc_login.go
+++ b/daemon/rpc_login.go
@@ -155,6 +155,8 @@ func (r *RPC) LoginOAuth2(ctx context.Context, in *pb.LoginOAuth2Request) (*pb.L
 	if in.GetType() == pb.LoginType_LoginType_SIGNUP {
 		eventType = events.LoginSignUp
 	}
+	// memorize what login type started: Login or Signup
+	r.initialLoginType = in.GetType()
 
 	r.events.User.Login.Publish(events.DataAuthorization{
 		DurationMs:   -1,
@@ -205,10 +207,11 @@ func (r *RPC) LoginOAuth2Callback(ctx context.Context, in *pb.LoginOAuth2Callbac
 			eventStatus = events.StatusFailure
 		}
 		r.events.User.Login.Publish(events.DataAuthorization{
-			DurationMs:   max(int(time.Since(lastLoginAttemptTime).Milliseconds()), 1),
-			EventTrigger: events.TriggerUser,
-			EventStatus:  eventStatus,
-			EventType:    loginType,
+			DurationMs:                 max(int(time.Since(lastLoginAttemptTime).Milliseconds()), 1),
+			EventTrigger:               events.TriggerUser,
+			EventStatus:                eventStatus,
+			EventType:                  loginType,
+			IsAlteredFlowOnNordAccount: in.GetType() != r.initialLoginType,
 		})
 		lastLoginAttemptTime = time.Time{}
 	}()
--- a/daemon/rpc_login_test.go
+++ b/daemon/rpc_login_test.go
@@ -1,0 +1,164 @@
+package daemon
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	daemonevents "github.com/NordSecurity/nordvpn-linux/daemon/events"
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/events/subs"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
+	testauth "github.com/NordSecurity/nordvpn-linux/test/mock/auth"
+	testcore "github.com/NordSecurity/nordvpn-linux/test/mock/core"
+)
+
+func TestLoginEvents(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	tests := []struct {
+		name            string
+		setup           func(*daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC
+		loginType1      pb.LoginType
+		expectedStatus1 pb.LoginStatus
+		expectedEvent1  events.DataAuthorization
+		expectedError1  bool
+		loginType2      pb.LoginType
+		expectedStatus2 pb.LoginStatus
+		expectedEvent2  events.DataAuthorization
+		expectedError2  bool
+	}{
+		{
+			name: "login attempt then login success",
+			setup: func(em *daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC {
+				return &RPC{
+					consentChecker: &mock.AnalyticsConsentCheckerMock{ConsentCompleted: true},
+					ac:             &testauth.AuthCheckerMock{LoggedIn: false},
+					cm:             mock.NewMockConfigManager(),
+					credentialsAPI: &testcore.CredentialsAPIMock{},
+					events:         &daemonevents.Events{User: &daemonevents.LoginEvents{Login: em}},
+					authentication: &testcore.AuthenticationAPImock{TokenValue: "token"},
+					publisher:      &subs.Subject[string]{},
+					ncClient:       &mock.NotificationClientMock{},
+				}
+			},
+			loginType1:      pb.LoginType_LoginType_LOGIN,
+			expectedStatus1: pb.LoginStatus_SUCCESS,
+			expectedEvent1: events.DataAuthorization{
+				EventType:                  events.LoginLogin,
+				EventTrigger:               events.TriggerUser,
+				EventStatus:                events.StatusAttempt,
+				IsAlteredFlowOnNordAccount: false,
+			},
+			expectedError1:  false,
+			loginType2:      pb.LoginType_LoginType_LOGIN,
+			expectedStatus2: pb.LoginStatus_SUCCESS,
+			expectedEvent2: events.DataAuthorization{
+				EventType:                  events.LoginLogin,
+				EventTrigger:               events.TriggerUser,
+				EventStatus:                events.StatusSuccess,
+				IsAlteredFlowOnNordAccount: false,
+			},
+			expectedError2: false,
+		},
+		{
+			name: "signup attempt then login success",
+			setup: func(em *daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC {
+				return &RPC{
+					consentChecker: &mock.AnalyticsConsentCheckerMock{ConsentCompleted: true},
+					ac:             &testauth.AuthCheckerMock{LoggedIn: false},
+					cm:             mock.NewMockConfigManager(),
+					credentialsAPI: &testcore.CredentialsAPIMock{},
+					events:         &daemonevents.Events{User: &daemonevents.LoginEvents{Login: em}},
+					authentication: &testcore.AuthenticationAPImock{TokenValue: "token"},
+					publisher:      &subs.Subject[string]{},
+					ncClient:       &mock.NotificationClientMock{},
+				}
+			},
+			loginType1:      pb.LoginType_LoginType_SIGNUP,
+			expectedStatus1: pb.LoginStatus_SUCCESS,
+			expectedEvent1: events.DataAuthorization{
+				EventType:                  events.LoginSignUp,
+				EventTrigger:               events.TriggerUser,
+				EventStatus:                events.StatusAttempt,
+				IsAlteredFlowOnNordAccount: false,
+			},
+			expectedError1:  false,
+			loginType2:      pb.LoginType_LoginType_LOGIN,
+			expectedStatus2: pb.LoginStatus_SUCCESS,
+			expectedEvent2: events.DataAuthorization{
+				EventType:                  events.LoginLogin,
+				EventTrigger:               events.TriggerUser,
+				EventStatus:                events.StatusSuccess,
+				IsAlteredFlowOnNordAccount: true,
+			},
+			expectedError2: false,
+		},
+		{
+			name: "login attempt fail - consent not completed",
+			setup: func(em *daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC {
+				return &RPC{
+					consentChecker: &mock.AnalyticsConsentCheckerMock{ConsentCompleted: false},
+				}
+			},
+			loginType1:      pb.LoginType_LoginType_LOGIN,
+			expectedStatus1: pb.LoginStatus_CONSENT_MISSING,
+			expectedError1:  false,
+			loginType2:      pb.LoginType_LoginType_LOGIN,
+			expectedStatus2: pb.LoginStatus_CONSENT_MISSING,
+			expectedError2:  false,
+		},
+		{
+			name: "signup attempt fail - already logged in",
+			setup: func(em *daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC {
+				return &RPC{
+					consentChecker: &mock.AnalyticsConsentCheckerMock{ConsentCompleted: true},
+					ac:             &testauth.AuthCheckerMock{LoggedIn: true},
+				}
+			},
+			loginType1:      pb.LoginType_LoginType_SIGNUP,
+			expectedStatus1: pb.LoginStatus_ALREADY_LOGGED_IN,
+			expectedError1:  false,
+			loginType2:      pb.LoginType_LoginType_LOGIN,
+			expectedError2:  true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			eventsMock := &daemonevents.MockPublisherSubscriber[events.DataAuthorization]{}
+			r := tt.setup(eventsMock)
+			resp1, err := r.LoginOAuth2(context.Background(), &pb.LoginOAuth2Request{Type: tt.loginType1})
+			if tt.expectedError1 {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.True(t, tt.expectedStatus1 == resp1.Status)
+				if resp1.Status == pb.LoginStatus_SUCCESS {
+					assert.True(t, tt.expectedEvent1.EventType == eventsMock.Event.EventType)
+					assert.True(t, tt.expectedEvent1.EventStatus == eventsMock.Event.EventStatus)
+					assert.True(t, tt.expectedEvent1.EventTrigger == eventsMock.Event.EventTrigger)
+					assert.True(t, tt.expectedEvent1.IsAlteredFlowOnNordAccount == eventsMock.Event.IsAlteredFlowOnNordAccount)
+				}
+			}
+			resp2, err := r.LoginOAuth2Callback(context.Background(), &pb.LoginOAuth2CallbackRequest{
+				Token: "exchange-token",
+				Type:  tt.loginType2,
+			})
+			if tt.expectedError2 {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.True(t, tt.expectedStatus2 == resp2.Status)
+				if resp2.Status == pb.LoginStatus_SUCCESS {
+					assert.True(t, tt.expectedEvent2.EventType == eventsMock.Event.EventType)
+					assert.True(t, tt.expectedEvent2.EventStatus == eventsMock.Event.EventStatus)
+					assert.True(t, tt.expectedEvent2.EventTrigger == eventsMock.Event.EventTrigger)
+					assert.True(t, tt.expectedEvent2.IsAlteredFlowOnNordAccount == eventsMock.Event.IsAlteredFlowOnNordAccount)
+				}
+			}
+		})
+	}
+}
--- a/daemon/rpc_logout_test.go
+++ b/daemon/rpc_logout_test.go
@@ -7,33 +7,19 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
-	"github.com/NordSecurity/nordvpn-linux/core"
 	daemonevents "github.com/NordSecurity/nordvpn-linux/daemon/events"
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 	"github.com/NordSecurity/nordvpn-linux/daemon/recents"
 	"github.com/NordSecurity/nordvpn-linux/events"
 	"github.com/NordSecurity/nordvpn-linux/events/subs"
 	"github.com/NordSecurity/nordvpn-linux/internal"
-	"github.com/NordSecurity/nordvpn-linux/nc"
+	"github.com/NordSecurity/nordvpn-linux/test/mock"
 	mockconfig "github.com/NordSecurity/nordvpn-linux/test/mock/config"
 	testcore "github.com/NordSecurity/nordvpn-linux/test/mock/core"
+
 	"github.com/NordSecurity/nordvpn-linux/test/mock/networker"
 	testnorduser "github.com/NordSecurity/nordvpn-linux/test/mock/norduser/service"
 )
-
-type mockNC struct {
-	nc.NotificationClient
-}
-
-func (mockNC) Stop() error  { return nil }
-func (mockNC) Revoke() bool { return true }
-
-type mockApi struct {
-	core.CombinedAPI
-}
-
-func (mockApi) DeleteToken(token string) error { return nil }
-func (mockApi) Logout(token string) error      { return nil }
 
 func TestLogout_Token(t *testing.T) {
 	cfgManagerMock := newMockConfigManager()
@@ -44,7 +30,7 @@ func TestLogout_Token(t *testing.T) {
 		cm:             cfgManagerMock,
 		norduser:       &testnorduser.MockNorduserCombinedService{},
 		netw:           &networker.Mock{},
-		ncClient:       mockNC{},
+		ncClient:       &mock.NotificationClientMock{},
 		publisher:      &subs.Subject[string]{},
 		credentialsAPI: &testcore.CredentialsAPIMock{},
 		events: &daemonevents.Events{
--- a/daemon/rpc_test.go
+++ b/daemon/rpc_test.go
@@ -99,7 +99,7 @@ func testRPC() *RPC {
 		&testcore.CredentialsAPIMock{},
 		testNewCDNAPI(),
 		testNewRepoAPI(),
-		&mockAuthenticationAPI{},
+		&testcore.AuthenticationAPImock{},
 		"1.0.0",
 		daemonevents.NewEventsEmpty(),
 		func(config.Technology) (vpn.VPN, error) {
--- a/events/events.go
+++ b/events/events.go
@@ -172,11 +172,12 @@ const (
 )
 
 type DataAuthorization struct {
-	DurationMs   int
-	EventTrigger TypeEventTrigger
-	EventStatus  TypeEventStatus
-	EventType    TypeLoginType
-	Reason       ReasonCode
+	DurationMs                 int
+	EventTrigger               TypeEventTrigger
+	EventStatus                TypeEventStatus
+	EventType                  TypeLoginType
+	Reason                     ReasonCode
+	IsAlteredFlowOnNordAccount bool
 }
 
 type DataRequestAPI struct {
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -311,11 +311,19 @@ func (s *Subscriber) NotifyLogin(data events.DataAuthorization) error { // regul
 		mooseFn = moose.NordvpnappSendServiceQualityAuthorizationRegister
 	}
 
+	loginFlowAltered := moose.NordvpnappOptBoolNone
+	if data.EventStatus != events.StatusAttempt {
+		loginFlowAltered = moose.NordvpnappOptBoolFalse
+		if data.IsAlteredFlowOnNordAccount {
+			loginFlowAltered = moose.NordvpnappOptBoolTrue
+		}
+	}
+
 	if err := s.response(mooseFn(
 		int32(data.DurationMs),
 		eventTriggerDomainToInternalType(data.EventTrigger),
 		eventStatusToInternalType(data.EventStatus),
-		moose.NordvpnappOptBoolNone,
+		loginFlowAltered,
 		-1,
 		nil,
 	)); err != nil {
--- a/nc/nc_test.go
+++ b/nc/nc_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/events/subs"
 	"github.com/NordSecurity/nordvpn-linux/test/category"
 	cfgmock "github.com/NordSecurity/nordvpn-linux/test/mock"
-	"github.com/NordSecurity/nordvpn-linux/test/mock/core"
+	coremock "github.com/NordSecurity/nordvpn-linux/test/mock/core"
 	ncmock "github.com/NordSecurity/nordvpn-linux/test/mock/nc"
 	"github.com/stretchr/testify/assert"
 
@@ -163,7 +163,7 @@ func TestStartStopNotificationClient(t *testing.T) {
 			Client: &mockMqttClient,
 		}
 
-		credsFetcher := NewCredsFetcher(&core.CredentialsAPIMock{
+		credsFetcher := NewCredsFetcher(&coremock.CredentialsAPIMock{
 			NotificationCredentialsError: test.credentialsFetchError,
 		}, cfgManager)
 		notificationClient := NewClient(&clientBuilderMock,
@@ -243,7 +243,7 @@ func TestConnectionCancellation(t *testing.T) {
 			Client: &mockMqttClient,
 		}
 
-		credsFetcher := NewCredsFetcher(&core.CredentialsAPIMock{
+		credsFetcher := NewCredsFetcher(&coremock.CredentialsAPIMock{
 			NotificationCredentialsError: test.fetchCredentialsErr,
 		}, cfgManager)
 
--- a/test/mock/core/authenticate.go
+++ b/test/mock/core/authenticate.go
@@ -1,0 +1,18 @@
+package core
+
+import "github.com/NordSecurity/nordvpn-linux/core"
+
+type AuthenticationAPImock struct {
+	URL        string
+	TokenValue string
+	LoginError error
+	TokenError error
+}
+
+func (a *AuthenticationAPImock) Login(bool) (string, error) {
+	return a.URL, a.LoginError
+}
+
+func (a *AuthenticationAPImock) Token(string) (*core.LoginResponse, error) {
+	return &core.LoginResponse{Token: a.TokenValue}, a.TokenError
+}
--- a/test/mock/core/core.go
+++ b/test/mock/core/core.go
@@ -6,6 +6,8 @@ import (
 )
 
 type CredentialsAPIMock struct {
+	core.CombinedAPI
+
 	NotificationCredentialsResponse       core.NotificationCredentialsResponse
 	NotificationCredentialsRevokeResponse core.NotificationCredentialsRevokeResponse
 	NotificationCredentialsError          error
@@ -23,7 +25,12 @@ func (c *CredentialsAPIMock) NotificationCredentialsRevoke(appUserID string, pur
 }
 
 func (*CredentialsAPIMock) ServiceCredentials(token string) (*core.CredentialsResponse, error) {
-	return nil, nil
+	return &core.CredentialsResponse{
+		ID:                 1234,
+		Username:           "test-user",
+		Password:           "test-pass",
+		NordlynxPrivateKey: "test-key",
+	}, nil
 }
 
 func (*CredentialsAPIMock) TokenRenew(renewalToken string, idempotencyKey uuid.UUID) (*core.TokenRenewResponse, error) {
--- a/test/mock/notification_client.go
+++ b/test/mock/notification_client.go
@@ -1,0 +1,17 @@
+package mock
+
+type NotificationClientMock struct {
+	StartError   error
+	StopError    error
+	RevokeStatus bool
+}
+
+func (nc *NotificationClientMock) Start() error {
+	return nc.StartError
+}
+func (nc *NotificationClientMock) Stop() error {
+	return nc.StopError
+}
+func (nc *NotificationClientMock) Revoke() bool {
+	return nc.RevokeStatus
+}
--- a/test/mock/userconsent.go
+++ b/test/mock/userconsent.go
@@ -1,0 +1,11 @@
+package mock
+
+type AnalyticsConsentCheckerMock struct {
+	ConsentCompleted bool
+}
+
+func (*AnalyticsConsentCheckerMock) PrepareDaemonIfConsentNotCompleted() {}
+
+func (c *AnalyticsConsentCheckerMock) IsConsentFlowCompleted() bool {
+	return c.ConsentCompleted
+}


ci/test_deb.sh
```diff
@@ -74,7 +74,7 @@ fi
 # Disable the TUI loader indicator to prevent interference during automated tests
 export DISABLE_TUI_LOADER=1
 
-python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 180 --teardown-timeout 25 -o log_cli=true \
+python3 -m pytest -v -x -rsx --setup-timeout 60 --execution-timeout 600 --teardown-timeout 25 -o log_cli=true \
```
devzbysiu: Why are we changing timeout?

test/qa/conftest.py
```diff
@@ -292,6 +306,147 @@ def disable_remote_endpoint():
...
+
+    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
+    subprocess.run(f"sudo cp {daemon_path} {os.getcwd()}/tmp", shell=True, check=True, text=True, capture_output=True)
+
+    if os.environ.get("NORDVPN_TYPE") == "deb":
+        # For container
+        sed_command = ["sudo", "sed", "-i", f"1a export RC_USE_LOCAL_CONFIG=1", daemon_path]
+    else:
+        # For VM
+        sed_command = ["sudo", "sed", "-i", f'/$$Service$$/a Environment="RC_USE_LOCAL_CONFIG=1"']
```
devzbysiu: Shouldn't this be `/^\[Service\]/` to find the service section?

devzbysiu: Also, no path in the sed command here? :thinking:

test/qa/test_remote_config.py
```diff
@@ -245,3 +269,316 @@ def test_hash_modification_and_daemon_restart(
     assert daemon_log_reader.wait_for_messages(
         messages=RC_INITIAL_RUN_MESSAGES, cursor=cursor
     ), "Expected 'RC_INITIAL_RUN_MESSAGES' to appear in the daemon logs."
+
+
+@pytest.mark.skip(reason="Not implemented mock CDN")
+@pytest.mark.parametrize(
+    "tcid, error_message",
```
devzbysiu: Didn't we agree to have just docstrings for testing test cases like [here](https://github.com/NordSecurity/nordvpn-linux/pull/1058/files#diff-f5279f269048a9beb0c2226d7e86d99de18ae7c327243e7282696192480817c5R35) or this `tcid` is a different topic here?

test/qa/constants.py
devzbysiu, line 1: We have some constants in `lib./__init__.py` also. Maybe would be worth to have it in one place?

test/qa/conftest.py
```diff
@@ -292,6 +306,147 @@ def disable_remote_endpoint():
...
+
+    if os.environ.get("NORDVPN_TYPE") == "snap":
+        subprocess.run(f"sudo systemctl daemon-reload", shell=True, check=True)
+        subprocess.run(f'sudo snap restart {NORDVPND.get(os.environ.get("NORDVPN_TYPE"))}', shell=True, check=True)
+    else:
+        daemon.restart()
+
+
+@pytest.fixture
+def set_use_local_config_for_rc(daemon_log_reader):
```
devzbysiu: Big part of this fixture is the same as in `set_custom_timeout_for_rc_retry_scheme`. Would it make sense to extract to common function or not worth it?

test/qa/conftest.py
```diff
@@ -292,6 +306,147 @@ def disable_remote_endpoint():
...
+    if os.environ.get("NORDVPN_TYPE") == "deb":
+        # For container
+        sed_command = ["sudo", "sed", "-i", f"1a export RC_USE_LOCAL_CONFIG=1", daemon_path]
+    else:
+        # For VM
+        sed_command = ["sudo", "sed", "-i", f'/$$Service$$/a Environment="RC_USE_LOCAL_CONFIG=1"']
+
+    subprocess.run(sed_command)
+
+    if os.environ.get("NORDVPN_TYPE") == "snap":
```
devzbysiu: This `"NORDVPN_TYPE"` is repeated many times. It should be moved to a constant so we can control it from one place. Similarly literals like `"snap"` and `"deb"`. 
Or maybe this functionality of checking an installation type could be extracted to a function and reused

test/qa/conftest.py
```diff
@@ -292,6 +306,147 @@ def disable_remote_endpoint():
...
+def set_use_local_config_for_rc(daemon_log_reader):
+    """Fixture to set parameter RC_USE_LOCAL_CONFIG in config for not overwriting local config by remote.
+    (Log about downloading config still persist)
+    """
+    print("Setting 'use local config' for NordVPN daemon's rc")
+    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))
+
+    if not os.path.exists(daemon_path):
+        print(f"Daemon file does not exist. {daemon_path}")
+        pytest.skip("Unable to modify config file. File doesn't exist")
```
devzbysiu: It may be simpler to inject override file instead of modifying the original service file, so you can have file like:
```
[Service]
Environment=RC_LOAD_TIME_MIN=1
Environment=RC_USE_LOCAL_CONFIG=1
```
into `/etc/systemd/system/nordvpnd.service.d/override.conf` for deb and do 

```
sudo systemctl daemon-reload
sudo systemctl restart snap.nordvpn.nordvpnd.service
```

Then to reset it, we can just remove the override and reset the service

ci/test_deb.sh
pvraketa: It was due to the previous version of rc package when we had a minimum timeout of 4 minutes, will change it back to 180 sec

test/qa/conftest.py
pvraketa: image
But there is path

pvraketa: Returns error:
`sudo sed -i '/^\[Service\]/ Environment="RC_LOAD_MIN=1"' /usr/lib/systemd/system/nordvpnd.service
sed: -e expression #1, char 16: unknown command: `E'
`

test/qa/test_remote_config.py
pvraketa: In this case, we declare the variable "tcid", which stores the number of case, for later passing it into the docstring

test/qa/constants.py
pvraketa: Constants should be stored in a constants file for later importing them, moreover, we will move all constants from __init__ files to the constants file

test/qa/conftest.py
pvraketa: This variable stores the value of what package we have, so we need to compare it with some value to execute a specific command according to the case. How do you want to compare it?

test/qa/conftest.py
pvraketa: @keliramu could you confirm that it is a good solution?

test/qa/conftest.py
devzbysiu: The comparison is perfectly fine. I'm talking about repeated `NORDVPN_TYPE` string literal - so we could have a const for it and if this will change, then you have to change it only in one place.

But we can go even further and have functions like this or similar
```
if is_installed_as(SNAP):
...

if is_installed_as(DEB):
...

```

and do this env variable check there and reuse the function. So we have the logic in one place

test/qa/conftest.py
pvraketa:
```
def _set_custom_config_for_rc(daemon_log_reader, string_to_be_added: str) -> int:
    """
    Function to set custom config parameter for rc's config

    :param string_to_be_added:   Parameter with value to be set

    :return                      Position of cursor in daemon log file
    """
    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))
    if os.environ.get("NORDVPN_TYPE") == "deb":
        # For container
        sed_command = ["sudo", "sed", "-i", f"1a export {string_to_be_added}", daemon_path]
    else:
        # For VM
        sed_command = ["sudo", "sed", "-i", f'/$$Service$$/a Environment="{string_to_be_added}"', daemon_path]

    subprocess.run(sed_command)

    time_mark = daemon_log_reader.get_cursor()

    if os.environ.get("NORDVPN_TYPE") == "snap":
        subprocess.run(f"sudo systemctl daemon-reload", shell=True, check=True)
        subprocess.run(f'sudo snap restart {NORDVPND.get(os.environ.get("NORDVPN_TYPE"))}', shell=True, check=True)
    else:
        daemon.restart()

    return time_mark
```
and
```
@pytest.fixture
def set_custom_timeout_for_rc_retry_scheme(daemon_log_reader):
    """Fixture for setting a custom timeout for the NordVPN daemon's rc retry scheme."""
    print("Setting custom timeout for NordVPN daemon's rc retry scheme")
    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))

    if not os.path.exists(daemon_path):
        print(f"Daemon file does not exist. {daemon_path}")
        pytest.skip("Unable to modify config file. File doesn't exist")

    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
    subprocess.run(f"sudo cp {daemon_path} {os.getcwd()}/tmp", shell=True, check=True, text=True, capture_output=True)

    time_mark = _set_custom_config_for_rc(daemon_log_reader, string_to_be_added=f"RC_LOAD_TIME_MIN={RC_TIMEOUT}")

    if not daemon_log_reader.wait_for_messages(
        f"[Info] remote config download job time period: {RC_TIMEOUT}m0s", cursor=time_mark
    ):
        print("Service doesn't applied new time period.")

    yield

    subprocess.run(
        f'sudo cp {os.getcwd()}/tmp/{daemon_path.split("/")[-1]} {daemon_path}',
        check=True,
        capture_output=True,
        text=True,
        shell=True,
    )

    if os.environ.get("NORDVPN_TYPE") == "snap":
        subprocess.run(f"sudo systemctl daemon-reload", shell=True, check=True)
        subprocess.run(f'sudo snap restart {NORDVPND.get(os.environ.get("NORDVPN_TYPE"))}', shell=True, check=True)
    else:
        daemon.restart()
```
Agree?

test/qa/conftest.py
devzbysiu: I posted just the regex, you also need to specify command for `sed`, so in this case it's `a` for append, so the whole command is 
```
sudo sed -i '/^\[Service\]/a Environment="RC_LOAD_MIN=1"' /usr/lib/systemd/system/nordvpnd.service
```

test/qa/conftest.py
keliramu: it's good/valid approach for `systemd` environment, but for docker/initd based approach is good current implementation

keliramu: but also, need to think about multiple environments, as tests will be (to be confirmed) executed in VM/systemd based environment as well.

test/qa/conftest.py
```diff
@@ -303,3 +361,116 @@ def env():
...
+
+    if _is_installed_as(SNAP):
+        subprocess.run(f"sudo systemctl daemon-reload", shell=True, check=True)
+        subprocess.run(f"sudo {SNAP} restart {NORDVPND.get(SNAP)}", shell=True, check=True)
+    else:
+        daemon.restart()
+
+
+@pytest.fixture
+def stop_nordvpnd():
```
dfetti: Nit: maybe restart_nordvpnd

test/qa/conftest.py
```diff
@@ -303,3 +361,116 @@ def env():
...
+    """Fixture to stop nordvpnd before tests and start it after"""
+    daemon.stop()
+
+    yield
+
+    daemon.start()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def get_package_system():
```
dfetti: Shouldn't it announce the failure somehow? Or the tests will fail later when they can't find the env varaible?

dfetti (approved):

test/qa/conftest.py
pvraketa: They will just fail. It will show us that we don't have an installed package, or we have, but not for existing test cases (deb or snap)

test/qa/conftest.py
pvraketa: Well, it doesn't do a restart. We prepare nordvpnd for tests. There are a couple of tests that require nordvpnd to be stopped before the test. And after `yield` is start for return the state of an app as it was before the test to be independent

test/qa/constants.py
```diff
@@ -0,0 +1,4 @@
+DEB = "deb"
+SNAP = "snap"
+NORDVPND = {DEB: "nordvpnd.service", SNAP: "snap.nordvpn.nordvpnd.service"}
```
bbrutall: Please rename the dicts to reflect the proper data.  e.g. SERVICE_NAME/SERVICE_PATHS

test/qa/conftest.py
```diff
@@ -133,7 +186,17 @@ def _check_dns_resolution(domain, stop_event):
...
+        "filesize:10240",
+        "-b",
+        "files:2",
+        "-i",
+        "any",
+        "-w",
+        os.environ["WORKDIR"] + "/dist/logs/tshark_capture.pcap",
+    ]
     print("Starting tshark")
     process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
```
bbrutall, line 196: Just not to mix 2 different approaches of executing shell cmd, please replace with `sh` where it's possible.

test/qa/test_remote_config.py
```diff
@@ -9,22 +11,18 @@
...
-RC_LOCAL_MESSAGES = [
-    f"[Info] feature [meshnet] config loaded from: {LOCAL_CACHE_DIR}",
-    f"[Info] feature [libtelio] config loaded from: {LOCAL_CACHE_DIR}",
-    f"[Info] feature [nordvpn] config loaded from: {LOCAL_CACHE_DIR}",
-]
+RC_REMOTE_MESSAGE = f"[Info] feature [{{}}] remote config downloaded to: {LOCAL_CACHE_DIR}"
+RC_LOCAL_MESSAGE = f"[Info] feature [{{}}] config loaded from: {LOCAL_CACHE_DIR}"
+RC_REQUEST_MESSAGES = "Request:  GET https://downloads.nordcdn.com/apps/linux/config/dev/{}-hash.json"
+SERVICES_TO_BE_CHECK = ["nordvpn", "libtelio", "meshnet"]
+RC_REMOTE_MESSAGES = [RC_REMOTE_MESSAGE.format(service) for service in SERVICES_TO_BE_CHECK]
```
bbrutall: What is a reason to make it implicit? Please return the previous log patterns, and declare log patterns explicitly without formatting.

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+        - # Check date time of config files
+
+    :endsteps
+        - # Restore original config in daemon config file
+
+    :expected
+        - # Nordvpn daemon shows download related to remote config
+        - # Files are downloaded
+        - # After 1 min, in next iteration, nordvpn daemon still checks and shows download logs related to remote config
+        - # Files are downloaded/not downloaded according to remote config files
```
bbrutall: Please follow the docstring format introduced in test_remote_config.

test/qa/lib/log_reader.py
```diff
@@ -94,7 +92,7 @@ def wait_for_messages(
         :raises FileNotFoundError: If log file is not found.
         """
         start_time = time.time()
-        messages_to_find = list(messages)
+        messages_to_find = list(messages if type(messages) is list else [messages])
```
bbrutall: Not needed. If you need to verify only one message, just place one message pattern into the list.

test/qa/conftest.py
```diff
@@ -321,3 +380,118 @@ def env():
...
+@pytest.fixture
+def backup_restore_rc_config_files(nordvpnd_scope_function):  # noqa: ARG001
+    """Fixture to back up original config for remote config, and restore it after tests."""
+    os.makedirs(f"{os.getcwd()}/tmp", exist_ok=True)
+
+    print("Copy original remote monitoring config files as back up")
+    subprocess.run(f"sudo cp -r {LOCAL_CACHE_DIR} {os.getcwd()}/tmp", check=True, shell=True)
+    yield
+    print("Restore original remote monitoring config files")
+    subprocess.run(f"sudo cp -r {os.getcwd()}/tmp {LOCAL_CACHE_DIR} ", check=True, shell=True)
```
bbrutall: Also, please replace `subprocess.run` with `sh`

test/qa/conftest.py
```diff
@@ -55,16 +63,58 @@ def print_to_string(*args, **kwargs):
...
 
 
+def _is_installed_as(package_type: str) -> bool:
+    """
+    Function to verify if installed nordvpn package is installed as deb or snap.
+
+    :param package_type:    The type of the package to check (DEB or SNAP).
+    :returns:               True if the package is installed, False otherwise.
+    """
+    return os.environ.get("NORDVPN_TYPE") == package_type
```
bbrutall: This function is not needed. It's enough just to check if `package_type == DEB,` or SNAP

test/qa/conftest.py
```diff
@@ -321,3 +380,118 @@ def env():
...
+    """Fixture for setting a custom timeout for the NordVPN daemon's rc retry scheme."""
+    print("Setting custom timeout for NordVPN daemon's rc retry scheme")
+    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))
+
+    if not os.path.exists(daemon_path):
+        print(f"Daemon file does not exist. {daemon_path}")
+        pytest.skip("Unable to modify config file. File doesn't exist")
+
+    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
+    subprocess.run(f"sudo cp {daemon_path} {os.getcwd()}/tmp", shell=True, check=True, text=True, capture_output=True)
```
bbrutall: Just not to duplicate it, please define the variable with `"{os.getcwd()}/tmp/" `value

test/qa/conftest.py
```diff
@@ -321,3 +380,118 @@ def env():
...
+@pytest.fixture(scope="session", autouse=True)
+def get_package_system():
+    """Fixture to set in env variable system version of package"""
+    try:
+        dpkg_result = subprocess.run(["dpkg", "-l", "nordvpn"], capture_output=True, text=True)
+        if dpkg_result.returncode == 0 and "nordvpn" in dpkg_result.stdout:
+            os.environ["NORDVPN_TYPE"] = DEB
+    except FileNotFoundError:
+        # dpkg command not found - might not be a Debian-based system
+        pass
```
bbrutall: Add a warning here.

test/qa/conftest.py
```diff
@@ -321,3 +380,118 @@ def env():
...
+
+    try:
+        snap_result = subprocess.run(["snap", "list", "nordvpn"], capture_output=True, text=True)
+        if snap_result.returncode == 0 and "nordvpn" in snap_result.stdout:
+            os.environ["NORDVPN_TYPE"] = SNAP
+    except FileNotFoundError:
+        # snap command not found
+        pass
```
bbrutall: Also, plese add a warning here.

test/qa/conftest.py
```diff
@@ -321,3 +380,118 @@ def env():
...
+
+
+@pytest.fixture
+def backup_restore_rc_config_files(nordvpnd_scope_function):  # noqa: ARG001
+    """Fixture to back up original config for remote config, and restore it after tests."""
+    os.makedirs(f"{os.getcwd()}/tmp", exist_ok=True)
```
bbrutall: Define `"{os.getcwd()}/tmp"` like a separate variable.

test/qa/lib/daemon.py
```diff
@@ -59,31 +62,58 @@ def is_killswitch_on():
...
+
+            while not os.path.exists(socket_path) and (time.time() < start_time + max_wait):
+                print(f"Waiting for socket file {socket_path}...")
+                time.sleep(1)
+
+            if not os.path.exists(socket_path):
+                print(f"Socket file {socket_path} not created within timeout")
+
+            print("NordVPN service started successfully")
+
```
bbrutall: What is the reason to have this logic here? Do we really need it, as it seems like a workaround for smth?!

test/qa/constants.py
pvraketa: @bbrutall NORDVPND_SERVICE_NAME, agree?

bbrutall: yes

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+    wrong_permissions_files = []
+    for line in conf_files_data.stdout.splitlines():
+        if chmod_to_be_found not in line:
+            wrong_permissions_files.append(line)
+
+    assert (
+        not missed_services_config
+    ), f"Found files that do not have {chmod_to_be_found} chmod: {wrong_permissions_files}"
+
+    check_log_for_request_get_messages(daemon_log_reader)
```
bbrutall, line 599: Isn't it a duplicate of `test_remote_config_consumed_and_logged` ?
If you need to additionally verify `600` permission, so then just add only this verification to the existing test or create a new one with only permission verification.

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+        ), f"Couldn't found log {additional_log_verification}"
+
+    if value is False:
+        with pytest.raises(subprocess.CalledProcessError):
+            subprocess.run("nordvpn meshnet", shell=True, text=True, check=True)
+    elif value is None:
+        try:
+            subprocess.run("nordvpn meshnet", shell=True, text=True, check=True)
+        except subprocess.CalledProcessError:
+            pytest.fail(reason="nordvpn meshnet is not enabled")
```
bbrutall, line 773: Hard to read this code. Please make variables more readable and split code with a comment section with an explanation.

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+        - # Check daemon logs for error logs related to remote config
+
+    :endsteps
+        - # Restore original config in daemon config file
+
+    :expected
+        - # Nordvpn daemon shows error related to remote config
+        - # After 1 min, in next iteration, nordvpn daemon still checks and shows error related to remote config
+    """
+    first_time_mark = daemon_log_reader.get_cursor()
```
bbrutall: Do not use  first/second/third in var names as it says nothing. Please rename vars with explicit naming like `timestamp_before_smth... ``timestamp_after_smth...`

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+@pytest.mark.parametrize(
+    ("tcid", "error_message"),
+    [
+        pytest.param("LVPN-8452", "error: downloading main hash file:", id="no_cache"),
+        pytest.param(
+            "LVPN-8453", "failed downloading feature [ nordvpn ] remote config: downloading main file", id="no_config"
+        ),
+    ],
+)
+def test_remote_config_cdn_unavailable_(
```
bbrutall: typo

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+        - # Check daemon logs for request logs related to remote config
+
+    :endsteps
+        - # Restore original config in daemon config file
+
+    :expected
+        - # Nordvpn daemon shows download related to remote config
+        - # After 1 min, in next iteration, nordvpn daemon still checks and shows request logs related to remote config
+    """
+    command_for_conf_file = f'sudo find {LOCAL_CACHE_DIR} -type f -exec stat -c "%a %y %n" {{}} \\;'
```
bbrutall: Propose to rename it just to `cmd` or `command`

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+    chmod_to_be_found = "600"
```
bbrutall: Propose to rename it just to `expected_file_permission`

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+    missed_services_config = []
```
bbrutall: Propose to rename it just to `not_found_files` or `missing_files`

test/qa/test_remote_config.py
```diff
@@ -465,3 +486,313 @@ def test_local_config_usage_via_systemd_env(
...
+
+    assert not missed_services_config, f"Couldn't found download logs related to {missed_services_config}"
+
+    for service_config in SERVICES_TO_BE_CHECK:
+        if not daemon_log_reader.wait_for_messages(RC_LOCAL_MESSAGE.format(service_config), timeout=90):
+            missed_services_config.append(service_config)
+
+    assert not missed_services_config, f"Couldn't found download logs related to {missed_services_config}"
+
+    conf_files_data = subprocess.run(command_for_conf_file, shell=True, check=True, capture_output=True, text=True)
```
bbrutall: Please try to always name variables to reflect their expected values. In this case, `conf_files_data` var expects to have file names, correct? So better, to rename it to `found_files` or smth like this

test/qa/lib/daemon.py
```diff
@@ -59,31 +62,58 @@ def is_killswitch_on():
...
     # TODO: Replace with 'dpkg --purge nordvpn'
-    ssh_client.exec_command('sudo apt remove -y nordvpn')
+    ssh_client.exec_command("sudo apt remove -y nordvpn")
 
 
 def start():
     """Starts daemon and blocks until it is actually started."""
     if is_under_snap():
-        #sh.sudo.snap("start", "nordvpn")
+        # sh.sudo.snap("start", "nordvpn")
```
DoYouEvenCpp: I guess there's no need in keeping commented out code, right?
Can we have it remove here, and in other places?

test/qa/lib/daemon.py
```diff
@@ -59,31 +62,58 @@ def is_killswitch_on():
...
+
+            if result.returncode != 0:
+                print(f"Error starting NordVPN service: {result.stderr}")
+
+            # Wait for the socket file to appear
+            max_wait = 30  # seconds
+            start_time = time.time()
+            socket_path = "/run/nordvpn/nordvpnd.sock"
+
+            while not os.path.exists(socket_path) and (time.time() < start_time + max_wait):
```
DoYouEvenCpp: I'd suggest here two slight modifications:
- split the condition so the while tests only against socket_path presence
- add another if testing against elapsed time within loop body (and breaks when timeouts)


Also, I guess it's safer to use time.monotonic() for more robust time tracking.

eg.:

```
while not os.path.exists(socket_path):
   if timeout_hit:
      break; (maybe also print)
   ... rest of the code

File: test/qa/conftest.py [hunk 31]
```diff
@@ -330,3 +378,118 @@ def env():
...
+def set_custom_timeout_for_rc_retry_scheme(daemon_log_reader):
+    """Fixture for setting a custom timeout for the NordVPN daemon's rc retry scheme."""
+    print("Setting custom timeout for NordVPN daemon's rc retry scheme")
+    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))
+
+    if not os.path.exists(daemon_path):
+        print(f"Daemon file does not exist. {daemon_path}")
+        pytest.skip("Unable to modify config file. File doesn't exist")
+
+    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
```
DoYouEvenCpp: Where does it get removed?

test/qa/conftest.py
```diff
@@ -330,3 +378,118 @@ def env():
...
+    subprocess.run(
+        f'sudo cp {os.getcwd()}/tmp/{daemon_path.split("/")[-1]} {daemon_path}',
+        check=True,
+        capture_output=True,
+        text=True,
+        shell=True,
+    )
+
+    if _is_installed_as(SNAP):
+        subprocess.run("sudo systemctl daemon-reload", shell=True, check=True)
```
DoYouEvenCpp: Since more code is being added to increase the stability of testing framework, why not check returned value here (and where it makes sense), and at least run some brief validation of the `returncode` ?

test/qa/conftest.py
DoYouEvenCpp: So maybe let's call it pause_nordvpnd ?

test/qa/lib/daemon.py
pvraketa: It is commented in `main` since 2024, so not sure. Could we make a separate task for clean up? What do you think?

pvraketa: Created a ticket for that:
Id is 9221

test/qa/conftest.py
```diff
@@ -330,3 +378,118 @@ def env():
...
+        subprocess.run(f"sudo {SNAP} restart {NORDVPND_SERVICE_NAME.get(SNAP)}", shell=True, check=True)
+    else:
+        daemon.restart()
```
devzbysiu: This is repeated three times, should be put in a utility function and reused

test/qa/lib/daemon.py
```diff
@@ -99,14 +129,26 @@ def start_peer(ssh_client: ssh.Ssh):
...
     elif is_init_systemd():
-        #sh.sudo.systemctl.stop.nordvpnd()
+        # sh.sudo.systemctl.stop.nordvpnd()
         os.popen("sudo systemctl stop nordvpn").read()
     else:
         # call to init.d returns before the daemon is actually stopped
-        sh.sudo("/etc/init.d/nordvpn", "stop")
+        try:
+            sh.sudo("/etc/init.d/nordvpn", "stop")
+        except (subprocess.SubprocessError, FileNotFoundError, TimeoutError):
```
devzbysiu: Are those exceptions happening when `stop` is run? What is the reason for it? 
If this keeps happening, we need to find the reason, otherwise we'll be hiding the issue by killing the process below

test/qa/lib/daemon.py
```diff
@@ -120,12 +162,15 @@ def stop_peer(ssh_client: ssh.Ssh):
...
-    start()
+    try:
+        stop()
+        if is_init_systemd():
+            # The default limit is to allow 5 restarts in a 10sec period.
+            # There are 100ms intervals between restarts.
+            time.sleep(2)
+        start()
+    except sh.ErrorReturnCode as ex:
+        print(f"Got next error: {ex}")
```
devzbysiu: What should happen in this case? Will the test continue running with failed restart?

test/qa/lib/daemon.py
```diff
@@ -59,31 +62,60 @@ def is_killswitch_on():
...
+            socket_path = "/run/nordvpn/nordvpnd.sock"
+
+            while time.time() < (start_time + max_wait):
+                if os.path.exists(socket_path):
+                    break
+                print(f"Waiting for socket file {socket_path}...")
+                time.sleep(1)
+
+            if not os.path.exists(socket_path):
+                print(f"Socket file {socket_path} not created within timeout")
```
devzbysiu: What should happen here? You are printing that socket was not created, but it continues execution and below you are printing that service was started successfully

test/qa/lib/daemon.py
```diff
@@ -99,14 +131,26 @@ def start_peer(ssh_client: ssh.Ssh):
...
+            if is_running():
+                print("Service still running, trying sudo pkill...")
+                subprocess.run(["sudo", "pkill", "nordvpnd"], capture_output=True, check=False)
+            time.sleep(3)
+            if is_running():
+                print("Service still running, trying sudo pkill -9 ...")
+                subprocess.run(["sudo", "pkill", "-9", "nordvpnd"], capture_output=True, check=False)
+            time.sleep(3)
+            if is_running():
+                print("Failed to stop nordvpn")
```
devzbysiu: What should happen at this point?

test/qa/lib/daemon.py
```diff
@@ -99,14 +131,26 @@ def start_peer(ssh_client: ssh.Ssh):
...
         while is_running():
```
devzbysiu: When this loop will finish?

test/qa/conftest.py
```diff
@@ -64,16 +71,58 @@ def print_to_string(*args, **kwargs):
...
 def _print_with_timestamp(*args, **kwargs):
     # Get the current time and format it
-    timestamp = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    timestamp = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
     # Prepend the timestamp to the original print arguments
     _original_print(timestamp, *args, **kwargs)
     logging.log(data=print_to_string(timestamp, *args, **kwargs))
 
 
+def _is_installed_as(package_type: str) -> bool:
```
devzbysiu: I see that there is also `is_under_snap()` function, can it be reused?

test/qa/conftest.py
```diff
@@ -64,16 +71,58 @@ def print_to_string(*args, **kwargs):
...
+    """
+    Function to verify if installed nordvpn package is installed as deb or snap.
+
+    :param package_type:    The type of the package to check (DEB or SNAP).
+    :returns:               True if the package is installed, False otherwise.
+    """
+    return os.environ.get("NORDVPN_TYPE") == package_type
```
devzbysiu: `"NORDVPN_TYPE"` is repeated multiple times, should be put in constant to have single place for change

test/qa/conftest.py
```diff
@@ -330,3 +379,122 @@ def env():
...
+        print(f"Daemon file does not exist. {daemon_path}")
+        pytest.skip("Unable to modify config file. File doesn't exist")
+
+    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
+    subprocess.run(f"sudo cp {daemon_path} {os.getcwd()}/tmp", shell=True, check=True, text=True, capture_output=True)
+
+    time_mark = _set_custom_config_for_rc(daemon_log_reader, string_to_be_added=f"RC_LOAD_TIME_MIN={RC_TIMEOUT}")
+
+    if not daemon_log_reader.wait_for_messages(
+        f"[Info] remote config download job time period: {RC_TIMEOUT}m0s", cursor=time_mark
```
devzbysiu: Typo `m0s`? Or we have it in logs like that?

test/qa/conftest.py
```diff
@@ -64,16 +72,58 @@ def print_to_string(*args, **kwargs):
...
+    """
+    Function to verify if installed nordvpn package is installed as deb or snap.
+
+    :param package_type:    The type of the package to check (DEB or SNAP).
+    :returns:               True if the package is installed, False otherwise.
+    """
+    return os.environ.get("NORDVPN_TYPE") == package_type
+
+
+def _set_custom_config_for_rc(daemon_log_reader, string_to_be_added: str) -> int:
```
keliramu, line 84: imo, name for this function is confusing. What it does: it adds environment variable with value to a linux service (aka daemon). It is general and is not so much related to rc (is this remote config?) I suggest change this function name to e.g. `_set_daemon_variable`, you even use `daemon` in one of parameter's name, and inside function also have `daemon_path`

test/qa/test_remote_config.py
```diff
@@ -9,22 +11,18 @@
...
-
-RC_LOCAL_MESSAGES = [
-    f"[Info] feature [meshnet] config loaded from: {LOCAL_CACHE_DIR}",
-    f"[Info] feature [libtelio] config loaded from: {LOCAL_CACHE_DIR}",
-    f"[Info] feature [nordvpn] config loaded from: {LOCAL_CACHE_DIR}",
-]
+RC_REMOTE_MESSAGE = f"[Info] feature [{{}}] remote config downloaded to: {LOCAL_CACHE_DIR}"
+RC_LOCAL_MESSAGE = f"[Info] feature [{{}}] config loaded from: {LOCAL_CACHE_DIR}"
+RC_REQUEST_MESSAGES = "Request:  GET https://downloads.nordcdn.com/apps/linux/config/dev/{}-hash.json"
+SERVICES_TO_BE_CHECK = ["nordvpn", "libtelio", "meshnet"]
```
keliramu: naming fix: SERVICES_TO_BE_CHECK -> FEATURES_TO_CHECK
this is not `SERVICES` this is `FEATURES`

test/qa/conftest.py
pvraketa: 1m0s

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
     ), "The missing messages do not match the expected RC_REMOTE_MESSAGES."
+
+
+@pytest.mark.skip(reason="Not implemented mock CDN")
+@pytest.mark.parametrize(
+    ("tcid", "error_message"),
+    [
+        pytest.param("LVPN-8452", "error: downloading main hash file:", id="no_cache"),
+        pytest.param(
+            "LVPN-8453", "failed downloading feature [ nordvpn ] remote config: downloading main file", id="no_config"
```
keliramu: Log message now does not contain spaces e.g. `... [nordvpn] ...`

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+def test_remote_config_download_config_on_start(
+    initialized_app_with_remote_config,  # noqa: ARG001
+    set_custom_timeout_for_rc_retry_scheme,  # noqa: ARG001
+    clean_cache_files,  # noqa: ARG001
+    daemon_log_reader,
+):
+    """
+    :tcid       LVPN-8456
+
+    :details    Verify that app download config files, set them 600 chmod
```
keliramu: language fix: `set them 600 chmod` -> `set file permissions to 600`

test/qa/conftest.py
devzbysiu: Looks like only this:
```
if not daemon_log_reader.wait_for_messages(
        f"[Info] remote config download job time period: {RC_TIMEOUT}m0s", cursor=time_mark
    ):
        print("Service doesn't applied new time period.")
```
Is different in `set_custom_timeout_for_rc_retry_scheme`, the rest is repeated

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+    backup_restore_rc_config_files,  # noqa: ARG001
+    rc_config_manager,
+    daemon_log_reader,  # noqa: ARG001
+    set_custom_timeout_for_rc_retry_scheme,  # noqa: ARG001
+    set_use_local_config_for_rc,  # noqa: ARG001
+):
+    """
+    :tcid       {tcid}
+
+    :details    Verify that app change behavior according to config
```
keliramu: Description is not clear, what is the purpose of this test?

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+    :tcid       {tcid}
+
+    :details    Verify that app verifies local config files and download then with additional verification of remote config
+    :keywords   RC
+
+    :preconditions
+        - # Config files are available on CDN
+        - # No remote config files are cached on local disk
+        - # nordvpnd is stopped
+        - # Timeout in daemon service is set to 1 min
```
keliramu: What timeout? probably it is remote config load job time between job runs?

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+    """
+    :tcid       LVPN-8456
+
+    :details    Verify that app download config files, set them 600 chmod
+    :keywords   RC
+
+    :preconditions
+        - # Config files are available on CDN
+        - # No remote config files are cached on local disk
+        - # Timeout in daemon service is set to 1 min
```
keliramu: What timeout? probably it is remote config load job time between job runs?

test/qa/lib/daemon.py
pvraketa: It helps to find a root cause of failed restart. Original code was:
```
def restart():
    stop()
    if is_init_systemd():
        # The default limit is to allow 5 restarts in a 10sec period.
        # There are 100ms intervals between restarts.
        time.sleep(2)
    start()
```

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+    :tcid       {tcid}
+
+    :details    Verify that app handles the case, when remote files are not available in CDN and no cache files are locally
+    :keywords   RC
+
+    :preconditions
+        - # remote file is removed from CDN according to use case
+        - # No remote config files are cached on local disk
+        - # nordvpnd is stopped
+        - # Timeout in daemon service is set to 1 min
```
keliramu: What timeout? probably it is remote config load job time between job runs?

test/qa/lib/daemon.py
pvraketa: Will add raise TimeoutError

test/qa/lib/daemon.py
pvraketa: Will add raise subprocess.SubprocessError

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+    initialized_app_with_remote_config,  # noqa: ARG001
+    set_custom_timeout_for_rc_retry_scheme,  # noqa: ARG001
+    clean_cache_files,  # noqa: ARG001
+    pause_nordvpnd,  # noqa: ARG001
+    daemon_log_reader,
+):
+    """
+    :tcid       {tcid}
+
+    :details    Verify that app verifies local config files and download then with additional verification of remote config
```
keliramu: Review and fix description of the test - it's tangled and not clear what you want to say

test/qa/lib/daemon.py
pvraketa: When daemon will stop or raise will be raised

test/qa/conftest.py
pvraketa: My method uses commands, especially snap or deb approach, which is understandable and easier to use

test/qa/conftest.py
```diff
@@ -64,16 +72,58 @@ def print_to_string(*args, **kwargs):
...
+        # For VM
+        sed_command = ["sudo", "sed", "-i", f'/^\[Service\]/a Environment="{string_to_be_added}"', daemon_path]
+
+    subprocess.run(sed_command)
+
+    time_mark = daemon_log_reader.get_cursor()
+
+    if _is_installed_as(SNAP):
+        subprocess.run("sudo systemctl daemon-reload", shell=True, check=True)
+        subprocess.run(f"sudo {SNAP} restart {NORDVPND_SERVICE_NAME.get(SNAP)}", shell=True, check=True)
```
devzbysiu: This will run `sudo snap restart snap.nordvpn.nordvpnd.service` which is wrong. You can either run:
```
sudo snap restart nordvpn
```
or
```
sudo systemctl restart snap.nordvpn.nordvpnd.service
```

test/qa/conftest.py
pvraketa: The main purpose of this feature is to set a custom config for rc, to change behavior. That is all what for this method is created. If in the future, there will be a need to set parameters, not only for RC, we can rename it

test/qa/constants.py
```diff
@@ -0,0 +1,4 @@
+DEB = "deb"
+SNAP = "snap"
+NORDVPND_SERVICE_NAME = {DEB: "nordvpnd.service", SNAP: "snap.nordvpn.nordvpnd.service"}
+NORDVPND_FILE = {DEB: "/etc/init.d/nordvpn", SNAP: "/usr/lib/systemd/system/nordvpnd.service"}
```
devzbysiu: Is this service file correct for snap? On my system, snap is running from service file under `/etc/systemd/system/snap.nordvpn.nordvpnd.service`

test/qa/lib/daemon.py
```diff
@@ -59,31 +62,60 @@ def is_killswitch_on():
...
+                time.sleep(1)
+
+            if not os.path.exists(socket_path):
+                print(f"Socket file {socket_path} not created within timeout")
+
+            print("NordVPN service started successfully")
+
+        except (subprocess.SubprocessError, FileNotFoundError, TimeoutError) as e:
+            print(f"Exception while starting NordVPN service: {e}")
     while not is_running():
```
devzbysiu, line 97: When this one will finish?

test/qa/lib/daemon.py
pvraketa: It happen from time to time without an error message:
```
Got next error:
  RAN: /usr/bin/sudo /etc/init.d/nordvpn stop
  STDOUT:
  STDERR:
```
We already talked about that

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,312 @@ def test_local_config_usage_via_systemd_env(
...
+
+    :expected
+        - # Nordvpn daemon shows error related to remote config
+        - # After 1 min, in next iteration, nordvpn daemon still checks and shows error related to remote config
+    """
+    first_time_mark = daemon_log_reader.get_cursor()
+    daemon.start()
+
+    assert daemon_log_reader.wait_for_messages(
+        RC_REMOTE_MESSAGE.format(error_message), cursor=first_time_mark, timeout=90
```
devzbysiu: Why is this parametrized with `error_message`? `RC_REMOTE_MESSAGE` is `[Info] feature [{{}}] remote config downloaded to: {LOCAL_CACHE_DIR}`

test/qa/conftest.py
```diff
@@ -64,16 +72,58 @@ def print_to_string(*args, **kwargs):
...
+    """
+    daemon_path = NORDVPND_FILE.get(os.environ.get("NORDVPN_TYPE"))
+    if _is_installed_as(DEB):
+        # For container
+        sed_command = ["sudo", "sed", "-i", f"1a export {string_to_be_added}", daemon_path]
+    else:
+        # For VM
+        sed_command = ["sudo", "sed", "-i", f'/^\[Service\]/a Environment="{string_to_be_added}"', daemon_path]
+
+    subprocess.run(sed_command)
```
devzbysiu: Is this service change restored somewhere?

test/qa/conftest.py
pvraketa: Also, a different value to set
RC_USE_LOCAL_CONFIG=1
and 
RC_LOAD_TIME_MIN={RC_TIMEOUT}

test/qa/lib/daemon.py
```diff
@@ -99,14 +131,26 @@ def start_peer(ssh_client: ssh.Ssh):
...
     elif is_init_systemd():
-        #sh.sudo.systemctl.stop.nordvpnd()
+        # sh.sudo.systemctl.stop.nordvpnd()
         os.popen("sudo systemctl stop nordvpn").read()
     else:
         # call to init.d returns before the daemon is actually stopped
-        sh.sudo("/etc/init.d/nordvpn", "stop")
+        try:
+            sh.sudo("/etc/init.d/nordvpn", "stop")
+        except (subprocess.SubprocessError, FileNotFoundError, TimeoutError):
```
devzbysiu: Btw. is `sh` throwing `subprocess.*` exceptions at all?

test/qa/test_remote_config.py
pvraketa: Please suggest your description. How would you describe it?
Here is name of all tests
Make sure, that local "remote" config can be passed to app, is loaded and works
Verify, that feature is active to users, if `rollout` is not specified
Verify, that all app versions load the config, if field `app_version` has value *
Verify, that all app versions load the config, if field `app_version` is empty

test/qa/conftest.py
pvraketa: That will not work
`sudo systemctl restart snap.nordvpn.nordvpnd.service`
As there's no systemctl for snap, Ramunas mentioned it. And there is no nordvpn service for snap

test/qa/constants.py
pvraketa: Ramunas also confirmed the path. Please look at remote-config-dev-qa chat

test/qa/lib/daemon.py
pvraketa: Already answered in the comments from above

test/qa/test_remote_config.py
pvraketa: These tests are currently skipped due to not implemented mockCDN

test/qa/conftest.py
pvraketa: Plase take a look at 
def set_custom_timeout_for_rc_retry_scheme
    os.makedirs(f"{os.getcwd()}/tmp/", exist_ok=True)
    subprocess.run(f"sudo cp {daemon_path} {os.getcwd()}/tmp", shell=True, check=True, text=True, capture_output=True)

    time_mark = _set_custom_config_for_rc(daemon_log_reader, string_to_be_added=f"RC_LOAD_TIME_MIN={RC_TIMEOUT}")

test/qa/lib/daemon.py
pvraketa: Should be OSError, will change that

test/qa/test_remote_config.py
pvraketa: No, it collects data of files such as name, chmod, time and etc

test/qa/test_remote_config.py
pvraketa: missed_feature_config

test/qa/test_remote_config.py
pvraketa: No, chmod is part of a command, andthe  output returns a code, not read/write/execute

test/qa/test_remote_config.py
pvraketa: Not a clear variable. Current one is better

test/qa/conftest.py
pvraketa: For what line is it supposed to be?

test/qa/test_remote_config.py
pvraketa: You had 3 lines that have only a different feature name. So you just keep repeating them. This approach is not doing it

test/qa/test_remote_config.py
pvraketa: Please share with me the documentation where it is decided to use the format in test_remote_config. This approach contains more detailed information than in test_remote_config.

test/qa/lib/log_reader.py
pvraketa: And you will get this:
image

test/qa/conftest.py
pvraketa: Is that a standard of using sh? What are the benefits of sh over subprocess?

test/qa/conftest.py
pvraketa: It is needed. We set the type of package we are working on as an environment variable in the fixture. In your case, it is just comparing whether what we pass is DEB or SNAP. It doesn't make any sense

test/qa/conftest.py
pvraketa: What kind of warning are you expecting here? We determine what type of package we have, DEB or SNAP by trying commands

test/qa/conftest.py
pvraketa: It's simply adding 'tmp' to the standard command output. I don't think that such a path requires a variable

test/qa/conftest.py
pvraketa: Already answered in the comment from above

test/qa/conftest.py
pvraketa: Already answered in the comment from above

test/qa/test_remote_config.py
pvraketa: Could you provide the ticket number of your test cases to check a steps and verify which one was created first?

test/qa/test_remote_config.py
pvraketa: It is just simply first_time_attempts, second one and etc. I don't see anything wrong with that

test/qa/test_remote_config.py
pvraketa: Suggest a proposal

test/qa/lib/daemon.py
pvraketa: Yes, we need it

test/qa/constants.py
keliramu: You need to doublecheck the correct path: Snap-created systemd unit files are stored on disk inside the snap's own directory structure under /var/lib/snapd/ (or /snap/ in some cases) and not directly in the usual systemd unit file directories like /etc/systemd/system/ or /usr/lib/systemd/system/. However, at runtime, snapd dynamically generates and exposes these units to systemd, and the active symbolic links or runtime unit files appear in /run/systemd/system/.

test/qa/test_remote_config.py
```diff
@@ -448,3 +472,320 @@ def test_local_config_usage_via_systemd_env(
...
+            "invalid version constraint: improper constraint: y",
+            id="set_app_version_invalid_version",
+        ),
+    ],
+)
+def test_remote_config_change_local_meshnet_config_settings_(
+    tcid,  # noqa: ARG001
+    parameter,  # noqa: ARG001
+    value,  # noqa: ARG001
+    additional_log_verification,
```
keliramu: Fixtures should be named differently to recognize them from first look

test/qa/conftest.py
devzbysiu: Discussed, resolving

PR #1066, opened 2025-09-01
LVPN-8657: Migrate workflows

devzbysiu:
Changes:
- README.md describing the workflow for GUI + core
- fix .gitignore to not ignore GUI files:
  - added ignored previously CMakeLists.txt files
- update translations file

This PR contains also changes merged here https://github.com/NordSecurity/nordvpn-linux/pull/1071


.github/workflows/orchestrator.yml
```diff
@@ -0,0 +1,43 @@
...
+      core_changed: ${{ steps.decide.outputs.core_changed }}
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Decide
+        id: decide
+        run: python3 ci/detect_changes.py --gui-dir gui
```
edvyno1: have you looked into this? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore maybe possible to implement running workflow based on path changes

devzbysiu: Yeah, I actually did it like that initially, but I think I had a brain damage or something, because I did two workflows:
- core + GUI
- GUI

And then with `paths` and `paths-ignore` I couldn't make the GUI part run only once - either via core + GUI or via GUI only workflow.

But that's nonsense, we can have just core only and GUI only workflows. GUI runs always, core runs for everything except GUI only changes, no idea what I had in mind last time :man_shrugging:

bartoszWojciechO (approved):

edvyno1 (approved):

PR #1076, opened 2025-09-02
LVPN-9038: Fix remote config validation error

keliramu:


config/remote/config_test.go
```diff
@@ -377,21 +478,1061 @@ func TestHandleIncludeFiles(t *testing.T) {
...
+						"app_version": "*",
+						"weight": 1
+					}
+				]
+			}
+		]
+	}`
+	includeFileContent := `{"setting1": "value1"}`
+	// includeFileHash := hash([]byte(includeFileContent))
+	// includeFileHashJson := fmt.Sprintf(`{"hash": "%s"}`, includeFileHash)
```
devzbysiu: Leftovers?

devzbysiu (approved):

config/remote/config.go
```diff
@@ -157,10 +157,6 @@ func (f *Feature) load(sourcePath string, fr fileReader, jv validator) error {
 	}
 
 	mainJsonFileName := f.FilePath(sourcePath)
-	if err := internal.IsFileTooBig(mainJsonFileName); err != nil {
```
DoYouEvenCpp, line 160: This check is really not needed anymore?

config/remote/config_test.go
```diff
@@ -377,21 +478,1061 @@ func TestHandleIncludeFiles(t *testing.T) {
...
+	// Create a temporary directory just for the path validation tests
+	tempDir, err := os.MkdirTemp("", "test_path_validation_*")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// For path validation tests, use the actual temp directory
+			if test.name == "invalid source path" {
+				test.sourcePath = "/nonexistent/path"
```
DoYouEvenCpp: [nit] I personally see this as a bit of a code smell - keeping logic to update `test.sourcePath` for every single one test case except `invalid source path` sounds like a reason for a fixture, and creating tempDir over there, so the rest of tests could reuse it?

config/remote/config.go
keliramu: This check is done by `jsonFileReaderWriter.readFile`

config/remote/config_test.go
keliramu: Yes, indeed, removed.

config/remote/config_test.go
keliramu: Ok, good point, fixed/improved

DoYouEvenCpp (approved):

PR #1077, opened 2025-09-03
LVPN-5404: Save daemon logs when running the tests in snap

mariusSincovici:
* After the tests are executed save the daemon log into daemon.log
* Add support to use a custom vagrant box for running the tests
* Replace for libvirt sync to nfs to bring the daemon.log to the host
* When running the tests add the current user into the nordvpn group

### DIFF
--- a/ci/qa_run_tests.sh
+++ b/ci/qa_run_tests.sh
@@ -48,9 +48,9 @@ esac
 
 # check that the nordvpn group exists in the system and that the current user is part of it
 GROUP="nordvpn"
-if ! getent group "$GROUP" > /dev/null; then
+if ! getent group "${GROUP}" > /dev/null; then
 	# application installer must create the group
-	echo "Group '$GROUP' does not exist."
+	echo "Group '${GROUP}' does not exist."
 	exit 1
 fi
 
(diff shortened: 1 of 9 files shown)


ci/qa_run_tests.sh
```diff
@@ -46,8 +46,26 @@ case "${pattern}" in
 esac
 
 
+# check that the nordvpn group exists in the system and that the current user is part of it
+GROUP="nordvpn"
+if ! getent group "$GROUP" > /dev/null; then
```
keliramu: ${GROUP} is the safer, more versatile, and recommended form. Here and in following changed lines.

ci/qa_run_tests.sh
```diff
@@ -46,8 +46,26 @@ case "${pattern}" in
...
+	# application installer must create the group
+	echo "Group '$GROUP' does not exist."
```
keliramu: Imo, we can create missing group here if it is not yet created. Group is being created by deb/rpm pre-install script, but if trying to run from dev environment without prior installation, would be good to have group created here and not to break execution.

ci/snap_functions.sh
```diff
@@ -1,40 +1,39 @@
...
 snap_connect_interfaces() {
-    local SNAP_NAME=nordvpn
-
     # List all connections for the Snap
-    echo "Checking connections for Snap package: ${SNAP_NAME}"
-    connections=$(snap connections "${SNAP_NAME}")
+    echo "Make the snap connections"
+
+    wait_for_daemon
+    output=$(nordvpn status || true)
```
bartoszWojciechO: Maybe the suggested `snap connect` commands can be grepped out of the output here? Then this variable could be called something like `snap_connect_commands`, and this would be more clear.

bartoszWojciechO (approved):

ci/qa_run_tests.sh
mariusSincovici: Actually exactly for this reason I don't want to create the group, to test the pre install scripts . The group must be created by the pre-install scripts. So if the group is not created then the scripts are not working and it should fail here.

If someone tests with the app already installed, then when it was installed the group should have been created at the installation. Any other cases, since it is not the normal flow should report an error and group must be created manually.

ci/qa_run_tests.sh
mariusSincovici: done, I hope I didn't missed any.

ci/snap_functions.sh
mariusSincovici: done

keliramu (approved):

PR #1078, opened 2025-09-03
LVPN-4750: Listen for status events in systray

thegravi:


client/messages.go
```diff
@@ -1,12 +1,12 @@
...
 
 const (
-	ConfigMessage = "It seems there's an issue with the config file. If the issue persists, please contact our customer support."
-
-	LoginFailure       = "Username or password is not correct. Please try again."
-	LegacyLoginFailure = "We couldn't log you in. Make sure your credentials are correct. If you have MFA enabled, log in using the 'nordvpn login' command."
-	TokenLoginFailure  = "Token parameter value is missing." // #nosec
-	TokenInvalid       = "We couldn't log you in - the access token is not valid. Please check if you've entered the token correctly. If the issue persists, contact our customer support."
+	ConfigMessage                = "It seems there's an issue with the config file. If the issue persists, please contact our customer support."
+	RecentConnectionErrorMessage = "It seems there's an issue with the recent connections file. If the issue persists, please contact our customer support."
```
devzbysiu: Does this need a review from Copy team?

tray/tray.go
```diff
@@ -196,22 +184,100 @@ func (ti *Instance) configureDebugMode() {
...
+		ti.setSettings(st.SettingsChange)
+		// identify whether we need to also update a country list
+		ti.connSensor.Set(connectionSettings{
+			Obfuscated:      st.SettingsChange.Obfuscate,
+			Protocol:        st.SettingsChange.Protocol,
+			Technology:      st.SettingsChange.Technology,
+			VirtualLocation: st.SettingsChange.VirtualLocation,
+		})
+
+		if ti.connSensor.Detected() {
```
devzbysiu: Would be nice to have more specific context in the naming here, because by just looking on it, it's not obvious what was detected. So `connChangeSensor.Detected()` or `connSensor.ChangeDetected()` would be clearer IMO

tray/recents.go
```diff
@@ -34,18 +67,18 @@ func makeDisplayLabel(conn *pb.RecentConnectionModel) string {
 		return conn.SpecificServerName
 
 	case config.ServerSelectionRule_GROUP:
-		return strings.ReplaceAll(conn.Group.String(), "_", " ")
+		return formatGroupTitle(conn.Group)
```
devzbysiu, line 73: Now this will produce `""` if the group is not in the map, previously it was using group name. 
I think it would be safer to do fallback to group name in `formatGroupTitle` if the group is not in the map. Or even use `switch` over the group instead of map

tray/recents.go
```diff
@@ -34,18 +67,18 @@ func makeDisplayLabel(conn *pb.RecentConnectionModel) string {
...
 
 	case config.ServerSelectionRule_COUNTRY_WITH_GROUP:
-		if conn.Group != config.ServerGroup_UNDEFINED && conn.Country != "" {
-			group := strings.ReplaceAll(conn.Group.String(), "_", " ")
-			return fmt.Sprintf("%s (%s)", group, conn.Country)
+		if conn.Group == config.ServerGroup_UNDEFINED || conn.Country == "" {
+			return ""
```
devzbysiu, line 78: Could this be part of the the `formatGroupTitle`?

tray/recents.go
```diff
@@ -89,24 +104,42 @@ func connectByConnectionModel(ti *Instance, model *pb.RecentConnectionModel) boo
 		return ti.connect("", "")
 
 	case config.ServerSelectionRule_CITY:
+		if model.City == "" {
+			return false
```
devzbysiu: How is this bool result used? I see only one place and it's ignored there :thinking:

daemon/recents/model.go
```diff
@@ -12,4 +13,5 @@ type Model struct {
 	SpecificServerName string                     `json:"specific-server-name"`
 	SpecificServer     string                     `json:"specific-server"`
 	ConnectionType     config.ServerSelectionRule `json:"connection-type"`
+	ServerTechnologies []core.ServerTechnology    `json:"server-technology"`
```
devzbysiu: Shouldn't it be `json:"server-technologies"`?

tray/expo_backoff.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+type BackoffConfig struct {
+	// InitialDelay time for the first retry.
+	// Will defaul to 1 second if non-positive value is provided.
```
devzbysiu: [typo] `defaul`

tray/expo_backoff.go
```diff
@@ -0,0 +1,92 @@
+package tray
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+type BackoffConfig struct {
```
devzbysiu, line 10: We already have `network.ExponentialBackoff()`, maybe would be good to use single mechanism across the codebase if possible

If it doesn't make sense, then here is some idea for simpler backoff https://commaok.xyz/post/simple-backoff/. May be useful

tray/expo_backoff.go
```diff
@@ -0,0 +1,92 @@
...
+	// InitialDelay time for the first retry.
+	// Will defaul to 1 second if non-positive value is provided.
+	InitialDelay time.Duration
+
+	// MaxDelay maximum time between attempts.
+	// Will default to 1 minutes if non-positive value ir provided.
```
devzbysiu: [typo] `ir provided`

tray/recents.go
```diff
@@ -116,3 +149,60 @@ func connectByConnectionModel(ti *Instance, model *pb.RecentConnectionModel) boo
...
 	return false
 }
+
+type recentConnectionsManager struct {
+	mu          sync.RWMutex
+	connections []RecentConnection
+	client      pb.DaemonClient
+}
+
+// newRecentConnectionsManager created new recent VPN connection manager
```
devzbysiu: `created` -> `creates`?

tray/state_listener.go
```diff
@@ -0,0 +1,105 @@
...
+	return &stateListener{
+		client:     client,
+		queue:      make(chan *pb.AppState, 10),
+		onDataFunc: onDataFunc,
+	}
+}
+
+func (l *stateListener) Start() {
+	if l.cancelFunc != nil {
+		log.Printf("%s %s Already listening to daemon events", logTag, internal.WarningPrefix)
```
devzbysiu: `\n` missing, here and in other places

client/messages.go
thegravi: Yes, I'm assembling other messages that need review to make a single ticket.

tray/tray.go
devzbysiu: Btw. isn't this always redrawing now even if the changes are not detected? I don't see the reset of `changed` field

daemon/rpc_connect.go
```diff
@@ -353,4 +335,73 @@ func determineTargetServerGroup(server *core.Server, parameters ServerParameters
 	return ""
 }
 
+// buildRecentConnectionModel creates a recent connection model
+func buildRecentConnectionModel(
```
mariusSincovici, line 351: Wouldn't make sense for the function to return also error, for the default case? And when it is error, then it is not added into `r.recentVPNConnStore.Add(`

daemon/rpc_get_recent_connections.go
```diff
@@ -17,24 +19,41 @@ func (r *RPC) GetRecentConnections(
...
-	// limit results if value is specified
-	if in.Limit != nil {
-		limit := int(*in.Limit)
-		if limit > 0 && limit < len(values) {
-			values = values[:limit]
+	var cfg config.Config
+	if err := r.cm.Load(&cfg); err != nil {
+		return &pb.RecentConnectionsResponse{
+			Connections: nil,
+		}, fmt.Errorf("reading config for recent connections: %w\n", err)
```
mariusSincovici: error message should not include `\n`.
Why not return `nil, error..`?

daemon/rpc_get_recent_connections.go
```diff
@@ -17,24 +19,41 @@ func (r *RPC) GetRecentConnections(
...
-		rcValues[i] = &pb.RecentConnectionModel{
-			Country:            v.Country,
-			City:               v.City,
-			SpecificServer:     v.SpecificServer,
-			SpecificServerName: v.SpecificServerName,
-			Group:              v.Group,
-			CountryCode:        v.CountryCode,
-			ConnectionType:     v.ConnectionType,
+	// limit results if value is specified
+	if in.Limit != nil {
```
mariusSincovici: [small] `int(in.GetLimit())` returns `0` when is nil.

tray/menu.go
```diff
@@ -116,287 +160,275 @@ func addQuitItem(ti *Instance) {
...
+	if ti == nil {
+		return
+	}
+	if ti.state.vpnStatus != pb.ConnectionState_CONNECTED {
+		return
+	}
+
 	if serverName := ti.state.serverName(); serverName != "" {
-		label := labelServerPrefix + serverName
+		label := fmt.Sprintf("%s%s", labelServerPrefix, serverName)
```
mariusSincovici: [maybe small] I cannot add comments on the labels, but it is a bit strange, some labels end with space while others don't.
If someone would change the messages would need to know which is which.

tray/menu.go
```diff
@@ -116,287 +160,275 @@ func addQuitItem(ti *Instance) {
...
+
+func buildConnectToItem(ti *Instance) {
+	if ti == nil {
+		return
+	}
 	connectionSelector := systray.AddMenuItem(labelConnectionSelection, tooltipConnectionSelection)
 
 	ti.state.mu.RLock()
 	countries := append([]string(nil), ti.state.connSelector.countries...)
+	recentConnections := ti.recentConnections.GetRecentConnections()
```
mariusSincovici, line 295: here is not `slices.Clone(`, but in other places it is. Also from what I see the array is not changed only the entire variable is replaced, so maybe it works in all the places without clone? :thinking:

tray/monitor.go
```diff
@@ -35,14 +51,29 @@ func (ti *Instance) ping() bool {
 		}
 	}
 
-	return ti.updateDaemonConnectionStatus(daemonError)
+	ti.redraw(ti.updateDaemonConnectionStatus(daemonError))
+	if daemonError != "" {
+		return fmt.Errorf(daemonError)
```
mariusSincovici: If there is no format try to use `errors.New` instead.

tray/monitor.go
```diff
@@ -161,16 +188,16 @@ func (ti *Instance) updateSettings() bool {
 		ti.state.notificationsStatus = newNotificationsStatus
 
 		if newNotificationsStatus == Enabled {
-			defer ti.notifyForce("Notifications for NordVPN turned on")
+			defer ti.notifyForce(labelNotificationsOn)
```
mariusSincovici: This(if and else) are not displayed anymore when changing the tray on/off from CLI.

tray/monitor.go
```diff
@@ -35,14 +51,29 @@ func (ti *Instance) ping() bool {
 		}
 	}
 
-	return ti.updateDaemonConnectionStatus(daemonError)
+	ti.redraw(ti.updateDaemonConnectionStatus(daemonError))
```
mariusSincovici: The notifications that the daemon cannot be found are also not displayed anymore.
And if the daemon process is killed the menu is still displayed without any errors. Clicking on the menu items has no errors/effect.

tray/tray.go
```diff
@@ -196,22 +184,100 @@ func (ti *Instance) configureDebugMode() {
...
+			DefaultBackoffConfig(),
+			func(ctx context.Context) error {
+				return ti.ping()
+			},
+		)
+
+		// wait until we know whether tray should be enabled
+		err := RetryWithBackoff(
+			ctx,
+			BackoffConfig{MaxDelay: time.Second, MaxRetries: 10},
```
mariusSincovici: what happens after 10 retries, what is it displayed?

tray/tray.go
```diff
@@ -196,22 +184,100 @@ func (ti *Instance) configureDebugMode() {
...
+				if ti.initialDataLoadChan != nil {
+					select {
+					case ti.initialDataLoadChan <- struct{}{}:
+						close(ti.initialDataLoadChan)
+						ti.initialDataLoadChan = nil
+					default:
+						// Channel already being processed or closed
+					}
+				}
+				cancel()
```
mariusSincovici: probably not needed, since is called in defer.

tray/state_listener.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c *stateListener) handleAppState(ctx context.Context) {
+	for {
+		select {
+		case item := <-c.queue:
+			c.onDataFunc(item)
+
+		case <-ctx.Done():
+			log.Printf("%s %s exiting systray\n", logTag, internal.InfoPrefix)
```
mariusSincovici: If the daemon crashes then tray is not updated anymore.
Also could you change the log message, because the same message is for OnExit()

tray/tray.go
thegravi: Well, the object itself has a specific name, but I can also change this one.

tray/recents.go
thegravi: If the group is missing, we would not display the entry at all; therefore, I do not see the need to have this fallback mechanism.

tray/recents.go
thegravi: Nope, because the following line combines two parts. If the group is empty (hence unknown value), but the country is not, then the result of the label would be ` (Country)` and not `""`, which would force entry to be displayed.
The main purpose of this check is to validate whether we can build a valid label or not.
The best I can do here is to use a group variable after applying the `formatGroupTitle` helper.

tray/recents.go
thegravi: Was keeping a historical vibe with other functions, technically, there is no use here.

tray/expo_backoff.go
thegravi: We do, but that one is very specific/hardcoded for its use case. I can try to replace the existing one with a new one in another PR. I've borrowed the `jitter` param from that link.

daemon/rpc_get_recent_connections.go
thegravi: Oppsies

tray/menu.go
thegravi: In specialty servers PR they are cloned. Forgot to clone this one

tray/monitor.go
thegravi: `nordvpn set notify on/off` status notifications is working on my machine.
Are you referring to tray on/off status notifications form CLI?
If so, then it did not have the implementation for CLI: https://github.com/NordSecurity/nordvpn-linux/blob/9cbfd2734223b18bcddf57b5f394ff995af2ec73/tray/monitor.go#L163

mariusSincovici: It's ok, the notifications were not displayed because I've killed the daemon and then the nothing was observed.
https://github.com/NordSecurity/nordvpn-linux/pull/1078#discussion_r2330094800

tray/monitor.go
```diff
@@ -161,16 +188,16 @@ func (ti *Instance) updateSettings() bool {
 		ti.state.notificationsStatus = newNotificationsStatus
 
 		if newNotificationsStatus == Enabled {
-			defer ti.notifyForce("Notifications for NordVPN turned on")
+			defer ti.notifyForce(labelNotificationsOn)
 			defer log.Println(internal.InfoPrefix, "Notifications for NordVPN turned on")
```
mariusSincovici: `labelNotificationsOn` and `labelNotificationsOff`?

tray/connection_filter.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Set sets connection related settings
+func (s *connectionSettingsChangeSensor) Set(settings connectionSettings) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.settings != settings {
+		s.settings = settings
+		s.changed = true
```
mariusSincovici: Once set to true it will always be true, until a reset is made. Because of this at every settings changed the countries/recent will be rebuild.

Wouldn't make sense not to keep this as a member and `set` to return bool true when it was changed.

tray/connection_filter.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Detected returns whether settings have changed since the last check
+func (s *connectionSettingsChangeSensor) Detected() bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.changed
+}
+
+// Reset resets the sensor to its initial state
+func (s *connectionSettingsChangeSensor) Reset() {
```
mariusSincovici: Is this used?

tray/monitor.go
```diff
@@ -35,14 +51,29 @@ func (ti *Instance) ping() bool {
...
-	return ti.updateDaemonConnectionStatus(daemonError)
+	ti.redraw(ti.updateDaemonConnectionStatus(daemonError))
+	if daemonError != "" {
+		return fmt.Errorf(daemonError)
+	}
+
+	return nil
+}
+
+func (ti *Instance) update() {
```
mariusSincovici, line 94: won't this generate lots of notifications if multiple functions are failing or if some are failing but others work?
This is somehow related, some functions call `updateDaemonConnectionStatus` on gRPC error and other don't.

tray/recents.go
```diff
@@ -89,24 +104,42 @@ func connectByConnectionModel(ti *Instance, model *pb.RecentConnectionModel) boo
...
+		}
 		group_str := strings.ReplaceAll(model.Group.String(), "_", " ")
 		return ti.connect(model.CountryCode, group_str)
 
 	case config.ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP:
+		if model.SpecificServer == "" || model.Group == config.ServerGroup_UNDEFINED {
+			return false
+		}
 		group_str := strings.ReplaceAll(model.Group.String(), "_", " ")
 		return ti.connect(model.SpecificServer, group_str)
```
mariusSincovici, line 149: for `ServerSelectionRule_NONE` would it make sense to log a message?

tray/monitor.go
thegravi: They will be buffered by the OS and only the last one would be visible on tray itself.

protobuf/daemon/state.proto
```diff
@@ -28,6 +28,10 @@ message AccountModification {
   optional string expires_at = 1;
 }
 
+message VersionHealthStatus {
+    int32 status_code = 1;
```
mariusSincovici, line 32: Maybe it should include also the version number?

tray/menu.go
```diff
@@ -57,57 +57,101 @@ const (
 	tooltipSettings            = "Configure application preferences"
 	tooltipNotifications       = "Toggle desktop notifications"
 	tooltipTrayIcon            = "Show or hide tray icon"
+
+	// System messages
+	msgShutdownNotification = "Shutting down norduserd. To restart the process, run the \"nordvpn set tray on command\"."
```
mariusSincovici: wrong location for second quotes: 
`\"nordvpn set tray on command\"` => `\"nordvpn set tray on\" command`

tray/monitor.go
```diff
@@ -17,32 +22,153 @@ import (
...
+
+var (
+	daemonServiceNotActive        = errors.New("daemon service not active")
+	daemonConnectivityCheckPeriod = time.Second * 5
+)
+
+func (ti *Instance) handleVersionHealthChange(health *pb.VersionHealthStatus) {
 	daemonError := ""
+	switch int64(health.StatusCode) {
+	case internal.CodeOffline:
```
mariusSincovici, line 38: Is this ever sent for version API?

tray/monitor.go
mariusSincovici: They are not combined and now at every daemon restart `systemctl restart nordvpnd` results in 2 notifications, login success + reconnect to the daemon

tray/tray.go
```diff
@@ -196,22 +186,137 @@ func (ti *Instance) configureDebugMode() {
...
 
+func (ti *Instance) onDaemonStateEvent(item *pb.AppState) {
+	switch st := item.GetState().(type) {
+	case *pb.AppState_Error:
+		log.Printf("%s %s Received daemon error state\n", logTag, internal.ErrorPrefix)
+		ti.updateDaemonConnectionStatus(internal.ErrDaemonConnectionRefused.Error())
+
+	case *pb.AppState_ConnectionStatus:
+		ti.updateVpnStatus()
+		ti.updateRecentConnections()
```
mariusSincovici, line 201: [for later] it might be good to have an event on its own, because this will fetch the recent connections at every status changed, connecting, connected, disconnected.

tray/monitor.go
```diff
@@ -17,32 +22,153 @@ import (
...
+	case internal.CodeOutdated:
+		daemonError = cli.ErrUpdateAvailable.Error()
+	case internal.CodeSuccess:
+		daemonError = ""
+	default:
+		// For unknown status codes, assume success (no error)
+		daemonError = ""
+	}
+
+	ti.redraw(ti.updateDaemonConnectionStatus(daemonError))
```
mariusSincovici: This message will never remain because every 5 seconds will be replaced by the daemon monitoring.

tray/monitor.go
```diff
@@ -17,32 +22,153 @@ import (
...
+	}
+
+	status := strings.TrimSpace(output)
+	status = strings.ToLower(status)
+
+	return checkServiceActivity(status)
+}
+
+// startDaemonConnectivityMonitor starts periodic daemon connectivity monitoring
+func (ti *Instance) startDaemonConnectivityMonitor(ctx context.Context) {
```
mariusSincovici: Why not use instead the gRPC connection to monitor status changes?

daemon/job_appversion.go
```diff
@@ -1,11 +1,20 @@
...
+	"github.com/NordSecurity/nordvpn-linux/daemon/state"
+	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/coreos/go-semver/semver"
 )
 
-func JobVersionCheck(dm *DataManager, api *RepoAPI) func() {
+func JobVersionCheck(dm *DataManager, api *RepoAPI, statePublisher *state.StatePublisher) func() {
 	return func() {
+		// Always publish version status at the end
+		defer publishVersionHealthStatus(dm, statePublisher)
```
mariusSincovici: This can result in flickering/incorrect status being displayed into the tray because if there is an error at fetching the data, then the update available is remove `dm.SetVersionData(vdata.version, false)`. That will result in a notification being sent to the tray with `CodeSuccess`.

tray/monitor.go
thegravi: Nah, just like in CLI. Kept it for "same handling in case source would provide it one day". Would gladly even remove.

daemon/state/state.go
```diff
@@ -125,6 +125,16 @@ func (s *StatePublisher) NotifySubscriptionChanged(e *pb.AccountModification) er
 	return nil
 }
 
+func (s *StatePublisher) NotifyVersionHealth(healthStatus *pb.VersionHealthStatus) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	log.Printf(internal.DebugPrefix+" notifying about version health change: status_code=%d", healthStatus.StatusCode)
```
devzbysiu: [small] missing `\n`

daemon/recents/recent_connections.go
```diff
@@ -43,16 +47,28 @@ func (r *RecentConnectionsStore) Get() ([]Model, error) {
...
 	return slices.IndexFunc(list, func(m Model) bool {
-		return m == model
+		return m.Country == model.Country &&
+			m.City == model.City &&
+			m.Group == model.Group &&
+			m.CountryCode == model.CountryCode &&
+			m.SpecificServerName == model.SpecificServerName &&
+			m.SpecificServer == model.SpecificServer &&
+			m.ConnectionType == model.ConnectionType &&
+			slices.Equal(m.ServerTechnologies, model.ServerTechnologies)
```
devzbysiu, line 71: Is it possible that the slices will have the same values, but in different order?

daemon/rpc_connect.go
```diff
@@ -286,31 +287,17 @@ func (r *RPC) connect(
...
+			serverTechs = append(serverTechs, v.ID)
 		}
 
-		// do not add anything unrelated to connection type
-		if recentModel.ConnectionType == config.ServerSelectionRule_COUNTRY ||
-			recentModel.ConnectionType == config.ServerSelectionRule_COUNTRY_WITH_GROUP {
-			recentModel.City = ""
+		recentModel, err := buildRecentConnectionModel(serverTechs, event, parameters)
+		if err != nil {
+			log.Printf("%s Failed to build recent VPN connection model: %s", internal.WarningPrefix, err)
```
devzbysiu: [small] missing `\n`

daemon/recents/recent_connections.go
thegravi: The order is unaltered and comes from the server. I will sort it additionally just to be sure this does not happen.

PR #1079, opened 2025-09-03
LVPN-8888: Implement automation tests for RC feature - Part 1

bbrutall:
- added additional print for 'useOnlyLocalConfig'
- added 'test_restart_causes_daemon_to_consume_local_files'
- added 'test_equivalence_of_local_and_remote_config_files'
- added 'test_meshnet_feature_availability_based_on_remote_config'
- added 'test_local_config_usage_via_systemd_env'

### DIFF
--- a/config/remote/remote.go
+++ b/config/remote/remote.go
@@ -167,7 +167,7 @@ func (c *CdnRemoteConfig) Load() error {
 
 	useOnlyLocalConfig := internal.IsDevEnv(c.appEnvironment) && os.Getenv(envUseLocalConfig) != "" // forced load from disk?
 	if useOnlyLocalConfig {
-		log.Printf("%s Ignoring remote config, using only local", internal.InfoPrefix)
+		log.Printf("%s Ignoring remote config, using only local\n", internal.InfoPrefix)
 	} else {
 		if needReload, err = c.download(); err != nil {
 			return fmt.Errorf("downloading remote config: %w", err)
--- a/test/qa/lib/daemon.py
+++ b/test/qa/lib/daemon.py
@@ -2,7 +2,6 @@
 import glob
 import os
 import socket
-import subprocess
 import time
 
 import sh
@@ -202,12 +201,7 @@ def get_env() ->str:
 
     :return: the active environment (DEV or PROD)
     """
-    result = subprocess.run(
-        ["nordvpn", "--version"],
-        check=True,
-        stdout=subprocess.PIPE,
-        text=True
-    )
-    if Env.DEV in result.stdout:
+    result = sh.nordvpn("--version")
+    if Env.DEV in result:
         return Env.DEV
     return Env.PROD
--- a/test/qa/lib/remote_config_manager.py
+++ b/test/qa/lib/remote_config_manager.py
@@ -1,7 +1,7 @@
 import os
 import json
-import subprocess
 import requests
+import sh
 from lib.remote_config import RemoteConfig
 
 from lib.daemon import Env
@@ -32,7 +32,7 @@ class RemoteConfigManager:
         if not os.path.exists(self.cache_dir):
             print(f"Directory '{self.cache_dir}' does not exist.")
             return False
-        subprocess.run(["sudo", "chmod", "-R", "775", self.cache_dir], check=True)
+        sh.sudo.chmod("-R", "775", self.cache_dir)
         print(f"Permissions set recursively to 775 for {self.cache_dir}.")
         return True
 
@@ -47,10 +47,12 @@ class RemoteConfigManager:
         """
         self.set_permissions_cache_dir()
         config_files = os.listdir(self.cache_dir)
-        full_file_paths = [
-            os.path.join(self.cache_dir, filename) for filename in config_files if
-            os.path.isfile(os.path.join(self.cache_dir, filename))
-        ]
+
+        full_file_paths = []
+        for filename in config_files:
+            full_path = os.path.join(self.cache_dir, filename)
+            if os.path.isfile(full_path):
+                full_file_paths.append(full_path)
 
         print(f"Local files: {full_file_paths}")
         return full_file_paths
@@ -128,9 +130,11 @@ class RemoteConfigManager:
 
         if not REMOTE_DIR:
             raise ValueError("REMOTE_DIR environment variable is not set.")
+        print(f"REMOTE_DIR: {REMOTE_DIR}")
 
         if not REMOTE_FILES:
             raise ValueError("REMOTE_FILES environment variable is not set.")
+        print(f"REMOTE_FILES: {REMOTE_FILES}")
 
         remote_files = REMOTE_FILES.split(",")
         remote_files = [file for file in remote_files if file.startswith(env)]
@@ -160,10 +164,10 @@ class RemoteConfigManager:
 
         :param file_path: The path to the file to delete.
 
-        :raises subprocess.CalledProcessError: If file deletion fails.
+        :raises sh.ErrorReturnCode: If file deletion fails.
         """
         print(f"Deleting file: {file_path}")
         try:
-            subprocess.run(["sudo", "rm", file_path], check=True)
-        except subprocess.CalledProcessError as e:
+            sh.sudo.rm(file_path)
+        except sh.ErrorReturnCode as e:
             print(f"Got an error while deleting {file_path}: {e}")
--- a/test/qa/test_remote_config.py
+++ b/test/qa/test_remote_config.py
@@ -3,6 +3,7 @@ import subprocess
 import uuid
 
 import pytest
+import sh
 
 from lib import daemon
 from lib.log_reader import LogReader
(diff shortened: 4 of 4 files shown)


test/qa/test_remote_config.py
```diff
@@ -41,7 +45,56 @@ def initialized_app_with_remote_config(
...
+
+    This fixture injects 'export RC_USE_LOCAL_CONFIG=1' into the 'nordvpnd' service file, simulating enabling the local config usage for tests.
+    After the test, it removes the injected line and reloads the systemd daemon to restore the initial system state.
+    """
+    service_path = "/etc/init.d/nordvpn"
+
+    # Print original service file for reference
+    print(
+        "Service original:\n"
+        + subprocess.run(
```
edvyno1: why not use in built python features instead of doing everything with bash commands via `subprocess.run`?

test/qa/lib/daemon.py
```diff
@@ -188,3 +193,19 @@ def get_status_data() -> dict:
...
     formatted_pairs = {(key.lower(), value.strip()) for key, value in colon_separated_pairs}
     return dict(formatted_pairs)
+
+def get_env() ->str:
+    """
+    Detects and returns the active environment (DEV or PROD) based on the NordVPN version output.
+
+    :return: the active environment (DEV or PROD)
+    """
+    result = subprocess.run(
```
edvyno1: the rest of the file is using `sh` library to run bash commands in regards to nordvpn, maybe it makes sense to either use `sh` for all or `subprocess` for all and not mix and match

test/qa/lib/daemon.py
```diff
@@ -188,3 +193,19 @@ def get_status_data() -> dict:
     colon_separated_pairs = (element.split(':') for element in lines)
     formatted_pairs = {(key.lower(), value.strip()) for key, value in colon_separated_pairs}
     return dict(formatted_pairs)
+
+def get_env() ->str:
```
edvyno1, line 198: with the current implementation we are only going to be testing with prod configs during the release pipeline, do we not want to test prod configs with every run and only do dev environment config testing if there is something new to test?

test/qa/test_remote_config.py
```diff
@@ -245,3 +335,148 @@ def test_hash_modification_and_daemon_restart(
...
+            assert (
+                expected_error not in output
+            ), f"Command {command} output is not as expected for enabled meshnet. Output:\n{output}"
+        else:
+            assert (
+                expected_error in output
+            ), f"Command {command} output is not as expected for disabled meshnet. Output:\n{output}"
+
+
+@pytest.mark.local_config_only
```
edvyno1: please document how to properly run only marked tests in the case where we would trigger from config repo

config/remote/remote.go
```diff
@@ -142,12 +142,13 @@ func (c *CdnRemoteConfig) Load() error {
 	needReload := false
 
 	useOnlyLocalConfig := internal.IsDevEnv(c.appEnvironment) && os.Getenv(envUseLocalConfig) != "" // forced load from disk?
-	if !useOnlyLocalConfig {
+	if useOnlyLocalConfig {
+		log.Printf("%s Ignoring remote config, using only local", internal.InfoPrefix)
```
mariusSincovici: Use `log.Println(internal.InfoPrefix, msg..)` or add `\n` at the end of the message.

test/qa/lib/remote_config_manager.py
```diff
@@ -46,10 +48,11 @@ def get_local_files(self):
         self.set_permissions_cache_dir()
         config_files = os.listdir(self.cache_dir)
         full_file_paths = [
-            os.path.join(self.cache_dir, filename) for filename in config_files
+            os.path.join(self.cache_dir, filename) for filename in config_files if
```
mariusSincovici: Maybe this should go into a function that returns the needed info, because now `os.path.join(self.cache_dir, filename)` is duplicated and if one changes the other also needs to be updated and are bigger chances to miss.

test/qa/test_remote_config.py
```diff
@@ -1,31 +1,35 @@
...
 RC_LOCAL_MESSAGES = [
-    "[Info] feature [meshnet] config loaded from: /var/lib/nordvpn/conf",
-    "[Info] feature [libtelio] config loaded from: /var/lib/nordvpn/conf",
-    "[Info] feature [nordvpn] config loaded from: /var/lib/nordvpn/conf",
+    f"[Info] feature [meshnet] config loaded from: {LOCAL_CACHE_DIR}",
+    f"[Info] feature [libtelio] config loaded from: {LOCAL_CACHE_DIR}",
+    f"[Info] feature [nordvpn] config loaded from: {LOCAL_CACHE_DIR}",
 ]
 
+RC_USE_LOCAL_CONFIG_MESSAGE = ["[Info]Ignoring remote config, using only local"]
```
mariusSincovici: I think this is not the same as [logged message](https://github.com/NordSecurity/nordvpn-linux/pull/1079/files#diff-12a8e33ecaa5812ac633bceb003e7de5dd938ab47dd43b16b045bd1e9ba8745dR146). In logs it contains `space` between `[Info]Ignore`?. :thinking:

test/qa/test_remote_config.py
```diff
@@ -135,9 +188,46 @@ def test_local_hash_files_removal_and_daemon_restart(
...
+
+    daemon.restart()
+
+    res, not_found = daemon_log_reader.wait_for_messages(
+        messages=RC_INITIAL_RUN_MESSAGES,
+        cursor=cursor,
+        return_not_found=True,
+    )
+
+    assert set(not_found) == set(
```
mariusSincovici, line 209: shouldn't also `res` be checked?

test/qa/test_remote_config.py
```diff
@@ -41,7 +45,56 @@ def initialized_app_with_remote_config(
...
+
+    # Print service file after modification
+    print(
+        "Service after:\n"
+        + subprocess.run(
+            ["cat", service_path], check=True, stdout=subprocess.PIPE, text=True
+        ).stdout
+    )
+
+    yield
```
mariusSincovici, line 73: shouldn't the daemon be restarted after config it was changed, before and after yield?
Same shouldn't `systemctl daemon-reload` be executed also after yield?

test/qa/test_remote_config.py
```diff
@@ -41,7 +45,56 @@ def initialized_app_with_remote_config(
...
+    print(
+        "Service original:\n"
+        + subprocess.run(
+            ["cat", service_path], check=True, stdout=subprocess.PIPE, text=True
+        ).stdout
+    )
+
+    # Insert environment variable into systemd service file
+    subprocess.run(
+        ["sudo", "sed", "-i", r"1a export RC_USE_LOCAL_CONFIG=1", service_path]
```
DoYouEvenCpp: Is this somewhere restored to default state once test (fixture) is executed?
I guess while running this on a host machine, this should get restored after tests are executed.

test/qa/test_remote_config.py
```diff
@@ -41,7 +45,56 @@ def initialized_app_with_remote_config(
...
+    """
+    Pytest fixture to temporarily inject 'RC_USE_LOCAL_CONFIG=1' environment variable into the 'nordvpnd' service file.
+
+    This fixture injects 'export RC_USE_LOCAL_CONFIG=1' into the 'nordvpnd' service file, simulating enabling the local config usage for tests.
+    After the test, it removes the injected line and reloads the systemd daemon to restore the initial system state.
+    """
+    service_path = "/etc/init.d/nordvpn"
+
+    # Print original service file for reference
+    print(
```
DoYouEvenCpp: Why plain `print` and not logging library?

test/qa/test_remote_config.py
bbrutall: Replaced with `sh`

test/qa/lib/daemon.py
bbrutall: Replaced with `sh`. Thanks!

test/qa/lib/daemon.py
bbrutall: dev configs - for dev builds, prod - for prod builds.

test/qa/test_remote_config.py
bbrutall: Will be introduced in a separate PR.

config/remote/remote.go
bbrutall: Done!

test/qa/test_remote_config.py
bbrutall: Done!

test/qa/test_remote_config.py
bbrutall: Makes sense like an additional verification, but not critical. I've already added it.

test/qa/test_remote_config.py
bbrutall: Will be introduced in a separate PR.

test/qa/test_remote_config.py
bbrutall: Done!

test/qa/test_remote_config.py
bbrutall: Yes, in the teardown after the `yield` keyword.

test/qa/lib/remote_config_manager.py
bbrutall: Could you please elaborate? Not clear where is a duplication if it's a way how to use list comprehension.

keliramu (approved):

PR #1080, opened 2025-09-04
LVPN-9045: Implement 'specialty server' selection for systray

thegravi:
Changes:
- SysTray now allows selecting any specialty server (a.k.a group) based on connection setting (tech, proto, etc.)


tray/tray.go
```diff
@@ -74,6 +75,21 @@ func (cp *ConnectionSelector) listCountries(client pb.DaemonClient) ([]string, e
 	return out, nil
 }
 
+func (cp *ConnectionSelector) listSpecialtyServers(client pb.DaemonClient) ([]string, error) {
+	resp, err := client.Groups(context.Background(), &pb.Empty{})
+	if err != nil {
+		return nil, err
+	}
+	log.Println("GROUPS:", resp)
```
DoYouEvenCpp: Looks like a leftover. :wink: 

If not, then please make it a proper log (eg. add severity, make it non all-capitalized, etc.).

tray/tray.go
```diff
@@ -74,6 +75,21 @@ func (cp *ConnectionSelector) listCountries(client pb.DaemonClient) ([]string, e
...
+	result := sortedConnections(resp.Servers)
+
+	cp.mu.Lock()
+	cp.specialtyServers = result
+	out := slices.Clone(cp.specialtyServers)
```
DoYouEvenCpp: Is this extra `out` here needed because fetching list of servers could be triggered either via incoming `DaemonStateEvent` or by the user interaction?

If that's the case then I'm a bit puzzled why `listCountries` and `listSpecialityServers` update internal state of `ConnectionSelector`.  I guess (and this might be a bit of a nit here, sorry! :wink: ) it's either about a bit off names used (`list` implies that it simply... well,  list thingies and doesn't update internal state) or should we keep basic getter methods (that would return simply list of servers) and other methods that actually do update to the internal bits of the `ConnectionSelecector`.

Or am I trying to overengineer here something?

tray/tray.go
mariusSincovici: leftover?

DoYouEvenCpp (approved):

tray/tray.go
```diff
@@ -211,7 +212,10 @@ func (ti *Instance) onDaemonStateEvent(item *pb.AppState) {
 		changed = ti.updateDaemonConnectionStatus(internal.ErrDaemonConnectionRefused.Error())
 
 	case *pb.AppState_ConnectionStatus:
-		changed = ti.updateVpnStatus() || ti.updateRecentConnections()
+		log.Println("new connection status:", st.ConnectionStatus.GetState())
```
devzbysiu: Missing log level?

daemon/rpc_logout.go
```diff
@@ -20,6 +22,10 @@ func (r *RPC) Logout(ctx context.Context, in *pb.LogoutRequest) (*pb.Payload, er
 		DisconnectFunc:               r.DoDisconnect,
 	})
 
+	if err := r.recentVPNConnStore.Clean(); err != nil {
+		log.Printf("%s %s failed to clean recent connections on logout: %v", "[rpc]", internal.WarningPrefix, err)
```
DoYouEvenCpp: [nit] Missing new line

DoYouEvenCpp (approved):

devzbysiu (approved):

PR #1083, opened 2025-09-04
LVPN-9039: fix exponential backoff for remote config on start

keliramu:


dfetti (approved):

config/remote/remote.go
```diff
@@ -117,14 +117,45 @@ func isNetworkRetryable(err error) bool {
...
+		}
+
+		// check for specific server errors
+		if errors.Is(currentErr, core.ErrServerInternal) ||
+			errors.Is(currentErr, core.ErrTooManyRequests) {
+			return true
+		}
+
+		// unwrap to check the next error in the chain
+		currentErr = errors.Unwrap(currentErr)
```
mariusSincovici: This is not needed, because `errors.As` already goes thru the entire error tree.

keliramu: Ok, indeed, removed loop, thx

DoYouEvenCpp (approved):

PR #1085, opened 2025-09-08
LVPN-8154: add more fields and filters data to the API request event

bartoszWojciechO:


request/publishing.go
```diff
@@ -22,7 +27,70 @@ func NewPublishingRoundTripper(
...
+}
+
+func getParameter(parameter string, values url.Values) string {
+	parameterValue, ok := values[parameter]
+	if !ok {
+		return ""
+	}
+
+	if len(parameterValue) != 1 || parameterValue[0] == "" {
+		log.Println(internal.WarningPrefix, "invalid value of limit parameter in the api call URL")
```
DoYouEvenCpp: Can we add the `parameter` name here, so in case of debugging it's a bit more clear what param was actually empty (invalid)?

request/publishing.go
```diff
@@ -22,7 +27,70 @@ func NewPublishingRoundTripper(
...
+		if len(value) != 0 && value[0] != "" {
+			queryString = fmt.Sprintf("%s=%s", query, value[0])
+		} else {
+			queryString = query
+		}
+
+		if isFilter {
+			filters = append(filters, queryString)
+		}
+		if isField {
```
DoYouEvenCpp: [ultra petty nit] can one string be simultaneously filter and field? If that's not the case then maybe we can `else if` here so the other if check isn't performed?

request/publishing.go
```diff
@@ -22,7 +27,70 @@ func NewPublishingRoundTripper(
...
+		limit:   limit,
+		offset:  offset,
+	}
+}
+
 func (rt *PublishingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	parameters := getURLParameters(req.URL)
+
 	startTime := time.Now()
 	rt.publisher.Publish(events.DataRequestAPI{
```
DoYouEvenCpp, line 95: Don't we need to provide the newly added parameters to the attempt request as well?

request/publishing.go
```diff
@@ -22,7 +27,70 @@ func NewPublishingRoundTripper(
...
+
+	return fmt.Sprintf("%s=%s", parameter, parameterValue[0])
+}
+
+func getURLParameters(url *url.URL) urlParameters {
+	filters := []string{}
+	fields := []string{}
+	queryMap := url.Query()
+	for query, value := range queryMap {
+		isFilter := strings.Contains(query, "filters")
```
mariusSincovici: Shouldn't be `HasPrefix` instead of `Contains`, to ensure it is not matching if something contains in the middle?

PR #1086, opened 2025-09-09
    LVPN-6748: add ARP ignore setting

bartoszWojciechO:
Attacker can detect whether user is connected to the VPN by utilizing
    ARP. App will now set net.ipv4.conf.all.arp_ignore to 1 by default. This
    was added as a setting, because it could break connectivity in some
    non-standard cases(for example with virtualization).

    Because net.ipv4.conf.all.arp_ignore can have multiple values configured
    by the user, SysctlSetter implementation was configured to always
    restore old value configured by the user. Previously it was restoring to
    the default value assumed by the app.


cmd/daemon/main.go
```diff
@@ -467,11 +467,14 @@ func main() {
 			kernel.NewSysctlSetter(
 				forwarder.Ipv4fwdKernelParamName,
 				1,
-				0,
 			)),
 		cfg.FirewallMark,
 		cfg.LanDiscovery,
 		ipv6.NewIpv6(),
+		cfg.ARPIgnore.Get(),
+		kernel.NewSysctlSetter(
```
mariusSincovici, line 475: When running in docker, does it require changes to the help pages? https://support.nordvpn.com/hc/en-us/articles/20465811527057-How-to-build-the-NordVPN-Docker-image

daemon/rpc_set_arp_ignore.go
```diff
@@ -0,0 +1,49 @@
...
+	if cfg.ARPIgnore.Get() == in.Enabled {
+		return &pb.Payload{
+			Type: internal.CodeNothingToDo,
+		}, nil
+	}
+
+	if err := r.netw.SetARPIgnore(in.Enabled); err != nil {
+		log.Println(internal.ErrorPrefix, "failed to set ARP ignore:", err)
+		return &pb.Payload{
+			Type: internal.CodeConfigError,
```
mariusSincovici: [small] Shouldn't this be a different error? Like for example kill switch.

kernel/sysctl_setter.go
```diff
@@ -44,6 +42,7 @@ func (s *SysctlSetterImpl) Set() error {
 		}
 
 		if values[s.paramName] != s.desiredValue {
+			s.unwantedValue = values[s.paramName]
```
mariusSincovici: [small] Not made in the PR, but shouldn't be `previousValue`| `initialValue` or something?

config/config.go
```diff
@@ -54,6 +54,7 @@ type Config struct {
 	RCLastUpdate    time.Time           `json:"rc_last_update,omitempty"`
 	// Indicates whether the virtual servers are used. True by default
 	VirtualLocation TrueField `json:"virtual_location,omitempty"`
+	ARPIgnore       TrueField `json:"arp_discovery,omitempty"`
```
devzbysiu: Maybe the json field should be named `arp_ignore` also?

cmd/daemon/main.go
bartoszWojciechO: Docker connected to the VPN will run fine with `arp-ignore` enabled. This is for even more exotic use cases, like routing to a container or something.

networker/networker.go
```diff
@@ -1603,3 +1633,29 @@ func (netw *Combined) GetConnectionParameters() (vpn.ServerData, bool) {
 
 	return netw.vpnet.GetConnectionParameters()
 }
+
+// SetARPIgnore sets arp ignore to the desired value if VPN connection is active. Networker will set arp ignore
+// ccordingly upon subsequent connections. Setting arp ignore to value previously configured is a noop.
```
edvyno1: [typo] ccordingly -> accordingly

networker/networker_test.go
```diff
@@ -1916,3 +2050,138 @@ func TestExitNodeLanAvailability(t *testing.T) {
...
+		},
+		{
+			name:                      "unset ignore is not applied to kernel when VPN is not active",
+			vpn:                       mock.WorkingInactiveVPN{},
+			currentARPIgnoreKernel:    true,
+			currentARPIgnoreNetworker: false,
+			targetARPIgnore:           false,
+			expectedARPIgnore:         true,
+			errExpected:               false,
+		},
```
devzbysiu: Isn't this a duplicate of the one above?

kernel/sysctl_setter.go
```diff
@@ -13,22 +13,20 @@ type SysctlSetter interface {
 }
 
 type SysctlSetterImpl struct {
-	paramName     string
-	desiredValue  int
-	unwantedValue int
-	changed       bool
+	paramName    string
+	desiredValue int
+	intialValue  int
```
devzbysiu: [typo] `intial`

cli/cli_set_arp_ignore.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+	if err != nil {
+		return formatError(err)
+	}
+
+	switch resp.Type {
+	case internal.CodeConfigError:
+		return formatError(ErrConfig)
+	case internal.CodeARPIgnoreError:
+		return formatError(internal.ErrUnhandled)
```
devzbysiu, line 35: Hmm, so we'll return generic error here in case of arp ignore issue? 🤔

test/qa/test_settings.py
```diff
@@ -515,3 +515,32 @@ def test_lan_discovery_on_off():
 
     assert "LAN Discovery has been successfully set to 'disabled'." in sh.nordvpn.set("lan-discovery", "off")
     assert not settings.is_lan_discovery_enabled()
+
+
+def test_set_arp_ignore():
```
devzbysiu: This could also assert on the settings first, before connecting, so it will catch potential regression of default value of this setting

networker/networker_test.go
```diff
@@ -1916,3 +2050,129 @@ func TestExitNodeLanAvailability(t *testing.T) {
...
+		{
+			name:                      "set is always applied to kernel even if already set when VPN is acive",
```
devzbysiu: [typo] `acive`, also below

cli/cli_set_arp_ignore.go
bartoszWojciechO: It's an unexpected when we fail to set the kernel parameter, so generic error is best we can give to the user.

test/qa/test_settings.py
bartoszWojciechO: Actually this test had to be removed, since docker doesn't allow setting most kernel parameters(it derives them from the host if I remember correctly). So this test would result in some flakiness.

PR #1087, opened 2025-09-10
LVPN-9051: Rework flush on disconnect to include mangle table

edvyno1:


daemon/firewall/iptables/iptables.go
```diff
@@ -22,6 +22,8 @@ const (
 	defaultComment = "nordvpn"
 )
 
+var usedIPTables = [...]string{"mangle", "filter"}
```
thegravi, line 25: [small] Why do we need a fixed array here? Isn't slice enough?

edvyno1: is there much of a difference in this scenario? just made it a static size because it's not being changed ever, if we add more tables, we will just add more elements to the array

PR #1088, opened 2025-09-11
LVPN-9058: Add snap permissions screen

dfetti:
When the GUI app starts it fetches the snap permissions. 
If any permission required is missing it will show a screen with the commands that the user must run to provide us the needed snap permissions.


gui/lib/daemon/error_screen.dart
```diff
@@ -247,3 +156,22 @@ ErrorData _dataForAccountError(WidgetRef ref) {
     ),
   );
 }
+
+CopyField _buildCopyFieldForSocketNotFound() {
+  if (Platform.environment.containsKey('SNAP_NAME')) {
+    return CopyField(items: [CopyItem(command: "sudo snap start nordvpn")]);
+  } else {
```
devzbysiu: [small] `else` can be skipped here

gui/lib/daemon/error_screen.dart
```diff
@@ -247,3 +156,22 @@ ErrorData _dataForAccountError(WidgetRef ref) {
     ),
   );
 }
+
+CopyField _buildCopyFieldForSocketNotFound() {
+  if (Platform.environment.containsKey('SNAP_NAME')) {
```
devzbysiu: Could this check be reused? I see the same thing is used in `snap_permissions_provider.dart`

gui/lib/data/providers/snap_permissions_provider.dart
```diff
@@ -0,0 +1,94 @@
...
+  "system-observe",
+  "hardware-observe",
+];
+
+// Observer of the snap permissions must implement this
+abstract class SnapPermissionsObserver {
+  void onPermissionsChanged();
+}
+
+// ConsentStatus is a provider that controls the ConsentScreen
```
devzbysiu: Copy Paste? :smile:

gui/lib/snap/snap_screen.dart
```diff
@@ -0,0 +1,87 @@
...
+import 'package:nordvpn/internal/urls.dart';
+import 'package:nordvpn/theme/app_theme.dart';
+import 'package:nordvpn/theme/error_screen_theme.dart';
+import 'package:nordvpn/widgets/copy_field.dart';
+import 'package:nordvpn/widgets/dynamic_theme_image.dart';
+import 'package:nordvpn/widgets/full_screen_scaffold.dart';
+import 'package:nordvpn/widgets/loading_button.dart';
+import 'package:nordvpn/widgets/rich_text_markdown_links.dart';
+
+// ConsentScreen - represents the entire consent screen in which also the
```
devzbysiu: Copy Paste :smile:

gui/lib/data/providers/consent_status_provider.dart
```diff
@@ -12,21 +13,27 @@ enum ConsentLevel { none, acceptedAll, essentialOnly }
...
   }
 
   void _registerNotifications() {
-    final notification = ref.read(appStateProvider);
-    notification.addSettingsObserver(this);
-    ref.onDispose(() => _dispose(notification));
+    final appNotification = ref.read(appStateProvider);
+    final snapNotification = ref.read(snapPermissionsProvider.notifier);
+    appNotification.addSettingsObserver(this);
+    snapNotification.addSnapObserver(this);
```
devzbysiu, line 27: Currently, `snapPermissionsProvider` is auto disposable. When it gets disposed by riverpod and later recreated, you are not re-registering observer on this new instance :thinking: , I think it's working now because `snapPermissionsProvider` is also used in router which is not disposed so `snapPermissionsProvider` is also not disposed, but this can be problematic later, or even in tests.
I think the best would be to make `SnapPermissions` non auto-disposable with `keepAlive: true`

gui/lib/data/providers/snap_permissions_provider.dart
```diff
@@ -0,0 +1,94 @@
...
+  void _notifySnapChanged() {
+    if (_snapObservers.isEmpty) {
+      return;
+    }
+    for (final observer in _snapObservers) {
+      observer.onPermissionsChanged();
+    }
+  }
+
+  static bool _isSnapContext() {
```
devzbysiu: Does it have to be `static`? Similarly with `_getMissingPermissions` below

gui/lib/widgets/copy_field.dart
```diff
@@ -0,0 +1,103 @@
...
+final class CopyItem {
+  final String command;
+  final String? description;
+  const CopyItem({required this.command, this.description});
+}
+
+final class CopyField extends StatelessWidget {
+  final List<CopyItem> items;
+
+  const CopyField({required this.items});
```
devzbysiu: Now, when `CopyField` is public, constructor should have `key` param

gui/lib/snap/snap_screen.dart
```diff
@@ -0,0 +1,87 @@
...
+            ],
+          ),
+        ),
+      ),
+    );
+  }
+
+  List<CopyItem> _buildNeededCommands(List<String>? missingPermissions) {
+    String commands = "";
+    missingPermissions?.forEach((item) {
```
devzbysiu: [small] this can be
```
final commands = missingPermissions
      .map((item) => "sudo snap connect nordvpn:$item")
      .join("\n");
```
to avoid `commands` reassignment and having to prepend it before next command. Not a big deal, so leaving it up to you

gui/lib/snap/snap_screen.dart
```diff
@@ -0,0 +1,87 @@
...
+              ),
+              const Expanded(child: SizedBox.shrink()),
+            ],
+          ),
+        ),
+      ),
+    );
+  }
+
+  List<CopyItem> _buildNeededCommands(List<String>? missingPermissions) {
```
devzbysiu: [small] maybe we could return just single `CopyItem` here to make it clear that this actually produces single item (and then wrap it in a list above where `CopyField` is used)

gui/lib/widgets/copy_field.dart
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+
+  Widget _textArea(
+    AppTheme appTheme,
+    CopyFieldTheme copyFieldTheme,
+    String text,
+  ) {
+    final lines = max('\n'.allMatches(text).length, 1);
+    final maxLines = min(4, lines);
```
devzbysiu, line 78: I would suggest to leave a note why `4`, so someone from "outside" will know

gui/lib/data/providers/snap_permissions_provider.dart
dfetti: I removed static from _getMissingPermissions, but I left it static so I can reuse it easily from outside.

DoYouEvenCpp (approved):

PR #1090, opened 2025-09-12
LVPN-9096: Create symlink for sqlite in posttrans script

mariusSincovici:
For RPM packets are 2 different scripts that are executed at package install/update: `postinstall` and `posttrans`. 
They are executed in different times while installing. The `posttrans` is executed the latest, after the update/install process takes place.  `postinstall` is executed before new package replaces the files of the old one with its own.
Since sqlite symlink is not part of the new package the file is deleted after `postinstall` is executed. Creating it into `posttrans` ensures that the file is not deleted.

### DIFF
--- a/contrib/scriptlets/rpm/post
+++ b/contrib/scriptlets/rpm/post
@@ -27,13 +27,6 @@ helpMessage="If you need help using the app, use the command 'nordvpn --help'."
 # update configuration and shared library cache for linker to find .so files
 echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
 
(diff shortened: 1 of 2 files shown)


contrib/scriptlets/rpm/post
```diff
@@ -34,11 +34,6 @@ echo "/usr/lib/nordvpn" > /etc/ld.so.conf.d/nordvpn.conf
 # literally nothing) fixes the issue.
 ldconfig
```
edvyno1: this one is no longer needed as well I assume as it's not called in this scriptlet anymore afterwards, also move the comment to the posttrans file

PR #1091, opened 2025-09-12
LVPN-8374: Migrate internal libraries

keliramu:


ci/docker/ruster/Dockerfile
```diff
@@ -2,6 +2,13 @@ FROM debian:bullseye-slim
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+COPY --from=ghcr.io/nordsecurity/uniffi-generators:v0.28.3-4 /bin/uniffi-bindgen /bin
+COPY --from=ghcr.io/nordsecurity/uniffi-generators:v0.28.3-4 /bin/uniffi-bindgen-cs /bin
+COPY --from=ghcr.io/nordsecurity/uniffi-generators:v0.28.3-4 /bin/uniffi-bindgen-go /bin
```
devzbysiu, line 6: Do we need all or just `go` is enough?

ci/docker/ruster/Dockerfile
```diff
@@ -79,6 +89,25 @@ RUN set -eux; \
     ldconfig; \
     rm -rf /var/lib/apt/lists/*
 
+RUN ARCH=$(uname -m) && \
```
devzbysiu, line 90: [small] maybe this could be moved to bash script and run here, to avoid those `\` everywhere

ci/test.sh
```diff
@@ -21,8 +21,16 @@ parallel=""
 if [ "${1:-""}" = "full" ]; then
 	# Apply moose patch in case compiling with moose
 	trap '${WORKDIR}/ci/remove_private_bindings.sh moose/events; ${WORKDIR}/ci/remove_private_bindings.sh moose/worker;' EXIT
-	source "${WORKDIR}"/ci/add_private_bindings.sh moose/events ./third-party/moose-events/moosenordvpnappgo/v16
-	source "${WORKDIR}"/ci/add_private_bindings.sh moose/worker ./third-party/moose-worker/mooseworkergo/v16
+	source "${WORKDIR}"/ci/add_private_bindings.sh moose/events ./third-party/moose-events/moosenordvpnappgo/v17
+	source "${WORKDIR}"/ci/add_private_bindings.sh moose/worker ./third-party/moose-worker/mooseworkergo/v17
+
+	# TODO/FIXME: workarround for the moose problem, remove this code after moose resolves this situation
```
bartoszWojciechO: typo: workarround => workaround

devzbysiu (approved):

ci/docker/ruster/Dockerfile
keliramu: This part is copy-pasted from libtelio build docker and I would prefer to leave this on the first level and visible/not hidden for later to be able to compare quickly if libtelio would change something.

ci/docker/ruster/Dockerfile
keliramu: Removed

ci/test.sh
keliramu: fixed.

PR #1110, opened 2025-09-25
LVPN-9130 Build deb and rpm packages for GUI in parallel to core

DoYouEvenCpp:
Changes under this PR:
- scripts for building RPM/DEB packages are now also crafting GUI packages
- adjusted `ci/test_reproducible_builds.sh` to handle multiple packages 
- adjusted `ci/nfpm/build_packages_resources.sh` so it now also builds GUI packages
  - it simply reuses existing scripts, with some extra variables
  - for now this step is conditionally enabled only if run under docker container
- minor updates to make the whole machinery working properly after open-sourcing GUI
- GUI package is now versioned on par with core-application

Tested locally/on CI pipelines.


ci/nfpm/build_packages_resources.sh
```diff
@@ -86,3 +86,46 @@ mv "${WORKDIR}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}"
...
+  # Build GUI package using the existing GUI build script
+  # Set auxiliary environment variables for the GUI script
+  export SKIP_DIST_CLEAN="true"  # Don't clean dist dir since we're integrating
+  export CUSTOM_SOURCE_DIR="${WORKDIR}/bin/${ARCH}/gui"  # Use our built GUI binaries
+
+  # Convert aarch64 to arm64 for GUI script compatibility
+  GUI_ARCH="${ARCH}"
+  if [[ "${ARCH}" == "aarch64" ]]; then
+    GUI_ARCH="arm64"
+  fi
```
devzbysiu, line 113: This could be done with this `ARCHS_FLUTTER` map

ci/nfpm/build_packages_resources.sh
```diff
@@ -86,3 +86,39 @@ mv "${WORKDIR}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}"
...
+  (cd gui && source scripts/build_package.sh)
+
+  # Move the generated GUI package to the expected location
+  mv gui/dist/"${PKG_TO_BUILD}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}/"
+  
+else
+  if [[ -z "${ENABLE_GUI_BUILD:-}" ]]; then
+    echo "Skipping GUI package build - ENABLE_GUI_BUILD not set (required for GUI builds)"
+  elif [[ -z "${ARCHS_FLUTTER[$ARCH]:-}" ]]; then
+    echo "Skipping GUI package build - architecture $ARCH not supported by Flutter"
```
thegravi: Shouldn't this raise an error? We want to build a GUI for this arch, but it's not possible, so why leave it to just "oops"? Shouldn't we know the exact parameters for each build?

ci/nfpm/build_packages_resources.sh
```diff
@@ -86,3 +86,39 @@ mv "${WORKDIR}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}"
...
+  echo "Building GUI package for Flutter-supported architecture: $ARCH (GUI build)"
+  
+  cleanup() {
+    local file="${WORKDIR}/gui/pubspec.yaml"
+    if [ -f "${file}.bak" ]; then
+      mv -f "${file}.bak" "${file}"
+      echo "Reverted changes to ${file}"
+    fi
+  }
+  trap cleanup EXIT ERR INT TERM
```
thegravi, line 102: If this script is sourced (cd gui & ./build_packages_resources.sh), not executed as its own process — that means the trap is registered in the parent shell (the one that sourced it) and not just while this script runs.
As a result, `cleanup()` may also run later when the parent shell exits or receives a signal, even long after this script has finished. In that case, god knows what can be "recovered" :) 

Don't know the background of how it's invoked, but just keep in mind.

ci/nfpm/build_packages_resources.sh
```diff
@@ -86,3 +86,39 @@ mv "${WORKDIR}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}"
...
+
+  # Build GUI package using the existing GUI build script
+  # Set auxiliary environment variables for the GUI script
+  export SKIP_DIST_CLEAN="true"  # Don't clean dist dir since we're integrating
```
thegravi, line 106: When multiple architectures or package types are built in a single CI run. May residual files leak across builds? Or are artifacts not cached for other jobs?

ci/nfpm/build_packages_resources.sh
```diff
@@ -86,3 +86,39 @@ mv "${WORKDIR}"/*."${PKG_TO_BUILD}" "${APP_DIR}/${PKG_TO_BUILD}"
...
+  export CUSTOM_SOURCE_DIR="${WORKDIR}/bin/${ARCH}/gui"  # Use our built GUI binaries
```
thegravi, line 107: What if this dir does not exist?

magefiles/mage.go
```diff
@@ -322,6 +322,8 @@ func buildPackageDocker(ctx context.Context, packageType string, buildFlags stri
 	env["WORKDIR"] = dockerWorkDir
 	env["ENVIRONMENT"] = string(internal.Development)
 	env["PACKAGE"] = devPackageType
+	//TODO (LVPN-9228) remove usage of ENABLE_GUI_BUILD variable, once the way of building is unified
+	env["ENABLE_GUI_BUILD"] = "1"
```
thegravi, line 326: Can we for now just add other mage targets with just gui enabled on top? And pass this flag as a build-flag?

ci/nfpm/build_packages_resources.sh
DoYouEvenCpp: Yeah, in fact I was in two minds about keeping those diagnostics.
Now, with your comment and some hindsight I'm leaning towards removing them completely (yes, GUI is now supported only for aarch/amd64 builds solely).

DoYouEvenCpp: Removed with the latest update.

ci/nfpm/build_packages_resources.sh
DoYouEvenCpp: It's created on binaries compilation stage, and it's not created when GUI is not built. Which is guarded by `ENABLE_GUI_BUILD`.
Meaning, it should be all good (but agree, this is not the best design to depend on other scripts!).

magefiles/mage.go
DoYouEvenCpp: Been thinking about that, but I guess adding an extra variable is way less effort to invest. Given the fact, we at least would like to revisit mage targets (and possibly modify/remove some of the currently existing ones).

ci/nfpm/build_packages_resources.sh
DoYouEvenCpp: This is kept separate

ci/nfpm/build_packages_resources.sh
DoYouEvenCpp: That's a fair point and good observation!
However in the very context we use this script it's all good.

dfetti (approved):

thegravi (approved):

PR #1111, opened 2025-09-26
LVPN-9227: Fix an error on stopping nordvpn service when process does…

bbrutall:
- Added proper stop for nordvpn service

### DIFF
--- a/test/qa/lib/daemon.py
+++ b/test/qa/lib/daemon.py
@@ -96,19 +96,29 @@ def start_peer(ssh_client: ssh.Ssh):
         time.sleep(1)
 
 
-def stop():
-    """Stops the daemon and blocks until it is actually stopped."""
-    if is_under_snap():
-        #sh.sudo.snap("stop", "nordvpn")
-        os.popen("sudo snap stop nordvpn").read()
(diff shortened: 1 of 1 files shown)


test/qa/lib/daemon.py
```diff
@@ -96,19 +96,29 @@ def start_peer(ssh_client: ssh.Ssh):
...
-        os.popen("sudo snap stop nordvpn").read()
-    elif is_init_systemd():
-        #sh.sudo.systemctl.stop.nordvpnd()
-        os.popen("sudo systemctl stop nordvpn").read()
-    else:
-        # call to init.d returns before the daemon is actually stopped
-        sh.sudo("/etc/init.d/nordvpn", "stop")
-        while is_running():
-            time.sleep(1)
+def stop(timeout: int = 10):
```
thegravi, line 99: Is this random or an estimated value after some testing?

bbrutall: It's a default value to wait for the running process to stop. The previous implementation had an infinite `while` loop without proper exiting from it. Now, `TimeoutError` is added if the process is not stopped after the timeout is reached

PR #1112, opened 2025-09-26
LVPN-9204: Minimize rc rollout events

keliramu:
Have `rc_rollout` event emitted only once per feature per app run, and reset on remote config changes downloaded.


config/remote/analytics.go
```diff
@@ -48,7 +78,23 @@ func (rca *RemoteConfigAnalytics) EmitJsonParseFailureEvent(client, featureName
...
 
 func (rca *RemoteConfigAnalytics) EmitPartialRolloutEvent(client, featureName string, frg int, rolloutPerformed bool) {
-	rca.publisher.Publish(
-		*NewRolloutEvent(rca.userRolloutGroup, client, featureName, frg, rolloutPerformed).
-			ToDebuggerEvent())
+	rca.mu.Lock()
+	defer rca.mu.Unlock()
+
+	if !rca.eventAlreadyEmitted(featureName, Rollout) {
+		rca.publisher.Publish(
```
devzbysiu: Is there some risk here that we are publishing while holding mutex? 
We could have only the check inside the mutex and setting the flag if needed and then publishing outside of the mutex, something like:
> :warning: Not tested

```go
func (rca *RemoteConfigAnalytics) EmitPartialRolloutEvent(client, feature string, frg int, rolloutPerformed bool) {
  rca.mu.Lock()
  if rca.eventFlags.has(feature, Rollout) {
    rca.mu.Unlock()
    return
  }
  rca.eventFlags.set(feature, Rollout)
  rca.mu.Unlock()

  rca.publisher.Publish(
    *NewRolloutEvent(rca.userRolloutGroup, client, feature, frg, rolloutPerformed).
      ToDebuggerEvent())
}
```
Or not really worth it?

config/remote/analytics.go
```diff
@@ -48,7 +78,23 @@ func (rca *RemoteConfigAnalytics) EmitJsonParseFailureEvent(client, featureName
...
+			*NewRolloutEvent(rca.userRolloutGroup, client, featureName, frg, rolloutPerformed).
+				ToDebuggerEvent())
+		rca.eventFlags.set(featureName, Rollout)
+	}
```
DoYouEvenCpp: [nit] what about a reversed logic order?  Should be a bit easier to follow the logic in here, about the event flags, I guess.

```suggestion
	if rca.eventAlreadyEmitted(featureName, Rollout) {
		return
	}
	rca.publisher.Publish(
		*NewRolloutEvent(rca.userRolloutGroup, client, featureName, frg, rolloutPerformed).
			ToDebuggerEvent())
	rca.eventFlags.set(featureName, Rollout)
```

config/remote/analytics_deduplication_test.go
```diff
@@ -0,0 +1,594 @@
...
+			eventCount := len(collectedEvents)
+			mu.Unlock()
+
+			assert.Equal(t, tt.expectedEvents, eventCount, "Expected %d events but got %d", tt.expectedEvents, eventCount)
+		})
+	}
+}
+
+// TestClearEventFlags_Scenarios tests various scenarios with ClearEventFlags
+func TestClearEventFlags_Scenarios(t *testing.T) {
```
DoYouEvenCpp, line 465: Can you extend this suite with a TC where a concurrent `ClearEventFlags` calls are being made?

config/remote/analytics_deduplication_test.go
```diff
@@ -0,0 +1,594 @@
...
+						// These should not be deduplicated
+						analytics.EmitDownloadEvent(ClientCli, feature)
+					}
+				}(i)
+			}
+
+			wg.Wait()
+
+			// Give a small delay for events to be processed
+			time.Sleep(10 * time.Millisecond)
```
DoYouEvenCpp: This might potentially lead to flaky tests - why not synchronise here using channels?

config/remote/analytics.go
keliramu: Yes, good catch! fixed.

config/remote/analytics.go
keliramu: Well, ok, changed to have publish outside of the lock - but it does not change the essence.

config/remote/analytics_deduplication_test.go
keliramu: Ok, good observation, fixed - eliminated sleep.

config/remote/analytics_deduplication_test.go
keliramu: Yes, added extra UT

PR #1123, opened 2025-10-02
Tray enhancements

thegravi:
This merge targets the release branch instead of main to avoid introducing upcoming release changes prematurely. The branch was rebased on earlier changes before the decision to postpone, so its contents will be included in the next release. The final target remains main after the release is completed.

Changes:
- All connection parameters are provided in lower case, not to clash with daemon's interpretation
- Tray startup/teardown mechanism made more robust
- Separated the daemon's state handling to be more readable

### DIFF
--- a/tray/daemon_state_handler.go
+++ b/tray/daemon_state_handler.go
@@ -15,7 +15,7 @@ func (ti *Instance) onDaemonStateEvent(item *pb.AppState) {
 	case *pb.AppState_ConnectionStatus:
 		changed = ti.handleConnectionStatusState(st)
 	case *pb.AppState_LoginEvent:
-		changed = ti.handleLoginEventState(st)
+		changed = ti.handleLoginEventState()
 	case *pb.AppState_SettingsChange:
 		changed = ti.handleSettingsChangeState(st)
 	case *pb.AppState_UpdateEvent:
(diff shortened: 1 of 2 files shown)


tray/daemon_state_handler.go
```diff
@@ -0,0 +1,101 @@
...
+// handleConnectionStatusState handles the connection status state from the daemon.
+func (ti *Instance) handleConnectionStatusState(st *pb.AppState_ConnectionStatus) bool {
+	log.Printf("%s %s New connection status: %s\n",
+		logTag,
+		internal.InfoPrefix,
+		st.ConnectionStatus.GetState(),
+	)
+	changed := ti.updateVpnStatus()
+	changed = ti.updateRecentConnections() || changed
+	return changed
```
bartoszWojciechO, line 49: Maybe it could be `return ti.updateVpnStatus() || ti.updateRecentConnections()`?

thegravi: Unfortunately, it cannot. If `ti.updateVpnStatus()` evaluates to `true`, then `ti.updateRecentConnections()` is not even executed. Kind of "short-circuit" case.

bartoszWojciechO (approved):

keliramu (approved):

PR #1128, opened 2025-10-06
LVPN-9293: Implement detect package manager

keliramu:


internal/package_manager.go
```diff
@@ -0,0 +1,112 @@
...
+		return true
+	}
+
+	// Check for Debian/Ubuntu specific files
+	if FileExists("/etc/debian_version") {
+		return true
+	}
+
+	// Check for Ubuntu specific files
+	if FileExists("/etc/lsb-release") {
```
DoYouEvenCpp: I'm aware `lsb-release` is something specific for Ubuntu here, but wouldn't be more robust to use standarized one, `/etc/os-release` instead?

internal/package_manager.go
```diff
@@ -0,0 +1,112 @@
+package internal
```
DoYouEvenCpp: I personally find this name, `package_manager`, a bit misleading. It doesn't actually manage anything - what about `package_manager_detector` (slightly longer, I know...) or anything similar?

internal/package_manager.go
```diff
@@ -0,0 +1,112 @@
...
+// PackageManager represents the type of package manager
+type PackageManager string
+
+const (
+	// PackageManagerUnknown represents an unknown or unsupported package manager
+	PackageManagerUnknown PackageManager = "unknown"
+	// PackageManagerDEB represents DEB-based package managers (APT/DPKG)
+	PackageManagerDEB PackageManager = "deb"
+	// PackageManagerRPM represents RPM-based package managers (YUM/DNF/Zypper)
+	PackageManagerRPM PackageManager = "rpm"
```
DoYouEvenCpp: [nit] a small inconsistency for packages names. Sometimes it's all capitalized letters (DEB/RPM) and something it's camel case (Deb, Rpm). I guess it would be nice to have it aligned.

internal/package_manager.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// DetectPackageManager detects whether the system uses DEB or RPM package management
```
DoYouEvenCpp:
```suggestion
// DetectPackageManager detects whether the system uses DEB or RPM package management.
// DEB-based systems are checked first and take priority over RPM-based systems.
```

keliramu: Added.

internal/package_manager.go
keliramu: You are right, I see it now as well, I change it to `package_detector` (file) and change to `PackageType` (instead of `PackageManager`)

internal/package_manager.go
keliramu: Ok, i've checked now: `lsb-release` is kinda legacy, and `os-release` is standardized modern way. Made to check `os-release` first, then `lsb-release`.

internal/package_manager.go
keliramu: Unified.

daemon/job_appversion.go
```diff
@@ -22,14 +22,16 @@ func JobVersionCheck(dm *DataManager, api *RepoAPI) func() {
 		// Get info from repo and convert data to strings
 		var versionStrings []string
 		switch api.packageType {
```
thegravi, line 25: Can we use predefined values for deb/ rpm there (`PackageType`), not to have hardcoded strings?

daemon/versions_test.go
```diff
@@ -61,3 +63,209 @@ func TestParseRpmVersions(t *testing.T) {
 	sort.Strings(parsed)
 	assert.EqualValues(t, expected, parsed)
 }
+
+func TestParseRpmVersions_Unit(t *testing.T) {
```
thegravi, line 67: [small] As I recall, we have an internal agreement on writing a single test per function, and not using several t.Run() in the same function, just to keep the same style of tests in the repo/

daemon/versions_test.go
```diff
@@ -61,3 +63,209 @@ func TestParseRpmVersions(t *testing.T) {
...
+	category.Set(t, category.Unit)
+
+	tests := []struct {
```
thegravi, line 70: Can you also add a test that has `nordvpn-gui` package in xml tag?

internal/package_detector.go
```diff
@@ -0,0 +1,110 @@
...
+			if err == nil && len(data) > 0 {
+				content := string(data)
+				for _, keywrd := range osKeywords {
+					if strings.Contains(content, keywrd) {
+						return true
+					}
+				}
+			}
+		}
+	}
```
thegravi, line 72: [nit] Can we have less nesting here?

internal/package_detector_test.go
```diff
@@ -0,0 +1,122 @@
...
+	if result {
+		// At least one of these should be true
+		hasRpmIndicator := IsCommandAvailable("rpm") ||
+			IsCommandAvailable("yum") ||
+			IsCommandAvailable("dnf") ||
+			IsCommandAvailable("zypper") ||
+			FileExists("/etc/redhat-release") ||
+			FileExists("/etc/SuSE-release") ||
+			FileExists("/etc/SUSE-brand") ||
+			FileExists("/etc/os-release")
```
thegravi: This one is not used in `TestIsDebBasedSystemIntegration()`

thegravi (approved):

internal/package_detector_test.go
keliramu: Right, added.

internal/package_detector.go
keliramu: Right, minimized nesting

daemon/versions_test.go
keliramu: Right, added

daemon/job_appversion.go
keliramu: Right, fixed

daemon/versions_test.go
keliramu: Multiple t.Run()s refactored into table-based test

PR #1132, opened 2025-10-10
LVPN-9316: Develop recent connections widget

thegravi:
Changes:
- Introducing "Recent connections" widget for GUI

Debt (for next PR):
- Does not support region-based specialty groups yet
- Incorrect VPN status for region-based specialty servers and country-based connections

Demo:
image

### DIFF
--- a/ci/docker/protochecker/docker-compose.yml
+++ b/ci/docker/protochecker/docker-compose.yml
@@ -1,7 +1,17 @@
+# Variables for reusability
+x-common-variables: &common-variables
+  WORKSPACE_DIR: /app
+  GEN_STATUS_DIR: /app/gen_status
+  GUI_PB_DIR: /app/gui/lib/pb
+  GUI_PB_BACKUP_DIR: /app/gui/lib/pb.backup
+  DAEMON_PB_DIR: daemon/pb/
+
 services:
-  # Generates Go protobuf files for the daemon and tray app
+  # Generates Go protobuf files for the daemon
   app-protobuf-gen:
     image: ghcr.io/nordsecurity/nordvpn-linux/generator:1.4.2
+    environment:
+      <<: *common-variables
     volumes:
       - ../../..:/app
     working_dir: /app
@@ -12,8 +22,8 @@ services:
         trap "exit 0" TERM
         ci/generate_protobuf.sh
         echo "app-protobuf-gen script finished successfully."
-        mkdir -p /app/gen_status
-        touch /app/gen_status/app_done
+        mkdir -p $${GEN_STATUS_DIR}
+        touch $${GEN_STATUS_DIR}/app_done
         sleep infinity & wait
     healthcheck:
       test: ["CMD", "test", "-f", "/app/gen_status/app_done"]
@@ -25,6 +35,8 @@ services:
   # Note: The generation script removes gui/lib/pb, so we backup before generation
   gui-protobuf-gen:
     image: ghcr.io/nordsecurity/nordvpn-linux/protobuf_dart-3.8.1:1.0.1
+    environment:
+      <<: *common-variables
     volumes:
       - ../../..:/app
     working_dir: /app
@@ -33,13 +45,13 @@ services:
       - -c
       - |
         trap "exit 0" TERM
-        rm -f /app/gen_status/gui_done
+        rm -f $${GEN_STATUS_DIR}/gui_done
         # Backup existing files before generation (script removes the directory)
-        cp -r /app/gui/lib/pb /app/gui/lib/pb.backup 2>/dev/null || true
+        cp -r $${GUI_PB_DIR} $${GUI_PB_BACKUP_DIR} 2>/dev/null || true
         gui/scripts/generate_protobuf.sh
         echo "gui-protobuf-gen script finished successfully."
-        mkdir -p /app/gen_status
-        touch /app/gen_status/gui_done
+        mkdir -p $${GEN_STATUS_DIR}
+        touch $${GEN_STATUS_DIR}/gui_done
         sleep infinity & wait
     healthcheck:
       test: ["CMD", "test", "-f", "/app/gen_status/gui_done"]
@@ -52,6 +64,8 @@ services:
   protochecker:
     image: alpine/git
     entrypoint: ["/bin/sh", "-c"]
+    environment:
+      <<: *common-variables
     volumes:
       - ../../..:/app
     working_dir: /app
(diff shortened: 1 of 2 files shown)


gui/lib/data/models/recent_connections.dart
```diff
@@ -0,0 +1,112 @@
+// import 'package:nordvpn/data/models/server_group_extension.dart';
+// import 'package:nordvpn/pb/daemon/common.pb.dart';
```
devzbysiu: This should be removed

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'package:nordvpn/data/models/connect_arguments.dart';
+import 'package:nordvpn/data/models/server_group_extension.dart';
+import 'package:nordvpn/data/providers/recent_connections_controller.dart';
+import 'package:nordvpn/data/models/recent_connections.dart';
+import 'package:nordvpn/i18n/string_translation_extension.dart';
+import 'package:nordvpn/vpn/server_list_item_factory.dart';
+import 'package:nordvpn/theme/app_theme.dart';
+import 'package:nordvpn/internal/images_manager.dart';
+
+class RecentConnectionsList extends ConsumerWidget {
```
devzbysiu: [small] Add `final` if it's not designed to be extended

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+  final Function(ConnectArguments) onSelected;
+  const RecentConnectionsList({super.key, required this.onSelected});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final appTheme = context.appTheme;
+    final recentConnections = ref.watch(recentConnectionsControllerProvider);
+    final serverFactory = ServerListItemFactory(imagesManager: ImagesManager());
```
devzbysiu: This is already registered in service locator [here](https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/service_locator.dart#L18-L18). You can see how this is injected for example [here](https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/widgets/searchable_servers_list.dart#L50). This way gives us option to overwrite this dependency for testing purposes and at the same time it's transparent when you use it in widgets.

The way you have now is coupling the widget to this item factory and to images manager which makes it hard to test and prevents us from writing negative tests

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+    return recentConnections.when(
+      loading: () => const Center(child: CircularProgressIndicator()),
+      error: (err, stack) => Center(child: Text('Error: $err')),
+      data: (connections) {
+        if (connections.isEmpty) {
+          return const SizedBox.shrink();
+        }
+
+        return Padding(
+          padding: const EdgeInsets.symmetric(vertical: 8.0),
```
devzbysiu: There is a `verticalSpaceSmall` for such cases defined in [here](https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/theme/theme.dart#L263-L263)

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              Text("Recent connections".tr(), style: appTheme.caption),
```
devzbysiu: This is not the right way to use translations. You should define this in [this](https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/i18n/en/ui.i18n.json) file and run `rps generate translations`, then you can use `t.ui.<defined-key>` like for example [here](https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/widgets/bin_button.dart#L17-L17)

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+              const SizedBox(height: 8.0),
```
devzbysiu: You can use `spacing` property of `Column` for this, and this value should also come from `AppTheme` - `verticalSpaceSmall` should be ok

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+                },
+              ),
+            ],
+          ),
+        );
+      },
+    );
+  }
+
+  Widget addModelBasedIcon(RecentConnection conn) {
```
devzbysiu, line 65: Is this used? :thinking:

gui/lib/vpn/recent_connections_list.dart
```diff
@@ -0,0 +1,73 @@
...
+    if (conn.isSpecialtyServer) {
+      final specialtyType = conn.group.toSpecialtyType();
+      if (specialtyType == null) {
+        return const Icon(Icons.history);
+      }
+      return ImagesManager().forSpecialtyServer(specialtyType);
```
devzbysiu: Don't construct this directly in the widget - too much coupling, hard to test. This is also registered in service locator so you can inject it in the constructor. Also this is registered as a singleton and now you are creating more instances.
Also in the line below

gui/lib/vpn/recent_server_list_item.dart
```diff
@@ -0,0 +1,166 @@
...
+  Widget build(BuildContext context) {
+    final isSpecialtyServer =
+        model.group != ServerGroup.UNDEFINED &&
+        model.group != ServerGroup.STANDARD_VPN_SERVERS;
+
+    final image = _buildItemImage(isSpecialtyServer);
+    final title = _buildItemTitle(context, isSpecialtyServer);
+    final connectArgs = _buildItemConnectArgs(isSpecialtyServer);
+
+    return CustomExpansionTile(
```
devzbysiu, line 40: Will there be a case that the recent server list item will expand? :thinking: This expansion tile was done for the case where country has multiple cities

gui/lib/vpn/server_item_image.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:nordvpn/data/models/vpn_status.dart';
+import 'package:nordvpn/data/providers/vpn_status_controller.dart';
+import 'package:nordvpn/theme/app_theme.dart';
+import 'package:nordvpn/theme/servers_list_theme.dart';
+import 'package:nordvpn/widgets/loading_indicator.dart';
+import 'package:nordvpn/widgets/padded_circle_avatar.dart';
+
+class ServerItemImage extends ConsumerWidget {
```
devzbysiu: [small] If it's not designed to be extended, add `final` keyword - it's not a big deal, but it's a signal for other devs that this was not planned with extension in mind

gui/lib/vpn/server_list_item_factory.dart
devzbysiu, line 1: A lot of unused imports in this file after moving out server item image

gui/lib/vpn/servers_list_card.dart
```diff
@@ -95,6 +96,7 @@ final class _ServersListCardState extends State<ServersListCard> {
       buttonText: t.ui.retry,
       onPressed: () async {
         await ref.read(serversListControllerProvider.notifier).refetch();
+        return;
```
devzbysiu: Not needed

gui/lib/vpn/recent_server_list_item.dart
thegravi: It will not. Just wanted to reuse something instead of duplicating some part of the code.

gui/lib/vpn/recent_connections_list.dart
thegravi: Forgot to remove this for another PR. This will determine which icon to use based on a recent connection model.

gui/lib/data/models/recent_connections.dart
```diff
@@ -0,0 +1,110 @@
...
+
+      case ServerSelectionRule.SPECIFIC_SERVER:
+        return specificServerName;
+
+      case ServerSelectionRule.GROUP:
+        return specialtyServer;
+
+      case ServerSelectionRule.COUNTRY_WITH_GROUP:
+        final ss = specialtyServer;
+        if (ss == "" || country == "") {
```
bartoszWojciechO: Can ternary operator be used in this case, to make it consistent with line 65? Or maybe the other way around.

dfetti (approved):

PR #1133, opened 2025-10-13
LVPN-8332 killswitch during upgrade

edvyno1:


contrib/scriptlets/rpm/preun
```diff
@@ -23,6 +23,27 @@ case "$1" in
             ;;
         esac
     ;;
+    1)
+        case "$ENV" in
+        init)
+            /etc/init.d/nordvpn upgrade-stop &>/dev/null || :
```
keliramu, line 29: for `rpm` case this cannot be used as inside `upgrade-stop` fn uses `start-stop-daemon` which is utility provided by the `dpkg` package (core Debian package manager) only.

edvyno1: that is true, we also have the base stop case which is calling `/etc/init.d/nordvpn stop`, that also does not work as it uses the same start-stop-daemon, what should be done in this case, leave as is or remove the code which does not work?

keliramu: there is no direct match for `rpm` based systems, maybe `kill / killall`: for scripts needing to manually start or stop processes, but also need to check which package provides this and set corresponding deps in `nfpm` template.

test/qa/lib/network.py
```diff
@@ -209,7 +209,11 @@ def is_not_available(retry=5) -> bool:
 
     # If assert below fails, and you are running Kill Switch tests on your machine, inside of Docker,
     # set DNS in resolv.conf of your system to anything else but 127.0.0.53
-    return not is_internet_reachable(retry=retry, ip_address="8.8.8.8") and not _is_dns_resolvable(retry=retry)
+    internet = not is_internet_reachable(retry=retry, ip_address="8.8.8.8")
+    dnsavail = not _is_dns_resolvable(retry=retry)
+    print(internet)
```
keliramu: Is this debug print? or is it supposed to provide info on any run? will it just print `true` or `false`? need more info on each print e.g. `Internet reachable: true / false`

test/qa/test_killswitch.py
```diff
@@ -13,7 +14,8 @@
 
 pytestmark = pytest.mark.usefixtures("nordvpnd_scope_module", "collect_logs")
 
-
+PROJECT_ROOT = os.environ['WORKDIR']
+DEB_PATH = glob.glob(f'{PROJECT_ROOT}/dist/app/deb/*amd64.deb')[0]
```
keliramu: Will this work with `CLI` and `GUI` packages?

test/qa/test_update.py
```diff
@@ -11,6 +11,7 @@
 from test_connect import connect_base_test, disconnect_base_test
 
 PROJECT_ROOT = os.environ['WORKDIR']
+DEB_PATH = glob.glob(f'{PROJECT_ROOT}/dist/app/deb/*amd64.deb')[0]
```
keliramu: Will this work with CLI and GUI packages?

snap/hooks/pre-refresh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+kill -s USR1 $(ps aux | grep -e "snap.*nordvpn" | awk '{print $2}' | head -1)
```
keliramu: If we are sure we select only our process, then why send signal only to the first one? what if multiple processes will be found by search pattern? will it always choose the right one as first?

keliramu: Maybe just remove `head -1`

networker/networker.go
```diff
@@ -552,14 +553,14 @@ func (netw *Combined) blockTraffic() error {
 	// block PREROUTING & POSTROUTING
 	return netw.fw.Add([]firewall.Rule{
 		{
-			Name:       "drop IPv4",
+			Name:       "drop-IPv4",
```
keliramu: Have reusable constants for names as well, use them in tests as well

test/qa/lib/network.py
edvyno1: used this to debug ,reverting to normal

test/qa/test_killswitch.py
edvyno1: edited to include only cli package

test/qa/test_update.py
edvyno1: edited

snap/hooks/pre-refresh
edvyno1: :+1: removed

networker/networker.go
edvyno1: applied

cmd/daemon/main.go
```diff
@@ -729,14 +729,16 @@ func main() {
...
 	}
-
 	if cfg.Mesh {
 		go rpc.StartAutoMeshnet(meshService, network.ExponentialBackoff)
 	}
 
 	// Graceful stop
 
-	internal.WaitSignal()
+	// internal.WaitSignal()
```
keliramu: Cleanup needed?

edvyno1: good spot, cleaned up some other unnecessary comments as well

test/qa/test_killswitch.py
```diff
@@ -226,3 +228,32 @@ def printdir():
         assert res == "exists", f"File {os.path} should exist after kill-switch was enabled"
 
 
+# This test assumes being run on docker
+def test_killswitch_on_after_update():
+    # Mocking ps to pretend as if we are in an initd system
+    sh.sudo.mv("/usr/bin/ps", "/usr/bin/pso")
```
keliramu, line 234: Need restore

edvyno1: added :+1:

daemon/firewall/firewall.go
```diff
@@ -60,9 +63,17 @@ func (fw *Firewall) Add(rules []Rule) error {
 			}
 			fw.publisher.Publish(fmt.Sprintf("replacing existing rule %s", rule.Name))
 		}
+		// dont add if already added
+		alreadyAddedRules, err := fw.current.GetActiveRules()
```
bartoszWojciechO: Do we need to fetch the rules on each loop iteration? Would it work to fetch them once, before the loops starts?

daemon/firewall/firewall.go
```diff
@@ -60,9 +63,17 @@ func (fw *Firewall) Add(rules []Rule) error {
...
+		if err != nil {
+			log.Printf("%v, unable to get already active rules: %v", internal.WarningPrefix, err)
+		}
 
-		if err := fw.current.Add(rule); err != nil {
-			return NewError(fmt.Errorf("adding %s: %w", rule.Name, err))
+		if !slices.Contains(alreadyAddedRules, rule.Name) &&
```
bartoszWojciechO, line 71: Will the proper ordering of rules be maintained with such approach?

daemon/firewall/iptables/iptables.go
```diff
@@ -455,3 +464,51 @@ func connectionStateToString(state firewall.ConnectionState) string {
 	}
 	return ""
 }
+
+func (ipt *IPTables) GetActiveRules() ([]string, error) {
+	ipt.Lock()
+	defer ipt.Unlock()
+	re, err := regexp.Compile(`--comment \S*`)
```
bartoszWojciechO, line 472: Won't this also match non-nord rules with a comment?

daemon/firewall/firewall.go
bartoszWojciechO: Got some weir behavior with those steps:

1. Add some subnet to an allowlist.
2. Connect.
3. Copy allowlist rules.
4. Disconnect.
5. Add allowlist rules back into the firewall manually.
6. Connect.

The allowlist rule was placed after the final DROP rule, making it ineffective. But this is quite artificial repro, I doubt that we'd have an situation where only a single rule remains, most likely it will be all or nothing.

daemon/firewall/firewall.go
```diff
@@ -60,9 +63,17 @@ func (fw *Firewall) Add(rules []Rule) error {
...
+			!slices.Contains(alreadyAddedRules, rule.SimplifiedName) {
+			if err := fw.current.Add(rule); err != nil {
```
bartoszWojciechO, line 73: Maybe in a case where we find that a rule already exists it could be added to the rules map(without running any actual iptables commands)? Then it would be properly deleted in `Disable` method.

edvyno1: yes, that is actually the case already, the if guard clause is protecting the adding with actual iptables commands, but the following line 
`fw.rules.Add(rule)` is still ran regardless of if it was added via iptables or not, adding it into memory to be tracked as already added, it should properly be deleted when turning off the killswitch, unless you mean something else

daemon/firewall/iptables/iptables.go
edvyno1: that is true, this needs to be changed to only parse through rules that have the nordvpn comment to begin with

daemon/firewall/firewall.go
edvyno1: yeah this approach is not super error resilient, if such edge cases happens we would be in trouble, but we do have the fallback of flushing on every disconnect(unless the killswitch is on), I did not want to try and redesign the system, seeing as we will be swapping to nftables relatively soon, this was mostly aimed at to help with the rules while upgrading, when we are only expecting the killswitch rules to stay, which conveniently are also inserted first

daemon/firewall/firewall.go
edvyno1: yeah most likely it would be enough to just fetch the active rules before the loop, will change

daemon/firewall/firewall.go
bartoszWojciechO: I agree, no point in changing anything atm for an edge case.

daemon/firewall/firewall.go
bartoszWojciechO: When testing the allowlist case I found one rule remaining after disconnection. I ran:
```
sudo iptables -t mangle -A PREROUTING -s 192.168.10.0/24 -i wlp0s20f3 -m comment --comment nordvpn -m comment --comment allowlist_subnets -j ACCEPT
sudo iptables -t mangle -A PREROUTING -s 192.168.10.0/24 -i enp0s31f6 -m comment --comment nordvpn -m comment --comment allowlist_subnets -j ACCEPT
```
Then connected, then disconnected. But only the rule for `wlp0s20f3` remained. Might be worth checking out.

daemon/firewall/iptables/iptables.go
edvyno1: added if clause to only parse through if line contains `defaultComment`

daemon/firewall/firewall.go
edvyno1: fixed this with https://github.com/NordSecurity/nordvpn-linux/pull/1133/commits/47db9f27aa14913f2199e3a214c4c59e141a7ef8

.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.3.2
+      triggered-ref: killswitch_upgrade_tester_bump
```
edvyno1: fix before merging

PR #1137, opened 2025-10-14
LVPN-8093: Bump go to 1.25.3

DoYouEvenCpp:
This PR brings go v1.25.3

Other significant changes under this PR:
- update `golanglint-ci` configuration file to match >= v.2.x
  - this is breaking change, new config file structure is not compatible with linters versioned < v2.x
  - fix a few UTs after go bump
- update workflow action configuration to match updated version of go/golang linter


.github/workflows/core_ci.yml
```diff
@@ -62,8 +62,7 @@ jobs:
 
       - uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
         with:
-          # TODO: LVPN-8373: change this back to go-version-file after Go update
```
devzbysiu, line 65: Can't we address this now with https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file ?
Then we wouldn't have to duplicate the version everywhere

.github/workflows/codeql.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
             - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
               with:
-                  go-version: "1.22"
+                  go-version: "1.25.3"
```
devzbysiu: Can this also use `go-version-file`?

devzbysiu (approved):

.github/workflows/codeql.yml
DoYouEvenCpp: Good catch. Fixed!

PR #1125, opened 2025-10-02
LVPN-7482: Update error messages

dfetti:

PR #1138, opened 2025-10-15
LVPN-7482: Update error messages in GUI

devzbysiu:
Changes:
- Update error messages modified in https://github.com/NordSecurity/nordvpn-linux/pull/1125 also in GUI
- Move two errors to `internal/errors.go` to have it in one place and add warn comment with info that those error messages are important from GUI perspective

### DIFF
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -45,14 +45,22 @@ var (
 	ErrTagDoesNotExist         = errors.New(TagNonexistentErrorMessage)
 	ErrGroupDoesNotExist       = errors.New(GroupNonexistentErrorMessage)
 	ErrDoubleGroup             = errors.New(DoubleGroupErrorMessage)
-	// ErrAlreadyLoggedIn is returned on repeated logins
-	ErrAlreadyLoggedIn = errors.New("You're already logged in")
-	// ErrNotLoggedIn is returned when the caller is expected to be logged in
-	// but is not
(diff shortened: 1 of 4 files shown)


internal/errors.go
```diff
@@ -45,14 +45,22 @@ var (
...
+	// https://github.com/NordSecurity/nordvpn-linux/blob/main/gui/lib/data/repository/daemon_status_codes.dart#L49-L49
+
+	// ErrAlreadyLoggedIn is returned on repeated logins
+	ErrAlreadyLoggedIn = errors.New("You're already logged in")
+	// ErrNotLoggedIn is returned when the caller is expected to be logged in
+	// but is not
+	ErrNotLoggedIn = errors.New("You're not logged in")
+	// ErrMissingExchangeToken is returned when login was successful but
+	// there is not enough data to request the token
+	ErrMissingExchangeToken = errors.New("The exchange token is missing. Please try logging in again. If the issue persists, contact our customer support.")
```
DoYouEvenCpp, line 65: nit: errors should not end with a punctuation mark nor begin with a non-capital letters.

https://staticcheck.dev/docs/checks#ST1005

DoYouEvenCpp (approved):

thegravi (approved):

internal/errors.go
devzbysiu: The whole error message is not appropriate, but we are using errors for user messages in few places for legacy reasons

PR #1140, opened 2025-10-15
LVPN-9294: Add handle login flow change

keliramu:


events/moose/moose.go
```diff
@@ -311,11 +311,16 @@ func (s *Subscriber) NotifyLogin(data events.DataAuthorization) error { // regul
 		mooseFn = moose.NordvpnappSendServiceQualityAuthorizationRegister
 	}
 
+	loginFlowAltered := moose.NordvpnappOptBoolFalse
+	if data.IsAlteredFlowOnNordAccount {
+		loginFlowAltered = moose.NordvpnappOptBoolTrue
+	}
+
```
thegravi: When login is attempted, we would need to set it to `NordvpnappOptBoolNone`.

thegravi (changes requested):

events/moose/moose.go
keliramu: Fixed

daemon/rpc_login_test.go
```diff
@@ -0,0 +1,164 @@
...
+func TestLoginEvents(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	tests := []struct {
+		name            string
+		setup           func(*daemonevents.MockPublisherSubscriber[events.DataAuthorization]) *RPC
+		loginType1      pb.LoginType
+		expectedStatus1 pb.LoginStatus
+		expectedEvent1  events.DataAuthorization
+		expectedError1  bool
```
DoYouEvenCpp, line 28: So, by looking on `LoginOAuth2` (which never returns an error) and how the tests are constructed looks like this can be removed, along with the respective if condition to run error/noerror assertion.

daemon/rpc_login_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			eventsMock := &daemonevents.MockPublisherSubscriber[events.DataAuthorization]{}
+			r := tt.setup(eventsMock)
+			resp1, err := r.LoginOAuth2(context.Background(), &pb.LoginOAuth2Request{Type: tt.loginType1})
+			if tt.expectedError1 {
+				assert.Error(t, err)
+			} else {
```
DoYouEvenCpp, line 136: This is always NoError here (a bit of a tautology here I'd say - on the other hand if the `LoginOAuth2` starts reporting errors here some day, it might be useful to catch this still in tests)
```suggestion
```

keliramu: Good observation, but I would stay in test by following method's semantics.

PR #1146, opened 2025-10-16
LVPN-9370: fix linter issues after bump

DoYouEvenCpp:
A strand of https://github.com/NordSecurity/nordvpn-linux/pull/1137 where golang got updated to v1.25.3. This enforces to update also the linter tool (to v2.4.0).

Here, with this PR all of newly reported issues are either suppressed or fixed.

### DIFF
--- a/cli/cli_mesh_peers.go
+++ b/cli/cli_mesh_peers.go
@@ -280,6 +280,7 @@ func (c *cmd) MeshPeerDenyRouting(ctx *cli.Context) error {
 		return formatError(err)
 	}
 
+	//TODO - LVPN-9412: reavaulate error handling
 	resp, _ := c.meshClient.DenyRouting(
 		context.Background(),
 		&pb.UpdatePeerRequest{


networker/networker.go
```diff
@@ -1081,11 +1081,11 @@ func (netw *Combined) refresh(cfg mesh.MachineMap) error {
 	netw.cfg = cfg
 
 	var err error
-	if err = netw.defaultMeshBlock(cfg.Machine.Address); err != nil {
```
edvyno1, line 1084: is this intentional? I don't see any changes related to MeshMap, why are all the `cfg.Machine` calls truncated?

DoYouEvenCpp: Yes, this is because [Machine](https://github.com/NordSecurity/nordvpn-linux/blob/skubiak/LVPN-9370-fix-linter-issues-after-update/core/mesh/peer.go#L13) type follows the struct embedding pattern here, thus linter started to notify this could be reduced.

If on the other hand, this is not something we'd like to maintain I can restore the previous version, and basically suppress it here.

edvyno1: I would maybe be for keeping the full name for clarity's sake, but it's up to you, thanks for the explanation

daemon/vpn/openvpn/config.go
```diff
@@ -74,15 +74,16 @@ func generateConfigFile(protocol config.Protocol, serverIP netip.Addr, obfuscate
...
+		return fmt.Errorf("reading ovpn template file: %w", err)
 	}
 
 	out, err := generateConfig(serverIP, identifier, template)
 	if err != nil {
 		return fmt.Errorf("generating OpenVPN config: %w", err)
 	}
 
-	if err := addExtraParameters(out, serverIP, protocol); err != nil {
+	out, err = addExtraParameters(out, serverIP, protocol)
```
devzbysiu: So the result of `addExtraParameters` was basically not used and now it is being used 🤔 , are there any differences in the configuration?

daemon/job_servers.go
```diff
@@ -43,8 +43,8 @@ func JobServers(dm *DataManager, api core.ServersAPI, validate bool) func() erro
 		dist := distance(
 			geoInfoData.Insights.Latitude,
 			geoInfoData.Insights.Longitude,
-			servers[0].Locations[0].Country.City.Latitude,
-			servers[0].Locations[0].Country.City.Longitude,
+			servers[0].Locations[0].Latitude,
```
devzbysiu: I'm not sure about this 🤔, I think it's less obvious what's happening with this change. Also keep in mind that we for sure have instances of wrong usage of the embedding in the code

And in this case, this is multiple levels of skipped embeddings to get to `City` struct via `Country` struct and this `Country` struct has two embeddings, so this may be annoying to work with.

Also, there is a foot gun with those embeddings which is embedding shadowing - if I add field `Latitude` to struct `Country` now, then the line above will automatically lead to this new field instead of field `Latitude` in `City` struct and the compiler won't complain so being explicit with those embeddings is usually better default

meshnet/jobs.go
```diff
@@ -45,7 +52,7 @@ func JobRefreshMeshMap(s *Server) func() error {
 func JobRefreshMeshnet(s *Server) func() error {
 	return func() error {
 		// ignore what is returned, try to do it here as light as possible
-		_, _ = s.RefreshMeshnet(nil, nil)
+		_, _ = s.RefreshMeshnet(&NoopContext{}, nil)
```
devzbysiu: For this, we can pass `context.Background()` - it is basically empty context.
Eventually `context.WithTimeout()` could be passed here also to avoid job hanging on `RefreshMeshnet`, but maybe let's not change too much to avoid unexpected regression, I would go with `context.Background()` here

networker/networker.go
devzbysiu: Agree with @edvyno1, especially here when the context is not obvious - with `cfg.Address` - it's not clear what `Address` it is by just looking on the code

tray/monitor.go
```diff
@@ -439,11 +439,9 @@ func messageForDaemonError(err error) string {
...
 
 func (ti *Instance) updateDaemonConnectionStatus(errorMessage string) bool {
-	daemonAvailable := false
-
-	if (errorMessage == "") || (errorMessage == cli.ErrUpdateAvailable.Error()) {
-		daemonAvailable = true
-	}
+	daemonAvailable := func() bool {
+		return (errorMessage == "") || (errorMessage == cli.ErrUpdateAvailable.Error())
+	}()
```
devzbysiu: Hmm 🤔 , what's the point of doing it this way? Coulnd't we just do this?
```go
	daemonAvailable := errorMessage == "" || errorMessage == cli.ErrUpdateAvailable.Error()

```

DoYouEvenCpp: Indeed, not sure why I snapped there such a over complicated "solution".  :smile:

cli/cli_mesh_peers.go
```diff
@@ -280,7 +280,7 @@ func (c *cmd) MeshPeerDenyRouting(ctx *cli.Context) error {
 		return formatError(err)
 	}
 
-	resp, err := c.meshClient.DenyRouting(
+	resp, _ := c.meshClient.DenyRouting(
```
keliramu, line 284: imo, it's not safe to leave `err` not checked - if someday grpc  method implementation would add return error and `resp` would be undefined we would get `panic`. And who know if these cases are covered with tests.

DoYouEvenCpp: Agree, no error should be left unhandled.
Create an internal ticket, to follow-up on this.

devzbysiu (approved):

PR #1149, opened 2025-10-17
Implement protobuf generated files checker

thegravi:
Changes:
- Added a new CI job to verify that all protobuf-generated files are up to date with the current source code

### DIFF
--- a/.github/workflows/protochecker.yml
+++ b/.github/workflows/protochecker.yml
@@ -2,6 +2,7 @@ name: Protobuf Checker
 
 on:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   protochecker:
--- a/ci/docker/protochecker/docker-compose.yml
+++ b/ci/docker/protochecker/docker-compose.yml
@@ -36,9 +36,10 @@ services:
       - generated-repo:/code/app
     working_dir: /code/app
     user: "${UID:-1000}:${GID:-1000}"
+
   protochecker:
     image: alpine/git
-    command: [ "diff", "--quiet" ]
+    command: [ "diff", "--name-only", "--exit-code" ]
     depends_on:
       app-protobuf-gen:
         condition: service_completed_successfully


ci/docker/protochecker/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -eux
```
DoYouEvenCpp, line 2: [nit] I guess this should be our standard for bash scripts:

```suggestion
set -euxo pipefail
```

ci/docker/protochecker/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      setup:
+        condition: service_completed_successfully
+    volumes:
+      - generated-repo:/work
+    working_dir: /work
+    user: "${UID:-1000}:${GID:-1000}"
+
+  gui-protobuf-gen:
+    image: ghcr.io/nordsecurity/nordvpn-linux/protobuf_dart-3.8.1:1.0.1
+    command: "gui/scripts/generate_protobuf.sh"
```
DoYouEvenCpp, line 31: FYI - there's a task to have a single point of protobuf generation, happening in this day and age.

ci/docker/protochecker/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    depends_on:
+      setup:
+        condition: service_completed_successfully
+    volumes:
+      - generated-repo:/code/app
+    working_dir: /code/app
+    user: "${UID:-1000}:${GID:-1000}"
+  protochecker:
+    image: alpine/git
+    command: [ "diff", "--quiet" ]
```
DoYouEvenCpp: [suggestion]
Why not use git diff here?

With `git diff --name-only` we'd be able to produce nice diagnostics here, which might be a bit more user-friendly at CI checks stage?

.github/workflows/protochecker.yml
```diff
@@ -0,0 +1,28 @@
+name: Protobuf Checker
```
DoYouEvenCpp, line 1: Care to add `workflow_dispatch` usage here, so it might be triggered manually on any branch (not upon PRs only)?

ci/docker/protochecker/docker-compose.yml
thegravi: Is it really necessary to combines those? What if there is a need to only generate for a single component?

ci/docker/protochecker/entrypoint.sh
thegravi: Since there are no other scripts in this pipeline, I omitted this. Script is isolated in the Docker env only. Also, I'm running this su sh, so it does not support this option.

ci/docker/protochecker/docker-compose.yml
DoYouEvenCpp: Yeah, it's kinda on par with the overall (desired) structure of the building scripts. Anyway, alignments will be done once handling this unification.

Issue #1142, opened 2025-10-15
libxml2.so.2 => not found


Samoonfire:
Good evening,

Today after completing a dup on openSUSE Tumbleweed, it removed the libxml2.so.2 that NordVPN CLI requires to run.

The current version on Tumbleweed is libxml2-16.

```
ldd /usr/sbin/nordvpnd
	linux-vdso.so.1 (0x00007f49f1988000)
	libmoosenordvpnapp.so => /usr/lib/nordvpn/libmoosenordvpnapp.so (0x00007f49efe00000)
	libmooseworker.so => /usr/lib/nordvpn/libmooseworker.so (0x00007f49efc63000)
	libsqlite3.so.0 => /lib64/glibc-hwcaps/x86-64-v3/libsqlite3.so.3.50.4 (0x00007f49efad2000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f49f195c000)
	libquench.so => /usr/lib/nordvpn/libquench.so (0x00007f49ef600000)
	libtelio.so => /usr/lib/nordvpn/libtelio.so (0x00007f49ee800000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f49f1957000)
	libxml2.so.2 => not found
	libc.so.6 => /lib64/libc.so.6 (0x00007f49ee400000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f49f1928000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f49f0105000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f49f1923000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f49f198a000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f49f191e000)
```

Are there any plans to address this? Some searches online seem to suggest Arch and Manjaro users had this issue a few months back and were able to fix it by installing a legacy package of libxml2.

Currently I'm having to use a workaround of using OpenVPN with the config files downloaded from Nord.

Please let me know if you require any further information.

lourdas:
Hi, I'm using Gentoo and since the upgrade of libxml2 to 2.14.6, the nordvpnd daemon fails to start. Do you have an estimate of when you will provide a newer build that is linked against libxml2 2.14.x?

I created a ebuild for my system with the help of chatgpt and after many hours of tinkering, the daemon fails. Here is a snippet of the log:

[...]

PR #1158, opened 2025-10-22
LVPN-9381: Replace gokogiri,ratago with call to xsltproc

devzbysiu:
Instead of using xml and xslt libraries to transform config for openvpn, use `xsltproc` binary. Those libraries require oldish `libxml2` which is updated on newer distributions and the old one is not available for installation.

Changes:
- remove `libxml2` from dependencies, add `xsltproc` - for deb, rpm and snap
- remove `github.com/jbowtie/gokogiri/xml`, `github.com/jbowtie/ratago/xslt` libs
- delegate xsl transformation to `xsltproc` binary

This addresses https://github.com/NordSecurity/nordvpn-linux/issues/1142

### DIFF
--- a/.github/workflows/ci-gitlab.yml
+++ b/.github/workflows/ci-gitlab.yml
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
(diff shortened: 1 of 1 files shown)


dfetti (approved):

TheNexter:
This patch is very important, on ubuntu 25.10, nordvpn is not possible to install

bartoszWojciechO (approved):

PR #1161, opened 2025-10-23
Handling edge cases and race conditions (protochecker job)

thegravi:
Changes:
- Implements handling custom protobuf generation logic from the GUI image
- Adds image caching to improve build efficiency
- Coordinates service startup to prevent race conditions during generation
- Reporting what exact files are outdated

### DIFF
--- a/config/group.pb.go
+++ b/config/group.pb.go
@@ -7,11 +7,10 @@
 package config
 
 import (
-	reflect "reflect"
-	sync "sync"
-
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
+	sync "sync"
 )
 
 const (
@@ -50,6 +49,7 @@ var (
 		7:  "ANTI_DDOS",
 		9:  "DEDICATED_IP",
 		11: "STANDARD_VPN_SERVERS",
+		13: "NETFLIX_USA",
 		15: "P2P",
 		17: "OBFUSCATED",
 		19: "EUROPE",
@@ -65,6 +65,7 @@ var (
 		"ANTI_DDOS":                        7,
 		"DEDICATED_IP":                     9,
 		"STANDARD_VPN_SERVERS":             11,
+		"NETFLIX_USA":                      13,
 		"P2P":                              15,
 		"OBFUSCATED":                       17,
 		"EUROPE":                           19,
--- a/gui/lib/pb/daemon/set.pb.dart
+++ b/gui/lib/pb/daemon/set.pb.dart
@@ -30,6 +30,9 @@ class SetAutoconnectRequest extends $pb.GeneratedMessage {
     $core.String? serverGroup,
   }) {
     final result = create();
+    if (enabled != null) result.enabled = enabled;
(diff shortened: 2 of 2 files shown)


.github/workflows/protochecker.yml
```diff
@@ -15,16 +15,40 @@ jobs:
 
       - name: Set up Docker
         uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
+        with:
+          driver-opts: |
+            image=moby/buildkit:v0.17.3
+            network=host
+
+      - name: Cache Docker layers
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
```
keliramu, line 24: Comment is for line no12: I see controversy - here at line no24 we pin action by very exact version i.e. by commit SHA ID, but on line no12 we use `ubuntu-latest` which could lead to failures when latest becomes incompatible and we will fail having reproducible outcome of this job. Maybe would be good to pin ubuntu image to exact version as well?

ci/docker/protochecker/entrypoint.sh
```diff
@@ -1,8 +1,80 @@
...
+	echo "Comparing generated GUI files with committed versions..."
+
+	# Compare files excluding google directories (symlinks)
+	BACKUP_FILES=$(cd "${GUI_PB_BACKUP_DIR}" && find . -type f ! -path '*/google/*' | sort)
+	GENERATED_FILES=$(cd "${GUI_PB_DIR}" && find . -type f ! -path '*/google/*' | sort)
+
+	# First check if the file lists match (let's us skip file-by-file comparison)
+	if [ "$BACKUP_FILES" != "$GENERATED_FILES" ]; then
+		echo "ERROR: Generated GUI protobuf files differ from committed versions!"
+		echo "File list mismatch detected."
```
keliramu, line 30: Would it be possible to show what is the difference in file list?

thegravi: Yes, it is, but if there were several new files, then the logs would be 3 papyrus long.
Is there a need for this, since we are only interested in whether the files are outdated?

.github/workflows/protochecker.yml
```diff
@@ -8,23 +8,47 @@ permissions: {}
...
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Prepare directories
+        run: |
+          sudo rm -rf gen_status || true
```
DoYouEvenCpp, line 33: [nit] is `|| true` bit really needed here? 
(after all we call rm with  -f which shall discard any error anyway?)

ci/docker/protochecker/entrypoint.sh
keliramu: That worst case situation should not be frequent, but when happens would be good to see where is the problem from the first look without re-executing something again.

.github/workflows/protochecker.yml
thegravi: Mostly yes, but there are several cases where it can still yield non-zero value. One of those is the lack of permissions.

PR #1164, opened 2025-10-24
LVPN-9419: Add recent connection only after connection change

thegravi:
Changes:
* A new entry for the most recent server connection is now created when the user disconnects from a server or connects to a different one. Previously, the entry was added immediately upon connection.

### DIFF
--- a/daemon/recent_connections_rpc_helper.go
+++ b/daemon/recent_connections_rpc_helper.go
@@ -9,7 +9,7 @@ import (
 func StorePendingRecentConnection(
 	store *recents.RecentConnectionsStore,
 	eventPublisher func(events.DataRecentsChanged)) {
-	exists, recentModel := store.GetPending()
+	exists, recentModel := store.PopPending()
 	if !exists {
 		return
 	}
--- a/daemon/recent_connections_rpc_helper_test.go
+++ b/daemon/recent_connections_rpc_helper_test.go
@@ -88,7 +88,7 @@ func TestStorePendingRecentConnection_BasicBehavior(t *testing.T) {
 
 			// Verify pending was cleared if expected
 			if tt.expectedCleared {
-				exists, empty := store.GetPending()
+				exists, empty := store.PopPending()
 				assert.False(t, exists)
 				assert.True(t, empty.IsEmpty())
 			}
@@ -375,7 +375,7 @@ func TestStorePendingRecentConnection_ErrorHandling(t *testing.T) {
 
 			// Verify pending was cleared
 			if tt.expectedCleared {
-				exists, empty := store.GetPending()
+				exists, empty := store.PopPending()
 				assert.False(t, exists)
 				assert.True(t, empty.IsEmpty())
 			}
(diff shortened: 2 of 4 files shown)


daemon/recent_connections_rpc_helper.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/NordSecurity/nordvpn-linux/events"
+)
+
+// StorePendingRecentConnection stores current pending recent connection to memory
+func StorePendingRecentConnection(
+	store *recents.RecentConnectionsStore,
+	eventPublisher func(events.DataRecentsChanged)) {
+	err, recentModel := store.GetPending()
+	if err != nil {
+		return
```
bartoszWojciechO, line 14: Is error expected in this case? If not, can we have a log here?

thegravi: Right now, it means that pending does not contain anything. Perhaps I will change err to a boolean to just have this simpler approach of exists or not.

DoYouEvenCpp: Then I guess this is a right way to go for, currently this error reported in case of nothing  is pending is a bit of a  YAGNI here.

daemon/recent_connections_rpc_helper_test.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/daemon/recents"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	mockconfig "github.com/NordSecurity/nordvpn-linux/test/mock/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStorePendingRecentConnection_WithPendingConnection(t *testing.T) {
```
DoYouEvenCpp: Could this be converted into one larger table-driven test, with the next two ones?

daemon/recents/recent_connections.go
```diff
@@ -113,6 +114,28 @@ func (r *RecentConnectionsStore) Add(model Model) error {
...
 }
 
+// AddPending stores a recent connection model as pending to be added later
+func (r *RecentConnectionsStore) AddPending(model Model) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.pendingConnection = model
+}
+
+// GetPending retrieves and clears the pending recent connection model
```
DoYouEvenCpp: [nit] `GetPending` doesn't doesn't imply the content is actually a `swap` (leaving it with an empty one) method. Shall we rename it maybe?

daemon/recents/recent_connections.go
```diff
@@ -113,6 +114,28 @@ func (r *RecentConnectionsStore) Add(model Model) error {
...
+func (r *RecentConnectionsStore) GetPending() (error, Model) {
+	r.mu.Lock()
```
DoYouEvenCpp, line 126: Assuming this method still reads  and _overwrites_ here does it make sense to change the `mu` to `sync.RWMutext` and distinguish how we lock here for read only and r/w opreations? (to not block other readers on r/o ops here, for example)

DoYouEvenCpp (approved):

PR #1170, opened 2025-10-28
LVPN-9315: Updated support for specialty servers

thegravi:
**Added**
* Add view-only support for *regions as specialty servers* in UI to be compatible with connections originated from the CLI or Tray.
* Updated specialty server list by including "Obfuscated servers"

**Fixed**
* Improved matching logic for recent connections (avoid duplicates by rule-based exclusion).

### DIFF
--- a/daemon/recents/recent_connections.go
+++ b/daemon/recents/recent_connections.go
@@ -78,16 +78,8 @@ type SearchOptions struct {
 func searchOptionsForConnectionType(connType config.ServerSelectionRule) SearchOptions {
 	opts := SearchOptions{}
 
-	switch connType {
-	case config.ServerSelectionRule_SPECIFIC_SERVER,
-		config.ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP:
-		// No exclusions
-	case config.ServerSelectionRule_COUNTRY,
-		config.ServerSelectionRule_COUNTRY_WITH_GROUP,
-		config.ServerSelectionRule_GROUP,
-		config.ServerSelectionRule_NONE,
-		config.ServerSelectionRule_RECOMMENDED,
-		config.ServerSelectionRule_CITY:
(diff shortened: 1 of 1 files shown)


gui/test/country_names_test.dart
```diff
@@ -37,6 +37,8 @@ void main() async {
 
   test(
     'All cities, country codes and names from API exist in translation files',
+    // disable test for now, because list is changing and names are not translated anyway
```
dfetti, line 40: We still need to keep that list updated and I don't think we have other indicator than this failing test. So my suggestion is to keep the testcase active and add the missing cities when the test fails.

thegravi: I took this from the main to have a working pipeline. This will be "fixed" when the feature branch is merged back to the main.

daemon/recents/recent_connections.go
```diff
@@ -60,17 +60,76 @@ func (r *RecentConnectionsStore) Get() ([]Model, error) {
...
+	ExcludeConnectionType     bool
+	ExcludeServerTechnologies bool
+	ExcludeIsVirtual          bool
+}
+
+// searchOptionsForConnectionType returns SearchOptions based on the connection type.
+func searchOptionsForConnectionType(connType config.ServerSelectionRule) SearchOptions {
+	opts := SearchOptions{}
+
+	switch connType {
```
bartoszWojciechO: Wouldn't something like:
```
if connType != config.ServerSelectionRule_SPECIFIC_SERVER && conType != config.ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP {
// ...
}
```
be more clear?

thegravi: Definitely. I've started with small 2-switch-cases, and later linter kicked in, resulting into this monstrocity :D

PR #1174, opened 2025-10-30
LVPN-8805: Add widget keys

dfetti:
We should add key to all our widgets so we can find them by key in the integration tests.
This PR adds keys just to the widgets used in our current integration tests.
It also formats the code, step which I forgot to do when adding the adaptive scaffold.

### DIFF
--- a/ci/test_install_rpm.sh
+++ b/ci/test_install_rpm.sh
@@ -4,7 +4,7 @@ set -euxo pipefail
 case "$1" in
     centos) yum -y install yum-utils createrepo ;;
     fedora) dnf -y install dnf-plugins-core createrepo ;;
-    opensuse) zypper -n install curl createrepo_c ;;
+    opensuse) zypper refresh && zypper -n install curl createrepo_c ;;
     *) echo "Can't recognise the OS" && exit 1 ;;
 esac
 


gui/lib/settings/autoconnect_settings.dart
```diff
@@ -189,7 +199,7 @@ final class _ConnectNowButton extends ConsumerWidget {
     final isEnabled =
         isUpdating || _isConnectedToSelectedLocation(vpnStatus, savedLocation);
     return LoadingElevatedButton(
-      key: ValueKey(isEnabled),
```
devzbysiu, line 192: Hmm, I'm wondering why this had bool as an argument, because it suggests that it was important to rebuild the widget on `isEnabled` change 🤔 , now we have it stable no matter the value of `isEnabled`

devzbysiu (approved):

gui/lib/settings/autoconnect_settings.dart
mariusSincovici: I think it wasn't changing the state disable/enable, but maybe in the meantime something changed and is not important anymore 🤔 
does it work also fine when the connection is made from the CLI and the button needs to refresh itself?

devzbysiu: Yeah, yeah it does work

PR #1176, opened 2025-10-31
LVPN-9295: Implement login failure event exceptions

keliramu:
Implement login failure event exceptions

### DIFF
--- a/daemon/atomic_login_type.go
+++ b/daemon/atomic_login_type.go
@@ -7,58 +7,58 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 )
 
-// AtomicLoginType manages login type and timing information with thread-safe access.
+// atomicLoginType manages login type and timing information with thread-safe access.
 // It uses a single RWMutex for synchronization of both the login type value and timestamp.
-type AtomicLoginType struct {
+type atomicLoginType struct {
 	loginType            pb.LoginType
 	lastLoginAttemptTime time.Time
 	mu                   sync.RWMutex
 }
 
-func NewAtomicLoginType() *AtomicLoginType {
-	return &AtomicLoginType{
+func NewAtomicLoginType() *atomicLoginType {
+	return &atomicLoginType{
 		loginType: pb.LoginType_LoginType_UNKNOWN,
 	}
 }
 
-func (a *AtomicLoginType) Get() pb.LoginType {
+func (a *atomicLoginType) get() pb.LoginType {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
 	return a.loginType
 }
 
-func (a *AtomicLoginType) Set(loginType pb.LoginType) {
+func (a *atomicLoginType) set(loginType pb.LoginType) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	a.loginType = loginType
 }
 
-func (a *AtomicLoginType) Reset() {
+func (a *atomicLoginType) reset() {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	a.loginType = pb.LoginType_LoginType_UNKNOWN
 	a.lastLoginAttemptTime = time.Time{}
 }
 
-func (a *AtomicLoginType) IsUnknown() bool {
-	return a.Get() == pb.LoginType_LoginType_UNKNOWN
+func (a *atomicLoginType) isUnknown() bool {
+	return a.get() == pb.LoginType_LoginType_UNKNOWN
 }
 
-func (a *AtomicLoginType) WasStarted() bool {
-	return !a.IsUnknown()
+func (a *atomicLoginType) wasStarted() bool {
+	return !a.isUnknown()
 }
 
-func (a *AtomicLoginType) IsAltered(loginType pb.LoginType) bool {
-	return a.Get() != loginType
+func (a *atomicLoginType) isAltered(loginType pb.LoginType) bool {
+	return a.get() != loginType
 }
 
-func (a *AtomicLoginType) SetLoginAttemptTime(t time.Time) {
+func (a *atomicLoginType) setLoginAttemptTime(t time.Time) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	a.lastLoginAttemptTime = t
 }
 
-func (a *AtomicLoginType) GetLoginAttemptTime() time.Time {
+func (a *atomicLoginType) getLoginAttemptTime() time.Time {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
 	return a.lastLoginAttemptTime
--- a/daemon/rpc_login.go
+++ b/daemon/rpc_login.go
@@ -55,7 +55,7 @@ func (r *RPC) loginWithToken(token string) (payload *pb.LoginResponse, retErr er
 	})
 
 	// check if previous login/signup process was started
-	if r.initialLoginType.WasStarted() {
+	if r.initialLoginType.wasStarted() {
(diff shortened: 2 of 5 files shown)


daemon/rpc_login.go
```diff
@@ -27,28 +25,24 @@ func (r *RPC) LoginWithToken(ctx context.Context, in *pb.LoginWithTokenRequest)
 		}, nil
 	}
 
+	if in.GetToken() == "" {
```
mariusSincovici, line 26: No events?

daemon/rpc_login.go
```diff
@@ -267,3 +309,15 @@ func (r *RPC) IsLoggedIn(ctx context.Context, _ *pb.Empty) (*pb.IsLoggedInRespon
 	loggedIn, _ := r.ac.IsLoggedIn()
 	return &pb.IsLoggedInResponse{IsLoggedIn: loggedIn}, nil
 }
+
+func (r *RPC) lockGetInitialLoginType() pb.LoginType {
```
mariusSincovici: I don't think it makes sense to have `lock` into the function name. 
Because this is a bit of "internal" implementation. Maybe in the future it will be replace with atomic or something else without locks.

daemon/rpc_login.go
mariusSincovici, line 109: If load fails, then `eventReason` is not initialized

mariusSincovici, line 126: Also for save `eventReason` remains `0`

daemon/rpc_login.go
```diff
@@ -155,18 +162,38 @@ func (r *RPC) LoginOAuth2(ctx context.Context, in *pb.LoginOAuth2Request) (*pb.L
...
-	r.initialLoginType = in.GetType()
 
-	r.events.User.Login.Publish(events.DataAuthorization{
-		DurationMs:   -1,
-		EventTrigger: events.TriggerUser,
-		EventStatus:  events.StatusAttempt,
-		EventType:    eventType,
-	})
+	// check if previous login/signup process was started
+	if r.lockGetInitialLoginType() != pb.LoginType_LoginType_UNKNOWN {
```
mariusSincovici: [small] This is a duplicated also into the login with token, if it is not complicating things, then it would be good to be in one place

daemon/rpc_login.go
```diff
@@ -61,6 +55,19 @@ func (r *RPC) loginCommon(customCB customCallbackType) (payload *pb.LoginRespons
...
+		r.events.User.Login.Publish(events.DataAuthorization{
+			DurationMs:   -1,
+			EventTrigger: events.TriggerUser,
+			EventStatus:  events.StatusFailure, // emit failure event, but continue
+			EventType:    events.LoginLogin,
+			Reason:       events.ReasonUnfinishedPrevLogin,
+		})
+	}
+
+	var eventReason events.ReasonCode
```
mariusSincovici: It would be better if this is initialized with some value from the enum. Or if needed to add one into the enum.

daemon/rpc_login.go
mariusSincovici, line 287: No error value set for `eventReason`

daemon/rpc_login.go
keliramu: it's missing user input situation - yes, no events.

daemon/rpc_login.go
thegravi: `ReasonNotSpecified`

daemon/rpc_login.go
keliramu: Fixed

keliramu: Fixed

keliramu: Fixed

daemon/rpc_login.go
keliramu: Fixed/improved

daemon/rpc_login.go
keliramu: Improved

daemon/atomic_login_type.go
```diff
@@ -0,0 +1,43 @@
...
+	"sync/atomic"
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+)
+
+type AtomicLoginType struct {
+	value atomic.Int32
+}
+
+func NewAtomicLoginType(initialValue ...pb.LoginType) *AtomicLoginType {
```
mariusSincovici: The parameter is a bit misleading, because more elements can be provided, but actually only the first is used.
Is this used in the code, except for tests, otherwise I would suggest to remove it to prevent future confusion?

mariusSincovici (approved):

daemon/rpc_login.go
```diff
@@ -27,28 +25,24 @@ func (r *RPC) LoginWithToken(ctx context.Context, in *pb.LoginWithTokenRequest)
 		}, nil
 	}
 
+	if in.GetToken() == "" {
+		return &pb.LoginResponse{
+			Type: internal.CodeTokenLoginFailure,
+		}, nil
+	}
```
thegravi, line 30: This check if covered by the following if statement. Can we reuse it in a way?

daemon/rpc_login.go
```diff
@@ -15,8 +15,6 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/session"
 )
 
-type customCallbackType func() (*core.LoginResponse, *pb.LoginResponse, error)
-
 var lastLoginAttemptTime time.Time
```
thegravi: [small] A bit out of scope, but this is not a thread-safe var to use here. It could lead to incorrect metric durations.

daemon/atomic_login_type.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (a *AtomicLoginType) Reset() {
+	a.Set(pb.LoginType_LoginType_UNKNOWN)
+}
+
+func (a *AtomicLoginType) IsUnknown() bool {
+	return a.Get() == pb.LoginType_LoginType_UNKNOWN
+}
+
+func (a *AtomicLoginType) WasStarted() bool {
```
thegravi: In context of `NewAtomicLoginType` is not clear what would this method do.

daemon/atomic_login_type.go
```diff
@@ -0,0 +1,43 @@
...
+	return !a.IsUnknown()
+}
+
+func (a *AtomicLoginType) IsAltered(loginType pb.LoginType) bool {
```
thegravi: This is just a purely difference in taste, but I would suggest using `isChanged`, as we use similar naming in codebase.

daemon/rpc_login.go
keliramu: These are two different cases:
1. user has not given input value;
2. user has given invalid values;
Messages are different to the user, it's already in place, would not change that in the sake of some insignificant code beautification.

daemon/atomic_login_type.go
keliramu: In the context of daemon login functionality, this name is related to the login event flag `isFlowAltered` - I would not generalize it to broader code context, but rather stay with daemon login flow context.

daemon/atomic_login_type.go
keliramu: This is daemon login flow context (not general code context, this type is inside daemon package intentionally) and all concepts are tightly related to defined/documented login flow definitions to keep relevance and readability later who would need to support it.

daemon/atomic_login_type.go
keliramu: Removed

daemon/rpc_login.go
keliramu: Yes, right, improved - added into `AtomicLoginType`

PR #1177, opened 2025-10-31
LVPN-9452: Fix connect tests for snap env

pvraketa:
Changed disabling of the network for the init systemd env

### DIFF
--- a/ci/test_deb.sh
+++ b/ci/test_deb.sh
@@ -41,4 +41,15 @@ if ! sudo grep -q "export GOCOVERDIR=${GOCOVERDIR}" "/etc/init.d/nordvpn"; then
     trap revert_go_cov EXIT INT TERM
 fi
 
+if [[ -n ${RC_FILES:-} ]]; then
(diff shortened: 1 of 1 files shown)


test/qa/lib/network.py
```diff
@@ -236,33 +236,37 @@ def is_disconnected(retry=5) -> bool:
...
-def stop() -> str:
+def stop() -> dict:
     """Returns default_gateway to be used when starting network again."""
-    default_gateway = None
+    default_gateway = {}
     for line in sh.ip.route().split('\n'):
         if line.startswith('default'):
-            default_gateway = line.split()[2]
+            default_gateway[line.split()[4]]=line.split()[2]
     assert default_gateway is not None
```
keliramu: This assert has to be adjusted to the line's nr260 (old nr258) change

pvraketa:
```suggestion
    assert default_gateway, f"Couldn't find default gateway. Ip route are next: {sh.ip.route()}"
```
Agree?

keliramu: Yes, good

PR #1181, opened 2025-11-03
LVPN-8370: Add display protocol telemetry sending

edvyno1:


events/moose/moose.go
```diff
@@ -690,14 +690,22 @@ func (s *Subscriber) OnTelemetry(metric telemetry.Metric, value any) error {
...
-			// set 'x11' metric
+			if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateDisplayProtocol("x11")); err != nil {
+				return fmt.Errorf("setting display protocol: %w", err)
+			}
 		case telemetrypb.DisplayProtocol_DISPLAY_PROTOCOL_UNKNOWN:
 		default:
-			// set 'unknown' metric (e.g. NordvpnappUnsetContextDeviceDisplayProtocol)
+			if err := s.response(moose.NordvpnappUnsetContextApplicationNordvpnappConfigCurrentStateDisplayProtocol()); err != nil {
+				return fmt.Errorf("unsetting display protocol: %w", err)
+			}
```
thegravi, line 707: Perhaps we could set this to `unknown`, since it currently represents an unsupported value. If we later observe increased usage with this value, we can adjust the metric to accept the raw protocol value instead.

thegravi (approved):

events/moose/moose.go
edvyno1: changed

events/moose/moose.go
```diff
@@ -690,14 +690,22 @@ func (s *Subscriber) OnTelemetry(metric telemetry.Metric, value any) error {
 		// TODO: missing moose metric support (e.g. NordvpnappSetContextDeviceDisplayProtocol)
```
mariusSincovici: Is the TODO still needed?

thegravi: nope

edvyno1: removed :+1:

PR #1182, opened 2025-11-04
LVPN-9497: Tag most important tests related to Meshnet

ggediminass:
We need to tag most important tests related to Meshnet, so we can easily run them separately.

### DIFF
--- a/test/pytest.ini
+++ b/test/pytest.ini
@@ -1,0 +1,5 @@
+
+# content of pytest.ini
+[pytest]
+markers =
+    core_meshnet: mark a test as a core test.


test/qa/conftest.py
```diff
@@ -52,6 +52,8 @@ def pytest_configure(config):
 
     :param config: The pytest config object, which holds command-line options and internal state.
     """
+    config.addinivalue_line("markers", "core_meshnet: marks core meshnet tests")
```
bbrutall: what is a reason to add markers here?  You've already introduced them in the pytest.ini.  This change would only be necessary if no pytest.ini file exists.

test/pytest.ini
```diff
@@ -0,0 +1,6 @@
+
+# content of pytest.ini
+[pytest]
+markers =
+    core: mark a test as a core test.
+    serial
```
bbrutall: What is the purpose of `serial` mark?

test/pytest.ini
```diff
@@ -0,0 +1,6 @@
+
+# content of pytest.ini
+[pytest]
+markers =
+    core: mark a test as a core test.
```
bbrutall: iIt should be `core_meshnet`

test/qa/conftest.py
ggediminass: Good catch. My mistake here. I could not make markers to work over `pytest.ini`, so decided to add them here, and forgot to remove `pytest.ini` change :smile:

test/pytest.ini
ggediminass: You are right in the other comment, this has to be `core_meshnet`

test/qa/conftest.py
ggediminass: Introduced changes in `pytest.ini`, removed `config.addinivalue_line`

test/qa/conftest.py
```diff
@@ -52,6 +52,7 @@ def pytest_configure(config):
 
     :param config: The pytest config object, which holds command-line options and internal state.
     """
+
```
bbrutall: No empty line needed

PR #1183, opened 2025-11-04
LVPN-9304: Disable tray communication to fileshare socket when mesh off

dfetti:
I found 3 places where there is communication with the fileshare socket:
- The tray, that keeps the connection active while it's running. The most "dangerous" one, which I already fixed so the connection is established only when the meshnet is enabled.
- The norduser, which holds the management system for fileshare, which tries to communicate over it only to send the stop command, which closes the fileshare process. Now it tries to connect and send the close command only if the socket exists so we get rid of the not found error when the fileshare is already disabled.
-  The cli, which connects to fileshare to be able to execute the fileshare commands. The trick here is that if we want to avoid the connection when meshnet is off, we will need to do an aditional rpc call for every command to fetch the settings and see if the meshnet is enabled which doesn’t make sense in my opinion.

Decided with @devzbysiu  to call it a day just with the tray fixes.

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -93,8 +93,8 @@ func main() {
 	clientIDMetadataInterceptor := clientid.NewInsertClientIDInterceptor(pb.ClientID_CLI)
 
 	loaderInterceptor := cli.LoaderInterceptor{}
-	//nolint:staticcheck
-	conn, err := grpc.Dial(
+
+	conn, _ := grpc.NewClient(
 		DaemonURL,
 		// Insecure credentials are OK because the connection is completely local and
 		// protected by file permissions
@@ -104,8 +104,7 @@ func main() {
 		grpc.WithChainStreamInterceptor(loaderInterceptor.StreamInterceptor,
 			clientIDMetadataInterceptor.SetMetadataStreamInterceptor),
 	)
-	//nolint:staticcheck
-	fileshareConn, err := grpc.Dial(
+	fileshareConn, _ := grpc.NewClient(
 		fileshare_process.FileshareURL,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithUnaryInterceptor(loaderInterceptor.UnaryInterceptor),
--- a/cmd/norduser/main.go
+++ b/cmd/norduser/main.go
@@ -23,8 +23,6 @@ import (
 	childprocess "github.com/NordSecurity/nordvpn-linux/child_process"
 	"github.com/NordSecurity/nordvpn-linux/clientid"
 	daemonpb "github.com/NordSecurity/nordvpn-linux/daemon/pb"
-	"github.com/NordSecurity/nordvpn-linux/fileshare/fileshare_process"
-	filesharepb "github.com/NordSecurity/nordvpn-linux/fileshare/pb"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	meshpb "github.com/NordSecurity/nordvpn-linux/meshnet/pb"
 	"github.com/NordSecurity/nordvpn-linux/norduser"
@@ -85,21 +83,7 @@ func startTray(quitChan chan<- norduser.StopRequest) {
 	}
 	ReportTelemetry(conn, ReportOnStart, false)
 
-	//nolint:staticcheck
-	fileshareConn, err := grpc.Dial(
-		fileshare_process.FileshareURL,
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-	)
-
-	var fileshareClient filesharepb.FileshareClient
-	if err == nil {
-		fileshareClient = filesharepb.NewFileshareClient(fileshareConn)
-	} else {
-		log.Println(internal.ErrorPrefix, "Error connecting to the NordVPN fileshare daemon:", err)
-		return
-	}
-
-	ti := tray.NewTrayInstance(client, fileshareClient, quitChan)
+	ti := tray.NewTrayInstance(client, quitChan)
 	ti.Start()
 
 	topLevelCtx, topLevelCancelFunc := context.WithCancel(context.Background())
--- a/tray/actions.go
+++ b/tray/actions.go
@@ -14,7 +14,6 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/cli"
 	"github.com/NordSecurity/nordvpn-linux/client"
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
-	filesharepb "github.com/NordSecurity/nordvpn-linux/fileshare/pb"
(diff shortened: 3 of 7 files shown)


tray/actions.go
```diff
@@ -252,10 +251,7 @@ func (ti *Instance) setNotify(flag bool) bool {
 	case internal.CodeSuccess:
 	}
 
-	_, err = ti.fileshareClient.SetNotifications(context.Background(), &filesharepb.SetNotificationsRequest{Enable: flag})
-	if err != nil {
-		log.Printf("%s Setting fileshare notifications %s error: %s", internal.ErrorPrefix, flagText, err)
-	}
+	ti.fileshare.SetNotifications(flag, flagText)
```
mariusSincovici: [small] Is it really needed to pass `flagText` for logging?

tray/tray.go
```diff
@@ -145,6 +148,39 @@ type trayState struct {
 	mu                   sync.RWMutex
 }
 
+type FileshareManager struct {
```
mariusSincovici: Could you move this into a separate file?

tray/monitor.go
```diff
@@ -241,6 +241,8 @@ func (ti *Instance) setSettings(settings *pb.Settings) bool {
 		return false
 	}
 
+	ti.fileshare.UpdateFileshareConnection(settings.Meshnet)
```
mariusSincovici, line 250: At logout, fileshare continues to run, even if the meshnet is disabled.

tray/fileshare_manager.go
```diff
@@ -0,0 +1,45 @@
...
+		if err == nil {
+			fs.fileshareClient = filesharepb.NewFileshareClient(fileShareConn)
+		} else {
+			log.Println(internal.ErrorPrefix, "Error connecting to the NordVPN fileshare daemon:", err)
+		}
+	}
+}
+
+func (fs *FileshareManager) SetNotifications(flag bool) {
+	if fs.fileshareClient != nil {
```
devzbysiu: [small] you can treat this as a precondition, then we can minimize nesting by doing 
```
if fs.fileshareClient == nil {
  return
}
```

Also would be good to at least log something in this case, because if someone calls this method when `fileshareClient` is `nil` then something is wrong 😄

tray/fileshare_manager.go
```diff
@@ -0,0 +1,45 @@
...
+		_, err := fs.fileshareClient.SetNotifications(context.Background(), &filesharepb.SetNotificationsRequest{Enable: flag})
```
devzbysiu: [small] you can combine this with the `if` like 
```
if _, err := fs.fileshareClient.SetNotifications(...); err != nil {
   ...
}
```

tray/fileshare_manager.go
```diff
@@ -0,0 +1,45 @@
...
+		if err != nil {
+			log.Printf("%s Setting fileshare notifications %s error: %s", internal.ErrorPrefix, getFlagText(flag), err)
```
devzbysiu: I see we already had it like that so far, but pls add `\n` to the message

cmd/norduser/main.go
```diff
@@ -85,21 +83,7 @@ func startTray(quitChan chan<- norduser.StopRequest) {
 	}
 	ReportTelemetry(conn, ReportOnStart, false)
 
-	//nolint:staticcheck
-	fileshareConn, err := grpc.Dial(
```
bartoszWojciechO, line 89: Is fileshare tied to tray now? What happens in non-gui environments?

dfetti: No, I just changed the place where the connection (that is used inside tray) is initiated.

cmd/cli/main.go
```diff
@@ -94,7 +94,7 @@ func main() {
 
 	loaderInterceptor := cli.LoaderInterceptor{}
 	//nolint:staticcheck
```
DoYouEvenCpp: IIRC, now this `nolint` suppression might be removed?

tray/fileshare_manager.go
```diff
@@ -0,0 +1,47 @@
...
+	"log"
+
+	"github.com/NordSecurity/nordvpn-linux/fileshare/fileshare_process"
+	filesharepb "github.com/NordSecurity/nordvpn-linux/fileshare/pb"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type FileshareManager struct {
```
DoYouEvenCpp, line 14: From testability perspective, shouldn't this be based on an interface?

tray/fileshare_manager.go
```diff
@@ -0,0 +1,47 @@
...
+	fileshareClient filesharepb.FileshareClient
+}
+
+func (fs *FileshareManager) UpdateFileshareConnection(meshnetEnabled bool) {
```
DoYouEvenCpp, line 24: Can you please provide docs  for those exported methods?

tray/fileshare_manager.go
dfetti: Is it really worth to do that right now? I would honestly skip it at this moment.

DoYouEvenCpp: OK, then I guess we can have a follow-up ticket to cover it at more convenient time.

tray/monitor.go
thegravi, line 152: [nit] Perhaps we can merge these 2, eliminating repetitive check on `!loggedIn`?

tray/tray.go
```diff
@@ -168,10 +167,10 @@ func (state *trayState) serverName() string {
 	return vpnServerName
 }
 
-func NewTrayInstance(client pb.DaemonClient, fileshareClient filesharepb.FileshareClient, quitChan chan<- norduser.StopRequest) *Instance {
+func NewTrayInstance(client pb.DaemonClient, quitChan chan<- norduser.StopRequest) *Instance {
 	obj := &Instance{
 		client:            client,
-		fileshareClient:   fileshareClient,
+		fileshare:         FileshareManager{fileshareClient: nil},
```
thegravi: [small] Can we have a "constructor" here? This would also hide internal assignments.

tray/fileshare_manager.go
```diff
@@ -0,0 +1,53 @@
...
+type FileshareManager struct {
+	fileshareClient filesharepb.FileshareClient
+}
+
+func NewFileshareManager() FileshareManager {
+	return FileshareManager{fileshareClient: nil}
+}
+
+// Updates the fileshare gRPC connection based on the meshnetEnabled status
+func (fs *FileshareManager) UpdateFileshareConnection(meshnetEnabled bool) {
```
devzbysiu, line 24: Doc comments on methods should start with method name, so `UpdateFileshareConnection updates ...`, similar in other places. Not a hard rule, but it's widely accepted convention in go

PR #1184, opened 2025-11-05
LVPN-8034: Add Terms section to the GUI

thegravi:
**Added**
* New **“Terms”** section in **Settings**, containing legal information.
* (Internal) Introduced `IconLink` component that enables icons to function as part of a clickable link.

**Fixed**
* Incorrectly selected `text-accent` theme color

<details>
<summary>Screenshots</summary>
image
image
</details>


gui/lib/settings/terms_screen.dart
```diff
@@ -0,0 +1,62 @@
...
+      (t.ui.termsOfService, termsOfServiceUrl),
+      (t.ui.autoRenewalTerms, autoRenewalTermsUrl),
+      (t.ui.privacyPolicy, privacyPolicyUrl),
+    ];
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Text(t.ui.termsAgreementDescription, style: context.body),
+        SizedBox(height: 16),
```
devzbysiu: This should use `verticalSpaceMedium`

Also, you probably could use `spacing` prop of `Column`

gui/lib/settings/terms_screen.dart
```diff
@@ -0,0 +1,62 @@
...
+      ],
+    );
+  }
+
+  Widget _buildLinkEntry(BuildContext context, String title, Uri link) {
+    return Row(
+      children: [
+        Expanded(child: Text(title, style: context.body)),
+        IconLink(
+          title: "Read more",
```
devzbysiu: This should go through translations

gui/lib/settings/terms_screen.dart
```diff
@@ -0,0 +1,62 @@
...
+    final links = [
+      (t.ui.termsOfService, termsOfServiceUrl),
+      (t.ui.autoRenewalTerms, autoRenewalTermsUrl),
+      (t.ui.privacyPolicy, privacyPolicyUrl),
+    ];
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Text(t.ui.termsAgreementDescription, style: context.body),
```
devzbysiu: For this line and other places also.

So far we are using theme values by following this pattern:
```
final appTheme = context.appTheme;

...

style: appTheme.body
```

The reasoning behind this is that theme dependencies are more explicit and clear what part of the theme the widget depends on by just looking on a code.

We could discuss this approach also, but then it should be decided and updated everywhere to be consistent

gui/lib/widgets/link.dart
```diff
@@ -18,15 +20,23 @@ final class Link<T> extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final theme = context.appTheme;
+    return _buildClickableArea(context, _buildContent(context));
```
devzbysiu: This could be inlined, no reason to have more jumps in code considering that `_buildClickableArea` is short

gui/lib/widgets/link.dart
```diff
@@ -37,3 +47,41 @@ LaunchableUri _toLaunchable<T>(T uri) {
...
+  IconLink({
+    super.key,
+    required super.title,
+    required super.uri,
+    super.size,
+    DynamicThemeImage? icon,
+    String? iconPath,
+  }) : assert(
+         (icon != null) ^ (iconPath != null),
+         'Exactly one of icon or iconPath must be provided',
```
devzbysiu: Is `icon` approach used somewhere? 🤔 I see only `iconPath` used

gui/lib/widgets/link.dart
```diff
@@ -37,3 +47,41 @@ LaunchableUri _toLaunchable<T>(T uri) {
...
+       _icon = icon ?? DynamicThemeImage(iconPath!);
+
+  @override
+  Widget _buildContent(BuildContext context) {
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      crossAxisAlignment: CrossAxisAlignment.center,
+      children: [
+        super._buildContent(context),
+        SizedBox(width: AppSpacing.spacing2),
```
devzbysiu: You can use `spacing` prop of `Row`

devzbysiu: Also don't use `AppSpacing` directly, it should be exposed via theme if it's not done yet

gui/integration_test/e2e_tests/legal_information.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:nordvpn/i18n/strings.g.dart';
+
+import '../../test/utils/test_helpers.dart';
```
DoYouEvenCpp, line 4: Can we extract test_helpers into an internal package so similar imports would be less fragile in the future?
(possibly, a question to @devzbysiu here)

gui/test/utils/legal_information_screen_handle.dart
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  String currentBreadcrumbLabel() {
+    final finder = currentNavigationBreadcrumb();
+    final widget = app.tester.widget<Breadcrumb>(finder);
+    return widget.name;
+  }
+
+  bool hasTermsAgreementDescription() {
+    return find.text(t.ui.termsAgreementDescription).evaluate().isNotEmpty;
```
devzbysiu: Please coordinate with @dfetti - we are in the process of changing the way we find widgets for testing. Specifically those kind of changes - https://github.com/NordSecurity/nordvpn-linux/pull/1174/files#diff-d5aac2e0a95b913f18794806d61332ff8d25b34b9dc1099f5104a5fd960018efR49 , less coupling with the implementation details

gui/test/utils/legal_information_screen_handle.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:nordvpn/i18n/strings.g.dart';
+import 'package:nordvpn/settings/navigation.dart';
+
+import 'finders.dart';
+import 'screen_handle.dart';
+
+final class LegalInformationScreenHandle extends ScreenHandle {
```
devzbysiu, line 6: Isn't part of this class a duplicate of `AccountScreenHandle`?

gui/integration_test/e2e_tests/legal_information.dart
devzbysiu: Could you elaborate more? I'm not sure I'm getting the idea :smile:

DoYouEvenCpp: So instead of `import '../../test/utils/test_helpers.dart';` we'd able to import this way:
`import 'package:nordvon/test/utils/helpers.dart`;`

My gripe was about using relative paths that gets easily broken.

mariusSincovici: Yes, it would be nicer to have that as a separate package.
I just think that should go as a separate task, because it will change more files + add some more ones.

devzbysiu: Sorry for the delay. Yeah, it could be done, but then I think those test utils would have to be under `lib/` and not under `test` (`integration_test/` also won't work), so not sure if it makes sense to put test related code into non-test directory. 
The code won't be included in the final app, so that's not an issue - the only thing here is this test code in non-test dir. Then considering that this relative path would be an issue only when we would move the code (which we are not planning), I'm not sure if it makes sense to spend time on this. I wouldn't do that, but I'm also not objecting, so if you prefer it, we can have a task for it and address it

gui/test/utils/legal_information_screen_handle.dart
dfetti: Yes, take a look at the approach from this [PR](https://github.com/NordSecurity/nordvpn-linux/pull/1174)

gui/lib/widgets/link.dart
thegravi: I did it the same way the menu icon builder did. It supported both cases but used only one. Consistency, I would say :D

gui/test/utils/legal_information_screen_handle.dart
thegravi: The account page will be modified after this PR. Some parts will be removed, others added.

gui/lib/widgets/link.dart
devzbysiu: Aah, ok. Then maybe we could remove it in both places, because it's a dead code now :man_shrugging:

PR #1192, opened 2025-11-07
LVPN-8949: use correct event to notify application is used for the first time

DoYouEvenCpp:
Changes under this PR:
- used correct event to notify when application is used for the first time after installation
- the mechanism of triggering this event once and only once is mostly preserved
- removed some not needed bits from `events/firstopen/firstopen_event.go`
   - analytics event is now trigged directly instead of using for that purpose subscriber-notification mechanism (which is less efficient and not needed in that particular case)
   - adjusted UTs of `firstopen_event`

### DIFF
--- a/daemon/rpc_connect_test.go
+++ b/daemon/rpc_connect_test.go
@@ -405,18 +405,28 @@ func TestRPCConnect(t *testing.T) {
 				if test.setup != nil {
 					test.setup(rpc)
 				}
+				firstOpenEventCouter := 0
+				firstOpenListener := func(any) error {
+					firstOpenEventCouter++
+					return nil
+				}
+				rpc.events.Service.FirstTimeOpened.Subscribe(firstOpenListener)
 				server := &mockRPCServer{}
 				err := rpc.Connect(&pb.ConnectRequest{
 					ServerGroup: test.serverGroup,
 					ServerTag:   test.serverTag,
 				}, server)
+
 				switch test.resp {
 				case internal.CodeConnected:
 					assert.NoError(t, err)
+					assert.Equal(t, firstOpenEventCouter, 1)
 				case 0:
 					assert.ErrorIs(t, internal.ErrUnhandled, err)
+					assert.Equal(t, firstOpenEventCouter, 0)
 				default:
 					assert.Equal(t, test.resp, server.msg.Type)
+					assert.Equal(t, firstOpenEventCouter, 0)
 				}
 			})
 		}


events/firstopen/firstopen_event.go
```diff
@@ -10,69 +9,45 @@ import (
...
+		return
+	}
+	n := FirstOpenNotifier{
+		cm:                     cm,
+		emitFirstTimeOpenEvent: emitFirstTimeOpenEvent,
 	}
+	deviceLocation.Subscribe(n.notifyOnceAppJustInstalled)
 }
 
 // NotifyOnceAppJustInstalled checks whether this is a fresh installation,
```
thegravi: [nit]
```suggestion
// notifyOnceAppJustInstalled checks whether this is a fresh installation,
```

events/firstopen/firstopen_event.go
```diff
@@ -10,69 +9,45 @@ import (
...
-	guard        *sync.Once
+	cm                     *config.FilesystemConfigManager
+	emitFirstTimeOpenEvent func() error
 }
 
-// RegisterNotifier sets up a `FirstOpenNotifier` using the shared `defaultGuard`.
-// If `cm.NewInstallation` is true and we haven’t yet published, it subscribes
+// RegisterNotifier sets up a `FirstOpenNotifier`.
+// If `cm.NewInstallation` is true and the first open event is not yet published, it subscribes
 // [firstopen.NotifyOnceAppJustInstalled] to the device location events stream.
```
thegravi: [nit]
```suggestion
// [firstopen.notifyOnceAppJustInstalled] to the device location events stream.
```

events/firstopen/firstopen_event.go
```diff
@@ -10,69 +9,45 @@ import (
...
-	published    atomic.Bool    // true once we’ve actually published
+	published atomic.Bool // true only when first open event was actually published
 )
 
 // FirstOpenNotifier publishes a one-time `first_open` event when
 // `cm.NewInstallation` is true and the device location event fires.
-// It uses a `sync.Once` guard plus an `atomic.Bool` to enforce exactly-once
-// behavior.
+// It uses an `atomic.Bool` to enforce exactly-once behavior.
 type FirstOpenNotifier struct {
```
thegravi: [small] This can be internal, since it's only used in this file.

events/firstopen/firstopen_event.go
```diff
@@ -10,69 +9,45 @@ import (
...
-		ItemType:      "button",
-		ItemValue:     "first_open",
-		FormReference: "daemon",
-	})
+	if err := i.emitFirstTimeOpenEvent(); err != nil {
+		return err
+	}
 	published.Store(true)
+	return nil
 }
```
thegravi: What if...imagine 2 goroutines call `notifyOnceAppJustInstalled()` at nearly the same time, because...well, reasons, or perhaps it's just this notifier got subscribed twice by mistake or notified twice. However, then both go routines would pass and emit the same event eventually. 

This can be used to make check and set really atomic.

```go
func (i *FirstOpenNotifier) notifyOnceAppJustInstalled(_ core.Insights) error {
	if !i.cm.NewInstallation || !published.CompareAndSwap(false, true) {
		return nil
	}
	return i.emitFirstTimeOpenEvent()
}
```

events/firstopen/firstopen_event.go
```diff
@@ -10,69 +9,40 @@ import (
...
+	deviceLocation.Subscribe(n.notifyOnceAppJustInstalled)
 }
 
-// NotifyOnceAppJustInstalled checks whether this is a fresh installation,
+// notifyOnceAppJustInstalled checks whether this is a fresh installation,
 // and, if so, publishes a `first_open` event exactly once.
-func (i *FirstOpenNotifier) notifyOnceAppJustInstalled(_ core.Insights) error {
-	if !i.cm.NewInstallation {
+func (i *firstOpenNotifier) notifyOnceAppJustInstalled(_ core.Insights) error {
+	if !i.cm.NewInstallation || !published.CompareAndSwap(false, true) {
```
keliramu, line 44: `cm.NewInstallation` is not protected in case of concurrent access - do we care?

DoYouEvenCpp: It actually is protected (in the config manager), however I'll fill-in a ticket to fix possible issues with the way how NewInstallation is updated.

events/firstopen/firstopen_event.go
```diff
@@ -1,78 +1,47 @@
...
-// This method is protected by `sync.Once` to guarantee a single emission.
-func (i *FirstOpenNotifier) publishFirstOpen() {
-	i.uiItemsClick.Publish(events.UiItemsAction{
-		ItemName:      "first_open",
-		ItemType:      "button",
-		ItemValue:     "first_open",
-		FormReference: "daemon",
-	})
-	published.Store(true)
+	return i.emitFirstTimeOpenEvent()
```
bartoszWojciechO: Can we have this error wrapped? Would make it easier to trace.

daemon/rpc_connect.go
```diff
@@ -310,6 +310,7 @@ func (r *RPC) connect(
 		}
 	}
 	r.events.Service.Connect.Publish(event)
+	r.events.Service.FirstTimeOpened.Publish(struct{}{})
```
devzbysiu, line 315: Could you add comment here why we are publishing first time opened while connecting? It may be confusing in few months

events/firstopen/firstopen_event.go
```diff
@@ -1,78 +1,47 @@
...
 
 var (
-	defaultGuard = &sync.Once{} // ensures `publishFirstOpen` runs at most once
-	published    atomic.Bool    // true once we’ve actually published
+	published atomic.Bool // true only when first open event was actually published
 )
 
-// FirstOpenNotifier publishes a one-time `first_open` event when
+// firstOpenNotifier publishes a one-time `first_open` event when
 // `cm.NewInstallation` is true and the device location event fires.
```
devzbysiu: Now this `and the device location event fires.` is not true anymore

events/firstopen/firstopen_event.go
```diff
@@ -1,78 +1,47 @@
...
+	cm                     *config.FilesystemConfigManager
+	emitFirstTimeOpenEvent func() error
 }
 
-// RegisterNotifier sets up a `FirstOpenNotifier` using the shared `defaultGuard`.
-// If `cm.NewInstallation` is true and we haven’t yet published, it subscribes
-// [firstopen.NotifyOnceAppJustInstalled] to the device location events stream.
+// RegisterNotifier sets up a `firstOpenNotifier`.
+// If `cm.NewInstallation` is true and the first open event is not yet published, it subscribes
+// [firstopen.notifyOnceAppJustInstalled] to the device location events stream.
```
devzbysiu: `to the device location events stream` - this also isn't true now

events/firstopen/firstopen_event_test.go
```diff
@@ -104,60 +104,39 @@ func Test_ConcurrentInstallEvents_OnlyOnePublish(t *testing.T) {
 
 	cm := &config.FilesystemConfigManager{NewInstallation: true}
 	sub := &stubSubscriber{}
-	pub := &stubPublisher{}
-	guard := &sync.Once{}
-
-	registerNotifier(cm, sub, pub, guard)
+	count := 0
+	testNotifier := func() error {
```
devzbysiu: Maybe this `testNotifier` could be moved to one place and reused?

config/manager.go
```diff
@@ -217,6 +217,7 @@ func (f *FilesystemConfigManager) load(c *Config) error {
 	// always init with default settings and override later with the values from the file
 	*c = *newConfig(f.machineIDGetter)
 
+	// TODO(LVPN-9628): revisist how this mechanism works
```
devzbysiu: [typo] `revisist`

daemon/rpc_connect.go
devzbysiu, line 315: Would it be possible, without too much hustle, to make a test for this in `rpc_connect.go`?

PR #1193, opened 2025-11-07
LVPN-9513: Fix worker init

devzbysiu:
Changes:
- moose worker is initialized with `false` (don't send) only when consent is not completed yet
- moose tracker is initialized with with `true` (send all events) only when consent is granted

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -145,7 +145,7 @@ func (s *Subscriber) Init(httpClient http.Client) error {
 	if cfg.AnalyticsConsent == config.ConsentUndefined {
 		canSendEvents = false
 	}
-	log.Println(internal.DebugPrefix, "allow moose to send events:", canSendEvents)
+	log.Println(internal.InfoPrefix, "[moose] configure to send events:", canSendEvents)
 
 	client := worker.NewHttpClientContext(s.currentDomain)
 	client.Client = httpClient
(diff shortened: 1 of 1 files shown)


events/moose/moose.go
```diff
@@ -139,7 +140,12 @@ func (s *Subscriber) Init(httpClient http.Client) error {
 	var batchSize uint32 = 20
 	compressRequest := true
 
-	sendEvents := cfg.AnalyticsConsent == config.ConsentGranted
+	// can we send events at all? - if consent is undefined (not completed) - don't send
+	canSendEvents := true
+	if cfg.AnalyticsConsent == config.ConsentUndefined {
+		canSendEvents = false
+	}
+	log.Println(internal.DebugPrefix, "events blocked:", canSendEvents)
```
DoYouEvenCpp: nit: this debug is a bit misleading.
With events enabled, it will log `events blocked: true` which I guess is not what might help with debugging :wink: 

```suggestion
	log.Printf(internal.DebugPrefix, "[moose] events %sblocked\n", func() string {
        if canSendEvents {
            return "un"
        }
        return ""
    }()
```

thegravi: [small] Would help adding some "analytics" word there and there. Also why are we leaving debug logs in the prod? They should be at least with info level.

events/moose/moose.go
```diff
@@ -148,20 +154,24 @@ func (s *Subscriber) Init(httpClient http.Client) error {
...
 		batchSize,
 		compressRequest,
 		&client,
 	))); err != nil {
 		return fmt.Errorf("starting worker: %w", err)
 	}
 
+	// can we send only essential or all?
+	sendAllEvents := cfg.AnalyticsConsent == config.ConsentGranted
+	log.Println(internal.DebugPrefix, "sending all events:", sendAllEvents)
```
thegravi: [small] Here also, too abstract log message. Also, debug -> info log level.

events/moose/moose.go
devzbysiu: Good catch! Changed it to make it clear what it means

DoYouEvenCpp (approved):

PR #1195, opened 2025-11-07
LVPN-9357: Make config migration to happen only once

devzbysiu:
Changes:
- use `sync.Once` to run config migration only once per process lifetime
- add test checking that config save is executed only once

### DIFF
--- a/test/mock/config_manager.go
+++ b/test/mock/config_manager.go
@@ -9,16 +9,18 @@ import (
 )
 
 type ConfigManager struct {
-	mu      sync.RWMutex
-	Cfg     *config.Config
-	SaveErr error
-	LoadErr error
-	Saved   bool
+	mu            sync.RWMutex
+	Cfg           *config.Config
+	SaveCallCount int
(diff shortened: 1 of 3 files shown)


test/mock/config_manager.go
```diff
@@ -73,6 +75,7 @@ func (m *ConfigManager) SaveWith(fn config.SaveFunc) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	m.SaveCallCount += 1
```
thegravi, line 78: Why do we need this? Is it only for tests?

devzbysiu: Yes, it's mock config manager. This acts as a Spy to detect how many times `SaveWith` was called

thegravi:
Incorrect ticket key

test/mock/config_manager.go
thegravi: Didn't see that's a mock file, my bad

PR #1197, opened 2025-11-07
LVPN-9312: Set user consent level into context

keliramu:


events/moose/moose.go
```diff
@@ -92,6 +92,11 @@ func (s *Subscriber) changeConsentState(newState config.AnalyticsConsent) error
 
 	s.consent = newState
 
+	consentLevel := consentTypeToInternalType(s.consent)
+	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesConsentLevel(consentLevel)); err != nil {
+		return fmt.Errorf("setting user consent level: %w", err)
+	}
```
thegravi: Can we have a helper func for this? The same code is used in `Init` function also, jsut wrapped in if statement.

events/moose/moose.go
```diff
@@ -1079,3 +1091,16 @@ func deviceTypeToInternalType(deviceType sysinfo.SystemDeviceType) moose.Nordvpn
 
 	return dt
 }
+
+func consentTypeToInternalType(consent config.AnalyticsConsent) moose.NordvpnappConsentLevel {
+	var rc moose.NordvpnappConsentLevel
+	switch consent {
+	case config.ConsentGranted:
+		rc = moose.NordvpnappConsentLevelAnalytics
+	case config.ConsentDenied, config.ConsentUndefined:
```
mariusSincovici: Maybe this function can be changed to return errors, because it should not fallback for `config.ConsentUndefined` essentials.
Optionally, any other value than granted or denied should return error instead of falling back on the essentials.

keliramu: It's internal utility function, why complicate it - if changed the way to return error - then we would ignore that error anyway

events/moose/moose.go
keliramu: Improved

events/moose/moose.go
keliramu: Improved

PR #1201, opened 2025-11-10
LVPN-9523: Moved user home folder to tmpFS options

mariusSincovici:
To ensure that the dynamically create user in mage is owner of its HOME folder, the folder is set into container.HostConfig.Tmpfs. In this way it can be controlled who is its owner.


edvyno1 (approved):
tested it out, build:debDocker works fine now

bartoszWojciechO (approved):

PR #1203, opened 2025-11-10
LVPN-7282: Add services.expiresAt field

dfetti:


events/moose/moose.go
```diff
@@ -264,7 +264,7 @@ func (s *Subscriber) NotifyKillswitch(data bool) error {
 }
 
 func (s *Subscriber) NotifyAccountCheck(any) error {
-	return s.fetchSubscriptions()
+	return errors.Join(s.fetchSubscriptions(), s.fetchAndSetVpnServiceExpiration())
```
mariusSincovici, line 268: If the first fails does it make sense to do the second call? Or it is to fetch as much as possible?

events/moose/moose.go
```diff
@@ -798,6 +817,21 @@ func findOrder(p core.Payment, orders []core.Order) (core.Order, bool) {
 	return orders[index], true
 }
 
+func findVpnServiceExpiration(services core.ServicesResponse) (string, bool) {
+	ok := false
+	var vpnExpiry string
+	for _, service := range services {
```
mariusSincovici: The same find is made in auth.go. Wouldn't be possible to have it in one place only?
If not at least to reuse the server id instead of hardcoded value.

events/moose/moose.go
```diff
@@ -724,17 +724,36 @@ func (s *Subscriber) OnTelemetry(metric telemetry.Metric, value any) error {
...
 
+func (s *Subscriber) fetchAndSetVpnServiceExpiration() error {
+	services, err := s.ClientAPI.Services()
+	if err != nil {
+		return fmt.Errorf("fetching services: %w", err)
+	}
+
+	var serviceErr error
+	expiresAt, ok := findVpnServiceExpiration(services)
+	if !ok {
```
mariusSincovici: `expiresAt` is used when there is an error returned by `findVpnServiceExpiration`.
Does it need to be set to empty in this case? Or should the unset be called?

dfetti: Fixed it. Just not set anything if there is no value

events/moose/moose.go
dfetti: I opted for the option where we have as much data as possible. And these data come from different sources so it made sense to do the second anyway.

auth/auth.go
```diff
@@ -196,22 +196,30 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
...
 }
 
-// fetchSaveServices fetches services and updates data appropriately
-func (r *RenewingChecker) fetchSaveServices(userID int64, data *config.TokenData) error {
-	services, err := r.creds.Services()
-	if err != nil {
-		return err
-	}
-
+func FindVpnServiceExpiration(services core.ServicesResponse) (string, bool) {
```
DoYouEvenCpp: Can you please add (even a one-liner) brief comment about this method?

DoYouEvenCpp: Also, is there any particular reason to not use (string, error) here instead?
Elevating this to an error might be more flexible in possible error handling across multiple layers of the app (assuming, this is indeed an exported method, and will be used externally)

events/moose/moose.go
```diff
@@ -736,17 +737,43 @@ func (s *Subscriber) OnTelemetry(metric telemetry.Metric, value any) error {
...
+		return fmt.Errorf("no vpn service")
+	}
+
+	// We must use YYYY-MM format
+	parts := strings.Split(expiresAt, "-")
+	if len(parts) < 2 {
+		return fmt.Errorf("invalid date format")
+	}
+	date := fmt.Sprintf("%s-%s", parts[0], parts[1])
```
DoYouEvenCpp: Why not use https://pkg.go.dev/time#Parse combined with https://pkg.go.dev/time#Time.Date?

auth/auth.go
dfetti: Added comment and changed to error. No reasons there.

events/moose/moose.go
dfetti: Done that too. Good suggestion. Thanks!

auth/auth.go
```diff
@@ -196,22 +196,35 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
...
 			break
 		}
 	}
 
-	if vpnExpiry == "" {
+	if !ok {
+		return "", fmt.Errorf("vpn service not found")
+	}
+
+	return vpnExpiry, nil
```
DoYouEvenCpp: [nit] could be still simplified a bit here :wink: 

```
 for _, service := range services {
	// VPN service ID
	if service.Service.ID == VPNServiceID {
		return service.ExpiresAt, nil
	}
}
	
return "", fmt.Errorf("vpn service not found")

File: auth/auth.go [hunk 7]
```diff
@@ -196,22 +196,35 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
...
 }
 
-// fetchSaveServices fetches services and updates data appropriately
-func (r *RenewingChecker) fetchSaveServices(userID int64, data *config.TokenData) error {
-	services, err := r.creds.Services()
-	if err != nil {
-		return err
-	}
-
+// This will return the VPN service expiration date in the format YYYY-MM-DD HH:MM:SS
```
DoYouEvenCpp: [nit] might be also worth to mention when an error is returned, something like:
`FindVpnServiceExpiration returns VPN service expiration date in the format of YYY-MM-DD HH:MM:SS if there is matching VPN service ID, an error otherwise`

PR #1206, opened 2025-11-10
LVPN-9503: Update "Account" page

thegravi:
**Added**:
- "Manage subscription" and "Change password" links
- Information section about the date when the user account was created
- Information section about the date when the VPN subscription was activated
- `InternalLink`, `ExternalLink` objects for managing presentation of link based on outgoing URL.

**Fixed**:
- Incorrect UTM variable for NordLocker

**Removed**:
- All "legal" links to the Nord Account website from the "Account" page

<details>
<summary>Screenshot</summary>
image
</details>

### DIFF
--- a/gui/lib/settings/account_details_screen.dart
+++ b/gui/lib/settings/account_details_screen.dart
@@ -195,6 +195,10 @@ final class UserInfo extends StatelessWidget {
       return "";
     }
 
+    if (account.isSubscriptionExpired) {
+      return t.ui.subscriptionInactive;
+    }
+
     final date = _dateFormat.format(account.vpnExpirationDate!);
     return t.ui.subscriptionValidationDate(expirationDate: date);
   }
--- a/gui/lib/data/repository/account_repository.dart
+++ b/gui/lib/data/repository/account_repository.dart
@@ -36,7 +36,7 @@ final class AccountRepository {
   }
 
   Future<AccountResponse> accountInfo() async {
-    return await _client.accountInfo(AccountRequest(full: false));
+    return await _client.accountInfo(AccountRequest(full: true));
   }
 
   Future<TokenInfoResponse> tokenInfo() async {
--- a/gui/lib/i18n/en/ui.i18n.json
+++ b/gui/lib/i18n/en/ui.i18n.json
@@ -252,5 +252,6 @@
   "accountCreatedOn": "Account created: ${creation_date: String}",
   "manageSubscription": "Manage subscription",
   "changePassword": "Change password",
-  "productHub": "Product Hub"
+  "productHub": "Product Hub",
+  "subscriptionInactive": "Inactive"
 }
--- a/gui/lib/i18n/strings.g.dart
+++ b/gui/lib/i18n/strings.g.dart
@@ -4,7 +4,7 @@
 /// To regenerate, run: `dart run slang`
 ///
 /// Locales: 1
-/// Strings: 624
+/// Strings: 625
 
 // coverage:ignore-file
 // ignore_for_file: type=lint, unused_import
--- a/gui/lib/i18n/strings_en.g.dart
+++ b/gui/lib/i18n/strings_en.g.dart
@@ -52,515 +52,175 @@ class TranslationsCitiesEn {
 	final Translations _root; // ignore: unused_field
 
 	// Translations
-
-	/// en: 'Tirana'
 	String get tirana => 'Tirana';
-
-	/// en: 'Algiers'
 	String get algiers => 'Algiers';
-
-	/// en: 'Addis Ababa'
 	String get addis_ababa => 'Addis Ababa';
-
-	/// en: 'Andorra la Vella'
 	String get andorra_la_vella => 'Andorra la Vella';
-
-	/// en: 'Buenos Aires'
 	String get buenos_aires => 'Buenos Aires';
-
-	/// en: 'Yerevan'
 	String get yerevan => 'Yerevan';
-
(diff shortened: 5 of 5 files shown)


gui/lib/widgets/link_types.dart
```diff
@@ -0,0 +1,19 @@
...
+final class InternalLink<T> extends Link<T> {
+  InternalLink({super.key, required super.title, required super.uri});
+}
+
+/// A clickable link widget for third-party pages.
+///
+/// Extends [IconLink] to display a link with an external link icon.
+/// Use this for links to pages outside the company's ecosystem.
+/// The icon is automatically added to indicate an external destination.
+final class ExternalLink<T> extends IconLink<T> {
```
mariusSincovici: Is this used anywhere?

gui/lib/widgets/link_types.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:nordvpn/widgets/link.dart';
+
+/// A clickable link widget for first-party pages.
+///
+/// Extends [Link] to provide a simple text link without any icons.
+/// Use this for links to publicly accessible pages within the company's ecosystem.
+final class InternalLink<T> extends Link<T> {
```
mariusSincovici: I think the name is a bit misleading,  I've had the impression it opens something inside the application.

gui/lib/widgets/link_types.dart
mariusSincovici, line 1: [minor] Just for this might not be really needed 2 new classes. `Ljnk` class can have multiple constructors and then they would be something like `Link.some_name`.

gui/assets/images/external_link.svg
mariusSincovici, line 1: Is this used?

gui/lib/settings/account_details_screen.dart
```diff
@@ -68,138 +147,171 @@ final class UserInfo extends StatelessWidget {
...
+                        style: settingsTheme.otherProductsSubtitle,
+                      ),
+                    ],
+                  ),
+                ),
+              ),
+            ],
+          ),
+          // Row 2: Empty area offset + learn more link
+          Row(
```
mariusSincovici, line 294: Why can't this be the 3rd item into the above column? And then no calculations are required.

gui/lib/settings/account_details_screen.dart
```diff
@@ -43,14 +76,60 @@ final class AccountDetailsSettings extends ConsumerWidget {
   Widget _build(BuildContext context, WidgetRef ref, UserAccount? userAccount) {
     return SettingsWrapperWidget(
       itemsCount: 1,
-      stickyHeader: UserInfo(
-        key: AccountWidgetKeys.userInfo,
-        userAccount: userAccount,
-        onLogout: () => ref.read(accountControllerProvider.notifier).logout(),
+      itemBuilder: (_, _) => Column(
+        spacing: 48,
```
mariusSincovici: Maybe this should be added into the theme and reference it, like in design.

gui/lib/settings/account_details_screen.dart
thegravi: The icon is vertically centered based on the combined heights of the title and subtitle. If the "Learn mode" link is moved to the same row, a fixed offset would need to be calculated — but this offset would break if font sizes change, making the alignment no longer truly vertical according to the designs.
Also, by keeping the link on the second row and calculating only the horizontal offset from the left, we avoid that issue and remain unaffected by font styling changes.

daemon/rpc_account.go
```diff
@@ -137,6 +137,7 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
 		return nil, internal.ErrUnhandled
 	}
 
+	accountInfo.CreatedOn = currentUser.CreatedOn
```
edvyno1, line 140: since we have this information in the daemon, were there any talks about having this info show up in the CLI `nordvpn account` command too?

thegravi: Not yet.

gui/lib/widgets/link_types.dart
```diff
@@ -0,0 +1,16 @@
...
+///
+/// Use for links to pages within the company's ecosystem.
+final class FirstPartyLink<T> extends Link<T> {
+  FirstPartyLink({super.key, required super.title, required super.uri});
+}
+
+/// A clickable link for third-party websites.
+///
+/// Use for links to pages outside the company's ecosystem.
+final class ThirdPartyLink<T> extends IconLink<T> {
```
DoYouEvenCpp: [nit] Is this used anywhere, or already a YAGNI?

thegravi: By design, it will be used for links to 3rd-party websites. Which "could" appear in the near future?
So, this is, let's say, positively, a "waiting" placeholder to be used, since `FirstPartyLink` is already in use.

mariusSincovici: It would be good to try in the future not to add things not used/needed now, because they can change by the time are used(or might never be used). Also could result in some side effects, for example the `external_link` added image is changed, but will be missed so is not updated.

devzbysiu: Agree 100% with @mariusSincovici , writing code upfront is a form of a Boat Anchor anti-pattern :smile:. It may not be a big deal, but if there is no clear need to have it now you'll never make a mistake by not writing this code now

thegravi: Removed

PR #1208, opened 2025-11-11
Link with X11

mariusSincovici:


mariusSincovici:
Implemented in: https://github.com/NordSecurity/nordvpn-linux/pull/1226

PR #1211, opened 2025-11-12
LVPN-8217: Report app start time

dfetti:
It was decided to report the time that the daemon needs to be able to serve grpc requests

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -130,6 +130,8 @@ func initializeStaticConfig(machineID uuid.UUID) config.StaticConfigManager {
 }
 
 func main() {
+	appStartTime := time.Now()
+
 	// pprof
 	if internal.IsDevEnv(Environment) {
 		go func() {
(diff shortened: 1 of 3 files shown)


cmd/daemon/main.go
```diff
@@ -702,6 +704,11 @@ func main() {
 			log.Fatalf("Invalid predefined connection type: %s", ConnType)
 		}
 
+		appStartDuration := time.Since(appStartTime).Milliseconds()
+		if err := analytics.NotifyAppStartTime(appStartDuration); err != nil {
+			log.Printf("Error while setting app start time: %s\n", err)
```
bartoszWojciechO: Maybe add `internal.ErrorPrefix`.

events/moose/moose.go
```diff
@@ -683,6 +683,18 @@ func (s *Subscriber) NotifyDebuggerEvent(e events.DebuggerEvent) error {
 	return s.response(moose.NordvpnappSendDebuggerLoggingLog(e.JsonData, combinedPaths, nil))
 }
 
+func (s *Subscriber) NotifyAppStartTime(duration int64) error {
+	if duration > math.MaxInt32 || duration < math.MinInt32 {
```
DoYouEvenCpp: Maybe we should simply clamp it to <0, MaxInt32>? Everything that's outside of this scope, produces here an error

DoYouEvenCpp (approved):

PR #1214, opened 2025-11-13
LVPN-9557: GUI fastest label issue

thegravi:
**Fix**:
- Updated label

### DIFF
--- a/gui/lib/vpn/servers_list_card.dart
+++ b/gui/lib/vpn/servers_list_card.dart
@@ -159,7 +159,10 @@ final class _ServersListCardState extends State<ServersListCard> {
                         isScrollable: true,
                         tabs: [
                           Tab(text: t.ui.countries),
-                          Tab(text: t.ui.specialServers),
+                          Tab(
+                            key: ServerListWidgetKeys.specialtyServersTab,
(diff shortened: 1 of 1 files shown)


gui/lib/i18n/strings_en.g.dart
```diff
@@ -2587,7 +2584,6 @@ extension on Translations {
 			case 'ui.termsAgreementDescription': return 'By continuing to use this app, you agree to our terms and how we handle your data. To read the terms and privacy policy check the links below.';
 			case 'ui.readMore': return 'Read more';
 			case 'ui.recentConnections': return 'Recent connections';
-			case 'ui.obfuscatedSpecialtyServerSelectionTooltip': return 'In order to select obfuscated server please enable \'Obfuscation\' in the setting.';
```
DoYouEvenCpp, line 2594: I know this is a generated code, but is this change an essential bit of this PR?

DoYouEvenCpp (approved):

gui/lib/i18n/strings_en.g.dart
thegravi: This is some residual thingy that is no longer with us :|

PR #1215, opened 2025-11-14
LVPN-9567: emit disconnect event when user refreshes the VPN connection

bartoszWojciechO:


networker/networker.go
```diff
@@ -387,10 +391,13 @@ func (netw *Combined) restart(
 		log.Println(internal.WarningPrefix, err)
 	}
 
+	stopStartTime := time.Now()
 	err = netw.vpnet.Stop()
 	if err != nil {
+		disconnectCallback(stopStartTime, true, err)
 		return err
 	}
+	disconnectCallback(stopStartTime, true, nil)
```
DoYouEvenCpp: This could be slightly reduced:
```suggestion
	err = netw.vpnet.Stop()
	disconnectCallback(stopStartTime, true, err)
	if err != nil {
		return err
	}
```

DoYouEvenCpp: Is this expected that for both error and OK case we emit `true`, which then is converted to EventStatus equal to `StatusSuccess`?

networker/networker_test.go
```diff
@@ -1191,6 +1201,7 @@ func TestCombined_Reconnect(t *testing.T) {
 				config.NewAllowlist(nil, nil, nil),
 				[]string{"1.1.1.1"},
 				test.enableLan,
+				func(startTime time.Time, success bool, err error) {},
```
DoYouEvenCpp: [nit] how do you feel about defining here in this module, a variable that would hold this empty callback - and simply reuse if for all the tests?

events/events.go
```diff
@@ -210,3 +209,32 @@ type UiItemsAction struct {
...
+	if success {
+		status = StatusSuccess
+	}
+
+	d.eventTemplate.EventStatus = status
+	d.eventTemplate.Duration = time.Since(startTime)
+	d.eventTemplate.Error = err
+
+	d.publishFunc(d.eventTemplate)
+}
```
thegravi: Are we really confident here that `startTime` would always be greater than `Now()`?

bartoszWojciechO: As of now, `startTime` should always be a point in time before the callback is called/`Now()`. Do you think we should have a safeguard here?

networker/networker.go
bartoszWojciechO: Yeah, it should be `false` in the error case. I siplified it so that event status is now based on the value of `err`.

events/events.go
```diff
@@ -210,3 +210,32 @@ type UiItemsAction struct {
...
+
+func NewDisconnectSender(eventTemplate DataDisconnect, publishFunc func(DataDisconnect)) DisconnectSender {
+	return DisconnectSender{
+		eventTemplate: eventTemplate,
+		publishFunc:   publishFunc,
+	}
+}
+
+func (d *DisconnectSender) PublishDisconnect(startTime time.Time, err error) {
```
DoYouEvenCpp, line 234: [nit] Can you please add here a (event most brief) comment that this publish updated event status based on passed error?

events/events.go
bartoszWojciechO: Added a check

PR #1216, opened 2025-11-14
LVPN-9582: Update install scripts used in tests

mariusSincovici:
In the latest OS versions the package managers changed how local packages are configured to be installed. 
The updates are made to support the newer and the older package managers versions.

### DIFF
--- a/ci/test_install_rpm.sh
+++ b/ci/test_install_rpm.sh
@@ -1,11 +1,18 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
+# store the path where the local repository for the application is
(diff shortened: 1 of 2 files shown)


ci/test_install_rpm.sh
```diff
@@ -1,11 +1,16 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
+REPO_PATH="${REPO_DIR}"
```
mariusSincovici, line 5: Add comment why is variable needed

bartoszWojciechO (approved):

PR #1217, opened 2025-11-14
LVPN-9570: Add IP identifier to meshnet iptables rule comments

edvyno1:
All mesh related rules were commented as `-allow-rule-`. Because they are also added one by one, not in a single batch, this means that after one rule was added, the rest would no longer be added as they would have identical comments, this commit fixes that issue by giving each rule their associated IP in the comment, to differentiate between rules

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -196,6 +196,19 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
 	return dipServices, nil
 }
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.4.2
+      triggered-ref: LVPN-9570_mesh_fail_tests
```
edvyno1: fix before merge

PR #1219, opened 2025-11-17
LVPN-9563: delay moose initialisation until consent state is fully known

DoYouEvenCpp:
Changes under this PR:
- delayed exact moment when moose gets initialized, now it's done the when user provides consent decision
- moose is now doesn't keep internal state of the analytics, instead it uses values from the global config
- upon regular startup when the consent state is already known, moose gets initialized with the daemon startup, just like it's done now

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -814,6 +814,10 @@ func (s *Subscriber) fetchAndSetVpnServiceExpiration() error {
 	return nil
 }
 
+func (s *Subscriber) OnFirstOpen() error {
+	return s.response(moose.NordvpnappSendServiceQualityStatusFirstOpenApp(-1, nil))
+}
+
 func (s *Subscriber) fetchSubscriptions() error {
 	cfg, err := s.getConfig()
 	if err != nil {
--- a/.github/workflows/ci-gitlab.yml
+++ b/.github/workflows/ci-gitlab.yml
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.4.2
+      triggered-ref: v1.4.3
--- a/.github/workflows/core_ci.yml
+++ b/.github/workflows/core_ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Install build utilities
         run: |
           go install github.com/magefile/mage@9e91a03eaa438d0d077aca5654c7757141536a60 # v1.15.0
-          go install github.com/goreleaser/nfpm/v2/cmd/nfpm@2e47af64a3f33d32ec1395dbd56165c4feb2bc21 # v2.43.0
+          go install github.com/goreleaser/nfpm/v2/cmd/nfpm@d0d484200d974e7ac6d06ed6d5f02d82d4952b3f # v2.43.4
           go install github.com/google/go-licenses@5348b744d0983d85713295ea08a20cca1654a45e # v1.6.0
 
       - name: Build clean DEB package
--- a/ci/docker/packager/Dockerfile
+++ b/ci/docker/packager/Dockerfile
@@ -1,35 +1,17 @@
-FROM ghcr.io/nordsecurity/nordvpn-linux/golang:debian-bullseye-20250627
+FROM ghcr.io/nordsecurity/nordvpn-linux/golang:debian-bullseye-20251014@sha256:a5156248a5497a3b261bd5288abb3ddc4b0f8177803e06ea53f50d95bbda5c7e
 
-# Need to build nfpm with patch; expeting sometime
-# nfpm upstream will be fixed (issue submitted)
-RUN cd && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends file binutils && \
-    wget --progress=dot:giga "https://github.com/go-task/task/releases/download/v3.34.1/task_linux_amd64.deb" -O /tmp/task.deb && \
-    echo "c40af16a9bda90ffd55c50f72d8f5ad2e7a9a341dc1edf4154e42271463858ad /tmp/task.deb" | sha256sum -c - && \
-    dpkg -i /tmp/task.deb && rm /tmp/task.deb && \
-    git clone https://github.com/keliramu/nfpm.git && \
-    cd nfpm && \
-    git checkout ensure-reproducable-deb-new && \
-    task setup && \
-    task build && \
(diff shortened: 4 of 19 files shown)


daemon/rpc_set_analytics.go
```diff
@@ -24,9 +24,25 @@ func (r *RPC) SetAnalytics(ctx context.Context, in *pb.SetGenericRequest) (*pb.P
...
+		return c
+	}); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+		return &pb.Payload{
+			Type: internal.CodeConfigError,
+		}, nil
+	}
+
+	if err := r.analytics.Init(); err != nil {
+		log.Println(internal.ErrorPrefix, "moose failed to initialize with error:", err)
```
mariusSincovici: Why in this case the error doesn't result in failure?

daemon/rpc_set_analytics.go
```diff
@@ -24,9 +24,25 @@ func (r *RPC) SetAnalytics(ctx context.Context, in *pb.SetGenericRequest) (*pb.P
 		}
 	}
 
-	newConsentLevel := config.ConsentDenied
+	// moose requires consent status updated in the config before triggering initialization
+	if err := r.cm.SaveWith(func(c config.Config) config.Config {
```
mariusSincovici: If this is successful, but afterwards there is an error, the RPC call ends with error. But at the next call the moose call will not be reexecuted and returns nothing to do, because the result is already in settings.

The consent level should be deleted from moose and only use the one from the config.

daemon/userconsent.go
```diff
@@ -75,7 +75,9 @@ func NewConsentChecker(
 //   - save consent as completed and accepted, no consent flow for standard mode countries
 func (acc *AnalyticsConsentChecker) PrepareDaemonIfConsentNotCompleted() {
 	if acc.IsConsentFlowCompleted() {
-		// nothing to do
+		if err := acc.analytics.Init(); err != nil {
+			log.Println(internal.ErrorPrefix, "moose failed to initialize with error:", err)
```
keliramu, line 85: In case of failure here, would be nice to have unique error message going to the log just to know where this error took place. Now it is exactly the same as in `rpc_set_analytics`

events/moose/moose.go
```diff
@@ -49,55 +50,80 @@ type mooseSetConsentIntoContextFunc func(moose.NordvpnappConsentLevel) uint32
...
 	}
 
+	enabled := newState == config.ConsentGranted
+	log.Println(internal.InfoPrefix, LogComponentPrefix, "request to set consent level to", enabled)
 	if err := s.response(s.mooseConsentLevelFunc(enabled)); err != nil {
-		s.consent = config.ConsentDenied
 		return fmt.Errorf("setting new consent level: %w", err)
 	}
 
+	log.Println(internal.InfoPrefix, LogComponentPrefix, "update consent level into contenxt with new value", newState.String())
```
devzbysiu: [typo] contenxt

daemon/userconsent.go
```diff
@@ -75,7 +75,13 @@ func NewConsentChecker(
 //   - save consent as completed and accepted, no consent flow for standard mode countries
 func (acc *AnalyticsConsentChecker) PrepareDaemonIfConsentNotCompleted() {
```
devzbysiu, line 76: Doc comment needs to be updated now

daemon/userconsent.go
```diff
@@ -75,7 +75,13 @@ func NewConsentChecker(
...
 	if acc.IsConsentFlowCompleted() {
-		// nothing to do
+		var cfg config.Config
+		if err := acc.cm.Load(&cfg); err != nil {
+			log.Println(internal.ErrorPrefix, "failed to load config while running daemon preparation", err)
+		}
+		if err := acc.analytics.Init(cfg.AnalyticsConsent); err != nil {
```
devzbysiu, line 84: Maybe we should make `Init` more explicit in case of config load fails? Or maybe just load the config earlier and and pass it to `IsConsentFlowCompleted` and then use here for `Init` 🤔

daemon/rpc_test.go
```diff
@@ -109,15 +110,15 @@ func testRPC() *RPC {
 		&subs.Subject[string]{},
 		&mock.DNSGetter{Names: []string{"1.1.1.1"}},
 		nil,
-		&mockAnalytics{},
+		func() events.Analytics { return &analytics }(),
```
devzbysiu: Is inline `func` call here needed?

events/moose/moose.go
```diff
@@ -162,21 +188,16 @@ func (s *Subscriber) Init(httpClient http.Client) error {
...
 	compressRequest := true
 
-	// can we send events at all? - if consent is undefined (not completed) - don't send
-	canSendEvents := true
-	if cfg.AnalyticsConsent == config.ConsentUndefined {
-		canSendEvents = false
-	}
-	log.Println(internal.InfoPrefix, "[moose] configure to send events:", canSendEvents)
-
+	// moose is now started always after user specifies analytics consent
```
devzbysiu: Maybe this comment should be for `true` in line `#200` below?

devzbysiu (approved):

events/moose/moose.go
```diff
@@ -185,27 +206,29 @@ func (s *Subscriber) Init(httpClient http.Client) error {
...
-		s.EventsDbPath,
-		internal.IsProdEnv(s.BuildTarget.Environment),
+		s.eventsDbPath,
+		internal.IsProdEnv(s.buildTarget.Environment),
 		s,
 		s,
-		sendAllEvents,
+		s.canSendAllEvents.Load(),
 	)); err != nil {
 		if !strings.Contains(err.Error(), "moose: already initiated") {
```
thegravi, line 221: Aren't we also interested in order "generic" cases for this potential error?

events/moose/moose.go
DoYouEvenCpp: Good point, however I'll keep the comment here (I feel comments in-between parameters passed to a function do nothing but clutter it).
Will instead create a constant with a verbose name and will use it directly in the client init method.

events/moose/moose.go
DoYouEvenCpp: We are, actually. :wink:  but this is to be covered under `LVPN-9654`

daemon/userconsent.go
DoYouEvenCpp: Good point. This however feels like a separate story to investigate, and possibly enhance.
Created a follow-up ticket to cover it.

PR #1222, opened 2025-11-17
LVPN-9556: Duplicated recent connections

thegravi:
**Fixed**
- Resolved an issue where two connection models were treated as duplicates only if every field matched, including `server-technologies`.
- Resolved an issue where connections to the same underlying server (in obfuscated mode) were incorrectly treated as unique.

### DIFF
--- a/daemon/recent_connections_rpc_helper.go
+++ b/daemon/recent_connections_rpc_helper.go
@@ -26,7 +26,7 @@ func StorePendingRecentConnection(
 }
 
 // isRecentConnectionSupported returns true if server connection can be used for reconnection,
-// otherwise returns false
+// otherwise false
 func isRecentConnectionSupported(rule config.ServerSelectionRule) bool {
 	return rule != config.ServerSelectionRule_RECOMMENDED && rule != config.ServerSelectionRule_NONE
 }
--- a/daemon/recents/filter.go
+++ b/daemon/recents/filter.go
@@ -26,10 +26,10 @@ func NewFilter(target Model, candidates []Model) *Filter {
 	}
 }
 
-// WithoutSpecificServerFor excludes specific server fields from matching when the target
-// connection type is NOT one of the specified rules.
-// Use this to ignore specific server fields for location-based connections.
-func (f *Filter) WithoutSpecificServerFor(rules []config.ServerSelectionRule) *Filter {
+// WithSpecificServerOnlyFor configures the filter to match specific server fields (ID and name)
+// only when the target connection type is one of the specified rules.
+// For all other connection types, specific server fields are ignored during matching.
+func (f *Filter) WithSpecificServerOnlyFor(rules []config.ServerSelectionRule) *Filter {
 	matchesRule := slices.Contains(rules, f.target.ConnectionType)
 	if !matchesRule {
 		// Connection type is NOT in the list - exclude specific server fields
--- a/daemon/recents/filter_test.go
+++ b/daemon/recents/filter_test.go
@@ -5,171 +5,108 @@ import (
 
 	"github.com/NordSecurity/nordvpn-linux/config"
 	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestFilter_Apply_TechnologySuperset(t *testing.T) {
-	target := Model{
-		Country:            "Australia",
-		City:               "Sydney",
+// Helper to create a basic city connection model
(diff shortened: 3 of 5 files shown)


daemon/recent_connections_rpc_helper.go
```diff
@@ -18,3 +24,157 @@ func StorePendingRecentConnection(
 		eventPublisher(events.DataRecentsChanged{})
 	}
 }
+
+// isRecentConnectionSupported returns true if server connection can be used for reconnection,
+// otherwise returns false
```
DoYouEvenCpp: [ultra nit]
```suggestion
// isRecentConnectionSupported returns true if server connection can be used for reconnection,
// otherwise false
```

daemon/recents/filter.go
```diff
@@ -0,0 +1,114 @@
...
+// Calling this clears any previously set WithoutTechnologies filter to avoid conflicting states.
+func (f *Filter) WithTechnologies(serverTechs []core.ServerTechnology) *Filter {
+	f.excludeTechnologies = false
+	f.includeTechnologies = serverTechs
+	return f
+}
+
+// Apply applies all configured filters and returns all matching models
+func (f *Filter) Apply() []Model {
+	var result []Model
```
DoYouEvenCpp, line 62: [nit] is preallocation here is desired (or based on the number of candidate this is not a case)?

daemon/recents/filter_test.go
```diff
@@ -0,0 +1,293 @@
...
+	filter := NewFilter(target, candidates)
+	filter.WithoutSpecificServerFor([]config.ServerSelectionRule{
+		config.ServerSelectionRule_SPECIFIC_SERVER,
+		config.ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP,
+	})
+	filter.WithTechnologies(target.ServerTechnologies)
+
+	result := filter.Apply()
+
+	if assert.Len(t, result, 1, "expected 1 match") {
```
DoYouEvenCpp: I guess we don't have to combine if and assertion here - why not simply remove if, and allow test fail when this condition is not met?

daemon/recents/filter_test.go
```diff
@@ -0,0 +1,293 @@
+package recents
```
DoYouEvenCpp, line 1: First 5 tests can be (easily?) rewritten to a table-driven tests - care to go for it?

daemon/recents/filter_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+import (
+	"testing"
+
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFilter_Apply_TechnologySuperset(t *testing.T) {
```
DoYouEvenCpp: Tests are missing `category` values provided.

daemon/recents/recent_connections.go
```diff
@@ -235,3 +188,16 @@ func (r *RecentConnectionsStore) checkExistence() error {
 	}
 	return nil
 }
+
+// modelEquals compares two models for equality
+func (r *RecentConnectionsStore) modelEquals(a, b Model) bool {
```
DoYouEvenCpp: I guess this should be implemented directly for Model by itself and also exported from there. Unless any specific reason no to go for that option?

daemon/recents/filter.go
```diff
@@ -0,0 +1,114 @@
...
+	return &Filter{
+		target:     target,
+		candidates: slices.Clone(candidates),
+	}
+}
+
+// WithoutSpecificServerFor excludes specific server fields from matching when the target
+// connection type is NOT one of the specified rules.
+// Use this to ignore specific server fields for location-based connections.
+func (f *Filter) WithoutSpecificServerFor(rules []config.ServerSelectionRule) *Filter {
```
bartoszWojciechO: This name is a bit confusing. `WhithoutSpecificServerFor(rules)`, it implies imo that specific fields will be ignored for the server that matches the rules. But this method works the other way around, it transforms the filter so that specific fields are ignored for server that doesn't match the rules. Or maybe I misunderstood something?

daemon/recents/filter.go
thegravi: We cannot know in advance how many we would find, if any.

daemon/recents/recent_connections.go
thegravi: Right on spot

daemon/recents/filter.go
DoYouEvenCpp: Oh, right. We filter things out here, in the end.

PR #1224, opened 2025-11-18
LVPN-9560: "Recent connections" not updated immediately in GUI

thegravi:
**Fix**
- Recent connections were not immediately updated in the GUI on change to user settings, unless switched back and forth between tabs to invoke manual refresh.

### DIFF
--- a/gui/lib/data/providers/app_state_provider.dart
+++ b/gui/lib/data/providers/app_state_provider.dart
@@ -184,17 +184,16 @@ class AppStateChange {
     for (final observer in _settingsObservers) {
       observer.onSettingsChanged(appSettings);
     }
-    // for some user changes refresh servers list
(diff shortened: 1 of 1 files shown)


gui/lib/data/providers/app_state_provider.dart
```diff
@@ -219,6 +222,9 @@ class AppStateChange {
     });
   }
 
+  bool _shouldRefreshRecentConnectionsList(ApplicationSettings settings) =>
+      _shouldRefreshServersList(settings);
```
DoYouEvenCpp: [nit] Since `_shouldRefreshServersList` is used to detect changed server list and now also connection list, should we then rename it, to reflect its a bit more generic purpose?
Thing is I  can't come up with a good name here :smile:

PR #1226, opened 2025-11-19
LVPN-9595: Update flutter to 3.38.1

mariusSincovici:
Update flutter to 3.38.1 to fix issues with X11.
To prevent future issues on X11, link with X11 libraries and initialize its threads at the application startup.


gui/pubspec.yaml
```diff
@@ -28,14 +30,14 @@ dependencies:
   flutter_riverpod: ^2.4.10
   riverpod_annotation: ^2.3.4
   riverpod: ^2.5.1
-  protobuf: ^4.1.0
-  grpc: ^4.0.4
+  protobuf: ^5.1.0
```
thegravi, line 33: We would also need to update the protobuf version for Go

mariusSincovici: For this one probably not.
But we'll need to review the plugins versions. I'll check in a separate PR that, because it would require more changes.

gui/pubspec.yaml
```diff
@@ -8,8 +8,10 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 # more info into the build scripts
 version: 0.0.1
 
+
 environment:
-  sdk: ">=3.8.0 <4.0.0"
+  sdk: ">=3.10.0 <3.10.1"
```
mariusSincovici, line 13: Check .lock version why was not marked as changed.

mariusSincovici: File was added

PR #1227, opened 2025-11-19
LVPN-9271: Detect container environment

dfetti:
So far we used to detect only Desktop/Server. We must also detect when we run inside containers.
We agreed to use virtual_machine due to the moose restrictions.

### DIFF
--- a/sysinfo/device_type.go
+++ b/sysinfo/device_type.go
@@ -1,18 +1,22 @@
 package sysinfo
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
 type SystemDeviceType string
 
 const (
-	SystemDeviceTypeUnknown SystemDeviceType = "unknown"
-	SystemDeviceTypeDesktop SystemDeviceType = "desktop"
-	SystemDeviceTypeServer  SystemDeviceType = "server"
+	SystemDeviceTypeUnknown   SystemDeviceType = "unknown"
+	SystemDeviceTypeDesktop   SystemDeviceType = "desktop"
+	SystemDeviceTypeServer    SystemDeviceType = "server"
+	SystemDeviceTypeContainer SystemDeviceType = "container"
 )
 
 const (
@@ -31,6 +35,7 @@ type deviceTypeDetector interface {
 // desktop environment presence, and session type, in that order.
 func GetDeviceType() SystemDeviceType {
 	detectors := []deviceTypeDetector{
+		newContainerDetector(),
 		newSystemdTargetDetector(),
 		newGraphicalEnvDetector(),
 		newXDGSessionDetector(),
@@ -154,4 +159,72 @@ func newXDGSessionDetector() deviceTypeDetector {
 			return getDisplayProtocol(os.Getenv), nil
 		},
 	}
+}
+
+// -------------------------------------
+// Container Detector
+// -------------------------------------
+
+type containerDetector struct {
+	checkIfFileExists    func(string) bool
+	readEnv              func(string) string
+	readEnvironAndCgroup func() ([]byte, []byte)
+}
+
+func (d containerDetector) Get() (SystemDeviceType, error) {
+	if d.isDocker() || d.isKubernetes() || d.isLXC() {
+		return SystemDeviceTypeContainer, nil
+	}
+
+	return SystemDeviceTypeUnknown, nil
+}
+
+func newContainerDetector() deviceTypeDetector {
+	return &containerDetector{
+		checkIfFileExists:    internal.FileExists,
+		readEnv:              os.Getenv,
+		readEnvironAndCgroup: readEnvironAndCgroup,
+	}
+}
+
+func readEnvironAndCgroup() ([]byte, []byte) {
+	environ, err := os.ReadFile("/proc/1/environ")
+	if err != nil {
+		return []byte{}, []byte{}
+	}
+
+	cgroup, err := os.ReadFile("/proc/1/cgroup")
+	if err != nil {
+		return environ, []byte{}
+	}
+
+	return environ, cgroup
+}
+
+// isDocker detects docker environment looking for /.dockerenv file
+func (d containerDetector) isDocker() bool {
+	// If /.dockerenv exists we are running inside docker
+	return d.checkIfFileExists("/.dockerenv")
+}
+
+// isKubernetes detects Kubernetes environment looking for KUBERNETES_SERVICE_HOST env
+func (d containerDetector) isKubernetes() bool {
+	return d.readEnv("KUBERNETES_SERVICE_HOST") != ""
+}
+
+// isLXC detects traditional LXC or LXD containers via cgroup or environment variables.
+func (d containerDetector) isLXC() bool {
+	// Some LXC containers may have container=lxc env variable
+	if val := d.readEnv("container"); strings.ToLower(val) == "lxc" {
+		return true
+	}
+
+	// Check /proc/1/environ for container=lxc or container=lxd and /proc/1/cgroup for lxc or lxd
+	environ, cgroup := d.readEnvironAndCgroup()
+	if bytes.Contains(environ, []byte("container=lxc")) || bytes.Contains(environ, []byte("container=lxd")) ||
+		bytes.Contains(cgroup, []byte("lxc")) || bytes.Contains(cgroup, []byte("lxd")) {
+		return true
+	}
+
+	return false
 }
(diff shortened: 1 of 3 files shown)


sysinfo/device_type.go
```diff
@@ -155,3 +157,46 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+type containerDetector struct{}
+
+func (d containerDetector) Get() (SystemDeviceType, error) {
+	// If /.dockerenv exists we are running inside docker
+	if _, err := os.Stat("/.dockerenv"); err == nil {
+		return SystemDeviceTypeContainer, nil
+	}
+
+	// Check for specific keywords
+	if _, err := os.Stat("/proc/1/mountinfo"); err != nil {
```
devzbysiu: Is this condition correct? If we have error - we continue with reading `mountinfo`? 🤔

sysinfo/device_type.go
```diff
@@ -34,6 +35,7 @@ func GetDeviceType() SystemDeviceType {
 		newSystemdTargetDetector(),
 		newGraphicalEnvDetector(),
 		newXDGSessionDetector(),
+		newContainerDetector(),
```
devzbysiu: If we want't to have reliable detection of containers, we need to make sure that previous detectors won't trigger on containers, because otherwise, the code will never trigger `containerDetector` - it will stop on first detector that meets its requirements, so just make sure this can't happen. 

Maybe `containerDetector` should have highest priority? If we are in container, then that's it?

sysinfo/device_type.go
```diff
@@ -155,3 +157,46 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+				strings.Contains(s, "podman") ||
+				strings.Contains(s, "kubepods") ||
+				strings.Contains(s, "overlay") {
+				return SystemDeviceTypeContainer, nil
+			}
+		}
+	}
+
+	// Check if PID 1 is not systemd
+	// On a normal linux host it will be systemd
```
devzbysiu: What about non-systemd OSes? 

Also should we take into account containers with `systemd`? I think this is niche, so probably not worth spending too much time on such heuristic

sysinfo/device_type.go
```diff
@@ -155,3 +157,46 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+	if _, err := os.Stat("/proc/1/comm"); err != nil {
```
devzbysiu: Same here, is the condition correct?

sysinfo/device_type.go
```diff
@@ -155,3 +157,46 @@ func newXDGSessionDetector() deviceTypeDetector {
 		},
 	}
 }
+
+// -------------------------------------
+// Container Detector
+// -------------------------------------
+
+type containerDetector struct{}
```
devzbysiu: This should be covered with tests. We already have tests for other detectors

sysinfo/device_type.go
```diff
@@ -155,3 +157,46 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+
+func (d containerDetector) Get() (SystemDeviceType, error) {
+	// If /.dockerenv exists we are running inside docker
+	if _, err := os.Stat("/.dockerenv"); err == nil {
```
DoYouEvenCpp: [nit] IIRC, this is not something standardized by Docker thus not 100% reliable.
In fact this is just a first check from a series of them so nothing critical here I'd say, however just a heads up it might easily become a check that returns false all the time.

sysinfo/device_type.go
dfetti: It's my mistake and it seems I did it twice. Will fix it.

sysinfo/device_type.go
dfetti: Also discussed it with Marius and I think the best solution is to get rid of this check.

sysinfo/device_type.go
dfetti: On it. Sorry :(

sysinfo/device_type.go
devzbysiu: We can go with a different heuristic - we can check `/proc/1/cgroup` if it contains container related keywords, I think this technique is used by k8s, but can't find the code I was looking on some time ago :thinking:

sysinfo/device_type.go
dfetti: Agree on that, but for now, it seems a pretty good indicator

sysinfo/device_type.go
dfetti: I also found this method, but on the latest version Docker, there is no usefull information.

sysinfo/device_type.go
```diff
@@ -155,3 +157,42 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+
+func newContainerDetector() deviceTypeDetector {
+	return &containerDetector{
+		detectSession: func() (SystemDeviceType, error) {
+			// If /.dockerenv exists we are running inside docker
+			if _, err := os.Stat("/.dockerenv"); err == nil {
+				return SystemDeviceTypeContainer, nil
+			}
+
+			// Check for specific keywords
```
keliramu: Do we care detecting `lxc/lxd` linux containers? it's quite easy to add those into detection logic

sysinfo/device_type.go
```diff
@@ -10,9 +10,10 @@ import (
 type SystemDeviceType string
 
 const (
-	SystemDeviceTypeUnknown SystemDeviceType = "unknown"
-	SystemDeviceTypeDesktop SystemDeviceType = "desktop"
-	SystemDeviceTypeServer  SystemDeviceType = "server"
+	SystemDeviceTypeUnknown   SystemDeviceType = "unknown"
+	SystemDeviceTypeDesktop   SystemDeviceType = "desktop"
+	SystemDeviceTypeServer    SystemDeviceType = "server"
+	SystemDeviceTypeContainer SystemDeviceType = "container"
```
keliramu, line 19: Imo, would be good to know what container is being used - `docker` `podman` `lxc/lxd`.
Either second field could be used (if available), or in the same filed value with postfix e.g. `container-docker` `container-podman` `container-lxc` could be used.

sysinfo/device_type.go
```diff
@@ -155,3 +157,42 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+
+type containerDetector struct {
+	detectSession func() (SystemDeviceType, error)
+}
+
+func (d containerDetector) Get() (SystemDeviceType, error) {
+	return d.detectSession()
+}
+
+func newContainerDetector() deviceTypeDetector {
```
thegravi, line 182: We can additionally check for `hypervisor` in `/proc/cpuinfo"` for those cases when virtualization is used.

And these:
`os.Getenv("container") != ""`, `os.Getenv("KUBERNETES_SERVICE_HOST") != ""`
But need to check whether our daemon would even reach these.

sysinfo/device_type.go
dfetti: We can add these. Maybe you can help me with the euritsic

sysinfo/device_type.go
dfetti: The thing is, right now we are limited by the moose. We will report any container environment as virtual_machine. So there will make no sense for now to have specifics on the container side. After we will have this metric released, and if we see it is useful, we can request additional fields and get more specific data.

sysinfo/device_type.go
dfetti: Sure thing, let's chat about these.

sysinfo/device_type.go
```diff
@@ -155,3 +158,82 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+	return &containerDetector{
+		checkIfFileExists:    checkIfFileExists,
+		readEnv:              readEnv,
+		readEnvironAndCgroup: readEnvironAndCgroup,
+	}
+}
+
+func checkIfFileExists(path string) bool {
```
mariusSincovici: isn't this `internal.FileExists`?

sysinfo/device_type.go
```diff
@@ -155,3 +158,82 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+	if _, err := os.Stat(path); err == nil {
+		return true
+	}
+	return false
+}
+
+func readEnv(key string) string {
```
mariusSincovici: [small] Is it really needed since it just forwards to `o.Getenv`?

sysinfo/device_type.go
dfetti: Yes, forgot to update here too

sysinfo/device_type.go
dfetti: I'll remove it

sysinfo/device_type.go
```diff
@@ -155,3 +160,71 @@ func newXDGSessionDetector() deviceTypeDetector {
...
+
+// isKubernetes detects Kubernetes environment looking for KUBERNETES_SERVICE_HOST env
+func (d containerDetector) isKubernetes() bool {
+	return d.readEnv("KUBERNETES_SERVICE_HOST") != ""
+}
+
+// isLXC detects traditional LXC or LXD containers via cgroup or environment variables.
+func (d containerDetector) isLXC() bool {
+	// Some LXC containers may have container=lxc env variable
+	if val := d.readEnv("container"); val == "lxc" || val == "LXC" {
```
devzbysiu: [small] you can go with `strings.EqualFold()` to have case insensitive comparison. Or you can use `strings.ToLower()`

PR #1234, opened 2025-11-21
LVPN-9605: Fresh account without vpn service treat as if vpn service has expired

keliramu:
LVPN-9605: Now (temporarily) fresh account without vpn service treat as if vpn service has expired

### DIFF
--- a/gui/lib/data/providers/vpn_status_controller.dart
+++ b/gui/lib/data/providers/vpn_status_controller.dart
@@ -8,7 +8,7 @@ import 'package:nordvpn/data/providers/popups_provider.dart';
 import 'package:nordvpn/data/repository/daemon_status_codes.dart';
 import 'package:nordvpn/data/repository/vpn_repository.dart';
 import 'package:nordvpn/logger.dart';
-import 'package:nordvpn/pb/daemon/status.pb.dart' as pb;
+import 'package:nordvpn/pb/daemon/status.pb.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'vpn_status_controller.g.dart';
@@ -30,7 +30,7 @@ class VpnStatusController extends _$VpnStatusController
   Future<void> connect(ConnectArguments? args) =>
       _doAndShowPopup((vpn) => vpn.connect(args ?? ConnectArguments()));
 
-  Future<void> reconnect(pb.ConnectionParameters args) =>
+  Future<void> reconnect(ConnectionParameters args) =>
       _doAndShowPopup((vpn) => vpn.reconnect(args));
 
   Future<void> disconnect() => _doAndShowPopup((vpn) => vpn.disconnect());
@@ -70,7 +70,7 @@ class VpnStatusController extends _$VpnStatusController
   }
 
   @override
-  void onVpnStatusChanged(pb.StatusResponse status) {
+  void onVpnStatusChanged(StatusResponse status) {
     if (!state.hasValue) {
       // if there is no status already fetched it is not possible to construct the status
       logger.d("ignore VPN status changed because app doesn't have a status");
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -165,6 +166,10 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
 
 	if r.expChecker.IsExpired(data.ServiceExpiry) {
 		if err := r.fetchSaveServices(cfg.AutoConnectData.ID, &data); err != nil {
+			// TODO: now new account no vpn service is treated as if vpn service is expired
+			if strings.Contains(err.Error(), "vpn service not found") {
```
edvyno1, line 173: maybe worth defining as constant somehow or declare as specific error, checking if error contains specific string seems error prone if the returned string were to change in the future

auth/auth.go
```diff
@@ -165,6 +166,10 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
 
 	if r.expChecker.IsExpired(data.ServiceExpiry) {
 		if err := r.fetchSaveServices(cfg.AutoConnectData.ID, &data); err != nil {
+			// TODO: now new account no vpn service is treated as if vpn service is expired
```
DoYouEvenCpp: Something is off with this comment, maybe a missing `with`?

```suggestion
			// TODO: from now on, new account with no vpn service is treated as if vpn service is expired
```

internal/constants.go
```diff
@@ -99,7 +99,7 @@ const (
 
 	NordWhisperInterfaceName = "qtun"
 
-	MaxBytesLimit int64 = 1024*1024*10 + 1 // 10MB + 1 because we allow for values which are equal to the limit
+	MaxBytesLimit int64 = 1024*1024*20 + 1 // + 1 because we allow for values which are equal to the limit
```
DoYouEvenCpp, line 102: This probably is pending a rebase, correct?

auth/auth.go
DoYouEvenCpp: Ideally, that could live by a set of completely separate errors maybe? (but on the other hand, it feels like a another story as it'd be quite some work for sure)

internal/constants.go
keliramu: Cherry-picked, now rebased - but it's what it is on main right now.

auth/auth.go
keliramu: This change/fix is rushed for upcoming release and supposed to be temporary until full/complete fix. 
Clarified TODO comment. This string constant being checked is only partial error and is not supposed to be reused, so, moving it into separate constant is not worth the hassle.

auth/auth.go
keliramu: Clarified TODO comment.

edvyno1 (approved):

gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -57,6 +57,25 @@ class VpnStatusController extends _$VpnStatusController
       status = DaemonStatusCode.failedToConnectToVpn;
     }
 
+    if (status != DaemonStatusCode.success &&
```
mariusSincovici: This is not needed, because the problem has different root cause.

keliramu: Removed

gui/lib/data/providers/vpn_status_controller.dart
mariusSincovici, line 1: Do you need the changes from this file?
If they are that rerun code generation

gui/lib/data/repository/app_state_repository.dart
mariusSincovici, line 1: Run the code generation to update the also .g.dart file.

gui/lib/data/providers/vpn_status_controller.dart
keliramu: Removed

gui/lib/data/repository/app_state_repository.dart
keliramu: Done

PR #1235, opened 2025-11-21
LVPN-9559: GUI recents missing virtual flag

thegravi:
**Added**
- All virtual connections are now marked with the `Virtual` label

### DIFF
--- a/gui/lib/vpn/recent_server_list_item.dart
+++ b/gui/lib/vpn/recent_server_list_item.dart
@@ -86,7 +86,7 @@ class RecentServerListItem extends StatelessWidget {
           mainAxisAlignment: MainAxisAlignment.center,
           children: [
             specialtyTitle,
-            Text(subtitle, style: appTheme.caption),
+            Text(_maybeAddVirtualLabel(subtitle), style: appTheme.caption),
           ],
         );
       }


gui/lib/data/models/server_group_extension.dart
```diff
@@ -6,12 +6,22 @@ extension Convert on ServerGroup {
 
   ServerType? toSpecialtyType() {
     return switch (this) {
+      // groups
       ServerGroup.DOUBLE_VPN => ServerType.doubleVpn,
```
bartoszWojciechO: In `gui/lib/data/models/server_info.dart` on line 108 there is a map that does the same conversion, can this be unified?

gui/lib/vpn/recent_server_list_item.dart
```diff
@@ -101,6 +101,12 @@ class RecentServerListItem extends StatelessWidget {
       );
     }
 
+    Text maybeAddVirtualLabel(Text text) {
```
DoYouEvenCpp: [nit] `maybe` doesn't provide full picture when a virtual label is nadded.
Thus, what about:
```suggestion
    Text addVirtualLabelForVirtualConnection(Text text) {
```

On the other hand, this ` virtual` is now repeated which is also not great. So, leaving this as a `nit` only :smile:

gui/lib/vpn/recent_server_list_item.dart
devzbysiu: I know this is already merged, but I would suggest to move `RegExp` to `static final` to not recreate it with every build

devzbysiu: Also could you add example text against which this regex is supposed to be run? It makes reviewing easier and will be cleaner in few months when everyone will forget what's the case here

gui/lib/vpn/recent_server_list_item.dart
devzbysiu: There are few issues with this method:
- you construct `Text` widget argument even when it may not be needed
- it's not obvious that the text argument is the fallback
- you should not peek into `Text.data` and `Text.style` fields, it's rather: you should construct widgets from data, not fetch data from widgets, so current approach is not idiomatic flutter
- this has a footgun, it will show `null` on `text.data` if you call it with `Text.rich(..)` as an argument

I would suggest to have a helper method only for the text as this is the only thing that varies, the style is also the same for both cases

gui/lib/vpn/recent_server_list_item.dart
devzbysiu, line 39: This is already merged to feature branch, but this `connectArgs` is computed always, but it's not always used in line `#45` below

Those vars here could be inlined

gui/lib/vpn/recent_server_list_item.dart
```diff
@@ -111,7 +117,7 @@ class RecentServerListItem extends StatelessWidget {
         mainAxisAlignment: MainAxisAlignment.center,
         children: [
           Text(model.country, style: appTheme.body),
-          Text(model.city, style: appTheme.caption),
+          maybeAddVirtualLabel(Text(model.city, style: appTheme.caption)),
```
devzbysiu: Btw. so far we used this `ServerListItemFactory` to produce the item and this factory has logic to determine the `Virtual` label for existing cases. Here, in those changes, the logic for this is in the widget itself. Maybe we should consider moving this to factory and have the widget be just dummy container for data that's responsible only for displaying it nicely. It's easier to test this way.

Actually it's also comment for images - the `CustomListTile` we are producing so far in the factory is getting arguments what do display. `RecentServerListItem` is getting `ImagesManager` (but we also have this pattern in the codebase already so not the end of the world)

gui/lib/vpn/recent_server_list_item.dart
thegravi: Following PR has this updated with virtual bool  as an arg.

PR #1241, opened 2025-11-24
LVPN-9647: Upgrade nfpm

keliramu:
- Upgrade to latest nfpm code;
- Get rid of internal fix/hack for reproducible packages;
- Extract version date/time and use it for nfpm package internal files;

### DIFF
--- a/.github/workflows/ci-gitlab.yml
+++ b/.github/workflows/ci-gitlab.yml
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.4.2
+      triggered-ref: v1.4.3


bartoszWojciechO (approved):

ci/docker/packager/Dockerfile
```diff
@@ -1,35 +1,15 @@
...
-    task setup && \
-    task build && \
-    strip ./nfpm && \
-    file ./nfpm && \
-    ls -lh ./nfpm && \
-    ./nfpm --version
+ENV GOBIN=/usr/bin
 
-FROM ghcr.io/nordsecurity/nordvpn-linux/golang:debian-bullseye-20250627
+RUN go install github.com/goreleaser/nfpm/v2/cmd/nfpm@v2.43.4 && \
```
mariusSincovici: Could you bump the version to the pipeline, to keep them in sync?
https://github.com/NordSecurity/nordvpn-linux/blob/ebb55d669bba43065572d2e5fb090efdfc300422/.github/workflows/core_ci.yml#L76

ci/nfpm/build_packages_resources.sh
```diff
@@ -78,6 +78,8 @@ case "$PKG_TO_BUILD" in
 	;;
 esac
 
+# value must be given from outside
+echo "${VERSION_DATE}"
```
edvyno1, line 82: debug leftovers? or left here for clarity?

ci/docker/packager/Dockerfile
keliramu: Unified versions in packager docker file and github core ci

ci/nfpm/build_packages_resources.sh
keliramu: Added here as a mandatory check it is set - if not set, it will break out with error

keliramu: This variable is used in `template.yaml` if variable is not set - there it will not complain, so, added check here.

edvyno1: would be nice if you add this explanation as comment as it is not obvious what the intent of the echo was just by looking at it

keliramu: Updated comment

edvyno1 (approved):

PR #1242, opened 2025-11-24
LVPN-9635: Automate Meshnet smoke tests

pvraketa:
Added: 
 TCID for test cases
 smoke mark for test cases


test/qa/lib/meshnet.py
```diff
@@ -806,3 +807,24 @@ def delete_machines_by_identifier(token: str, identifiers: list | None = None) -
         except requests.RequestException as e:
             logging.log(f"Got an error during DELETE request for {identifier}: {e}")
     session.close()
+
+def get_ping_latency(hostname: str = "", qa_peer: bool = False, ssh_client: ssh.Ssh = None) -> float | None:
```
bartoszWojciechO: Was this test checked for flakiness? Not sure about meshnet, but in case of VPN first pings take more time. Maybe it would be safer to do a few pings and calculate an average?

test/qa/lib/meshnet.py
```diff
@@ -22,6 +22,7 @@
 TELIO_EXPECTED_RELAY_TO_DIRECT_TIME = 5.0
 TELIO_EXPECTED_RTT = 5.0
 TELIO_EXPECTED_PACKET_LOSS = 0.0
+PEER_EXPECTED_LATENCY = 12
```
bartoszWojciechO: What unit is this? Milliseconds?

test/qa/lib/meshnet.py
```diff
@@ -806,3 +807,24 @@ def delete_machines_by_identifier(token: str, identifiers: list | None = None) -
...
+    :hostname:  str             Hostname to ping
+    :qa_peer:   bool            Flag if ping should be executed from qa-peer container
+    :ssh_client:                Instance of ssh.Ssh class for qa_peer
+
+    :return:    float | None    Time of ping latency in case is time was found, None otherwise
+    """
+    if qa_peer:
+        ping_output = ssh_client.exec_command(f"ping -c 1 {hostname}")
+    else:
+        ping_output = sh_no_tty.ping("-c", "1", hostname)
```
mariusSincovici: ping must not be used because is unreliable.

test/qa/lib/meshnet.py
pvraketa: Yes

test/qa/lib/meshnet.py
pvraketa: The manual test case uses ping and has been in existence since 2022. The scope of this task is to automate the manual test case

test/qa/lib/meshnet.py
pvraketa: That's why we use <= 12 in the assert. I have been running it locally several times

test/qa/test_meshnet.py
```diff
@@ -182,9 +186,11 @@ def test_set_meshnet_off_when_logged_out(meshnet_allias):
     assert "You are not logged in." in ex.value.stdout.decode("utf-8")
 
 
+@pytest.mark.smoke
```
edvyno1, line 189: what is the difference between smoke mark and core_meshnet mark, aren't all smoke tests supposed to be in the core set?

test/qa/lib/meshnet.py
edvyno1: flakiness locally and flakiness in CI is very different, if we are planning to run these tests in day to day pipelines, they must be trustworthy and we mustn't get used to retrying these tests again, please do some testing for reliability in CI as well

bbrutall (approved):

ci/docker/qa-peer/Dockerfile
```diff
@@ -6,8 +6,11 @@ EXPOSE 22
 
 RUN apt-get update && \
     apt-get install -y openssh-server sudo iptables iputils-ping \
-        iproute2 procps ca-certificates libxml2 libidn2-0 zlib1g && \
+        iproute2 procps ca-certificates libxml2 libidn2-0 zlib1g sockperf curl && \
     echo "root:root" | chpasswd && \
+    # Adding license for sockperf below
+    mkdir -p /usr/share/licenses/sockperf/ && \
+    curl -s https://raw.githubusercontent.com/Mellanox/sockperf/sockperf_v2/copying > /usr/share/licenses/sockperf/LICENSE && \
```
devzbysiu, line 13: Let's pin the notice to specific git hash of version we need, also in tester image below

test/qa/test_meshnet.py
```diff
@@ -368,3 +382,33 @@ def test_login_mesh_on_set_defaults_logout_login_mesh_on():
...
+    """
+
+    peer_list = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list())
+    host_address = peer_list.get_this_device().ip
+    peer_address = peer_list.get_external_peer().ip
+
+    # Need time for changing from DERP server to direct connection
+    time.sleep(15)
+
+    assert meshnet.get_ping_latency(peer_address, ssh_client=ssh_client) <= meshnet.PEER_EXPECTED_LATENCY_MS, \
```
devzbysiu, line 410: Won't this break in case `None` is returned from `get_ping_latency`?

test/qa/lib/meshnet.py
```diff
@@ -806,3 +807,40 @@ def delete_machines_by_identifier(token: str, identifiers: list | None = None) -
...
+    :return:            float | None    Latency in milliseconds, None on failure
+    """
+    client_command = f"sockperf ping-pong -i {ip} -t {sockperf_duration}"
+    server_command = "nohup sockperf server > /dev/null 2>&1 & echo $!"
+    kill_server_command = "pkill -f 'sockperf server'"
+
+    if qa_peer:
+        sh_no_tty.bash("-c", server_command)
+        output = ssh_client.exec_command(client_command)
+        sh_no_tty.bash("-c", kill_server_command)
```
devzbysiu, line 831: This command for stopping should be done in `try .. finally` so we have confidence it's called in case of any error in commands before. Otherwise `sockperf` won't be stopped

test/qa/test_meshnet.py
```diff
@@ -368,3 +382,33 @@ def test_login_mesh_on_set_defaults_logout_login_mesh_on():
...
+        f"Expected latency is above: {meshnet.PEER_EXPECTED_LATENCY_MS}. Ping was done from {host_address}"
```
devzbysiu, line 411: The messages should probably say that measured latency is above XYZ, not expected

test/qa/lib/meshnet.py
pvraketa: What should be in the `finally` and in the `except` block?

PR #1243, opened 2025-11-26
Moving killswitch RC test to RC suite

edvyno1:
This is done because during new remote config testing, a full build repo pipeline is done, testing the local config files. Any test that tries to download remote config from cdn as a test condition will not work with these local tests. Instead of trying to exclude one test for the killswitch test suite under specific circumstances, it is easier to move the test to the remote_config tests file which does not work and is not run in this situation.

### DIFF
--- a/test/qa/test_remote_config.py
+++ b/test/qa/test_remote_config.py
@@ -3,6 +3,7 @@ import json
 import subprocess
 import time
 import uuid
+import os
 
 import pytest
 import sh
@@ -18,6 +19,7 @@ from lib.daemon import (
     set_rc_retry_custom_time,
 )
 from constants import RC_TIMEOUT
+from test_killswitch import MSG_KILLSWITCH_ON
 
 RC_INVALID_VERSION_MESSAGE = "invalid version constraint: improper constraint: y"
 RC_REMOTE_MESSAGES = [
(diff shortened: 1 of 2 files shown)


keliramu (approved):

test/qa/test_remote_config.py
```diff
@@ -774,3 +776,59 @@ def test_remote_config_change_local_meshnet_config_settings(
...
+    assert MSG_KILLSWITCH_ON in sh.nordvpn.set.killswitch("on")
+    assert daemon.is_killswitch_on()
+
+    sh.nordvpn.disconnect()
+    assert daemon.is_disconnected()
+    print("--- before manual files removal")
+    printdir()
+    # remove previously fetched files
+    # upon restart, they should be loaded again
+    os.system(f"sudo rm -rf {conf_dir}")
```
bbrutall, line 815: Please replace with `sh.sudo.rm(file_path) `as we agreed to use `sh` instead of `os`.

test/qa/test_remote_config.py
```diff
@@ -774,3 +776,59 @@ def test_remote_config_change_local_meshnet_config_settings(
...
+
+    # make sure the rc files are gone
+    for fname in expected_files:
+        path = os.path.join(conf_dir, fname)
+        res = os.popen(f"sudo test -f {path} && echo exists || echo missing").read().strip()
```
bbrutall, line 820: It can be replaced in python way like:
`Path(conf_dir).joinpath(fname) in list(Path(confdir).iterdir())`
or check if it does not exist:
`not Path(conf_dir).joinpath(fname).exists()`

test/qa/test_remote_config.py
```diff
@@ -774,3 +776,59 @@ def test_remote_config_change_local_meshnet_config_settings(
...
+        assert res == "missing", f"File {path} should not exist"
+
+    daemon.restart()
+    print("--- after restart")
+    printdir()
+
+    for _ in range(10):
+        time.sleep(1)
+        printdir()
```
bbrutall, line 829: Why do we need to print dir content 10 times? We can print it if the files are missing in the next verification

test/qa/test_remote_config.py
```diff
@@ -774,3 +776,59 @@ def test_remote_config_change_local_meshnet_config_settings(
...
+
+    for fname in expected_files:
+        path = os.path.join(conf_dir, fname)
+        res = os.popen(f"sudo test -f {path} && echo exists || echo missing").read().strip()
+        assert res == "exists", f"File {os.path} should exist after kill-switch was enabled"
```
bbrutall, line 834: It can be replaced in python way like:
`Path(conf_dir).joinpath(fname).exists()`

test/qa/test_remote_config.py
```diff
@@ -774,3 +776,59 @@ def test_remote_config_change_local_meshnet_config_settings(
...
+    Jira ID: LVPN-8626
+    """
+
+    conf_dir = "/var/lib/nordvpn/conf/"
+    expected_files = [
+        "libtelio.json",
+        "meshnet.json",
+    ]
+    def printdir():
+        os.system(f"cd {conf_dir}/../ && ls -la && ls {conf_dir} -a")
```
bbrutall, line 803: Do we really need this function here?

PR #1244, opened 2025-11-26
GUI "recents" factory update

thegravi:
**Changes**
* Moved recent widget/item construction to a dedicated factory to decouple it from the existing server-list factory, as one handles active connections while the other handles historical entries.

### DIFF
--- a/gui/lib/widgets/popups/popup.dart
+++ b/gui/lib/widgets/popups/popup.dart
@@ -43,9 +43,14 @@ abstract class Popup extends ConsumerWidget {
     return Row(
       mainAxisAlignment: MainAxisAlignment.spaceBetween,
       children: [
-        Row(
-          spacing: appTheme.horizontalSpace,
-          children: [if (leadingIcon != null) leadingIcon!, _title(appTheme)],
+        Expanded(
+          child: Row(
+            spacing: appTheme.horizontalSpace,
+            children: [
+              if (leadingIcon != null) leadingIcon!,
+              Flexible(child: _title(appTheme)),
+            ],
+          ),
         ),
         _closeIcon(context),
       ],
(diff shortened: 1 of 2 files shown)


gui/lib/data/models/recent_connections.dart
```diff
@@ -70,4 +74,11 @@ class RecentConnection {
   bool get isCountryBased {
     return !isSpecialtyServer;
   }
+
+  /// Extracts server ID from specific server name
+  /// e.g., "Lithuania #123" -> "#123"
+  String? get serverId {
+    final match = _serverIdRegex.firstMatch(specificServerName);
+    return match?[1];
+  }
```
DoYouEvenCpp, line 83: What in case the input is either illformed or doesn't match assumed format?
Specifically, won't be go here out of bounds?

gui/lib/widgets/popups/popup.dart
```diff
@@ -41,14 +41,16 @@ abstract class Popup extends ConsumerWidget {
...
-          Padding(
-            padding: EdgeInsets.only(right: appTheme.horizontalSpace),
-            child: leadingIcon!,
-          ),
         Expanded(
-          child: _title(appTheme),
+          child: Row(
+            spacing: appTheme.horizontalSpace,
+            children: [
+              if (leadingIcon != null) leadingIcon!,
```
devzbysiu, line 50: [small] If you are changing this then we can use `?leadingIcon` now if I remember correctly

gui/lib/data/models/recent_connections.dart
devzbysiu: If the match won't find anything, then `match` will be `null` and `[1]` won't happen thanks to `?` - `null` will be returned.

What's tricky here is that if the more capture groups will be added in regex, then index will not point to correct thing, this in fact could be written in a more defensive way or we could have test for this that would break in case of changes

PR #1245, opened 2025-11-26
GUI Recents wrap up fixes

thegravi:
Changes:
- i18n support for country/city
- If the "recents" widget receives an error from the daemon, we will not display it

### DIFF
--- a/gui/lib/vpn/server_list_item_factory.dart
+++ b/gui/lib/vpn/server_list_item_factory.dart
@@ -9,7 +9,6 @@ import 'package:nordvpn/data/models/vpn_status.dart';
 import 'package:nordvpn/i18n/strings.g.dart';
 import 'package:nordvpn/internal/images_manager.dart';
 import 'package:nordvpn/pb/daemon/config/group.pb.dart';
-import 'package:nordvpn/pb/daemon/server_selection_rule.pb.dart';
 import 'package:nordvpn/theme/app_theme.dart';
 import 'package:nordvpn/i18n/string_translation_extension.dart';
 import 'package:nordvpn/theme/servers_list_theme.dart';


gui/lib/vpn/server_list_item_factory.dart
```diff
@@ -9,6 +9,7 @@ import 'package:nordvpn/data/models/vpn_status.dart';
 import 'package:nordvpn/i18n/strings.g.dart';
 import 'package:nordvpn/internal/images_manager.dart';
 import 'package:nordvpn/pb/daemon/config/group.pb.dart';
+import 'package:nordvpn/pb/daemon/server_selection_rule.pb.dart';
```
mariusSincovici: [small] might not be needed

gui/lib/vpn/recent_connections_item_factory.dart
```diff
@@ -77,9 +75,12 @@ final class RecentConnectionsItemFactory {
     if (isSpecialtyServer) {
       var specialtyTitle = Text(model.specialtyServer, style: appTheme.body);
       if (model.country.isNotEmpty) {
-        var subtitle = model.country;
-        subtitle +=
-            " - ${model.city.isEmpty ? t.ui.fastestServer : model.city}";
+        final country = Country.fromCode(model.countryCode);
+        var subtitle = country.localizedName;
+        final city = model.city;
+        final cityLocalized = City(city).localizedName;
```
mariusSincovici: If city is empty `City(city)` asserts.
maybe something like `subtitle += " - ${city.isEmpty ? t.ui.fastestServer : City(city).localizedName}";`
or more changes required to store into the model Country and City objects instead of strings.

PR #1246, opened 2025-11-28
LVPN-9694: fix connect button states on connection refresh

bartoszWojciechO:


daemon/state/connection_info.go
```diff
@@ -123,14 +123,16 @@ func (c *ConnectionInfo) ConnectionStatusNotifyConnect(e events.DataConnect) err
...
 		State:      pb.ConnectionState_DISCONNECTED,
 		TunnelName: "",
 		StartTime:  nil,
 	}
-	c.setStatus(status, false)
-	c.internalNotif.Publish(events.DataConnectChangeNotif{Status: status})
+	if !e.IsRefresh {
+		c.setStatus(status, false)
+		c.internalNotif.Publish(events.DataConnectChangeNotif{Status: status})
+	}
```
DoYouEvenCpp: I'd slightly revert order here. At first test against `IsRefresh` (and if so, bail out early) and only then do all the rest of the logic:

```
	if e.IsRefresh {
		return nil
	}
	status := types.ConnectionStatus{
		State:      pb.ConnectionState_DISCONNECTED,
		TunnelName: "",
		StartTime:  nil,
	}
	c.setStatus(status, false)
	c.internalNotif.Publish(events.DataConnectChangeNotif{Status: status})
	return ni

File: events/events.go [hunk 2]
```diff
@@ -246,13 +247,14 @@ func (d *DisconnectSender) PublishDisconnect(startTime time.Time, err error) {
...
 	status := StatusFailure
-	if err != nil {
+	if err == nil {
 		status = StatusSuccess
 	}
 
 	d.eventTemplate.EventStatus = status
 	d.eventTemplate.Duration = time.Since(startTime)
 	d.eventTemplate.Error = err
+	d.eventTemplate.IsRefresh = true
```
DoYouEvenCpp, line 257: I guess this should be easily unit  tested. Can you please provide some tests to cover that scenario?

PR #1254, opened 2025-12-03
LVPN-7478: Use updated xor template for OVPN

edvyno1:


core/urls.go
```diff
@@ -131,7 +131,7 @@ const (
 	ovpnTemplateURL = "/configs/templates/ovpn/1.0/template.xslt"
 
 	// ovpnObfsTemplateURL defines url to ovpn obfuscated server template
-	ovpnObfsTemplateURL = "/configs/templates/ovpn_xor/1.0/template.xslt"
+	ovpnObfsTemplateURL = "/configs/templates/ovpn_xor/1.1/template.xslt"
```
devzbysiu, line 134: (I know it's draft, but there is also `data.sh` 😄 )

edvyno1: added there as well as edited the test case as well just to be sure

PR #1255, opened 2025-12-03
Fix get version date during build

keliramu:
When introduced VERSION_DATE for `nfpm` package builder to set internal file date/time, one edge case was missed - i.e. when pipeline is triggered from release pipeline e.g. `release/4.3.0` version is extracted from branch name and thus `prod` build is executed, and then attempt to get version date from git tag which is non existing. In case when version git tag is not available, get version date from git head. Same for another case when version date is attempted to get from changelog file - if it does not exists, get version date from git head.

### DIFF
--- a/ci/env.sh
+++ b/ci/env.sh
@@ -2,42 +2,54 @@
 # CI_COMMIT_TAG variable is predefined if run by CI/CD. In case you want to build qa or
 # prod builds locally, set these variables accordingly or use release/X.X.X branch for qa build
 # dev and qa builds contain hash in version name
+
+# VERSION_DATE is expected by nfpm in package template.yaml. For the same version, version date 
+# should be the same - to have reproducible builds/packages.
+# Version date-time is taken from git version tag timestamp or changelog file last modification 
+# timestamp, or, if those not available, use git HEAD commit timestamp.
+
 set -euxo pipefail
 
 NAME=nordvpn
-export NAME
-
 HASH=$(git rev-parse --short HEAD)
-export HASH
+
+format_date_utc() {
+  local date_input="$1"
+  # format date to ISO 8601 UTC format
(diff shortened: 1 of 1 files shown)


ci/env.sh
```diff
@@ -5,39 +5,40 @@
...
-  VERSION_DATE="$(git log -1 --format=%aI "${CI_COMMIT_TAG}" | xargs -I{} date -u -d "{}" +"%Y-%m-%dT%H:%M:%SZ")"
-  export VERSION_DATE
+  # version date should be always the same;
+  # check if the tag actually exists in git, else use HEAD's commit date
+  if git rev-parse "${CI_COMMIT_TAG}" >/dev/null 2>&1; then
+    VERSION_DATE="$(git log -1 --format=%aI "${CI_COMMIT_TAG}" | xargs -I{} date -u -d "{}" +"%Y-%m-%dT%H:%M:%SZ")"
+  else
+    echo "Warning: Git tag '${CI_COMMIT_TAG}' not found, using HEAD instead"
+    VERSION_DATE="$(git log -1 --format=%aI HEAD | xargs -I{} date -u -d "{}" +"%Y-%m-%dT%H:%M:%SZ")"
+  fi
```
DoYouEvenCpp, line 37: Can this get extracted into a separate method, and reused here, and below (so we won't have this much code repetition)?

keliramu: Yes, made code DRY

ci/env.sh
```diff
@@ -2,42 +2,54 @@
...
 NAME=nordvpn
-export NAME
-
 HASH=$(git rev-parse --short HEAD)
-export HASH
+
+format_date_utc() {
+  local date_input="$1"
+  # format date to ISO 8601 UTC format
+  echo "$date_input" | xargs -I{} date -u -d "{}" +"%Y-%m-%dT%H:%M:%SZ"
```
devzbysiu, line 19: [small] `${date_input}` and in other places where applicable

PR #1256, opened 2025-12-04
LVPN-7398: make xfail and skip decorators for meshnet tests to be conditionally enabled

DoYouEvenCpp:
This PR changes how meshnet tests are decorated with `xfail` or `skip`. Now it is taken into actual usage only when a specific env. variable is exported (eg. the `DISABLE_MESHNET_TESTS`)

From now on, meshnet tests will be fully scheduled on nigthly builds.

### DIFF
--- a/test/qa/test_meshnet_routing.py
+++ b/test/qa/test_meshnet_routing.py
@@ -134,7 +134,7 @@ def test_route_to_nonexistant_node():
 
     assert expected_message in ex.value.stdout.decode("utf-8")
 
-@pytest.mark.skipif(meshnet.is_meshnet_test_disabled_from_run(), reason="TOOD: LVPN-9459")
+@pytest.mark.skipif(meshnet.is_meshnet_test_disabled_from_run(), reason="TODO: LVPN-9459")
 def test_route_to_peer_status_valid():
     peer_hostname = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list()).get_external_peer().hostname
 


test/qa/test_meshnet_invite.py
```diff
@@ -1,8 +1,11 @@
+import os
 import pytest
 import sh
 
 from lib import login, meshnet, ssh
 from lib.shell import sh_no_tty
+
+DISABLE_MESHNET_TESTS = os.getenv("DISABLE_MESHNET_TESTS") is not None
```
edvyno1: can possibly add this one to lib/meshnet and import from there so you don't need to import os for every mesh test

test/qa/test_meshnet.py
```diff
@@ -182,6 +184,7 @@ def test_set_meshnet_off_when_logged_out(meshnet_allias):
     assert "You are not logged in." in ex.value.stdout.decode("utf-8")
 
 
+@pytest.mark.xfail(condition=DISABLE_MESHNET_TESTS, reason="Run only in nightly")
```
edvyno1: test previously was not marked as xfail, now is, so it's a flaky test that will be run and not xfail in our day to day pipeline due to core meshnet mark

is this the intention?

test/qa/test_meshnet.py
```diff
@@ -247,6 +250,7 @@ def test_permission_messages_error(permission, permission_state, expected_messag
     assert expected_message in ex.value.stdout.decode("utf-8")
 
 
+@pytest.mark.xfail(condition=DISABLE_MESHNET_TESTS, reason="Run only in nightly")
```
edvyno1: test previously not marked as xfail, now is, was there a decision on this?

test/qa/test_meshnet.py
```diff
@@ -1,17 +1,18 @@
...
 import sh
+import time
+
+from datetime import datetime
 
 import lib
 from lib import daemon, logging, login, meshnet, settings, ssh
 from lib.shell import sh_no_tty
 
+DISABLE_MESHNET_TESTS = os.getenv("DISABLE_MESHNET_TESTS") is not None
```
bbrutall: its better to replace it  with `bool(os.getenv("DISABLE_MESHNET_TESTS"))`

DoYouEvenCpp: Extracted to a helper method, into lib/meshnet

test/qa/test_meshnet.py
DoYouEvenCpp: Yes, this is because for RC pipelines it now gets enabled, and because of this test tends to flip.

bbrutall (approved):

PR #1258, opened 2025-12-04
LVPN-9723: prevent potential NC token request spam

bartoszWojciechO:
In the NC api token response, expiration date is provided in seconds, but daemon treated the value as nanoseconds. This resulted in incredibly short token lifetimes and a refresh request spam.

### DIFF
--- a/cmd/daemon/init_session_stores_test.go
+++ b/cmd/daemon/init_session_stores_test.go
@@ -488,7 +488,7 @@ func TestRenewalFunctions_NCCredentials(t *testing.T) {
 	assert.Equal(t, "nc-user", resp.Username)
 	assert.Equal(t, "nc-pass", resp.Password)
 	assert.Equal(t, "wss://nc.example.com", resp.Endpoint)
-	assert.Equal(t, time.Duration(3600) * time.Second, resp.ExpiresIn)
+	assert.Equal(t, time.Duration(3600)*time.Second, resp.ExpiresIn)
 }
 
 func TestGetTokenData_Success(t *testing.T) {


nc/creds_fetcher.go
```diff
@@ -84,7 +82,10 @@ func (cf *CredentialsGetter) GetCredentialsFromAPI() (config.NCData, error) {
 	ncData.Endpoint = resp.Endpoint
 	ncData.Username = resp.Username
 	ncData.Password = resp.Password
-	ncData.ExpirationDate = time.Now().Add(credentialsValidityPeriod)
+	// validity period is provided in seconds
+
```
devzbysiu: Not needed blank line

cmd/daemon/init_session_stores.go
```diff
@@ -286,11 +286,14 @@ func renewNCCredentials(confman config.Manager, api core.ClientAPI) session.NCCr
 			return nil, fmt.Errorf("getting nc credentials data: %w", err)
 		}
 
+		// validity period is provided in seconds
+		expiresIn := time.Duration(resp.ExpiresIn) * time.Second
```
DoYouEvenCpp, line 290: Can you please reflect this fix in UTs, as well?
Possibly, this could be helpful: https://github.com/NordSecurity/nordvpn-linux/blob/LVPN-9723-nc-spam/cmd/daemon/init_session_stores_test.go#L77

bartoszWojciechO: This was actually included in the UTs, just needed to adjust the time value to seconds.

PR #1261, opened 2025-12-08
LVPN-9721: make RC integration tests based on mocked I/O and HTTP calls

DoYouEvenCpp:
From time to time, during various CI runs some tests from the suite (specifically `TestGetUpdatedTelioConfig`) failed, most of the cases symptoms pointed to timing issues caused by actual I/O operations.

In order to mitigate it, this suite has been rewritten so it now uses solely mocked I/O and HTTP calls.

Changes under this PR:
- grouped some of the standalone methods from `internal` package under a common interface, which is reused in the RC implementation, so the code is actually testable - `RemoteConfigFileOps`
   - those method are now moved to remote package, as they naturally belong there (to the RC implementation)
- grouped `readFile` and `writeFile` methods (also used in internals of the RC implementation) under another interface - `FileStoreOps`

Changes in `config/remote/remote_test.go`:
 - removed actual HTTP calls from all the tests
   - currently there's no HTTP calls, instead CDN interface is mocked, to provide content of requested files
 - removed actual I/O calls from all the tests
   - in-memory OS akin implementation is provided to serve purposes of the I/O calls
 - changed category of some tests from `Unit` to `Integration`


internal/filesystem.go
```diff
@@ -105,6 +105,16 @@ func (rdl *runDirListener) Addr() net.Addr {
 	return rdl.listener.Addr()
 }
 
+// RemoteConfigFileOps aggregates into a common interface
+// a set of file-related operations utilized for the sake of remote-config feature functionality.
+type RemoteConfigFileOps interface {
```
mariusSincovici: [small] This seams to be now specific for RC, wouldn't make more sense to move it into the RC folder/package instead? Including `DefaultRCFileOperations`

config/remote/remote_test.go
```diff
@@ -173,41 +167,146 @@ func (e *timeoutError) Error() string   { return "timeout" }
...
+		files: make(map[string]mockedFileEntry),
+		mu:    sync.RWMutex{},
+	}
+}
+
+func (cdn *mockedCdn) GetRemoteFile(name string) ([]byte, error) {
+	if content, ok := cdn.cdnFiles[name]; ok {
+		return content, nil
+	}
+	return []byte{}, nil
```
mariusSincovici: Shouldn't this return error if file is not found?

config/remote/remote.go
```diff
@@ -33,10 +39,65 @@ type ConfigLoader interface {
...
+type RemoteConfigFileOps interface {
+	IsValidExistingDir(path string) (bool, error)
+	CleanupTmpFiles(targetPath, fileExt string) error
+	RenameTmpFiles(targetPath, fileExt string) error
+}
+
+type FileOpsDefaultImpl struct{}
+
+// WalkFiles iterate files by given extension and do specified action
+func (FileOpsDefaultImpl) WalkFiles(targetPath, fileExt string, actionFunc func(string)) error {
```
keliramu: Plz be consistent: other methods declared using pointer receivers - struct has no fields, so, it's best to have all methods declared using value receivers

config/remote/remote.go
```diff
@@ -33,10 +39,65 @@ type ConfigLoader interface {
...
-)
+// RemoteConfigFileOps aggregates into a common interface
+// a set of file-related operations utilized for the sake of remote-config feature functionality.
+type RemoteConfigFileOps interface {
+	IsValidExistingDir(path string) (bool, error)
+	CleanupTmpFiles(targetPath, fileExt string) error
+	RenameTmpFiles(targetPath, fileExt string) error
+}
+
+type FileOpsDefaultImpl struct{}
```
keliramu: `FileOpsDefaultImpl` is not used outside current package - no need to have it declared as public. This struct and all its methods

config/remote/remote.go
```diff
@@ -33,10 +39,65 @@ type ConfigLoader interface {
 	TryPreload()
 }
 
-const (
-	envUseLocalConfig   = "RC_USE_LOCAL_CONFIG"
-	defaultFeatureState = true
-)
+// RemoteConfigFileOps aggregates into a common interface
+// a set of file-related operations utilized for the sake of remote-config feature functionality.
+type RemoteConfigFileOps interface {
```
keliramu: Interface is not used outside - should not be public (and all its methods)

config/remote/config.go
```diff
@@ -59,7 +59,7 @@ type validator interface {
 
 // download main json file and check if include files should be downloaded,
 // return `true` if remote config was really downloaded.
-func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBasePath, targetPath string) (success bool, err error) {
+func (f *Feature) download(cdn fileReader, fw fileWriter, jv validator, cdnBasePath, targetPath string, fileIO RemoteConfigFileOps) (success bool, err error) {
```
keliramu: Very good - you have improved this dependency to be `injectable` 👍

config/remote/remote.go
```diff
@@ -46,6 +107,12 @@ type RemoteConfigNotifier interface {
 	RemoteConfigUpdate(RemoteConfigEvent) error
 }
 
+// FileStoreOps aggregates file relevant operations under a common interface
+type FileStoreOps interface {
```
keliramu: Should not be declared as public interface - not used outside current package

PR #1267, opened 2025-12-10
LVPN-9760: Fix `test_fileshare_transfer` failure at transfer progress checking

ggediminass:
`test_fileshare_transfer` is flaky at `assert transfer_progress_local.transfer_progress_valid`. Attempt to change logic, to address the ongoing issue.

### DIFF
--- a/test/qa/test_fileshare.py
+++ b/test/qa/test_fileshare.py
@@ -263,18 +263,34 @@ def test_fileshare_transfer(filesystem_entity: fileshare.FileSystemEntity, backg
     transfers_remote = ssh_client.exec_command("nordvpn fileshare list")
     assert "waiting for download" in fileshare.find_transfer_by_id(transfers_remote, peer_transfer_id)
 
-    transfer_progress_local = fileshare.TransferProgressValidationThread(local_transfer_id, fileshare.TransferState.UPLOADING, None)
-    transfer_progress_local.start()
-
-    transfer_progress_remote = fileshare.TransferProgressValidationThread(local_transfer_id, fileshare.TransferState.DOWNLOADING, ssh_client)
-    transfer_progress_remote.start()
-
     if path_flag:
         peer_filepath = "/tmp/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} --path {peer_filepath} {peer_transfer_id}")
     else:
         peer_filepath = "~/Downloads/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} {peer_transfer_id}")
+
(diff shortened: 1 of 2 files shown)


test/qa/test_fileshare.py
```diff
@@ -261,19 +261,35 @@ def test_fileshare_transfer(filesystem_entity: fileshare.FileSystemEntity, backg
...
     if path_flag:
         peer_filepath = "/tmp/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} --path {peer_filepath} {peer_transfer_id}")
     else:
         peer_filepath = "~/Downloads/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} {peer_transfer_id}")
 
+    for transfers_started in poll(
+        lambda: (
+            "uploading" in fileshare.find_transfer_by_id(
```
DoYouEvenCpp: Can't we reuse here `TransferState.UPLOADING` (and `DOWNLOADING` for the other leg here)?

test/qa/test_fileshare.py
```diff
@@ -261,19 +261,35 @@ def test_fileshare_transfer(filesystem_entity: fileshare.FileSystemEntity, backg
...
-    transfer_progress_remote.start()
-
     if path_flag:
         peer_filepath = "/tmp/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} --path {peer_filepath} {peer_transfer_id}")
     else:
         peer_filepath = "~/Downloads/"
         t_progress_interactive = ssh_client.exec_command(f"nordvpn fileshare accept {background_accept} {peer_transfer_id}")
 
+    for transfers_started in poll(
```
DoYouEvenCpp, line 273: Is there a possibility this will never break?
Can this extend test execution significantly?

ggediminass: I doubt that it can enter state where it will never break. We just check transfer lists here, which is a pretty quick and not time consuming call. Furthermore, when it comes to `poll` function itself, if not specified otherwise, the loop is repeated for three times only.

test/qa/test_fileshare.py
```diff
@@ -261,19 +261,35 @@ def test_fileshare_transfer(filesystem_entity: fileshare.FileSystemEntity, backg
...
+        lambda: (
+            fileshare.TransferState.UPLOADING in fileshare.find_transfer_by_id(
+                sh.nordvpn.fileshare.list(_tty_out=False),
```
mariusSincovici: [small] In test `_tty_out=False` should not be needed, unless the env variable is not working fine, or there is a bug in the code. 🤔

PR #1268, opened 2025-12-11
LVPN-9448: GUI release build logs only info level

devzbysiu:
Current release builds of GUI log only `INFO` level. This makes it harder to debug issues found by users. This PR addresses this.

Changes:
- switch missing translation logs to `TRACE` instead of `DEBUG`
- allow controlling log level via `NORDVPN_GUI_LOG_LEVEL` environment variable
- custom printer to prefix logs with log level - not available OOTB (see https://github.com/SourceHorizon/logger/issues/5)
- tests
- docs

### DIFF
--- a/gui/lib/logger.dart
+++ b/gui/lib/logger.dart
@@ -33,7 +33,7 @@ Future<void> setupLogger() async {
         : fileOutput;
   }
 
-  final logLevel = giveLogLevel();
+  final logLevel = determineLogLevel();
   logger = Logger(
     output: loggerOutput,
     filter: ProductionFilter(),
@@ -86,7 +86,7 @@ Future<void> _trimLogFileIfNeeded(
   }
 }
 
-Level giveLogLevel() {
+Level determineLogLevel() {
   if (kDebugMode) return Level.all;
   final env = Platform.environment;
   final logLevelStr = env["NORDVPN_GUI_LOG_LEVEL"];


gui/README.md
DoYouEvenCpp, line 1: Maybe we should point from the main README (https://github.com/NordSecurity/nordvpn-linux/blob/main/README.md) to this one?

gui/lib/logger.dart
```diff
@@ -80,3 +85,40 @@ Future<void> _trimLogFileIfNeeded(
     }
   }
 }
+
+Level giveLogLevel() {
```
DoYouEvenCpp: nit:
```suggestion
Level getLogLevel() {
```

gui/README.md
```diff
@@ -28,6 +28,38 @@ You can find everything related to building, testing and environment setup in [B
...
+NORDVPN_GUI_LOG_LEVEL=<level> nordvpn-gui
+```
+
+- for snap package:
+
+```bash
+NORDVPN_GUI_LOG_LEVEL=<level> nordvpn.nordvpn-gui
+```
+
+Supported levels:
```
DoYouEvenCpp: Can we also point out those values for supported levels are case insensitive?
Or do you think this is simply too detailed?

gui/README.md
devzbysiu: There is whole set of documentation changes that will happen (migration debt). We have a task for that and we'll see even if we need separate `README.md` for GUI or just another section in main readme and build.md etc (other clients don't have any readme and GUI should be no different), so it will be change anyway

gui/lib/logger.dart
devzbysiu: Didn't want to go with just `get...` because it's not just a getter :smile:

devzbysiu: Maybe `determineLogLevel()` or `computeLogLevel()`?

DoYouEvenCpp: Yeah, true.
Then maybe `evaulateLogLevel()`?  (anyway, this is just a nit 😄)

gui/README.md
devzbysiu: Yeah, I can mention this, not sure if this gives any value to users, but it won't hurt

DoYouEvenCpp: Yeah, this is also my feeling here. Not sure, if this gives any _significant_ value. Anyway, it's up to you to decide on that one.

devzbysiu: Done

gui/lib/logger.dart
devzbysiu: Done

dfetti (approved):

PR #1274, opened 2025-12-16
LVPN-9079: Use daemon as single source of truth for snap interfaces

dfetti:
This PR removes the snap interfaces detection from the GUI and relies on the interceptor error provided by the daemon to report the missing permissions to the user.

### DIFF
--- a/gui/lib/daemon/error_screen.dart
+++ b/gui/lib/daemon/error_screen.dart
@@ -129,7 +129,7 @@ ErrorData _dataForApplicationError(WidgetRef ref, ApplicationError error) {
 }
 
 ErrorData _dataForSnapInterfaces(WidgetRef ref, ApplicationError error) {
-  final missingPermissions = _extractMissingConnections(error.originalError);
+  final missingPermissions = extractMissingConnections(error.originalError);
 
   if (missingPermissions.isEmpty) {
     logger.w("_extractMissingConnections returned an empty list");
@@ -206,7 +206,7 @@ CopyField _buildCopyFieldForSocketNotFound() {
   );
 }
 
-List<String> _extractMissingConnections(Object? error) {
+List<String> extractMissingConnections(Object? error) {
   if (error is! GrpcError) return const [];
 
   if (error.details != null) {
--- a/gui/lib/data/providers/grpc_connection_controller.dart
+++ b/gui/lib/data/providers/grpc_connection_controller.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:grpc/grpc.dart';
+import 'package:nordvpn/daemon/error_screen.dart';
 import 'package:nordvpn/data/models/application_error.dart';
 import 'package:nordvpn/grpc/grpc_service.dart';
 import 'package:nordvpn/grpc/error_handling_interceptor.dart';
@@ -75,10 +76,21 @@ class GrpcConnectionController extends _$GrpcConnectionController {
           error is ApplicationError,
           "$error must be of type ApplicationError",
         );
-        if (state is AsyncError<bool> &&
-            error is ApplicationError &&
-            state.error is ApplicationError &&
-            error.code == (state.error as ApplicationError).code) {
+        final appError = error as ApplicationError;
+        if (state is! AsyncError<bool>) break;
+        if (state.error is! ApplicationError) break;
+        final stateAppError = state.error as ApplicationError;
+        if (appError.code != stateAppError.code) break;
+        if (appError.code != AppStatusCode.snapInterfaces) {
+          return;
+        }
+        final oldSet = extractMissingConnections(
+          stateAppError.originalError,
+        ).toSet();
+        final newSet = extractMissingConnections(
+          appError.originalError,
+        ).toSet();
+        if (oldSet.length == newSet.length && oldSet.containsAll(newSet)) {
           return;
         }
         break;
--- a/gui/lib/data/providers/grpc_connection_controller.g.dart
+++ b/gui/lib/data/providers/grpc_connection_controller.g.dart
@@ -7,7 +7,7 @@ part of 'grpc_connection_controller.dart';
 // **************************************************************************
 
 String _$grpcConnectionControllerHash() =>
-    r'f437e3ec982b4060afb9d2e8779d593714cf76fe';
+    r'aa3418eef300c541dc1194e95f513993c1d5600f';
 
 /// See also [GrpcConnectionController].
 @ProviderFor(GrpcConnectionController)
--- a/gui/lib/vpn/servers_list_card.dart
+++ b/gui/lib/vpn/servers_list_card.dart
@@ -42,6 +42,7 @@ final class ServersListCard extends StatefulWidget {
   final ImagesManager imagesManager;
   final ServerListItemFactory itemFactory;
   final Future<void> Function(ConnectArguments) onSelected;
+
   /// Callback for when a recent connection is selected.
   /// If null, falls back to [onSelected].
   final Future<void> Function(ConnectArguments)? onRecentSelected;
--- a/gui/test/data/models/connect_arguments_test.dart
+++ b/gui/test/data/models/connect_arguments_test.dart
@@ -9,11 +9,16 @@ void main() {
   group('toUIEventItemValue', () {
     test('returns ITEM_VALUE_UNSPECIFIED for empty arguments', () {
       final args = ConnectArguments();
-      expect(args.toUIEventItemValue(), UIEvent_ItemValue.ITEM_VALUE_UNSPECIFIED);
+      expect(
+        args.toUIEventItemValue(),
+        UIEvent_ItemValue.ITEM_VALUE_UNSPECIFIED,
+      );
     });
 
     test('returns COUNTRY when only country is specified', () {
-      final args = ConnectArguments(country: Country(code: 'US', name: 'United States'));
+      final args = ConnectArguments(
+        country: Country(code: 'US', name: 'United States'),
+      );
       expect(args.toUIEventItemValue(), UIEvent_ItemValue.COUNTRY);
     });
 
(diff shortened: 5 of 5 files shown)


gui/lib/router/router.dart
```diff
@@ -124,7 +114,7 @@ final class RedirectState extends ChangeNotifier {
   // Otherwise don't do redirect.
   AppRoute? route(Uri uri) {
     // Global redirects
-    if (missingSnapPermissions) return AppRoute.missingSnapPermissions;
+    //if (missingSnapPermissions) return AppRoute.missingSnapPermissions;
```
thegravi: Leftovers?

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -188,6 +189,11 @@ class GrpcConnectionController extends _$GrpcConnectionController {
         case StatusCode.unavailable:
           return _mapErrorMessageToStatus(error.message);
 
+        case StatusCode.permissionDenied:
+          if ((error.message?.contains("Snap permissions") ?? false)) {
```
thegravi: [small] This relies on fragile string matching. If the daemon error message changes, the GUI won't detect snap permission errors.

gui/lib/daemon/error_screen.dart
```diff
@@ -174,3 +191,21 @@ CopyField _buildCopyFieldForSocketNotFound() {
...
 }
+
+List<String> _extractMissingConnections(Object? error) {
+  if (error is! GrpcError) return const [];
+
+  if (error.details != null) {
+    for (var detail in error.details!) {
+      if (detail is Any && detail.typeUrl.endsWith('ErrMissingConnections')) {
+        try {
+          return ErrMissingConnections.fromBuffer(detail.value).missingConnections;
```
thegravi: [small] Can `detail.value` be null?

dfetti: If detail is Any, the value will be a List<int> which should either contain data or be empty, but not null as far as I know.

gui/lib/data/providers/grpc_connection_controller.dart
dfetti: I updated it to match just "snap" on the lowercase converted message.
As far as I checked there is no other part of the app that uses StatusCode.permissionDenied as a grpc error so technically should be enough just the code check, but I added the string matching just to be extra sure is a snap error.

gui/lib/daemon/error_screen.dart
```diff
@@ -26,6 +28,17 @@ final class ErrorScreen extends ConsumerWidget {
 
     if (connectionProvider case AsyncError(:final error)) {
       logger.i("connection error $error");
+
+      // Check for snap permissions error
+      if (error is ApplicationError && error.code == AppStatusCode.snapInterfaces) {
```
mariusSincovici: When I'm saving, this gets formatted (rps format?) 🤔

gui/lib/daemon/error_screen.dart
```diff
@@ -174,3 +191,21 @@ CopyField _buildCopyFieldForSocketNotFound() {
...
+List<String> _extractMissingConnections(Object? error) {
+  if (error is! GrpcError) return const [];
+
+  if (error.details != null) {
+    for (var detail in error.details!) {
+      if (detail is Any && detail.typeUrl.endsWith('ErrMissingConnections')) {
+        try {
+          return ErrMissingConnections.fromBuffer(detail.value).missingConnections;
+        } catch (_) {
+          return const [];
```
mariusSincovici, line 221: Could you log the exception?

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -188,6 +189,12 @@ class GrpcConnectionController extends _$GrpcConnectionController {
         case StatusCode.unavailable:
           return _mapErrorMessageToStatus(error.message);
 
+        case StatusCode.permissionDenied:
+          final message = error.message?.toLowerCase() ?? "";
+          if ((message.contains("snap"))) {
```
mariusSincovici: Are all the `()` needed?

gui/lib/snap/snap_screen.dart
```diff
@@ -83,4 +86,8 @@ final class SnapScreen extends ConsumerWidget {
         "";
     return CopyItem(command: commands);
   }
+
+  static bool isSnapContext() {
```
mariusSincovici: Would it be possible to have a separate file with snap helper functions? Maybe more will come in the future.

gui/test/router/router_test.dart
mariusSincovici, line 1: Since the mocks were not updated, how can this be tested? 🤔

gui/lib/daemon/error_screen.dart
```diff
@@ -26,6 +29,20 @@ final class ErrorScreen extends ConsumerWidget {
 
     if (connectionProvider case AsyncError(:final error)) {
       logger.i("connection error $error");
+
+      // Check for snap permissions error
```
devzbysiu: Can't we do snap error the same way as other errors? 

So we would have `return _displayError(_dataForSnapPermissionsError(error));` in `build` and then `_dataForSnapPermissionsError` will figure out missing permissions and return `ErrorData` which is used by `_displayError`.

gui/lib/snap/snap_helpers.dart
```diff
@@ -0,0 +1,7 @@
+import 'dart:io';
+
+final class SnapHelpers {
```
devzbysiu, line 3: [small] I would suggest to add private constructor if this is only util class

gui/lib/daemon/error_screen.dart
```diff
@@ -174,3 +197,24 @@ CopyField _buildCopyFieldForSocketNotFound() {
...
+
+  if (error.details != null) {
+    for (var detail in error.details!) {
+      if (detail is Any && detail.typeUrl.endsWith('ErrMissingConnections')) {
+        try {
+          return ErrMissingConnections.fromBuffer(
+            detail.value,
+          ).missingConnections;
+        } catch (_) {
+          logger.e("Failed to parse ErrMissingConnections");
```
devzbysiu: Any reason to not log the exception itself?

gui/integration_test/e2e_tests/snap_test.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../test/utils/finders.dart';
+import '../../test/utils/test_helpers.dart';
+
+void runSnapErrorScreenTests() async {
+  group("test snap error screen", () {
```
devzbysiu, line 28: I would suggest to also add assertions for checking that the missing permissions are displayed as this is the most important part of this screen. Now, if something is broken and permissions are not displayed, the test will still pass as long as it has a title

gui/test/utils/finders.dart
```diff
@@ -235,3 +235,7 @@ Finder serviceSubscriptionInfoFinder() {
 Finder accountInfoFinder() {
   return find.byKey(AccountWidgetKeys.accountInfo);
 }
+
+Finder snapErrorScreenText() {
+  return find.text(t.ui.snapScreenTitle);
```
devzbysiu: I think we should go with `find.ByKey(...)` for all new stuff

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -153,6 +153,7 @@ class GrpcConnectionController extends _$GrpcConnectionController {
   }
 
   void _onGrpcInterceptorError(ErrorGrpc error) {
+    logger.e("Intercepted an error $error");
```
devzbysiu: [small] suggestion: maybe this should be `logger.w(...)` because we are handling some errors and then `logger.e(...)` when error is unhandled?

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -109,7 +109,7 @@ class GrpcConnectionController extends _$GrpcConnectionController {
     state = const AsyncData(true);
   }
 
-  Future<void> _pingDaemon() async {
+  Future<void> pingDaemon() async {
```
devzbysiu: I would suggest to leave this private and have separate `retry()` method which under the hood calls `_pingDaemon` to be consistent with other retry mechanisms we have in app. 

Also if you move snap screen to be handled similarly as other errors (this [comment](https://github.com/NordSecurity/nordvpn-linux/pull/1274/changes#r2752775303)) , then you can use `retryCallback` thing of `ErrorData`

gui/lib/router/routes.dart
```diff
@@ -106,7 +105,6 @@ List<RouteBase> configureRoutes() {
     _route(AppRoute.errorScreen, const ErrorScreen()),
     _route(AppRoute.login, const LoginScreen()),
     _route(AppRoute.consentScreen, const ConsentScreen()),
-    _route(AppRoute.missingSnapPermissions, const SnapScreen()),
```
devzbysiu, line 109: If this is no longer used, can it be removed also from enum in `routes.dart`?

gui/integration_test/e2e_tests/snap_test.dart
dfetti: Of course. The test was so basic to not waste time on it until the approach was validated. I'll improve it.

bartoszWojciechO (approved):

gui/lib/data/providers/grpc_connection_controller.dart
devzbysiu, line 74: [small] You can simplify this with preconditions check and early exit, also set comparison can be simplified, something like that:

```suggestion
      case AsyncError(:final error):
        assert(
          error is ApplicationError,
          "$error must be of type ApplicationError",
        );

        final appError = error as ApplicationError;

        if (state is! AsyncError<bool>) break;
        if (state.error is! ApplicationError) break;
        final stateAppError = state.error as ApplicationError;
        if (appError.code != stateAppError.code) break;

        if (appError.code != AppStatusCode.snapInterfaces) {
          return;
        }

        final oldSet = extractMissingConnections(stateAppError.originalError).toSet();
        final newSet = extractMissingConnections(appError.originalError).toSet();
        if (oldSet.length == newSet.length && oldSet.containsAll(newSet)) {
          return;
        }

        break;
```

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -75,7 +80,18 @@ class GrpcConnectionController extends _$GrpcConnectionController {
             error is ApplicationError &&
             state.error is ApplicationError &&
             error.code == (state.error as ApplicationError).code) {
-          return;
+          if (error.code == AppStatusCode.snapInterfaces) {
+            final old_list = extractMissingConnections(
```
devzbysiu: Wong naming convention for dart 😄 , too much rust? hehe

dfetti: Auch :)

PR #1284, opened 2025-12-18
LVPN-9787: Remove need for two tokens while downloading deps

edvyno1:
Reverts changes that were done to stay compatible with half migrated libraries


ci/check_dependencies.sh
mariusSincovici: Maybe remove this if is not needed?

mariusSincovici: Actually is line 20 😄

thegravi (approved):

PR #1293, opened 2026-01-07
LVPN-9442: resolv.conf based DNS management service detection

bartoszWojciechO:
resolv.conf DNS management service detection was addedd in order to ensure that the daemon will use the configuration method supported by the OS.

If a comment in resolv.conf matches a service, that service will be used to configure the DNS.

If no comment is recognized, the daemon will read a symlink target. If the symlink target matches some service, that service will be used to configure the DNS.

If resolv.conf is not a symlink or symlink target is not recognized, daemon will attempt to set the DNS using the best available method, based on a priority list: 1, systemd-resolved 2, direct changes in resolv.conf.

### DIFF
--- a/daemon/dns/dns.go
+++ b/daemon/dns/dns.go
@@ -28,11 +28,11 @@ const (
 type dnsManagementService int
 
 const (
-	systemdResolved dnsManagementService = iota + 1
-	unknown
+	unknown dnsManagementService = iota
+	systemdResolved
 )
 
-var ErrDNSNotSet = fmt.Errorf("DNS unsetter not set")
+var ErrDNSNotSet = errors.New("DNS unsetter not set")
 
 // statingFilesystemHandle extends FilesystemHandle with wrappers for os.Stat and os.SameFile
 type statingFilesystemHandle interface {
--- a/daemon/dns/dns_test.go
+++ b/daemon/dns/dns_test.go
@@ -259,12 +259,11 @@ options edns0 trust-ad
 search home`)
 
 	tests := []struct {
-		name                      string
-		resolvconfFileContents    []byte
-		resolvconfLinkDestination string
-		setBySystemdResolved      bool
-		setByResolvconf           bool
-		resolvConfIsASymlink      bool
+		name                   string
+		resolvconfFileContents []byte
+		setBySystemdResolved   bool
+		setByResolvconf        bool
+		resolvConfIsASymlink   bool
 		// resolvConfStatErr is returned when running Stat for /etc/resolv.conf
 		resolvConfStatErr error
 		// systemdStubStatErr is returned when running Stat for the systemd-resolved stub of /etc/resolv.conf
@@ -311,7 +310,8 @@ search home`)
 			name:                   "resolv.conf manager is unknown and running stat on resolv.conf fails, systemd-resolved is not available, resolv.conf is used to set DNS",
 			resolvconfFileContents: unknownManager,
 			resolvConfStatErr:      fmt.Errorf("failed to stat"),
-			setBySystemdResolved:   true,
+			systemdResolvedSetErr:  fmt.Errorf("failed to set"),
+			setByResolvconf:        true,
 		},
 		{
 			name:                   "resolv.conf manager is unknown and running stat on systemd stub fails, systemd-resolved is available, systemd-resolved is used to set DNS",
@@ -323,7 +323,8 @@ search home`)
 			name:                   "resolv.conf manager is unknown and running stat on systemd stub fails, systemd-resolved is not available, resolv.conf is used to set DNS",
 			resolvconfFileContents: unknownManager,
 			systemdStubStatErr:     fmt.Errorf("failed to stat"),
-			setBySystemdResolved:   true,
+			systemdResolvedSetErr:  fmt.Errorf("failed to set"),
+			setByResolvconf:        true,
 		},
 		{
 			name:                   "manager is not recognized based on resolv.conf contents but the file links to systemd-resolved is used to set DNS",
@@ -375,7 +376,8 @@ search home`)
 
 			fs := newMockStatingFilesystemHandle(t)
 			fs.ReadErr = test.readErr
-			// fs.statErr = test.isSymlinkToTargetErr
+			fs.statErrors[resolvconfFilePath] = test.resolvConfStatErr
+			fs.statErrors[systemdResolvedLinkTarget] = test.systemdStubStatErr
 			fs.isSameFile = test.resolvConfIsASymlink
 			fs.AddFile(resolvconfFilePath, test.resolvconfFileContents)
 


daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
-type DefaultSetter struct {
+// setBasedOnComment sets DNS using the setter matching the comment specified in resolv.conf
+func (d *DNSServiceSetter) setBasedOnComment(iface string, nameservers []string) error {
+	resolvConfFileContents, err := d.filesystemHandle.ReadFile(resolvconfFilePath)
+	if err != nil {
+		return fmt.Errorf("reading resolv.conf file: %w", err)
+	}
+
+	resolvConfFileContentsStr := string(resolvConfFileContents)
+	if strings.Contains(string(resolvConfFileContentsStr), resolvdComment) {
```
thegravi: No need to cast

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+		log.Println(internal.InfoPrefix, dnsPrefix,
+			"configuring DNS with systemd-resolved, infered from resolv.conf comment")
```
thegravi: [small] typo `inferred`

daemon/dns/dns.go
```diff
@@ -7,12 +7,40 @@ import (
...
 	"github.com/NordSecurity/nordvpn-linux/events"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
+const (
+	dnsPrefix = "[DNS]"
+	// resolvconfFilePath defines path to resolv.conf file for DNS
+	resolvconfFilePath = "/etc/resolv.conf"
+	// resolvdComment is the comment used to mark resolv.conf managed by systemd-resolved.
+	resolvdComment = "# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8)."
```
thegravi: If this "format" universal to all distros? What if file contains multiple comments?

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+}
+
+// Set sets the DNS using the most appropriate method, it attempts to:
+//  1. Infer which method to use from the comment in /etc/resolv.conf
+//  2. If the above fails, infer which method to use based on /etc/resovl.conf link target
+//  3. If the above fails, it attempts to use best available method(see setUsingBestAvailable)
+//  4. If all of the above fail, it returns an error
+func (d *DNSServiceSetter) Set(iface string, nameservers []string) error {
+	if err := d.setBasedOnComment(iface, nameservers); err != nil {
+		log.Println(internal.ErrorPrefix, dnsPrefix, "failed to configure DNS based on the resolv.conf comment:", err)
```
devzbysiu: Maybe `internal.WarningPrefix` would be more appropriate here and below?

daemon/dns/dns.go
```diff
@@ -7,12 +7,40 @@ import (
...
+	"github.com/NordSecurity/nordvpn-linux/config"
 	"github.com/NordSecurity/nordvpn-linux/events"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
+const (
+	dnsPrefix = "[DNS]"
+	// resolvconfFilePath defines path to resolv.conf file for DNS
+	resolvconfFilePath = "/etc/resolv.conf"
+	// resolvdComment is the comment used to mark resolv.conf managed by systemd-resolved.
```
mariusSincovici: [small] isn't resolv`e`d?

Suggestion: wouldn't make sense instead of resolved to use systemd? Because also in tests is a bit confusing to read resolvd vs resolvConf 😄 ?

daemon/dns/dns.go
bartoszWojciechO: It's same comment for the supported distros. For non-standard cases we have additional heuristic of following the symlink and a failsafe of using the best available method.

In case of multiple comments, `systemd-resolved` has the priority. So, for example,  if we have encounter some unexpected situation where `resolv.conf` contains a comment for `resolved` and `NetworkManager`, we'd end up with `resolved`. And if the config fails, we'd fall back to the symlink detection method.

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+	d.unsetter = setter
+	return nil
+}
 
-4. In case the resolvconf command line utility fails, /etc/resolv.conf is
-backed up and modified directly by NordVPN.
-*/
-type DefaultSetter struct {
+// setBasedOnComment sets DNS using the setter matching the comment specified in resolv.conf
+func (d *DNSServiceSetter) setBasedOnComment(iface string, nameservers []string) error {
```
mariusSincovici: [optional] Wouldn't make sense to split this in 2 functions, for testing?
* one to read and return the type
* second to actually do the set


Same for `setBasedOnResolvConfLinkTarget`.

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+	log.Println(internal.ErrorPrefix, dnsPrefix,
+		"failed to detect DNS management service based on resolv.conf, attempting to use the best avaialable method")
+	if err := d.setUsingBestAvailable(iface, nameservers); err != nil {
+		return fmt.Errorf("failed to set DNS: %w", err)
+	}
+
+	return nil
+}
+
+// Uset unsets the DNS using the method that was used to set it
```
devzbysiu: Maybe this should specify that it unsets using the "family" of methods that were used to set? 🤔 

Or maybe go even further if it would make sense - to record which specific method from selected family was used and do unset by using only this method?

Right now, we have it like that:
1. We pick family of methods to set DNS
2. Then `Set` for this family is iterating over its methods and whichever worked first is stopping iteration.
3. Then `Unset` for this family is also iterating over those methods to unset.

We had it like that before, but maybe it would be worth doing it in more precise way - so in step 2. we record which method from this family was used and then `Unset` would use just this method?

Now (and as we had before) if we have 3 methods as part of the family and `Set` for this family finishes on second method (so first method didn't work, but second worked), then `Unset` will also try the first method to unset, even when we already know it wasn't used

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+}
+
+// setUsingBestAvailable sets DNS using the first setter in the bellow priority list:
+//  1. systemd-resolvd DBUS
+//  2. resolvctl utility
+//  3. resolvconf utility
+//  4. direct write to resovl.conf
+func (d *DNSServiceSetter) setUsingBestAvailable(iface string, nameservers []string) error {
+	if err := d.set(d.systemdResolvedSetter, iface, nameservers); err != nil {
+		log.Println(internal.ErrorPrefix, dnsPrefix,
```
mariusSincovici: [very small] this is not really an error message, because it is still possible to use resolvConfSetter

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
 	publisher events.Publisher[string]
 	methods   []Method
 }
 
-func NewSetter(publisher events.Publisher[string]) *DefaultSetter {
-	ds := DefaultSetter{
+func NewSetter(publisher events.Publisher[string], methods ...Method) *DNSMethodSetter {
+	ds := DNSMethodSetter{
 		publisher: publisher,
 		methods:   []Method{},
```
mariusSincovici: why not `methods:   methods` instead of append?

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
 	Name() string
 }
 
-/*
-DefaultSetter handleds DNS in this order:
+// DNSServiceSetter detects how OS is managing the DNS configuration and tries to set it using the appropriate method.
+type DNSServiceSetter struct {
```
mariusSincovici, line 70: Does this need to be thread safe, or something else is ensuring this, specially for `unsetter`?

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+	// systemdResolvedSetter sets DNS using the most desired method:
+	// 	1. systemd-resolved DBUS
+	// 	2. resolvectl exec call
+	systemdResolvedSetter Setter
+	// resolvconfSetter sets DNS using the most desired method:
+	//	1. resolvconf exec call
+	//	2. direct write to /etc/resolv.conf
+	resolvconfSetter Setter
+	unsetter         Setter
```
mariusSincovici, line 79: Does this handle the case where user had resolvconf cmd only. Connects to VPN and then installs systemd. 
This will probably result in no-net, so VPN disconnect is needed. Will the disconnect break the resolv.conf file?

daemon/dns/dns.go
```diff
@@ -7,12 +7,40 @@ import (
...
 )
 
+const (
+	dnsPrefix = "[DNS]"
+	// resolvconfFilePath defines path to resolv.conf file for DNS
+	resolvconfFilePath = "/etc/resolv.conf"
+	// resolvdComment is the comment used to mark resolv.conf managed by systemd-resolved.
+	resolvdComment = "# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8)."
+
+	resolvedLinkTarget = "../run/systemd/resolve/stub-resolv.conf"
```
devzbysiu: Wouldn't be safer to normalize the `os.Readlink` result to absolute path and compare absolute paths? 🤔

daemon/dns/dns_test.go
```diff
@@ -159,3 +204,173 @@ func Test_CheckForEntry(t *testing.T) {
...
+			s := DNSServiceSetter{
+				systemdResolvedSetter: &resolvedSetter,
+				resolvconfSetter:      &resolvconfSetter,
+				filesystemHandle:      fs,
+			}
+
+			err := s.Set("eth0", []string{"1.1.1.1"})
+			assert.ErrorIs(t, err, test.expectedSetErr, "Expected set error was not returned.")
+
+			if test.setByResolvd {
```
mariusSincovici: [small] To unify the ifs?
```
assert.Equal(t, test.setByResolvd, resolvedSetter.isSet, "DNS was not configured by the expected setter(resolved).")
assert.Equal(t, test.setByResolvconf, resolvconfSetter.isSet, "DNS was configured by the unexpected setter(resolvconf)")
```

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+		}
+		log.Println(internal.InfoPrefix, dnsPrefix, "DNS configured with systemd-resolved")
+		return nil
+	}
+
+	return fmt.Errorf("management service not recognized from resolv.conf link destination")
+}
+
+// setUsingBestAvailable sets DNS using the first setter in the bellow priority list:
+//  1. systemd-resolvd DBUS
```
devzbysiu: [typo] `systemd-resolvd` vs `systemd-resolved` (also below)

daemon/dns/dns.go
```diff
@@ -26,42 +54,164 @@ type Method interface {
...
+//  2. resolvctl utility
+//  3. resolvconf utility
+//  4. direct write to resovl.conf
```
devzbysiu, line 154: [typo] `resovl.conf` vs `resolv.conf`, also below

daemon/dns/dns.go
devzbysiu: Also this is almost identical to `setBasedOnResolvConfLinkTarget`, only the detection mechanism and logs are different, maybe would make sense to unify this?

daemon/dns/dns.go
bartoszWojciechO: The outer layer(networker) is thread safe, so this doesn't need to be thread safe until we start to use the setter somewhere else.

daemon/dns/dns.go
bartoszWojciechO: `ResolvConfFile` checks if the file was modified by us before unsetting. But maybe it would be safer to just attempt an unset with every setter, same as it was done previously. :thinking:

daemon/dns/dns.go
```diff
@@ -7,12 +7,48 @@ import (
...
+)
+
+type dnsManagementService int
+
+const (
+	systemdResolved dnsManagementService = iota + 1
+	unknown
+)
+
+var ErrDNSNotSet = fmt.Errorf("DNS not set")
```
devzbysiu: Maybe this could be more specific - like "Unset called, but unsetter is not set" or something similar?

daemon/dns/dns.go
mariusSincovici: If it is checked then I think is ok.

>attempt an unset with every setter

Not fully sure if it would solve anything. Because if a new DNS resolver is installed into the system, that would probably replace resolv.conf => no-net.

daemon/dns/dns.go
```diff
@@ -7,12 +7,52 @@ import (
...
+	systemdResolvedComment = "# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8)."
+
+	systemdResolvedLinkTarget = "/run/systemd/resolve/stub-resolv.conf"
+)
+
+type dnsManagementService int
+
+const (
+	systemdResolved dnsManagementService = iota + 1
+	unknown
```
devzbysiu: Shouldn't this be `0` so the default is `unknown` or do we need a distinction between `unknown` and `not set`? 🤔

daemon/dns/dns.go
```diff
@@ -7,12 +7,52 @@ import (
...
+)
+
+var ErrDNSNotSet = fmt.Errorf("DNS unsetter not set")
```
devzbysiu: Just fun fact - this  and below also, can be `errors.New()`, because there is no formatting here, but on the other hand, in future go versions this won't make a difference - https://antonz.org/go-1-26/#fmt-errorf 😄

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+
+	unknownManager := []byte(`# This is managed by an unknown manager.
+nameserver 127.0.0.53
+options edns0 trust-ad
+search home`)
+
+	tests := []struct {
+		name                      string
+		resolvconfFileContents    []byte
+		resolvconfLinkDestination string
```
devzbysiu: Is this one used?

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+		setBySystemdResolved      bool
+		setByResolvconf           bool
+		resolvConfIsASymlink      bool
+		// resolvConfStatErr is returned when running Stat for /etc/resolv.conf
+		resolvConfStatErr error
```
devzbysiu, line 268: This is set up in test cases, but I don't see it being used in test body 🤔

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+		// systemdStubStatErr is returned when running Stat for the systemd-resolved stub of /etc/resolv.conf
+		systemdStubStatErr      error
```
devzbysiu, line 270: This also - set up in test case, but not used in test body 🤔

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+			resolvConfIsASymlink:   false,
+			setBySystemdResolved:   true,
+		},
+		{
+			name:                   "resolv.conf manager is unknown and running stat on resolv.conf fails, systemd-resolved is available, systemd-resolved is used to set DNS",
+			resolvconfFileContents: unknownManager,
+			resolvConfStatErr:      fmt.Errorf("failed to stat"),
+			setBySystemdResolved:   true,
+		},
+		{
```
devzbysiu, line 309: Isn't this the same test case as the one above? 🤔

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+			name:                   "resolv.conf manager is unknown and running stat on resolv.conf fails, systemd-resolved is not available, resolv.conf is used to set DNS",
```
devzbysiu, line 310: Also it says `systemd-resolved is not available` but the `setBySystemdResolved` is set to `true` 🤔

daemon/dns/dns_test.go
```diff
@@ -159,3 +216,192 @@ func Test_CheckForEntry(t *testing.T) {
...
+			setBySystemdResolved:   true,
+		},
+		{
+			name:                   "resolv.conf manager is unknown and running stat on systemd stub fails, systemd-resolved is available, systemd-resolved is used to set DNS",
+			resolvconfFileContents: unknownManager,
+			systemdStubStatErr:     fmt.Errorf("failed to stat"),
+			setBySystemdResolved:   true,
+		},
+		{
+			name:                   "resolv.conf manager is unknown and running stat on systemd stub fails, systemd-resolved is not available, resolv.conf is used to set DNS",
```
devzbysiu, line 323: This one also says `systemd-resolved is not available` with `setBySystemdResolved: true`

daemon/dns/dns.go
```diff
@@ -26,42 +66,171 @@ type Method interface {
...
+	}
+
+	if err := d.set(d.resolvconfSetter, iface, nameservers); err != nil {
+		return fmt.Errorf("failed to configure DNS with resolv.conf: %w", err)
+	}
+
+	log.Println(internal.InfoPrefix, dnsPrefix, "DNS configured with resolv.conf")
+
+	return nil
+}
```
DoYouEvenCpp, line 171: Do we care some of the method to handle DNS update might be repeated (eg. when any of the handlers returns error, due to any reason), hence not only some methods are unnecessarily re-executed but also some log pollution is created here (eg. there might be duplicated warning/logs messages).

bartoszWojciechO: I think this is an edge case, normally if some method is detected then that method should also be available. Accounting for that would increase complexity, which is not worth it imo.

PR #1294, opened 2026-01-07
LVPN-9733:Optimize pipeline test job execution 

bbrutall:
- Improved dynamic_parametrize.py to support different input parameters
 - Applied improved dynamic_parametrize decorator to the 'test_combinations' and 'test_allowlist_port'

### DIFF
--- a/test/qa/lib/dynamic_parametrize.py
+++ b/test/qa/lib/dynamic_parametrize.py
@@ -2,9 +2,49 @@ import random
 from functools import wraps
 
 import pytest
+from itertools import product
 
 
-def _generate_params_with_ids(randomized_source, ordered_source, id_pattern=None, always_pair=None):
+def _is_list_of_lists(obj):
+    """
+    Check if the object is a list/tuple of lists/tuples.
+
+    :param obj: Object to check.
+    :return: True if obj is a list/tuple of lists/tuples, else False.
+    """
+    return bool(isinstance(obj, list | tuple) and obj and all(isinstance(x, list | tuple) for x in obj))
+
+
+def _flatten_param_tuple(combo):
+    """
+    Flatten any nested tuples/lists inside a combo into a single tuple.
+
+    :param combo: An iterable of parameters.
+    :return: A flattened tuple of values.
+    """
+    flat = []
+    for item in combo:
+        if isinstance(item, tuple | list):
+            flat.extend(item)
+        else:
+            flat.append(item)
+    return tuple(flat)
+
+
+def as_tuple(obj):
+    """
+    Converts the input to a tuple.
+
+    If the input is already a tuple or a list, it is converted to a tuple (if not already).
+    If the input is any other type, it is returned as a single-element tuple containing the input.
+
+    :param obj: An iterable of parameters. The input object to be converted.
+    :return: A tuple representation of the input.
+    """
+    return tuple(obj) if isinstance(obj, list | tuple) else (obj,)
+
+
+def _generate_params_with_ids(argnames, randomized_source, ordered_source, generate_all=False, id_pattern=None, always_pair=None, sample_size=None):
     """
     Generate test parameters and their associated unique identifiers (IDs).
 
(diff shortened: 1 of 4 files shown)


test/qa/test_combinations.py
```diff
@@ -29,8 +33,17 @@ def connect_base_test(group: str = (), name: str = "", hostname: str = ""):
...
-@pytest.mark.parametrize(("target_tech", "target_proto", "target_obfuscated"), lib.TECHNOLOGIES)
-@pytest.mark.parametrize(("source_tech", "source_proto", "source_obfuscated"), lib.TECHNOLOGIES)
+@dynamic_parametrize(
+    [
+        "source_tech", "source_proto", "source_obfuscated",
+        "target_tech", "target_proto", "target_obfuscated",
+    ],
+    ordered_source=[lib.TECHNOLOGIES],
+    randomized_source=[lib.TECHNOLOGIES],
+    cartesian_mode=(os.getenv("CI_PIPELINE_SCHEDULE_DESCRIPTION") or "").strip().lower() == "nightly",
```
DoYouEvenCpp: Can we extract `cartesian_mode=(os.getenv("CI_PIPELINE_SCHEDULE_DESCRIPTION") or "").strip().lower() == "nightly",` into some common helper method and reuse it accordingly?

test/qa/lib/dynamic_parametrize.py
```diff
@@ -2,62 +2,114 @@
...
 
-def _generate_params_with_ids(randomized_source, ordered_source, id_pattern=None, always_pair=None):
+def _is_list_of_lists(obj):
+    """
+    Check if the object is a list/tuple of lists/tuples.
+
+    :param obj: Object to check.
+    :return: True if obj is a list/tuple of lists/tuples, else False.
+    """
+    return isinstance(obj, list | tuple) and obj and all(isinstance(x, list | tuple) for x in obj)
```
DoYouEvenCpp: This returns `[]` for an empty list. And `()` for an empty tuple. Is it desired?

If so, this should be reflected in the doc string

test/qa/lib/dynamic_parametrize.py
```diff
@@ -2,62 +2,114 @@
...
-        remaining_randomized_data = [item for item in randomized_source if item != always_pair]
-    else:
-        remaining_randomized_data = list(randomized_source)
-
-    # For each ordered value, get a random item from remaining_randomized_data
-    for ordered_value in ordered_source:
-        randomized_value = random.choice(remaining_randomized_data)
-        add_params_and_ids(randomized_value, ordered_value)
-
+            ids.append("".join(str(v) for v in flat_combo))
```
DoYouEvenCpp, line 104: Maybe we should add a `-` here to increase readability? Something like that

```suggestion
            ids.append("-".join(str(v) for v in flat_combo))
```

test/qa/test_combinations.py
bbrutall: It's a few lines of code in the code, I'm not sure we need that as a separate function because `itertools.product` provides this implementation.

test/qa/lib/dynamic_parametrize.py
bbrutall: Fixed.

test/qa/lib/dynamic_parametrize.py
bbrutall: It is handled by the particular `id_pattern` for every test

test/qa/lib/dynamic_parametrize.py
```diff
@@ -2,62 +2,114 @@
...
     This function creates a combination of parameters from two data sources:
     randomized data and ordered data. Additionally, test case IDs
     are constructed to provide human-readable descriptions for test output.
 
+    :param argnames: A list of argument names for the test function.
     :param randomized_source:
         A list of items that will be used as randomized data.
     :param ordered_source:
         A list of ordered items.
+    :param cartesian_mode: If True, generate the full cross product (cartesian product) of all sources.
```
mariusSincovici: Why not, `generate_all` or something similar? 
Now one needs to check the documentation what `cartesian_mode` is.

bbrutall: It makes sense. Renamed.

edvyno1 (approved):

PR #1296, opened 2026-01-08
LVPN-9756: Gather token-renew-date metric

thegravi:
Implements tracking of the token renewal date and sends it to the moose analytics system.

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -69,6 +69,7 @@ type Subscriber struct {
 	canSendAllEvents           atomic.Bool
 	mux                        sync.RWMutex
 	isInitialized              bool
+	configChangeHandlers       []configChangeHandler
 }
 
 func NewSubscriber(
@@ -91,6 +92,11 @@ func NewSubscriber(
 		clientAPI:     clientAPI,
 		httpClient:    httpClient,
 		isInitialized: false,
+	}
+
+	// Add more handlers here as needed
+	sub.configChangeHandlers = []configChangeHandler{
+		sub.handleTokenRenewDateChange,
 	}
 	return sub
 }
@@ -433,13 +439,8 @@ func (s *Subscriber) OnConfigChanged(e config.DataConfigChange) error {
 		return nil
 	}
 
-	// Add more handlers here as needed
-	handlers := []configChangeHandler{
-		s.handleTokenRenewDateChange,
-	}
-
 	var errs []error
-	for _, handler := range handlers {
+	for _, handler := range s.configChangeHandlers {
 		if err := handler(e.PreviousConfig, e.Config); err != nil {
 			errs = append(errs, err)
 		}
@@ -449,8 +450,6 @@ func (s *Subscriber) OnConfigChanged(e config.DataConfigChange) error {
 }
 
 // handleTokenRenewDateChange updates moose context when token renewal date changes.
-// Note: The data team confirmed that tokenRenewDate field should always have a value,
-// so we only set it and never call unset.
 func (s *Subscriber) handleTokenRenewDateChange(prev, curr *config.Config) error {
 	currentDate := getTokenRenewDate(curr)
 	previousDate := getTokenRenewDate(prev)


config/config.go
```diff
@@ -72,6 +73,17 @@ func (c Config) withAnalyticsConsent(value AnalyticsConsent) Config {
...
 
+// DeepCopy creates a deep copy of the Config struct using JSON serialization.
+func (c Config) DeepCopy() (Config, error) {
+	var copy Config
+	data, err := json.Marshal(c)
+	if err != nil {
+		return copy, err
+	}
+	err = json.Unmarshal(data, &copy)
+	return copy, err
```
DoYouEvenCpp: nit: move copy declaration point when config is correctly marshalled.
```suggestion
	data, err := json.Marshal(c)
	if err != nil {
		return nil, err
	}
	var copy Config
	err = json.Unmarshal(data, &copy)
	return copy, err
```

config/manager.go
```diff
@@ -120,13 +123,15 @@ func (f *FilesystemConfigManager) SaveWith(fn SaveFunc) error {
 	// deadlock when conifg change subscriber tries to read the config with the same manager when the change is
 	// published. The assumption here is that publisher is protected with it's own lock.
 	caller := getCaller()
+	var previousCfg *Config
 	var c Config
```
DoYouEvenCpp, line 127: Care to rename it to `currentConfig`, please?

config/manager.go
```diff
@@ -138,6 +143,12 @@ func (f *FilesystemConfigManager) SaveWith(fn SaveFunc) error {
 		return err
 	}
 
+	// Deep copy the config before applying changes to preserve the previous state
+	prevCopy, copyErr := c.DeepCopy()
+	if copyErr == nil {
```
DoYouEvenCpp: In case of an error, shall we et least emit a warning that storing previous config failed due to... well, an error here?

config/manager.go
```diff
@@ -187,6 +200,12 @@ func (f *FilesystemConfigManager) Reset(preserveLoginData bool, disableKillswitc
 		return fmt.Errorf("loading old config: %w", retErr)
 	}
 
+	// Deep copy the current config to preserve the previous state
+	prevCopy, copyErr := current.DeepCopy()
+	if copyErr == nil {
```
DoYouEvenCpp: Similarly here, emit an warning/debug that the storing a copy of config failed?

config/manager_test.go
```diff
@@ -163,3 +163,94 @@ func TestConfigDefaultValues(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			copied, err := tt.config.DeepCopy()
+			require.NoError(t, err)
+
+			assert.Equal(t, tt.config.AutoConnectData.ID, copied.AutoConnectData.ID)
+			assert.Equal(t, len(tt.config.TokensData), len(copied.TokensData))
+			for id, td := range tt.config.TokensData {
+				assert.Equal(t, td.Token, copied.TokensData[id].Token)
+				assert.Equal(t, td.TokenRenewDate, copied.TokensData[id].TokenRenewDate)
+			}
```
DoYouEvenCpp: Can this get reduced to following?
```suggestion
			assert.Equal(t, tt.config, copied)
```

config/config.go
thegravi: Return signature is `(Config, error)`, therefore I cannot return `nil` there. Instead `Config{}` must be passed, which is exactly what I'm doing right now.

config/manager.go
thegravi: Renaming to `currentConfig` would make it confusing, since it's used in `defer` block which is executed after the function returns, but 1 step before this the config is updated, and this `previousCfg` actually becomes "previous".

config/config.go
DoYouEvenCpp: Oh, indeed. Overlooked that. Then let's disregard that one :smile:

config/manager.go
DoYouEvenCpp: Ok, good point.

config/config.go
```diff
@@ -72,6 +73,17 @@ func (c Config) withAnalyticsConsent(value AnalyticsConsent) Config {
 	return c
 }
 
+// DeepCopy creates a deep copy of the Config struct using JSON serialization.
+func (c Config) DeepCopy() (Config, error) {
+	var copy Config
+	data, err := json.Marshal(c)
```
DoYouEvenCpp: Missed this bit with my first review, sorry!
Let me ask it now: conducting a deep copy with json's marshalling procedures will omit empty members, when `omitempty` is applied. Let's assume `TokensData` are empty, and after a copy that becomes a `nil`. 

This might be a hard question to answer 🙃 but are we sure the code is prepared a `nil` instead of an empty container?
Maybe it would be safer to run manually deep copy procedure, and keep empty fields well, empty instead of nils?

cmd/daemon/events_linux.go
```diff
@@ -14,40 +14,41 @@ import (
...
+func (*dummyAnalytics) NotifyDebuggerEvent(events.DebuggerEvent) error    { return nil }
+func (*dummyAnalytics) NotifyRequestAPI(events.DataRequestAPI) error      { return nil }
+func (*dummyAnalytics) NotifyUiItemsClick(events.UiItemsAction) error     { return nil }
+func (*dummyAnalytics) NotifyHeartBeat(time.Duration) error               { return nil }
+func (*dummyAnalytics) NotifyDeviceLocation(core.Insights) error          { return nil }
+func (*dummyAnalytics) NotifyLANDiscovery(bool) error                     { return nil }
+func (*dummyAnalytics) NotifyVirtualLocation(bool) error                  { return nil }
+func (*dummyAnalytics) NotifyPostquantumVpn(bool) error                   { return nil }
+func (*dummyAnalytics) NotifyAppStartTime(int64) error                    { return nil }
+func (*dummyAnalytics) NotifyConfigChanged(config.DataConfigChange) error { return nil }
```
DoYouEvenCpp: Some time ago we've been discussing naming convention for those callbacks. What about `OnConfigChanged` here, we dropped this idea? 😉

cmd/daemon/events_linux.go
```diff
@@ -45,6 +45,7 @@ func (*dummyAnalytics) NotifyLANDiscovery(bool) error                  { return
 func (*dummyAnalytics) NotifyVirtualLocation(bool) error               { return nil }
 func (*dummyAnalytics) NotifyPostquantumVpn(bool) error                { return nil }
 func (*dummyAnalytics) NotifyAppStartTime(int64) error                 { return nil }
+func (*dummyAnalytics) OnConfigChanged(config.DataConfigChange) error  { return nil }
```
DoYouEvenCpp: (nit, sorry!) if you could arrange the group of `On...` alphabetically, I'd really grateful! 😄

config/manager_test.go
```diff
@@ -163,3 +162,108 @@ func TestConfigDefaultValues(t *testing.T) {
...
+	assert.Equal(t, first.AutoConnect, second.AutoConnect)
+	assert.Equal(t, first.TokensData[123].Token, second.TokensData[123].Token)
+
+	// Modify first copy
+	first.AutoConnect = false
+	first.TokensData[123] = TokenData{Token: "modified-token"}
+
+	// Verify second copy is unchanged (independent)
+	assert.True(t, second.AutoConnect, "second copy should be unchanged")
+	assert.Equal(t, "test-token", second.TokensData[123].Token, "second copy TokensData should be unchanged")
```
DoYouEvenCpp: This could get limited to:

```suggestion
	assert.NotEqual(t, first, second)
```
(with added assertion message, of course) correct?

config/manager.go
```diff
@@ -217,36 +233,78 @@ func (f *FilesystemConfigManager) load(c *Config) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// loadTwoCopies loads config into two independent structs by unmarshaling the same JSON bytes twice.
+// Returns isNewInstallation=true if the config file didn't exist, and 'second' is not populated in
+// that case.
+func (f *FilesystemConfigManager) loadTwoCopies(first *Config, second *Config) (isNewInstallation bool, err error) {
```
mariusSincovici: Wouldn't be easier to maintain, if there will just one `load` and an optional second parameter. When that parameter is `nil` nothing is made, otherwise populate that too.
So later `Load` can pass `nil` second param. And `Save` can pass with `&prev`.

config/manager.go
```diff
@@ -217,36 +233,78 @@ func (f *FilesystemConfigManager) load(c *Config) error {
...
+	return false, nil
+}
+
+// readDecryptedConfigJSON reads and decrypts the config file from disk.
+// Returns (decryptedJSON, isNewInstallation, error).
+// If isNewInstallation is true, the caller should save default config.
+func (f *FilesystemConfigManager) readDecryptedConfigJSON() ([]byte, bool, error) {
+	// TODO(LVPN-9628): revisit how this mechanism works
+	if !f.fsHandle.FileExists(f.location) {
+		f.NewInstallation = true
```
mariusSincovici, line 238: [small] Can't this value be checked instead of passing an extra param?

config/manager.go
```diff
@@ -217,36 +233,78 @@ func (f *FilesystemConfigManager) load(c *Config) error {
...
+}
+
+// loadTwoCopies loads config into two independent structs by unmarshaling the same JSON bytes twice.
+// Returns isNewInstallation=true if the config file didn't exist, and 'second' is not populated in
+// that case.
+func (f *FilesystemConfigManager) loadTwoCopies(first *Config, second *Config) (isNewInstallation bool, err error) {
+	// Initialize first with defaults
+	*first = *newConfig(f.machineIDGetter)
+
+	decryptedJSON, isNewInstall, err := f.readDecryptedConfigJSON()
```
mariusSincovici: `err` needs to be checked before using other variables returned by the function

config/manager_test.go
```diff
@@ -75,7 +74,7 @@ func TestConfigIsBackwardsCompatible(t *testing.T) {
 	category.Set(t, category.File)
 
 	salt, ok := os.LookupEnv("SALT")
-	require.True(t, ok)
+	assert.True(t, ok)
```
mariusSincovici: Does it make sense for the tests to continue if SALT is not present?

events/moose/moose.go
```diff
@@ -416,6 +429,72 @@ func (s *Subscriber) NotifyMFA(data bool) error {
...
+
+	return errors.Join(errs...)
+}
+
+// handleTokenRenewDateChange updates moose context when token renewal date changes.
+func (s *Subscriber) handleTokenRenewDateChange(prev, curr *config.Config) error {
+	currentDate := getTokenRenewDate(curr)
+	previousDate := getTokenRenewDate(prev)
+
+	if currentDate == "" || currentDate == previousDate {
```
mariusSincovici, line 457: if `currentDate` is empty, shouldn't `unsetTokenRenewDate` be called?

events/moose/moose.go
```diff
@@ -416,6 +429,72 @@ func (s *Subscriber) NotifyMFA(data bool) error {
...
+		return nil
+	}
+
+	renewDate, err := time.Parse(internal.ServerDateFormat, currentDate)
```
mariusSincovici: [small] in what timezone is the date sent? user/server/GMT

events/moose/moose.go
```diff
@@ -416,6 +429,72 @@ func (s *Subscriber) NotifyMFA(data bool) error {
...
+	if err != nil {
+		return nil
```
mariusSincovici: Any reason not to pass the error?

session/access_token_session_store_test.go
```diff
@@ -682,3 +682,47 @@ func TestAccessTokenSessionStore_Renew_ForceRenewal(t *testing.T) {
...
+			RenewToken: "ab78bb36299d442fa0715fb53b5e3e59",
+			ExpiresAt:  futureTime.Format(internal.ServerDateFormat),
+			UpdatedAt:  updatedAt.Format(internal.ServerDateFormat),
+		}, nil
+	}
+
+	store := session.NewAccessTokenSessionStore(cfgManager, errorRegistry, renewAPICall)
+	err := store.Renew()
+
+	assert.NoError(t, err)
```
DoYouEvenCpp: [nit] can be slightly reduced:

```suggestion
	assert.NoError(t, store.Renew())
```

config/manager_test.go
```diff
@@ -163,3 +163,107 @@ func TestConfigDefaultValues(t *testing.T) {
...
+		configLocation,
+		vaultLocation,
+		"",
+		NewMachineID(os.ReadFile, os.Hostname),
+		StdFilesystemHandle{},
+		publisher,
+	)
+
+	// First save - creates initial config
+	err := fs.SaveWith(func(c Config) Config {
```
DoYouEvenCpp: [nit] no need to create those `err` variables. assert can be call directly on the called function.

thegravi: It's a force of habit when adding/removing logic. And clean up happens...sometimes oops

events/moose/moose.go
mariusSincovici: For future: now, it must not be called `unsetTokenRenewDate` because it can break things

events/moose/moose.go
```diff
@@ -416,6 +423,70 @@ func (s *Subscriber) NotifyMFA(data bool) error {
...
+// OnConfigChanged handles config change events and updates moose context accordingly.
+// It dispatches to specific handlers based on which config fields have changed.
+func (s *Subscriber) OnConfigChanged(e config.DataConfigChange) error {
+	if e.Config == nil {
+		return nil
+	}
+
+	// Add more handlers here as needed
+	handlers := []configChangeHandler{
+		s.handleTokenRenewDateChange,
```
devzbysiu: [small] seems like this is the only handler without plans for adding more, but if you really need to loop over one value, then this `handlers` var could be set on `Subscriber`, during creation to not allocate on every `OnConfigChanged`

PR #1297, opened 2026-01-09
LVPN-9414: Add RecommendationUuid into connect and disconnect events

dfetti:
RecommendationUuid is returned as a header "X-Recommendation-UUID" by the RecommendedServers API.
When this is available it must be set in the connect events (it has a special field in the event) and in the disconnect events (here it doesn't have a special field so it is added in the context).

### DIFF
--- a/daemon/access/access_disconnect.go
+++ b/daemon/access/access_disconnect.go
@@ -15,6 +15,7 @@ type DisconnectInput struct {
 	Networker                  networker.Networker
 	ConfigManager              config.Manager
 	PublishDisconnectEventFunc func(events.DataDisconnect)
+	RecommendationUUID         string
 }
 
 // Disconnect disconnects the user from the current VPN server. Returning boolean indicates
@@ -47,6 +48,7 @@ func Disconnect(input DisconnectInput) (bool, error) {
 			ThreatProtectionLite: cfg.AutoConnectData.ThreatProtectionLite,
 			Duration:             time.Since(startTime),
 			Error:                err,
+			RecommendationUUID:   input.RecommendationUUID,
 		})
 	}()
 
--- a/daemon/rpc_disconnect.go
+++ b/daemon/rpc_disconnect.go
@@ -24,10 +24,19 @@ func (r *RPC) Disconnect(_ *pb.Empty, srv pb.Daemon_DisconnectServer) error {
 
 // DoDisconnect is the non-gRPC function for Disconect to be used directly.
 func (r *RPC) DoDisconnect() (bool, error) {
+	var recommendationUUID string
+	// Not sure if it can be nil in the real scenarios
+	if r.connectionInfo != nil {
+		recommendationUUID = r.connectionInfo.Status().RecommendationUUID
+	} else {
+		log.Println(internal.WarningPrefix, "connection info is nil and it shouldn't be")
+	}
+
 	wasConnected, err := access.Disconnect(access.DisconnectInput{
 		Networker:                  r.netw,
 		ConfigManager:              r.cm,
 		PublishDisconnectEventFunc: r.events.Service.Disconnect.Publish,
+		RecommendationUUID:         recommendationUUID,
 	})
 
 	if wasConnected {
--- a/daemon/servers.go
+++ b/daemon/servers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
+	"net/http"
 	"regexp"
 	"strings"
 	"time"
@@ -13,49 +14,27 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/config"
 	"github.com/NordSecurity/nordvpn-linux/core"
 	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/google/uuid"
 	"golang.org/x/exp/slices"
 )
 
 var tag = regexp.MustCompile(`^[a-z]{2}[0-9]{2,4}$`)
 var ErrDedicatedIPServer = fmt.Errorf("selected dedicated IP servers group")
 
-// PickServer by the specified criteria.
-func PickServer(
-	api core.ServersAPI,
-	countries core.Countries,
-	servers core.Servers,
-	longitude float64,
-	latitude float64,
-	tech config.Technology,
-	protocol config.Protocol,
-	obfuscated bool,
-	tag string,
-	groupFlag string,
-	allowVirtualServer bool,
-) (core.Server, bool, error) {
-	result, remote, err := getServers(
-		api,
-		countries,
-		servers,
-		longitude,
-		latitude,
-		tech,
-		protocol,
-		obfuscated,
-		tag,
-		groupFlag,
-		1,
-		allowVirtualServer,
-	)
-	if err != nil {
-		return core.Server{}, remote, err
-	}
+const recommendationUUIDHeader string = "X-Recommendation-Uuid"
 
-	// #nosec G404 -- not used for cryptographic purposes
-	return result[rand.Intn(len(result))], remote, nil
+type recommendationUUID string
+
+const emptyUuid = ""
+
+type serverSelection struct {
+	server             *core.Server
+	recommendationUUID recommendationUUID
+	remote             bool
 }
 
-func getServers(
+// PickServer by the specified criteria.
+func PickServer(
 	api core.ServersAPI,
 	countries core.Countries,
 	servers core.Servers,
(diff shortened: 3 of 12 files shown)


daemon/servers.go
```diff
@@ -229,16 +232,39 @@ func getServersRemote(
...
+		return nil, "", fmt.Errorf("recommended: empty list")
+	}
+
+	recommendationUuid := extractRecommendationUuid(header)
+
+	return servers, recommendationUuid, nil
+}
+
+func extractRecommendationUuid(h http.Header) string {
+	raw := h.Get("X-Recommendation-Uuid")
```
thegravi: [nit] Yes, it's case insensitive, but maybe let's write acronyms in caps? `X-Recommendation-UUID`

daemon/servers.go
```diff
@@ -229,16 +232,39 @@ func getServersRemote(
...
+
+	u, err := uuid.Parse(raw)
+	if err != nil {
+		log.Println(internal.WarningPrefix, "Invalid X-Recommendation-Uuid header: ", err)
+		return ""
+	}
+
+	if u.Version() != 4 {
+		log.Println(internal.WarningPrefix, "Invalid X-Recommendation-Uuid header, wrong version")
+		return ""
```
thegravi: [small] WDYT if we return formatted error from `extractRecommendationUuid` function back to `getServersRemote` and then we combine the output with additional log message which would give more context to the warning e.g., `"Failed to get recommended remote servers:", err`?

events/moose/moose.go
```diff
@@ -611,6 +611,14 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
 		))
 	}
 
+	if data.RecommendationUuid != "" {
+		if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateRecommendationUuid(data.RecommendationUuid)); err != nil {
+			// We can ignore setting the recommendation Uuid
+			// Sending the disconnect event is much more important
+			log.Println(internal.ErrorPrefix, "Failed to set RecommendationUuid into the moose context ", err)
```
thegravi: Since it does not return here, we should call it a warning, don't you think?

events/moose/moose.go
```diff
@@ -611,6 +611,14 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
...
+		}
+	}
```
thegravi: Shouldn't this be moved to `NotifyConnect` function?

thegravi (commented):
Would it be possible to cover some parts with unit tests?

events/moose/moose.go
dfetti: No, in NotifyConnect this recommendationUuid is passed as a parameter to the function that generates the moose event. In the case of NotifyDisconnect that parameter is missing so we have to set it to the context before generating the moose event and to unset it after the event is generated.

daemon/servers.go
dfetti: I hope I understood correctly this recommendation

daemon/servers.go
dfetti: This is the exact form of the header, I would rather keep that instead of the acronyms convention. 
But as you said, the comparison is case insensitive so it doesn't really matter.

daemon/servers.go
```diff
@@ -229,16 +232,38 @@ func getServersRemote(
...
+
+	recommendationUuid, err := extractRecommendationUuid(header)
+	if err != nil {
+		log.Println(internal.WarningPrefix, "failed to extract recommendation uuid from the http header", err)
+	}
+	return servers, recommendationUuid, nil
+}
+
+func extractRecommendationUuid(h http.Header) (string, error) {
+	raw := h.Get("X-Recommendation-Uuid")
```
DoYouEvenCpp: I guess we can elevate the header name to a const here.

daemon/servers.go
```diff
@@ -480,25 +505,25 @@ func selectServer(r *RPC, insights *core.Insights, cfg config.Config, tag string
 		switch {
 		case errors.Is(err, core.ErrUnauthorized):
 			if err := r.cm.SaveWith(auth.Logout(cfg.AutoConnectData.ID, r.events.User.Logout)); err != nil {
-				return nil, false, err
+				return nil, "", false, err
```
DoYouEvenCpp: What about instead of returning an empty string in all of the erroneous cases simply return the passed `recommendationUuid`?  The further connection will get anyway ditched, hence the recommendation wont get propagated anyway.
I'm totally fine with returning here just an empty String, but on the other hand returning always the same value for recommendation makes it more coherent. Leaving this for you to decide 😉

daemon/servers_test.go
```diff
@@ -817,3 +818,94 @@ func TestGetServerParameters(t *testing.T) {
...
+			name:                    "valid uuid",
+			headerKey:               "X-Recommendation-Uuid",
+			headerValue:             "c0b4c990-3000-457f-8b81-6850b8cdb54e",
+			expectedError:           false,
+			expectedRecommendedUuid: "c0b4c990-3000-457f-8b81-6850b8cdb54e",
+		},
+		{
+			name:                    "missing uuid",
+			headerKey:               "NOT-Recommendation-Uuid",
+			headerValue:             "random-value",
```
DoYouEvenCpp, line 858: This actually is the same as "bad uuid". Don't we have to test an empty uuid as well here?

daemon/servers.go
dfetti: Oh, you're right. It would be nicer.

daemon/servers_test.go
dfetti: The key is different so the header we are looking for will be missing. This case where is missing and where is empy are identical. In both cases header.Get("X-Recommendation-Uuid") will return "".

daemon/servers.go
dfetti: Yeah, good point

daemon/servers.go
```diff
@@ -229,16 +232,38 @@ func getServersRemote(
...
 	if len(servers) == 0 {
-		return nil, fmt.Errorf("recommended: empty list")
+		return nil, "", fmt.Errorf("recommended: empty list")
+	}
+
+	recommendationUuid, err := extractRecommendationUuid(header)
+	if err != nil {
+		log.Println(internal.WarningPrefix, "failed to extract recommendation uuid from the http header", err)
+	}
+	return servers, recommendationUuid, nil
```
mariusSincovici: `recommendationUuid` is used when there is an error. Could you set it to "" in that case?

daemon/servers.go
```diff
@@ -480,25 +505,25 @@ func selectServer(r *RPC, insights *core.Insights, cfg config.Config, tag string
...
-			return nil, false, err
+			return nil, "", false, err
 
 		case errors.Is(err, ErrDedicatedIPServer):
 			dedicatedIPServer, err := selectDedicatedIPServer(r.ac, serversList)
 			if err != nil {
-				return nil, false, err
+				return nil, "", false, err
 			}
 			server = *dedicatedIPServer
```
mariusSincovici: When this is executed, then `recommendationUuid` is used for err != nil.
Maybe set it to `""` when entering into the block

events/moose/moose.go
```diff
@@ -635,6 +643,12 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
 		return err
 	}
 
+	if data.RecommendationUuid != "" {
```
mariusSincovici: Maybe this should be moved to defer, after it is set, to ensure is removed?

dfetti: That's a good idea. I'll do that

daemon/servers.go
dfetti: Decided to leave empty strings to have it more verbose

daemon/servers.go
dfetti: Good catch, fixed the dedicated ip case

daemon/access/access_disconnect.go
```diff
@@ -15,6 +15,7 @@ type DisconnectInput struct {
 	Networker                  networker.Networker
 	ConfigManager              config.Manager
 	PublishDisconnectEventFunc func(events.DataDisconnect)
+	RecommendationUuid         string
```
devzbysiu: [small] should be `RecommendationUUID` according to `go` styleguides - also in other places

daemon/rpc_disconnect.go
```diff
@@ -24,10 +24,17 @@ func (r *RPC) Disconnect(_ *pb.Empty, srv pb.Daemon_DisconnectServer) error {
 
 // DoDisconnect is the non-gRPC function for Disconect to be used directly.
 func (r *RPC) DoDisconnect() (bool, error) {
+	var recommendationUuid string
+	// Not sure if it can be nil in the real scenarios
```
devzbysiu, line 28: [small] It should not be, but you can log a warn maybe

daemon/servers.go
```diff
@@ -13,49 +14,27 @@ import (
...
+type RecommendationUUID string
+
+const EmptyUuid = ""
 
-	// #nosec G404 -- not used for cryptographic purposes
-	return result[rand.Intn(len(result))], remote, nil
+type ServerSelection struct {
+	server             *core.Server
+	recommendationUuid RecommendationUUID
+	remote             bool
```
devzbysiu: [small] those can be private

daemon/servers.go
```diff
@@ -128,30 +108,38 @@ func getServers(
...
-	if count == 1 && len(ret) > 0 {
+	if len(selectedServers) > 0 {
 		// #nosec G404 -- not used for cryptographic purposes
 		rng := rand.New(rand.NewSource(time.Now().UnixNano()))
-		ret = []core.Server{ret[rng.Int63n(int64(len(ret)))]}
+		selectedServer = &selectedServers[rng.Int63n(int64(len(selectedServers)))]
+	} else {
+		// We were not guarded against this case before
+		// So I assume it should not happen, but better be safe
+		return ServerSelection{}, internal.ErrServerIsUnavailable
```
devzbysiu: Maybe you can check if we are able to cover this with tests

daemon/servers.go
```diff
@@ -229,16 +213,38 @@ func getServersRemote(
...
 	}
 
 	if len(servers) == 0 {
-		return nil, fmt.Errorf("recommended: empty list")
+		return nil, emptyUuid, fmt.Errorf("recommended: empty list")
+	}
+
+	recommendationUUID, err := extractRecommendationUUID(header)
+	if err != nil {
+		log.Println(internal.WarningPrefix, "failed to extract recommendation uuid from the http header", err)
```
DoYouEvenCpp: [nit]
```suggestion
		log.Println(internal.WarningPrefix, "failed to extract recommendation UUID from the HTTP header", err)
		log.Println(internal.WarningPrefix, "Fall back to an empty recommendation UUID")
```

PR #1298, opened 2026-01-12
LVPN-9802: Delay TP servers download

mariusSincovici:
Fetch TP servers async without blocking the daemon startup.
In case of an error, servers are fetch using `ExponentialBackoff` until it is successful.

### DIFF
--- a/daemon/dns/nameservers_test.go
+++ b/daemon/dns/nameservers_test.go
@@ -3,6 +3,8 @@ package dns
 import (
 	"errors"
 	"reflect"
+	"slices"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -14,8 +16,11 @@ import (
 )
 
 // helper function to return the needed format for the fetcher used in NewNameServers
-func wrapServersList(servers []string) func() (*core.NameServers, error) {
+func wrapServersList(servers []string, wg *sync.WaitGroup) func() (*core.NameServers, error) {
 	return func() (*core.NameServers, error) {
+		if wg != nil {
+			wg.Done()
+		}
 		return &core.NameServers{
 			Servers: servers,
 		}, nil
@@ -40,16 +45,22 @@ func TestNameservers(t *testing.T) {
 		expected             []string
 	}{
 		{
-			name:                 "ipv4",
+			name:                 "default DNS servers, TP=false",
 			threatProtectionLite: false,
 			initial:              defaultTpServers,
 			expected:             defaultServers,
 		},
 		{
-			name:                 "ipv4 threat protection lite",
+			name:                 "fetch TP list and return it",
 			threatProtectionLite: true,
 			initial:              defaultTpServers,
 			expected:             defaultTpServers,
+		},
+		{
+			name:                 "fetched servers are returned for TP servers",
+			threatProtectionLite: true,
+			initial:              []string{"1.2.3.4"},
+			expected:             []string{"1.2.3.4"},
 		},
 		{
 			name:                 "empty initial list",
@@ -62,9 +73,28 @@ func TestNameservers(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			servers := NewNameServers()
-			go servers.FetchTPServers(wrapServersList(test.initial), func(attempt int) time.Duration { return time.Millisecond })
-			nameservers := servers.Get(test.threatProtectionLite)
-			assert.ElementsMatch(t, test.expected, nameservers)
+
+			// before fetching the servers from the API check the default values
+			assert.ElementsMatch(t, defaultTpServers, servers.Get(true))
+			assert.ElementsMatch(t, defaultServers, servers.Get(false))
+
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go servers.FetchTPServers(wrapServersList(test.initial, &wg), func(attempt int) time.Duration {
+				assert.True(t, len(test.initial) == 0, "this must be called only when test.initial is empty")
+				return time.Minute
+			})
+
+			wg.Wait()
+
+			// retry several times to fetch the servers, because at first attempt internal members might not be stored
+			for retry := 0; retry < 2; retry++ {
+				if slices.Contains(servers.Get(test.threatProtectionLite), test.expected[0]) {
+					break
+				}
+				time.Sleep(time.Millisecond * 2)
+			}
+			assert.ElementsMatch(t, test.expected, servers.Get(test.threatProtectionLite))
 		})
 	}
 }
@@ -95,10 +125,9 @@ func TestNameserversRandomness(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			servers := NewNameServers()
-			go servers.FetchTPServers(wrapServersList(test.initial), func(attempt int) time.Duration { return time.Millisecond })
 
-			// give time to fetch the data, before checking it
-			time.Sleep(time.Millisecond * 5)
+			// fetch in blocking mode
+			servers.FetchTPServers(wrapServersList(test.initial, nil), func(attempt int) time.Duration { return time.Minute })
 
 			nameservers1 := servers.Get(test.threatProtectionLite)
 			nameservers2 := servers.Get(test.threatProtectionLite)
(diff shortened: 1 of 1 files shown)


core/helper_test.go
```diff
@@ -1,151 +1,82 @@
...
 	res := m.Run()
+
 	for _, server := range servers {
-		server.Shutdown(context.Background())
+		server.Close()
 	}
 	os.Exit(res)
 }
 
-func localServerPath(port int) string {
```
mariusSincovici, line 71: All this was moved to `test/mock/http_test_server.go` to be reusable.

.github/workflows/core_ci.yml
```diff
@@ -67,8 +67,8 @@ jobs:
 
       - name: Install Protoc
         uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
```
mariusSincovici, line 70: Based on the [documentation](https://github.com/arduino/setup-protoc?tab=readme-ov-file#usage) this is how the token is passed.
This is to prevent [error](https://github.com/NordSecurity/nordvpn-linux/actions/runs/20933811327/job/60151214079?pr=1298) with rate limit and to use authenticated requests.

daemon/dns/nameservers.go
```diff
@@ -14,37 +20,89 @@ const (
...
 }
 
-func NewNameServers(servers []string) *NameServers {
-	return &NameServers{servers}
+func NewNameServers(fetcher ServersFetcher, timeoutFn CalculateRetryDelayForAttempt) *NameServers {
+	n := &NameServers{}
+	if fetcher != nil && timeoutFn != nil {
+		// start async to fetch the TP server names
+		var wg sync.WaitGroup
+		wg.Add(1)
```
bartoszWojciechO: I wonder if wait group is needed here. :thinking: If I understand correctly, we only care that a goroutine is running in the background and it might possibly fetch new TP servers at some point in the future?

daemon/dns/nameservers.go
```diff
@@ -14,37 +20,89 @@ const (
...
 	return net.LookupIP(host)
 }
 
+// fetches the TP servers until is successful. It uses exponential backoff between retries
+func (n *NameServers) fetchTPServers(fetcher ServersFetcher, timeoutFn CalculateRetryDelayForAttempt) {
+	if fetcher == nil {
+		return
+	}
+
+	for i := 0; ; i++ {
```
bartoszWojciechO: [small] `i` could use some meaningfull name(maybe `try` or `retry`).

daemon/dns/nameservers.go
```diff
@@ -14,37 +20,89 @@ const (
...
+			// copy to ensure pointer is not later modified from outside
+			log.Println(internal.InfoPrefix, "TP servers updated to", servers.Servers)
+			s := slices.Clone(servers.Servers)
+			n.tpServers.Store(&s)
+
+			return
+		}
+
+		tryAfterDuration := timeoutFn(i)
+		log.Printf("%s failed to fetch TP servers. retry(%d) servers after %v: %v\n", internal.ErrorPrefix, i, tryAfterDuration, err)
```
bartoszWojciechO: Would it be worth it to add some kind of limiter for this log?

daemon/dns/nameservers.go
```diff
@@ -14,37 +20,89 @@ const (
...
 type NameServers struct {
-	// List of TP servers fetched from cloud
-	servers []string
+	// Pointer to the List of TP servers fetched from cloud
+	tpServers atomic.Pointer[[]string]
 }
 
-func NewNameServers(servers []string) *NameServers {
-	return &NameServers{servers}
+func NewNameServers(fetcher ServersFetcher, timeoutFn CalculateRetryDelayForAttempt) *NameServers {
```
bartoszWojciechO: [small] This might just be me, but the functions starting with `New(...)` imply a simple constructor. Starting a new thread seems pretty heavy for that. Maybe it would be better to rename the function? Or only add a comment?

daemon/dns/nameservers.go
devzbysiu: (btw. small, just FYI - if we need `WaitGroup`, you can use 
```go
wg.Go(func() { ... } )
``` 

instead of 
```go
wg.Add(1)
go func() { defer wg.Done() ...) }()
``` 
since go 1.25)

mariusSincovici: Yes, I've forgot to add a comment why I've added it. 
It is mainly for tests. Otherwise tests require some sleep after calling NewNameServers to have time to start goroutine.

There is also the possibility to split the function in create and start it from outside?

daemon/dns/nameservers.go
mariusSincovici: Any reason why to limit?
The log should be printed, after 3 failed attempts, at a big interval distance.

daemon/dns/nameservers.go
mariusSincovici: I'll try to split this in 2. To have the start called from outside.

daemon/dns/nameservers.go
bartoszWojciechO: If we have a comment then for simplicity I think it would be ok to leave it with the wait group. There's a small possibility that it will block when OS doesn't have enough resources to start new goroutines, but the app will have much bigger problems in this case. :smile: 

If it's easy to split or isloate the async part out then it would be good to go with that, but if it's complicated then wait group is fine imo.

daemon/dns/nameservers.go
bartoszWojciechO: Ok, backoff has quite large intervals indeed. Limitter is not necesary. :+1:

daemon/dns/nameservers.go
mariusSincovici: I've made the split. Give a better overview that there are 2 separate operations.

mariusSincovici: Nice :+1: . But not needed anymore.

bartoszWojciechO (approved):

daemon/dns/nameservers_test.go
```diff
@@ -28,30 +42,27 @@ func TestNameservers(t *testing.T) {
...
 		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			servers := NewNameServers(test.initial)
+			servers := NewNameServers()
+			go servers.FetchTPServers(wrapServersList(test.initial), func(attempt int) time.Duration { return time.Millisecond })
 			nameservers := servers.Get(test.threatProtectionLite)
 			assert.ElementsMatch(t, test.expected, nameservers)
```
DoYouEvenCpp: I'm worried (maybe not necessarily) about synchronisation in this test suite.
We fire a goroutine, and then immediately try to fetch servers.
Under the hood there's some synchronisation mechanism (atomic, etc.) that probably makes it all work robustly.

But my proposal is to maybe introduce here a wait group (embeed it into `wrapServersList`) and only when the group is synced fetch servers, and assert its content.

What do you think about such an approach?
(For `TestNameserversRandomness` there's a 5 Milisecond sleep added to sync. all the machinery, which I believe might also be the  case to base this on a wait group, instead of a sleep)

daemon/dns/nameservers_test.go
```diff
@@ -102,3 +118,41 @@ func TestNameserversRandomness(t *testing.T) {
...
+			return nil, errors.New("fail to fetch")
+		},
+		func(attempt int) time.Duration {
+			retries.Add(-1)
+			assert.True(t, retries.Load() >= 0, "called too many times")
+			return time.Millisecond
+		},
+	)
+
+	time.Sleep(time.Millisecond * 5)
```
DoYouEvenCpp: And here as well, I find synchronisation based on sleeps prone to error (most probably this might reveal occasionally, but still). Should we switch to a wait group instead?

daemon/dns/nameservers_test.go
mariusSincovici: I've made some changes, to have the `wg`, but also some retries. Because wg itself is not sufficient.
When wg is set into the fetch function, there are still operations that are happening outside + setting the atomic value. 
Because of this I've also added 2 retries, this will give sufficient time to sync the atomic var inside.


For the rest of the tests, I've removed the async fetching because it wasn't relevant to the tests.

daemon/dns/nameservers_test.go
mariusSincovici: I've removed the async fetching because it wasn't relevant to the tests.
There is a comment on the other topic why using just wg would not be sufficient.

PR #1303, opened 2026-01-15
LVPN-9861: Do not monitor network, if variable is set

ggediminass:
Please review Jira task for more details.

### DIFF
--- a/test/qa/conftest.py
+++ b/test/qa/conftest.py
@@ -119,6 +119,10 @@ def setup_check_internet_connection():
 
 @pytest.fixture(scope="session", autouse=True)
 def start_system_monitoring():
+    if os.getenv("SKIP_SYSTEM_MONITORING"):
+        print("Skipping monitoring...")
+        yield
+
     print("Run start_system_monitoring")
 
     # control running threads execution


test/qa/lib/network.py
```diff
@@ -156,6 +157,9 @@ def is_internet_reachable(ip_address="1.1.1.1", port=443, retry=5) -> bool:
 
 def is_internet_reachable_outside_vpn(retry=5, ip_address="1.0.0.1") -> bool:
     """Returns True when remote host is reachable by its public IP outside VPN tunnel."""
+    if os.getenv("NO_MONITORING"):
+        return True
```
bbrutall: Why `True` if we skip it?

test/qa/conftest.py
```diff
@@ -119,31 +119,35 @@ def setup_check_internet_connection():
...
+        yield
 
-    # stop monitoring after execution
-    stop_event.set()
-    for thread in threads:
-        thread.join()
+        # stop monitoring after execution
+        stop_event.set()
+        for thread in threads:
+            thread.join()
```
bbrutall: It's better not to put this code into the `else` statement block. Just add this code at the beginning:
```
  if os.getenv("NO_MONITORING"):
      print("Skipping monitoring...")
      return
```
Also, propose to make the name of  `NO_MONITORING`  env var explicit, like `SKIP_SYSTEM_MONITORING`. Because for now it sounds like smth general.

test/qa/lib/network.py
ggediminass: Just in case, if it would be unexpectedly called from somewhere else

ggediminass: Conclusion made, that this is not needed.

DoYouEvenCpp (approved):

Issue #79, opened 2023-09-20
Closest and best server in terms of ping


mbfakourii:
I need to connect to the closest and best server in terms of ping, but the command 
```nordvpn connect```
works randomly and connects to a random country.

my nordvpn settings
```
Technology: OPENVPN
Protocol: UDP
Firewall: enabled
Firewall Mark: 0xe1f1
Routing: enabled
Analytics: enabled
Kill Switch: disabled
Threat Protection Lite: disabled
Obfuscate: disabled
Notify: disabled
Auto-connect: disabled
IPv6: disabled
Meshnet: disabled
DNS: disabled
```

EmilijusS:
Hello. My guess is that the API requests to find the recommended server are being blocked (you could probably see that in the logs using `sudo journaltctl -e -u nordvpnd`), so as a fallback you get assigned a random server from the whole server list. In this case, if you want a closer server you will have to provide a country or a specific server manually as an argument to `nordvpn connect`.

mbfakourii:
Alternate address for
```https://api.nordvpn.com/v1/servers/recommendations?limit=20```
Does not exist?

piotrjurkiewicz:
As @EmilijusS wrote, API access is needed to determine the closest servers to connect. If you want a closer server you will have to provide a country or a specific server manually as an argument to `nordvpn connect`.

In some countries specific domain access blocking is being done by changing/removing entries for these domains in ISP DNS servers. So you can try to change your DNS server address to a public one (for example Cloudflare's `1.1.1.1`). Maybe it will help and make the API server accessible for you.

PR #1305, opened 2026-01-20
LVPN-9444: introduce support for DNS setting with nmcli tool

DoYouEvenCpp:
This change introduces support to set DNS addresses on request using nmcli tool.

DNS handling based on NetworkManager is now the second option when it comes for the selection algorithm, following systemd and preceding direct resolve.conf manipulation.

Tested with UTs and manual verification.

### DIFF
--- a/daemon/dns/dns_resolvenmcli.go
+++ b/daemon/dns/dns_resolvenmcli.go
@@ -148,7 +148,10 @@ func (nmcli *NMCli) getConnectionFromPhysicalInterfaces() ([]string, error) {
 			strings.Contains(fields[1], bridgeType) ||
 			strings.Contains(fields[1], gsmType) ||
 			strings.Contains(fields[1], cdmaType) {
-			conns = append(conns, strings.TrimSpace(fields[0]))
+			// for tersed output (eg. the -t flag) nmcli provides escaped colons in the connection names
+			// thus it is essential to unescape them here
+			connName := strings.ReplaceAll(fields[0], "\\:", ":")
+			conns = append(conns, strings.TrimSpace(connName))
 		}
 	}
 	return conns, nil
--- a/daemon/dns/dns_resolvenmcli_test.go
+++ b/daemon/dns/dns_resolvenmcli_test.go
@@ -74,15 +74,15 @@ func Test_DNSResolveNmCli_PhysicalInterfaceDeduction(t *testing.T) {
 		},
 		{
 			name: "connections with semicolon in names",
-			cmdOutput: `Wired connection_:  2:802-3-ethernet
+			cmdOutput: `Wired connection_\:  2:802-3-ethernet
 docker0:bridge
 lo:loopback
 virbr0:tun
 br-f78c0ce0d3eb:tun
-mpqemubr0:bridge
+mpqemubr0\:\: \::bridge
 vnet3:tun`,
 			cmdError:      nil,
-			expectedConns: []string{"Wired connection_:  2", "docker0", "mpqemubr0"},
+			expectedConns: []string{"Wired connection_:  2", "docker0", "mpqemubr0:: :"},
 		},
 	}
 


daemon/dns/dns.go
mariusSincovici, line 60: Could we have this interface moved into `filesystem.go` and then the FilesystemHandle implements this?

mariusSincovici, line 71: Could we just have functions in `filesystem.go` that do this?
The idea is that all over the app are places that use the internal functions and the OS. If we would like/need to do something that affects all `os.Stat` commands in the app then each needs to be found. But if there is a single place, then it will be much easier.

daemon/dns/dns.go
```diff
@@ -129,7 +148,6 @@ func (d *DNSServiceSetter) getManagementService() dnsManagementService {
 	if err != nil {
```
mariusSincovici: Something is wrong at this logic. For error is printed the message, but later is printed that it was determined and the `managmenetService` which has unknown value is returned.

daemon/dns/dns.go
mariusSincovici: Shouldn't this return unknown?

mariusSincovici: Are we sure is the best? 😄 
Maybe we are actually breaking the file if we don't recognize the its content.

What if instead of `setUsingBestAvailable` -> `trySettingWithAllAvailable` or something similar?

mariusSincovici: [very small] Maybe to use info level, it might help in the logs if level will be disabled 😄

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,84 @@
...
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Provided interface name or nameservers are empty")
+		return fmt.Errorf("empty interface name or no nameservers provided")
+	}
+	args := []string{"connection", "modify", iface, "ipv4.dns"}
+	args = append(args, nameservers...)
+	args = append(args, "ipv4.ignore-auto-dns", "yes")
+
+	// #nosec G204: input is properly validated
+	if out, err := exec.Command(execNmCli, args...).CombinedOutput(); err != nil {
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Setting DNS with nmcli failed:", strings.TrimSpace(string(out)))
```
mariusSincovici: This might be duplicated in the logs, until now if a function returns errors then the caller decides if it should be logged, otherwise can be duplicated messages.

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Unset removes the DNS configuration for the specified network interface using NetworkManager's nmcli tool.
+// Parameters:
+//   - iface: the network interface name to unset DNS for
+//
+// Returns an error if the nmcli command fails or if reloading the connection fails.
+func (nmcli *NMCli) Unset(iface string) error {
+	if iface == "" {
+		return nil
```
mariusSincovici: Why there is no error returned in this case?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,84 @@
...
+	return nmcli.reloadConnection(iface)
+}
+
+// Name returns name of the DNS management method.
+func (nmcli *NMCli) Name() string {
+	return "nmcli"
+}
+
+// reloadConnection restarts the network connection for the specified interface using nmcli tool.
+func (nmcli *NMCli) reloadConnection(iface string) error {
```
mariusSincovici: Since the function does restart, wouldn't be more clear to name it `restartConnection`?

daemon/dns/dns.go
DoYouEvenCpp: `unknown` is aggregated as one of the possible returned type by any of the methods called to determine available DNS set method..

But I guess I see the confusion here - I'll rearrange a bit how this method is structured to reflect more accurately its intention (and way it logs).

daemon/dns/dns.go
DoYouEvenCpp: Indeed, this should get fixed.

daemon/dns/dns.go
```diff
@@ -120,26 +133,28 @@ func (d *DNSServiceSetter) getManagementServiceBasedOnResolvconfLinkTarget() (dn
...
-		"couldn't determine management service based on resolv.conf comment:", err)
-
-	managmenetService, err = d.getManagementServiceBasedOnResolvconfLinkTarget()
-	if err != nil {
-		log.Println(internal.ErrorPrefix, dnsPrefix,
-			"couldn't determine management service based on resolv.conf link target:", err)
-		return unknown
+	} else {
+		log.Println(internal.WarningPrefix, dnsPrefix, "couldn't determine management service based on resolv.conf comment:", err)
+		managmenetService, err = d.getManagementServiceBasedOnResolvconfLinkTarget()
```
devzbysiu: [typo] `managmenetService`

daemon/dns/dns.go
```diff
@@ -120,26 +133,28 @@ func (d *DNSServiceSetter) getManagementServiceBasedOnResolvconfLinkTarget() (dn
...
+	if nmcliFileInfo, err := d.filesystemHandle.stat(networkManagerLinkTarget); err != nil {
+		log.Println(internal.WarningPrefix, dnsPrefix, "failed to stat NetworkManager resolv.conf:", err)
+	} else if d.filesystemHandle.sameFile(resolvConfFileInfo, nmcliFileInfo) {
+		return nmcliManaged, nil
+	}
+
 	return unknown, fmt.Errorf("link target not recognized")
 }
 
 func (d *DNSServiceSetter) getManagementService() dnsManagementService {
```
devzbysiu, line 161: [small/can-be-ignored] with early returns, we can get rid of excessive `if-else` nesting. Not a big deal though

```suggestion
func (d *DNSServiceSetter) getManagementService() dnsManagementService {
	managementService, err := d.getManagementServiceBasedOnResolvconfComment()
	if err == nil {
		log.Println(internal.InfoPrefix, dnsPrefix, "management service inferred from resolv.conf comment")
		return managementService
	}
	log.Println(internal.WarningPrefix, dnsPrefix, "couldn't determine management service based on resolv.conf comment:", err)

	managementService, err = d.getManagementServiceBasedOnResolvconfLinkTarget()
	if err == nil {
		log.Println(internal.InfoPrefix, dnsPrefix, "management service inferred from link target")
		return managementService
	}
	log.Println(internal.ErrorPrefix, dnsPrefix, "couldn't determine management service based on resolv.conf link target:", err)

	return managementService
}
```

daemon/dns/dns.go
```diff
@@ -120,26 +133,28 @@ func (d *DNSServiceSetter) getManagementServiceBasedOnResolvconfLinkTarget() (dn
...
-	log.Println(internal.WarningPrefix, dnsPrefix,
-		"couldn't determine management service based on resolv.conf comment:", err)
-
-	managmenetService, err = d.getManagementServiceBasedOnResolvconfLinkTarget()
-	if err != nil {
-		log.Println(internal.ErrorPrefix, dnsPrefix,
-			"couldn't determine management service based on resolv.conf link target:", err)
-		return unknown
+	} else {
+		log.Println(internal.WarningPrefix, dnsPrefix, "couldn't determine management service based on resolv.conf comment:", err)
```
devzbysiu: [typo] `resovl.conf`

daemon/dns/dns.go
```diff
@@ -154,20 +169,28 @@ func (d *DNSServiceSetter) set(setter Setter, iface string, nameservers []string
...
 //  1. systemd-resolved DBUS
 //  2. resolvctl utility
 //  3. resolv.conf utility
 //  4. direct write to resovl.conf
-func (d *DNSServiceSetter) setUsingBestAvailable(iface string, nameservers []string) error {
+func (d *DNSServiceSetter) setUsingAvailable(iface string, nameservers []string) error {
 	if err := d.set(d.systemdResolvedSetter, iface, nameservers); err != nil {
 		log.Println(internal.WarningPrefix, dnsPrefix,
-			"failed to configure DNS using systemd-resolved, attempting with resolv.conf")
+			"failed to configure DNS using systemd-resolved, attempting with nmcli")
```
devzbysiu: Maybe would be good to add `err` to the log. Same for line `#188` below

daemon/dns/dns.go
```diff
@@ -154,20 +169,28 @@ func (d *DNSServiceSetter) set(setter Setter, iface string, nameservers []string
 	return nil
 }
 
-// setUsingBestAvailable sets DNS using the first setter in the bellow priority list:
+// setUsingAvailable sets DNS using the first setter in the bellow priority list:
 //  1. systemd-resolved DBUS
 //  2. resolvctl utility
 //  3. resolv.conf utility
 //  4. direct write to resovl.conf
```
devzbysiu: This is not accurate now

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+//
+// Parameters:
+//   - iface: unused
+//   - nameservers: a set of DNS addresses to be used for the configuration
+//
+// Returns an error if:
+//   - the nmcli command fails to fetch connections related to physical interfaces
+//   - the nmcli command fails to execute
+//   - the connection reload fails
+func (nmcli *NMCli) Set(iface string, nameservers []string) error {
```
devzbysiu, line 55: [small/can-be-ignored] maybe `_ string` if `iface` is unused?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return nil
+}
+
+// Unset removes the DNS configuration for the specified network interface using NetworkManager's nmcli tool.
+// Parameters:
+//   - iface: unused
+//
+// Returns an error if the nmcli command fails or if reloading the connection fails.
+func (nmcli *NMCli) Unset(iface string) error {
```
devzbysiu: [small/can-be-ignored] same here with unused `iface`

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
+	if err != nil {
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Failed to get active connections upon UnsetDNS", err)
+		return fmt.Errorf("%w", err)
+	}
+	for _, con := range connections {
+		args := []string{nmCliConKey, "modify", con, nmCliIPv4DNSKey, ""}
+		args = append(args, nmCliIPIgnoreAutoDnsKey, "no")
```
mariusSincovici: What if the user had this set to `yes` before the VPN connection? 🤔

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+//   - the nmcli command fails to execute
+//   - the connection reload fails
+func (nmcli *NMCli) Set(iface string, nameservers []string) error {
+	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
+	if err != nil {
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Failed to get active connections upon SetDNS", err)
+		return fmt.Errorf("%w", err)
+	}
+	for _, con := range connections {
+		args := []string{nmCliConKey, "modify", con, nmCliIPv4DNSKey}
```
mariusSincovici: Why is it needed for all interfaces? 🤔

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return nil
+}
+
+// Name returns name of the DNS management method.
+func (nmcli *NMCli) Name() string {
+	return "nmcli"
+}
+
+// getConnectionFromPhysicalInterfaces retrieves a list of active physical network connection names
```
mariusSincovici: what if the connections are virtual?

daemon/dns/dns_resolvenmcli.go
DoYouEvenCpp: With some changes that happened in the meantime, it actually triggers `reload` thus I say we can keep the name as is.

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+		args := []string{nmCliConKey, "modify", con, nmCliIPv4DNSKey}
+		args = append(args, strings.Join(nameservers, ","))
+		args = append(args, nmCliIPIgnoreAutoDnsKey, "yes")
+
+		if out, err := nmcli.cmdExecutor(execNmCli, args...); err != nil {
+			log.Println(internal.WarningPrefix, nmCliPrintTag, "Setting DNS with nmcli failed:", strings.TrimSpace(string(out)))
+			return fmt.Errorf("setting dns with nmcli failed: %w", err)
+		}
+
+		if err := nmcli.reloadConnection(con); err != nil {
```
devzbysiu, line 83: Maybe we should undo the changes in case something goes wrong either with `modify` or `reload`? Currently, we'll modify some connections and then if modification of Nth connection fails, we just bail, leaving inconsistent state right? 
Similarly with `reload` - if some of the connections are reloaded, but reload of Nth fails, we bail

daemon/dns/dns.go
DoYouEvenCpp: I'll demangle all those I/O operations  under a separate PR

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+			strings.Contains(fields[1], cdmaType) {
+			conns = append(conns, strings.TrimSpace(fields[0]))
+		}
+	}
+	return conns, nil
+}
+
+// reloadConnection restarts the network connection for the specified interface using nmcli tool.
+func (nmcli *NMCli) reloadConnection(iface string) error {
+	reloadArgs := []string{nmCliConKey, "reload", iface}
```
devzbysiu: Is `reload` accepting connection name? Isn't it just reloading all configurations?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+			strings.Contains(fields[1], gsmType) ||
+			strings.Contains(fields[1], cdmaType) {
+			conns = append(conns, strings.TrimSpace(fields[0]))
+		}
+	}
+	return conns, nil
+}
+
+// reloadConnection restarts the network connection for the specified interface using nmcli tool.
+func (nmcli *NMCli) reloadConnection(iface string) error {
```
devzbysiu: This is rather connection identifier instead of  interface right?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+// Name returns name of the DNS management method.
+func (nmcli *NMCli) Name() string {
+	return "nmcli"
+}
+
+// getConnectionFromPhysicalInterfaces retrieves a list of active physical network connection names
+// from NetworkManager using nmcli. It filters connections by type, including only wireless,
+// ethernet, GSM, and CDMA connections (eg. physical ones).
+// Returns a slice of connection names and an error if the nmcli command fails.
+// Returns an empty slice upon any malformed output from the nmcli command.
```
devzbysiu: Hmm, it looks like we just skip malformed output and continue processing

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+//
+// Returns an error if:
+//   - the nmcli command fails to fetch connections related to physical interfaces
+//   - the nmcli command fails to execute
+//   - the connection reload fails
+func (nmcli *NMCli) Set(iface string, nameservers []string) error {
+	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
+	if err != nil {
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Failed to get active connections upon SetDNS", err)
+		return fmt.Errorf("%w", err)
```
devzbysiu: This doesn't add any context 😄, similarly in line `#79`

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,148 @@
...
+		args = append(args, strings.Join(nameservers, ","))
+		args = append(args, nmCliIPIgnoreAutoDnsKey, "yes")
+
+		if out, err := nmcli.cmdExecutor(execNmCli, args...); err != nil {
+			log.Println(internal.WarningPrefix, nmCliPrintTag, "Setting DNS with nmcli failed:", strings.TrimSpace(string(out)))
+			return fmt.Errorf("setting dns with nmcli failed: %w", err)
+		}
+
+		if err := nmcli.reloadConnection(con); err != nil {
+			return fmt.Errorf("failure %w", err)
```
devzbysiu: Maybe more context here would be beneficial? `failure` doesn't tell much

daemon/dns/dns_resolvenmcli_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_DSNResolveNmCli_PhysicalInterfaceDeduction(t *testing.T) {
```
devzbysiu: [typo] `DSN`

daemon/dns/dns_resolvenmcli.go
DoYouEvenCpp: On a side note,  I think we should `reload` only once (since it's an operation over all the connections) and call `up` on respective cons we modified.

daemon/dns/dns.go
DoYouEvenCpp: Addressed under https://github.com/NordSecurity/nordvpn-linux/pull/1345

DoYouEvenCpp: Addressed under https://github.com/NordSecurity/nordvpn-linux/pull/1345

daemon/dns/dns.go
```diff
@@ -170,21 +190,30 @@ func (d *DNSServiceSetter) set(setter Setter, iface string, nameservers []string
...
-func (d *DNSServiceSetter) setUsingBestAvailable(iface string, nameservers []string) error {
+//  3. NetworkManager nmcli tool
+//  4. resolv.conf utility
+//  5. direct write to resolv.conf
+func (d *DNSServiceSetter) setUsingAvailable(iface string, nameservers []string) error {
 	d.analytics.setManagementService(systemdResolved)
 	if err := d.set(d.systemdResolvedSetter, iface, nameservers); err != nil {
 		log.Println(internal.WarningPrefix, dnsPrefix,
-			"failed to configure DNS using systemd-resolved, attempting with resolv.conf")
+			"failed to configure DNS using systemd-resolved, attempting with nmcli: %w", err)
```
devzbysiu: `Println` is not supporting format strings , also `%w` is for wrapping errors, maybe you meant `%v`? If this was meant to be just a `log.Println`, you can pass the `err`.
Also below in line `#211`

Currently this prints
```
Feb 18 06:50:25 debian-13 nordvpnd[821]: 2026/02/18 06:50:25 [Warning] [DNS] failed to configure DNS using nmcli, attempting with resolv.conf, %w failed to set DNS: dns not set, no dns setting method is available
```

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// getConnectionFromPhysicalInterfaces retrieves a list of active physical network connection names
+// from NetworkManager using nmcli. It filters connections by type, including only wireless,
+// ethernet, GSM, and CDMA connections (eg. physical ones).
+// Any malformed output  is disregarded.
+//
+// Returns a slice of connection names and an error if the nmcli command fails.
+func (nmcli *NMCli) getConnectionFromPhysicalInterfaces() ([]string, error) {
+	cmd, err := nmcli.cmdExecutor(nmCliExecutable, "-t", "-f", "NAME,TYPE", "con", "show", "--active")
```
devzbysiu: Maybe this should be something else than `cmd`? I've got confused reading this output processing below 😄 , I thought this is command, but it's the result 😄

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return []string{}, fmt.Errorf("Failed to fetch active devices: %w", err)
+	}
+	var conns = []string{}
+	lines := strings.SplitSeq(string(cmd), "\n")
+	for line := range lines {
+		parts := strings.Split(line, ":")
```
devzbysiu, line 136: Should we worry about conn names with `:` inside? 🤔 , network manager allows such connections. Those `:` are escaped in the name, but we'll ignore it 🤔

daemon/dns/dns_resolvenmcli.go
```diff
@@ -0,0 +1,214 @@
...
+	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
+	if err != nil {
+		log.Println(internal.WarningPrefix, nmCliPrintTag, "Failed to get active connections upon SetDNS", err)
+		return fmt.Errorf("failed to get active connection upon SetDNS: %w", err)
+	}
+
+	originalStates := make(map[string]*connectionState)
+	for _, con := range connections {
+		state, err := nmcli.getConnectionState(con)
+		originalStates[con] = state
```
devzbysiu: Shouldn't the state be saved after error check? 🤔 
And then in the `if err != nil` block line below - shouldn't we have `nmcli.rollback(originalStates)`?

daemon/dns/dns_resolvenmcli.go
DoYouEvenCpp: I guess this works correctly here, the bits between lines 139-141 takes care of that. Also covered this with a UT, right there: https://github.com/NordSecurity/nordvpn-linux/blob/skubiak/LVPN-9444_introduce-dns-support-via-nmcli/daemon/dns/dns_resolvenmcli_test.go#L76

daemon/dns/dns.go
DoYouEvenCpp: A typical PEBKAC, a.k.a. a copy-paste error there. 😉

daemon/dns/dns_resolvenmcli.go
DoYouEvenCpp: Indeed, this is incorrect. Good catch!

PR #1306, opened 2026-01-20
LVPN-9863: Review and unify overlapping autotests

bbrutall:
- removed 'test_reconnect_matrix' and added its logic to the 'test_status_change_technology_and_protocol_reconnect'
- optimized tests from 'test_connect', 'test_autoconnect', 'test_fileshare' modules

### DIFF
--- a/test/qa/test_autoconnect.py
+++ b/test/qa/test_autoconnect.py
@@ -20,6 +20,9 @@ def autoconnect_base_test(group):
     daemon.wait_for_autoconnect()
     assert network.is_connected()
 
+    status_info = daemon.get_status_data()
+    assert "Connected" in status_info["status"]
+
     output = sh_no_tty.nordvpn.set.autoconnect.off()
     print(output)
     assert settings.MSG_AUTOCONNECT_DISABLE_SUCCESS in output
@@ -111,8 +114,15 @@ def test_autoconnect_to_standard_group(tech, proto, obfuscated, group):
     autoconnect_base_test(group)
 
 
-@pytest.mark.parametrize("group", lib.ADDITIONAL_GROUPS)
-@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER)
+@dynamic_parametrize(
+    [
+        "tech", "proto", "obfuscated", "group",
+    ],
+    ordered_source=[lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER],
+    randomized_source=[lib.ADDITIONAL_GROUPS],
+    generate_all=IS_NIGHTLY,
+    id_pattern="{tech}-{proto}-{obfuscated}-{group}",
+)
 def test_autoconnect_to_additional_group(tech, proto, obfuscated, group):
     """Manual TC: LVPN-6786"""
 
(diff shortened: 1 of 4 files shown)


test/qa/test_autoconnect.py
```diff
@@ -111,17 +111,31 @@ def test_autoconnect_to_standard_group(tech, proto, obfuscated, group):
...
+    ordered_source=[lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER],
+    randomized_source=[lib.ADDITIONAL_GROUPS],
+    generate_all=IS_NIGHTLY,
+    id_pattern="{tech}-{proto}-{obfuscated}-{group}",
+)
 def test_autoconnect_to_additional_group(tech, proto, obfuscated, group):
     """Manual TC: LVPN-6786"""
 
     lib.set_technology_and_protocol(tech, proto, obfuscated)
     autoconnect_base_test(group)
```
mariusSincovici, line 130: Could the tests check the status and that is correctly connected?

test/qa/test_fileshare.py
```diff
@@ -207,10 +209,20 @@ def check_files_status_sender():
...
+        "background_accept",
+    ],
+    ordered_source=[list(fileshare.FileSystemEntity)],
+    randomized_source=[[(background_send, path_flag, background_accept) for
+                        background_send, path_flag, background_accept in
+                        product([True, False], [True, False], ["", "--background"])]],
+    id_pattern="send_{filesystem_entity}-send_bg_{background_send}-accept_custom_{path_flag}-accept_bg_{background_accept}",
+    generate_all=IS_NIGHTLY,
+)
 def test_fileshare_transfer(filesystem_entity: fileshare.FileSystemEntity, background_send: bool, path_flag: str, background_accept: str):
```
devzbysiu: Shouldn't we have `path_flag: bool` here? 🤔 

Maybe we could also name it in more specific way, something like `is_custom_path` or something like that?

devzbysiu: [small] also if `background_accept` will be empty, then we'll have `...-accept_bg_` in the pattern

test/qa/test_combinations.py
```diff
@@ -52,14 +52,24 @@ def test_reconnect_matrix(
         source_obfuscated,
         target_obfuscated,
 ):
-    """Manual TC: LVPN-8674"""
-
+    """Manual TC: LVPN-8674, LVPN-8694"""
     lib.set_technology_and_protocol(source_tech, source_proto, source_obfuscated)
     connect_base_test()
```
mariusSincovici, line 57: Same here, to check the status?

bbrutall: Added status verification to the `connect_base_test`

test/qa/test_autoconnect.py
bbrutall: Added status verification to the `connect_base_test`

test/qa/test_fileshare.py
bbrutall: It makes sense. However, I wanted to avoid making significant changes to the existing names.

bbrutall: Yes. Added quotes for better readability.

devzbysiu (approved):

PR #1311, opened 2026-01-21
LVPN-9882: Update rules insertion

mariusSincovici:
A fix for when daemon and OS rules are out of sync

### DIFF
--- a/.github/workflows/ci-gitlab.yml
+++ b/.github/workflows/ci-gitlab.yml
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.4.5
+      triggered-ref: v1.4.8


daemon/firewall/firewall.go
```diff
@@ -59,30 +60,39 @@ func (fw *Firewall) Add(rules []Rule) error {
...
-			log.Printf("%s rule already exists: %v", internal.WarningPrefix, rule.Name)
+			existingRule = Rule{}
+			// Check if rule exists in the OS, but not in the app
+			existsInOS := slices.ContainsFunc(ruleNamesFromOS, func(ruleName string) bool {
+				return ruleName == rule.Name || ruleName == rule.SimplifiedName
+			})
+			if existsInOS {
+				// delete it from the firewall because later it will be inserted
+				log.Printf("%s rule already exists in OS: %v", internal.WarningPrefix, rule.Name)
+				if err := fw.current.Delete(rule); err != nil {
```
mariusSincovici, line 96: this needs to be improved

daemon/firewall/firewall.go
```diff
@@ -59,30 +60,39 @@ func (fw *Firewall) Add(rules []Rule) error {
...
 		} else {
-			log.Printf("%s rule already exists: %v", internal.WarningPrefix, rule.Name)
+			existingRule = Rule{}
+			// Check if rule exists in the OS, but not in the app
+			existsInOS := slices.ContainsFunc(ruleNamesFromOS, func(ruleName string) bool {
+				return ruleName == rule.Name || ruleName == rule.SimplifiedName
+			})
+			if existsInOS {
+				// delete it from the firewall because later it will be inserted
+				log.Printf("%s rule already exists in OS: %v", internal.WarningPrefix, rule.Name)
```
thegravi: [small] New line is missing

daemon/firewall/firewall.go
```diff
@@ -59,30 +60,39 @@ func (fw *Firewall) Add(rules []Rule) error {
...
+				if err := fw.current.Delete(rule); err != nil {
+					log.Printf("%s failed to delete rule from OS: %v", internal.ErrorPrefix, rule.Name)
```
thegravi: [small] New line is missing. Also why not log `err`?

daemon/firewall/firewall_test.go
```diff
@@ -199,6 +219,9 @@ func TestFirewallAdd(t *testing.T) {
 				_, ok := err.(*Error)
 				assert.True(t, ok)
 			}
+			if len(test.networkChanged) > 0 {
+				assert.Equal(t, test.hasError, fw.Add(test.networkChanged) != nil)
```
DoYouEvenCpp: [nit] I know this is few lines more, but maybe something like that?

```suggestion
				if test.hasError {
					assert.Error(t, fw.Add(test.networkChanged))
				} else {
					assert.NoError(t, fw.Add(test.networkChanged))
				}
```

this combination of `test.hasError` and testing against `nil` is a bit cryptic.  😉 

WDYT?

daemon/firewall/firewall_test.go
```diff
@@ -188,6 +190,24 @@ func TestFirewallAdd(t *testing.T) {
 			agent:    &mockAgent{},
 			hasError: false,
 		},
+		{
```
thegravi, line 193: Can we add more tests to assert what happens when: 
- `Delete()` fails for the stale OS rule?
- `Add()` fails after successful Delete()?

daemon/firewall/firewall.go
```diff
@@ -59,30 +60,39 @@ func (fw *Firewall) Add(rules []Rule) error {
...
-		if !slices.Contains(alreadyAddedRules, rule.Name) &&
-			!slices.Contains(alreadyAddedRules, rule.SimplifiedName) {
-			if err := fw.current.Add(rule); err != nil {
-				return NewError(fmt.Errorf("adding %s: %w", rule.Name, err))
-			}
 		} else {
-			log.Printf("%s rule already exists: %v", internal.WarningPrefix, rule.Name)
+			existingRule = Rule{}
+			// Check if rule exists in the OS, but not in the app
+			existsInOS := slices.ContainsFunc(ruleNamesFromOS, func(ruleName string) bool {
```
bartoszWojciechO, line 76: For simplicity, wouldn't it be easier to always attempt a rule delete in the OS, ignoring the errors? No need to get rules from the OS/check if rule already exists..

bartoszWojciechO (approved):

daemon/firewall/firewall.go
```diff
@@ -49,40 +49,67 @@ func NewFirewall(noop, working Agent, publisher events.Publisher[string], enable
...
+				return ruleName == rule.Name || ruleName == rule.SimplifiedName
+			})
+			if existsInOS {
+				// block traffic until rules are swapped
+				if trafficWasDropped == false {
+					blockRule := Rule{Name: "drop-all", Direction: TwoWay, Allow: false}
+					if err := fw.current.Add(blockRule); err != nil {
+						log.Printf("%s failed to temporary block traffic: %v\n", internal.ErrorPrefix, err)
```
edvyno1: [typo] temporary -> temporarily

daemon/firewall/firewall.go
mariusSincovici: Not anymore, because the logic changed based on this.

daemon/firewall/rule.go
```diff
@@ -111,6 +111,10 @@ func (r Rule) Equal(other Rule) bool {
 		r.SimplifiedName == other.SimplifiedName
 }
 
+func (r Rule) IsValid() bool {
```
thegravi: This is confusion when thinking about it more. It does not tell that the rule is valid, it tells that it is not empty. I would suggest naming it `IsSet()` or `IsNotEmpty()`.

mariusSincovici: Yeah, makes sense. Change to `IsEmpty`.

PR #1314, opened 2026-01-22
LVPN-9443: resolv conf debug events

bartoszWojciechO:
Adds `resolv.conf` based developer events. 

A debug event will be emitted whenever:
1. `resolv.conf` is used by the daemon to configure DNS.
2. Changes to `resolv.conf` are detected.

This commit also adds basics that will be used in further DNS debugger events(coming in next PRs).

### DIFF
--- a/daemon/dns/resolvconf_monitor_test.go
+++ b/daemon/dns/resolvconf_monitor_test.go
@@ -80,7 +80,7 @@ func Test_ResolvConfMonitoring(t *testing.T) {
 	}
 
 	resolvConfMonitor.start()
-	eventsChan <- fsnotify.Event{}
+	eventsChan <- fsnotify.Event{Op: fsnotify.Write}
 	checkResultFunc := func() bool {
 		return analyticsMock.getResolvConfEmitted()
 	}


daemon/dns/dns.go
```diff
@@ -75,16 +75,23 @@ type DNSServiceSetter struct {
...
+	unsetter          Setter
+	filesystemHandle  statingFilesystemHandle
+	analytics         analytics
+	resolvConfMonitor resolvConfMonitor
 }
 
-func NewDNSServiceSetter(publisher events.Publisher[string]) *DNSServiceSetter {
+func NewDNSServiceSetter(publisher events.Publisher[string],
+	debugPublisher events.PublishSubcriber[events.DebuggerEvent]) *DNSServiceSetter {
+	analytics := newDNSAnalytics(debugPublisher)
```
DoYouEvenCpp, line 87: This might actually make the testing there a bit more problematic, maybe we should leverage dependency injection here?

daemon/dns/resolvconf_monitor.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func newResolvConfMonitor(analytics analytics) resolvConfFileWatcherMonitor {
+	return resolvConfFileWatcherMonitor{
+		analytics:      analytics,
+		getWatcherFunc: getWatcher,
+	}
+}
+
+func getWatcher() (*fsnotify.Watcher, error) {
```
DoYouEvenCpp: This is almost 1:1 to what we use at norduser process monitor (see: https://github.com/NordSecurity/nordvpn-linux/blob/main/norduser/process_monitor.go#L181). Should be consider extracting this to a common place and reuse it?

DoYouEvenCpp: We can maybe consider reusing the whole functionality of watcher, not only specific bits, WDYT?

daemon/dns/resolvconf_monitor.go
```diff
@@ -0,0 +1,108 @@
...
+	r.doneChan = doneChan
+	log.Println(internal.InfoPrefix, dnsPrefix, "starting resolv.conf file watcher")
+	for {
+		select {
+		case _, ok := <-watcher.Events:
+			log.Println(internal.InfoPrefix, dnsPrefix, "resolv.conf overwrite detected")
+			if !ok {
+				return fmt.Errorf("file watcher closed")
+			}
+			r.analytics.emitResolvConfOverwrittenEvent()
```
DoYouEvenCpp: Wouldn't be beneficial for us to report also some actual errors upon DNS setting?

daemon/dns/dns.go
bartoszWojciechO: Yeah, this is changed in the upcoming PR.

daemon/dns/resolvconf_monitor.go
bartoszWojciechO: Last PR in the chain adds events for errors. It's this PR => DNS config success events => DNS config error events

daemon/dns/resolvconf_monitor.go
bartoszWojciechO: Extracted the builder function to a common function. 

As for the other parts of watcher functionality(i.e handling the events): the simple, common part of handling the events is the for loop select and handling the errors coming from watcher. Other parts are customized for the use case. There is also the key difference in the fact that norduser watcher loop runs indefinitely, while DNS loop needs to respect the context.

We could maybe comonize them by using function variables, but imo it's not worth the complexity, since the common part is so simple.

internal/filesystem.go
```diff
@@ -560,3 +561,24 @@ func isProcessRunning(executablePath string, readdir readdirFunc, readfile readf
 
 	return false, nil
 }
+
+func GetFileWatcher(pathsToMonitor ...string) (watcher *fsnotify.Watcher, err error) {
```
DoYouEvenCpp, line 566: Can you please provide a short comment for this method (since it's made public now)?

daemon/dns/analytics.go
```diff
@@ -0,0 +1,80 @@
...
+		log.Println(internal.DebugPrefix, dnsPrefix, "failed to serialize event json for resovl.conf overwrite:", err)
+		jsonData = []byte("{}")
+	}
+
+	debuggerEvent := events.NewDebuggerEvent(string(jsonData)).
+		WithKeyBasedContextPaths(events.ContextValue{
+			Path:  debuggerEventBaseKey + ".type",
+			Value: resolvConfOverwritten.String()}).
+		WithGlobalContextPaths("device.*",
+			"application.nordvpnapp.*",
```
DoYouEvenCpp, line 77: Possibly we can narrow here a bit data sent out - but for now, let's keep it this way. Once the specific format of events is known, we can always tweak it here.

daemon/dns/analytics.go
```diff
@@ -0,0 +1,80 @@
+package dns
```
DoYouEvenCpp, line 1: Can you please provide some (usual) documentation into this file?

daemon/dns/resolvconf_monitor.go
```diff
@@ -0,0 +1,88 @@
...
+			log.Println(internal.ErrorPrefix, dnsPrefix, "resolv.conf monitoring failed:", err)
+		}
+	}()
+}
+
+func (r *resolvConfFileWatcherMonitor) Stop() {
+	if r.cancelFunc != nil {
+		r.cancelFunc()
+		// wait for the monitor goroutine to finish
+		<-r.doneChan
```
DoYouEvenCpp: Correct me if I'm wrong, but this might still ends up with a deadlock.

Probably not gonna happen, but what if:
- `Start` runs
- the goroutine with `monitorResolvConf` is triggered, but hangs on the scheduler and never got actually ran, eg. leaving the channel still unitialised
- `Stop` gets triggered, and is executed before the resolvconf goroutine, thus is left locked on the channel

I guess we should elevate channel creation into actual body of the `Start`, and maybe possibly also check if channel got actually created upon processing the `Stop`.

What do you think?

DoYouEvenCpp: This probably also could get reproduced with following test:

```
func Test_ResolvConfMonitoringDoesNotDeadlock(t *testing.T) {
	category.Set(t, category.Unit)

	eventsChan := make(chan fsnotify.Event)
	errorChan := make(chan error)
	getMockWatcherFunc := func(...string) (*fsnotify.Watcher, error) {

		watcher, _ := fsnotify.NewWatcher()
		watcher.Events = eventsChan
		watcher.Errors = errorChan
		time.Sleep(time.Duration(time.Duration.Seconds(1))) //or any other value, significantly long enough
		return watcher, nil
	}

	analyticsMock := newAnalyticsMock()

	resolvConfMonitor := resolvConfFileWatcherMonitor{
		analytics:      &analyticsMock,
		getWatcherFunc: getMockWatcherFunc,
	}

	resolvConfMonitor.Start()
	resolvConfMonitor.Stop()
	assert.Equal(t, true, true) //whatever here...
}

Comment (bartoszWojciechO) [reply daemon/dns/resolvconf_monitor.go]: Moved channel creation out of the goroutine and added a timeout when stopping, should resolve the possible issues.

File: daemon/dns/resolvconf_monitor.go [hunk 8]
```diff
@@ -0,0 +1,96 @@
...
+	log.Println(internal.InfoPrefix, dnsPrefix, "starting resolv.conf file watcher")
+	for {
+		select {
+		case _, ok := <-watcher.Events:
+			log.Println(internal.InfoPrefix, dnsPrefix, "resolv.conf overwrite detected")
+			if !ok {
+				return fmt.Errorf("file watcher closed")
+			}
+			r.analytics.emitResolvConfOverwrittenEvent()
+			return nil
```
devzbysiu, line 62: Now it will stop emitting events after first one, is this intentional? 🤔

daemon/dns/resolvconf_monitor.go
```diff
@@ -0,0 +1,96 @@
...
+	defer func() {
+		if watcher != nil {
+			_ = watcher.Close()
+		}
+	}()
+
+	log.Println(internal.InfoPrefix, dnsPrefix, "starting resolv.conf file watcher")
+	for {
+		select {
+		case _, ok := <-watcher.Events:
```
devzbysiu: I think this is too wide 😄 , it catches even `touch /etc/resolv.conf`

daemon/dns/resolvconf_monitor.go
bartoszWojciechO: Yeah. The assumption here that first overwrite breaks our config, so sending more data would just make it harder to read. In essence those events are _resolv_conf_broken_by_overwrite_ rather than just overwrites.

devzbysiu: Ok, then shouldn't we close the channel here on exit also? :thinking:

PR #1316, opened 2026-01-22
LVPN-9804: Extending UI events (CLI client implementation)

thegravi:
Migrating CLI and Tray clients from the legacy client-id metadata approach to the new `ui-*` metadata format for UI event tracking.

### DIFF
--- a/uievent/client_interceptor.go
+++ b/uievent/client_interceptor.go
@@ -41,12 +41,7 @@ func (i *ClientInterceptor) UnaryInterceptor(
 		return invoker(ctx, method, req, reply, cc, opts...)
 	}
 
-	uiCtx := &UIEventContext{
-		FormReference: i.formReference,
-		ItemName:      itemName,
-		ItemType:      pb.UIEvent_CLICK,
-		ItemValue:     pb.UIEvent_ITEM_VALUE_UNSPECIFIED,
-	}
+	uiCtx := NewClickContext(i.formReference, itemName)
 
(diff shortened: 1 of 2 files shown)


cmd/cli/main.go
```diff
@@ -90,7 +90,7 @@ func main() {
 	}
 	log.SetOutput(fileLogger)
 
-	clientIDMetadataInterceptor := clientid.NewInsertClientIDInterceptor(pb.ClientID_CLI)
```
mariusSincovici, line 93: Is this class still used, the `clientid`? Since it was added new file for interceptors.

mariusSincovici (approved):

uievent/client_interceptor.go
```diff
@@ -0,0 +1,119 @@
...
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+
+	itemName := methodToItemName(method)
+	if itemName == pb.UIEvent_ITEM_NAME_UNSPECIFIED {
+		// Skip metadata attachment
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+
+	uiCtx := &UIEventContext{
```
bartoszWojciechO: Maybe we could have a constructor or some sort of builder function for this struct? Then if new fields are added we'd get compile errors. Would make it impossible to miss updating the construction in any of the two interceptors.

cmd/cli/main.go
thegravi: It's removed [here](https://github.com/NordSecurity/nordvpn-linux/pull/1318)

PR #1317, opened 2026-01-23
LVPN-9804: Extending UI events (GUI client implementation)

thegravi:
Migrates GUI from legacy `client-id` metadata to the new `ui-*` metadata format for UI event analytics, with added support for "recent connections".


gui/lib/data/models/connect_arguments.dart
```diff
@@ -65,6 +66,39 @@ final class ConnectArguments {
     return connectRequest;
   }
 
+  /// Converts the connection arguments to a UIEvent_ItemValue for analytics.
+  ///
+  /// Specialty groups (Double VPN, P2P, etc.) take priority over connection type.
+  ///  Falls back to CITY/COUNTRY for standard VPN connections.
+  UIEvent_ItemValue toUIEventItemValue() {
```
bartoszWojciechO, line 73: Is this supposed to be different between GUI and CLI? CLI seems to skip the `ItemValue`. Unless it's added in some other PR, I've checked/tested https://github.com/NordSecurity/nordvpn-linux/pull/1316/changes#diff-b6c886b234642f53d5dcaca9567323982180e6a2a12289ed6bafc99fb46da21e and this one.

protobuf/daemon/uievent.proto
```diff
@@ -13,6 +13,7 @@ message UIEvent {
     CLI = 1;
     TRAY = 2;
     HOME_SCREEN = 3;
+    GUI = 4; // temporary, to be replaced with more specific references
```
bartoszWojciechO, line 16: When is this used? `HOME_SCREEN` seems to be used in it's place 🤔

thegravi: For login/logout rpc calls.

gui/lib/data/models/connect_arguments.dart
thegravi: It was mostly tailored to recent-connections uievents. It happens so that for GUI general `connect` method adds this uievent context for both caller (simple connect  / recent connect).

PR #1319, opened 2026-01-27
LVPN-9901: Implement debugger events for allowlist

thegravi:
Added `allowlist` debugger events.

### DIFF
--- a/internal/analytics/result.go
+++ b/internal/analytics/result.go
@@ -1,0 +1,16 @@
+// Package analytics provides shared utilities for analytics events across the application.
+package analytics
+
+// Result constants for analytics events
+const (
+	ResultSuccess = "success"
+	ResultFailure = "failure"
+)
+
+// BoolToResult converts a boolean success value to a result string.
+func BoolToResult(success bool) string {
+	if success {
+		return ResultSuccess
+	}
+	return ResultFailure
+}


daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+// Event identification constants
+const (
+	Namespace         = "nordvpn-linux"
```
DoYouEvenCpp: This is now a duplicate of https://github.com/NordSecurity/nordvpn-linux/blob/main/config/remote/constants.go#L12
Not also this is being unified under: https://github.com/NordSecurity/nordvpn-linux/pull/1314/changes#diff-5f8133123ca0cdda11d0fe1474f10209566dab8c86e9a553d3136bbb55f89e0fR104

daemon/jobs.go
```diff
@@ -35,6 +36,9 @@ func (r *RPC) StartJobs(
 	statePublisher *state.StatePublisher,
 	heartBeatPublisher events.Publisher[time.Duration],
 ) {
+	// emit allowlist snapshot event on startup
+	r.emitAllowlistSnapshot()
```
devzbysiu: Does it make sense to have it in `jobs.go`?

daemon/rpc_set_allowlist.go
```diff
@@ -155,34 +147,92 @@ func getPortsInARange(start int64, stop int64) []int64 {
...
-		}, nil
+		// emit failure event
+		if event := createAllowlistOperationEvent(in, allowlist.OpAdd, false, code); event != nil {
+			r.emitAllowlistAnalytics(event)
+		}
+		return &pb.Payload{Type: code}, nil
+	}
+
+	resultCode := r.handleNewAllowlist(allowlistCfg)
+	success := resultCode == internal.CodeSuccess
```
DoYouEvenCpp, line 166: Maybe move `success` evaluation into body of `createAllowlistOperationEvent`  - we pass the result code there already, so we can reuse this info to conclude also the success boolean.

daemon/rpc_set_allowlist.go
```diff
@@ -73,55 +74,46 @@ func (r *RPC) getNewAllowlist(req *pb.SetAllowlistRequest, remove bool) (config.
 	var cfg config.Config
 	err := r.cm.Load(&cfg)
 	if err != nil {
-		log.Println(internal.ErrorPrefix, "reading config: %w", err)
+		log.Println(internal.ErrorPrefix, "reading config:", err)
 		return config.Allowlist{}, internal.CodeConfigError
 	}
 
-	allowlist := cfg.AutoConnectData.Allowlist
+	al := cfg.AutoConnectData.Allowlist
```
devzbysiu: [small] Does it make sense to have less readable name for this case? Also the scope of this var is whole function, so I would say it's more desirable to have more readable name here

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+const (
+	ProtoTCP  = "tcp"
+	ProtoUDP  = "udp"
+	ProtoBoth = "both"
+	ProtoNA   = "n/a"
+)
+
+// Results
+const (
+	ResultSuccess = "success"
```
DoYouEvenCpp: [nit] At some point we might want a shared set of analytics constants/utilities, since usage will probably grow over time.

no need to fix it now, though. 😉

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+		UDPPortCount: udpCount,
+		SubnetCount:  subnetCount,
+		TotalCount:   total,
+		IsEnabled:    total > 0,
+	}
+}
+
+// ToDebuggerEvent converts OperationEvent to a DebuggerEvent for moose publishing
+func (e *OperationEvent) ToDebuggerEvent() *events.DebuggerEvent {
+	jsonData, _ := json.Marshal(e)
```
devzbysiu: Any reason for ignoring error? Also for line 208 below. For RC we register the error, not sure if it's valuable in this case, but maybe we could log the error at least

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+	"application.nordvpnapp.platform",
+	// Related feature states
+	"application.nordvpnapp.config.user_preferences.local_network_discovery_allowed.value",
+	"application.nordvpnapp.config.user_preferences.kill_switch_enabled.value",
+	"application.nordvpnapp.config.user_preferences.meshnet_enabled.value",
+	"application.nordvpnapp.config.current_state.is_on_vpn.value",
+}
+
+// OperationEvent represents an allowlist add/remove/clear operation.
+type OperationEvent struct {
```
DoYouEvenCpp, line 60: For scenarios where we don't pass all the fields - do we want all of them to be still marshalable?
eg., the `omit empty`

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+	case internal.CodeFailure:
+		return "operation failed"
+	case internal.CodeConfigError:
+		return "configuration error"
+	case internal.CodePrivateSubnetLANDiscovery:
+		return "private subnet conflicts with LAN discovery"
+	case internal.CodeAllowlistInvalidSubnet:
+		return "invalid subnet format"
+	case internal.CodeAllowlistSubnetNoop:
+		return "subnet already exists or does not exist"
```
DoYouEvenCpp: This is ambiguous, can you please make it more precise (specifically, the second bit  - `or does not exist`)?

Or maybe something like:
```suggestion
		return "no changes made to subnet allowlist"
```

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+	case internal.CodeAllowlistPortOutOfRange:
+		return "port out of valid range (1-65535)"
+	case internal.CodeAllowlistPortNoop:
+		return "port already exists or does not exist"
```
DoYouEvenCpp: Same here, slightly ambiguous.

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+// Event identification constants
+const (
+	Namespace         = "nordvpn-linux"
+	Subscope          = "allowlist"
+	EventOperation    = Subscope + "_operation"
+	EventSnapshot     = Subscope + "_snapshot"
+	contextPathPrefix = "allowlist"
+)
+
+// Operation types
```
devzbysiu, line 24: `[you can ignore this]` not sure if it makes sense or it will introduce too much noise, but something to consider is to have specific types for those to have more type safety, but then you would have to go with converting it to string at the end, so not sure if it's worth it in this case

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+		return 0
+	}
+	return mask
+}
+
+type Error struct {
+	Code int64
+}
+
+func NewError(code int64) *Error {
```
devzbysiu: This is always used like `NewError(..).Error()` so you convert it to some type and then immediately to string. Maybe something to look into if it makes sense

devzbysiu: Btw. is it a good name for this if we can have Error with Code "success"? 😄

daemon/allowlist/analytics.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func boolToResult(success bool) string {
+	if success {
+		return ResultSuccess
+	}
+	return ResultFailure
+}
+
+func extractMask(cidr string) int64 {
```
mariusSincovici: Shouldn't this return -1 on error,  to see if there are errors in the string parsing?
It would also cover the case for `0.0.0.0/0` 😄

daemon/rpc_set_allowlist.go
```diff
@@ -155,34 +147,92 @@ func getPortsInARange(start int64, stop int64) []int64 {
 	return ports
 }
 
+// emitAllowlistAnalytics publishes an allowlist operation event to the debugger events channel
+func (r *RPC) emitAllowlistAnalytics(event *allowlist.OperationEvent) {
+	r.events.Debugger.DebuggerEvents.Publish(*event.ToDebuggerEvent())
+}
+
+// createAllowlistOperationEvent creates an analytics event based on the request type
+func createAllowlistOperationEvent(
```
devzbysiu: Maybe this should be in `allowlist/analytics.go`?

daemon/jobs.go
mariusSincovici: [small] Also if this is something that can be executed async maybe it can be moved on some separate goroutine, because this StartJobs is getting bigger and bigger and it is slowing down the startup.

daemon/rpc_set_allowlist.go
thegravi: A new package was called allowlist, hence the rename. Gonna make it more self explanatory 👍🏻.

daemon/rpc_set_allowlist.go
thegravi: I'd keep things as they are. Keeping the `success` determination at the caller side makes the business logic explicit and maintains separation between what "success" means (caller's decision) and how events are created (factory's responsibility). There are also cases where we explicitly pass `success = false`.

devzbysiu (approved):

PR #1322, opened 2026-01-29
LVPN-9915: Update detection for virtual interfaces

mariusSincovici:
Improve virtual interfaces detection, when the interface is deleted while they are enumerated.

### DIFF
--- a/daemon/device/device.go
+++ b/daemon/device/device.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
@@ -30,7 +29,6 @@ type SystemDeps interface {
 	Interfaces() ([]net.Interface, error)
 
 	// os
-	ReadDir(name string) ([]os.DirEntry, error)
 	FileExists(path string) bool
 }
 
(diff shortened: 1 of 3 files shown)


daemon/device/device.go
```diff
@@ -65,7 +71,9 @@ func listVirtual() ([]net.Interface, error) {
 	for _, file := range files {
 		dev, err := sysDepsImpl.InterfaceByName(file.Name())
 		if err != nil {
-			return nil, fmt.Errorf("retrieving network interface by name: %w", err)
+			log.Printf("fail to get virtual interface by name [%s]: %v\n", file.Name(), err)
```
devzbysiu: `internal.WarningPrefix`?

devzbysiu: `internal.WarningPrefix` missing?

daemon/device/device.go
```diff
@@ -29,7 +29,7 @@ type SystemDeps interface {
 	Interfaces() ([]net.Interface, error)
 
 	// os
-	ReadDir(name string) ([]os.DirEntry, error)
+	FileExists(path string) bool
```
DoYouEvenCpp: Does this have to be exported, seems used only in the context of device pacakge?

daemon/device/device.go
```diff
@@ -261,3 +239,8 @@ func isOutsideCapable(r netlink.Route) bool {
 
 	return false
 }
+
+// isPhysical - checks if the interface has a device attached to it
+func isPhysical(iface string) bool {
+	return sysDepsImpl.FileExists(filepath.Join("/sys/class/net", iface, "device"))
```
DoYouEvenCpp: Can't we use here directly internal.FileExists instead?

DoYouEvenCpp (approved):

PR #1324, opened 2026-01-29
LVPN-9054: Change library permissions

edvyno1:


ci/nfpm/template.yaml
```diff
@@ -132,8 +137,16 @@ contents:
...
-      mode: 0755
+    # File info for tree type sets permission level for whole
+    # dst directory, if this is ran as file info 0644
+    # the directory itself will not have executable permissions
+    # and the user process of norduserd will not be able to be run
+
+    # Ensure that libraries are packaged in with proper permissions
+    # during build/download stage
+
+    # file_info:
```
keliramu: Is this part of doc? or some leftover?

edvyno1: I kept it as a reminder to not use file_info unless it is handled properly, but I can just keep the documentation text and get rid of the commented yaml code

thegravi (approved):

PR #1325, opened 2026-02-02
Fix failing fileshare process management tests

bartoszWojciechO:
Update the tests to match for new comments when checking if fileshare is blocked/unblocked.

### DIFF
--- a/test/qa/lib/fileshare.py
+++ b/test/qa/lib/fileshare.py
@@ -436,12 +436,16 @@ def bind_port() -> socket.socket | None:
     return None
 
 
-def port_is_allowed(peer_address: str) -> bool:
+def check_fileshare_port_status(is_allowed: bool, peer_address: str) -> bool:
     for _ in range(3):
-        if is_port_allowed(peer_address):
+        if is_port_allowed(peer_address) == is_allowed:
             return True
         time.sleep(1)
     return False
+
+
+def port_is_allowed(peer_address: str) -> bool:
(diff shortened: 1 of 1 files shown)


test/qa/lib/fileshare.py
```diff
@@ -436,22 +436,22 @@ def bind_port() -> socket.socket | None:
...
 
-def is_port_allowed() -> bool:
+def is_port_allowed(peer_address: str) -> bool:
     rules = os.popen("sudo iptables -S").read()
-    return "49111 -m comment --comment nordvpn-meshnet -j ACCEPT" in rules
+    return f"49111 -m comment --comment nordvpn-meshnet -m comment --comment \"-allow-fileshare-rule-{peer_address}\" -j ACCEPT" in rules
 
 
-def port_is_blocked() -> bool:
+def port_is_blocked(peer_address: str) -> bool:
```
mariusSincovici, line 456: Is it really needed 2 functions that do the same? 😄 
Isn't this just `not port_is_allowed`?

edvyno1 (approved):

test/qa/lib/fileshare.py
bartoszWojciechO: I have fixed it by extracting the loop out into a common function that can be called with a parameter. Kept `port_is_blocked`/`port_is_allowed` functions to be called from the outside because it's more readable than `check_fileshare_port_status(True/False)`

PR #1328, opened 2026-02-04
LVPN-9762: dns debug context

bartoszWojciechO:


daemon/dns/dns.go
```diff
@@ -160,6 +176,7 @@ func (d *DNSServiceSetter) set(setter Setter, iface string, nameservers []string
 //  3. resolv.conf utility
 //  4. direct write to resovl.conf
```
thegravi: [nit] typo (pre-existing)

daemon/dns/dns.go
thegravi: [nit] type (pre-existing)

daemon/dns/analytics.go
```diff
@@ -4,63 +4,98 @@ import (
...
+}
+
 func (d *dnsAnalytics) emitResolvConfOverwrittenEvent() {
-	debuggerEvent := buildDebuggerEvent(resolvConfOverwritten)
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	debuggerEvent := buildDebuggerEvent(resolvConfOverwritten, d.managementService)
+
+	log.Printf("%s%s publishing event: %+v", internal.DebugPrefix, dnsPrefix, debuggerEvent)
```
keliramu, line 77: New line `\n` is missing, or `log` adds new line automatically?

daemon/dns/analytics.go
```diff
@@ -4,63 +4,98 @@ import (
...
 )
 
 func (e eventType) String() string {
 	switch e {
 	case resolvConfOverwritten:
 		return "resolvconf_overwritten"
+	case dnsConfigured:
+		return "dns_configured"
 	default:
 		return fmt.Sprintf("%d", e)
```
keliramu: Maybe need some keyword `undefined, id:` otherwise if with new type we missed to add new case here in analytics we would start seeing obscure number without meaning

keliramu (approved):

daemon/dns/analytics.go
thegravi, line 77: [small] Missing new line

bartoszWojciechO: `log.Printf` adds new line by default.

bartoszWojciechO: As you said, new line is added automatically.

daemon/dns/dns.go
```diff
@@ -151,15 +165,18 @@ func (d *DNSServiceSetter) set(setter Setter, iface string, nameservers []string
 	}
 
 	d.unsetter = setter
+	d.analytics.emitDNSConfiguredEvent()
```
DoYouEvenCpp, line 168: Don't we want to collect also info about failed set dns procedures?

daemon/dns/analytics.go
```diff
@@ -4,63 +4,98 @@ import (
...
 // analytics provides an interface for sending DNS related debugger events
 type analytics interface {
+	setManagementService(dnsManagementService)
 	emitResolvConfOverwrittenEvent()
+	emitDNSConfiguredEvent()
 }
 
 type dnsAnalytics struct {
-	debugPublisher events.Publisher[events.DebuggerEvent]
+	mu                sync.Mutex
```
DoYouEvenCpp, line 51: I think this is a bit of a code smell here. Maybe we should keep analytics stateless, and provide management service upon every event emit request?

What do you think about such a solution?

daemon/dns/resolvconf_monitor.go
```diff
@@ -55,7 +55,7 @@ func (r *resolvConfFileWatcherMonitor) monitorResolvConf(ctx context.Context, do
 				return fmt.Errorf("file watcher closed")
 			}
 
-			if e.Op == fsnotify.Write || e.Op == fsnotify.Remove {
+			if e.Op.Has(fsnotify.Write) || e.Op.Has(fsnotify.Remove) {
 				r.analytics.emitResolvConfOverwrittenEvent()
```
DoYouEvenCpp, line 59: Now since the analytics keeps also the information about the currently activated management service, isn't any potential synchronisation mechanism missing since potentially there's a scenario possible:
- dns sets management service
- dns records got updated, but the OS parses for a longer while updates to the resolve.conf
- user changes their mind and updates dns records again
  - the management service got updated
- OS finally flushes updates to the resolve.conf, but the event is emitted with updated management service introducing data inconsistency

daemon/dns/dns.go
DoYouEvenCpp, line 194: Is OK to start the monitor only when `setUsingBestAvailable` is actually called?

daemon/dns/dns.go
bartoszWojciechO: Coming in next PR

daemon/dns/dns.go
bartoszWojciechO: Yeah, we only care about `resolv.conf` overwrites if DNS is unmanaged.

daemon/dns/analytics.go
bartoszWojciechO: We can store the state in the outer layer and pass it in the function call as you say, with such approach we should be able to get rid of this extra synchronization. I'll fix it in next PR if that's ok, would be much simpler with merge conflicts.

daemon/dns/resolvconf_monitor.go
bartoszWojciechO: Added a preemptive stop to the resolv.conf monitor when setting the DNS. This should resolve the issue and prevent any other issues, the file should not be monitored when the config is being applied.

daemon/dns/analytics.go
DoYouEvenCpp: Sure, sounds like a good plan.

PR #1336, opened 2026-02-11
LVPN-9923: TP active devices data is inaccurate

devzbysiu:
Context:
- TP Lite metric is taken from `UserPreferences` field in the context
- it has to be set to either `false` or `true`, can't be `null`
- before the change, this value is set in the context only when user toggles the setting, if user doesn't change it, it will be `null`

Changes:
- introduce `mooseSetTPLiteUserPrefFunc` type and setter
- move `mooseXYZ` functions initialization to constructor - otherwise, we can get Nil Pointer Dereference, when event was published, but `Init` was not yet called
- set TP Lite User Preference in context during moose initialization - so it's always either `true` or `false`, but never `null`
- change order of setting TP Lite in `User Preferences` vs `Current State` in `NotifyThreatProtectionLite` - to avoid early exit when setting `Current State` fails (`User Preferences` should always be set)
- setting TP Lite to on unsets Custom DNS and vice-versa
- add comments explaining `User Preferences` vs `Current State` use
- formatting

### DIFF
--- a/events/moose/moose.go
+++ b/events/moose/moose.go
@@ -133,7 +133,7 @@ func (s *Subscriber) getConfig() (config.Config, error) {
 func (s *Subscriber) changeConsentState(newState config.AnalyticsConsent) error {
 	cfg, err := s.getConfig()
 	if err != nil {
-		return err
+		return fmt.Errorf("loading config for consent state change: %w", err)
 	}
 
 	// the same state requested, no-op
@@ -149,7 +149,7 @@ func (s *Subscriber) changeConsentState(newState config.AnalyticsConsent) error 
 
 	log.Println(internal.InfoPrefix, LogComponentPrefix, "update consent level into context with new value", newState.String())
 	if err := setUserConsentLevelIntoContext(s, newState); err != nil {
-		return err
+		return fmt.Errorf("updating consent level in moose context: %w", err)
 	}
 	s.canSendAllEvents.Store(newState == config.ConsentGranted)
 
@@ -200,7 +200,7 @@ func (s *Subscriber) Init(consent config.AnalyticsConsent) error {
 
 	cfg, err := s.getConfig()
 	if err != nil {
-		return err
+		return fmt.Errorf("loading config during Init: %w", err)
 	}
 
 	err = s.updateEventDomain()
@@ -266,7 +266,7 @@ func (s *Subscriber) Init(consent config.AnalyticsConsent) error {
 	}
 
 	if err := setUserConsentLevelIntoContext(s, consent); err != nil {
-		return err
+		return fmt.Errorf("setting user consent level during Init: %w", err)
 	}
 
 	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappVersion(s.buildTarget.Version)); err != nil {
@@ -344,7 +344,10 @@ func (s *Subscriber) Stop() error {
 }
 
 func (s *Subscriber) NotifyKillswitch(data bool) error {
-	return s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesKillSwitchEnabledValue(data))
+	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesKillSwitchEnabledValue(data)); err != nil {
+		return fmt.Errorf("setting kill switch preference (enabled=%v): %w", data, err)
+	}
+	return nil
 }
 
 func (s *Subscriber) NotifyAccountCheck(any) error {
@@ -352,7 +355,10 @@ func (s *Subscriber) NotifyAccountCheck(any) error {
 }
 
 func (s *Subscriber) NotifyAutoconnect(data bool) error {
-	return s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesAutoConnectEnabledValue(data))
+	if err := s.response(moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesAutoConnectEnabledValue(data)); err != nil {
+		return fmt.Errorf("setting auto-connect preference (enabled=%v): %w", data, err)
+	}
+	return nil
 }
 
 func (s *Subscriber) NotifyDefaults(any) error {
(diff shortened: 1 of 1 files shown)


events/moose/moose.go
```diff
@@ -92,8 +96,12 @@ func NewSubscriber(
 		clientAPI:     clientAPI,
 		httpClient:    httpClient,
 		isInitialized: false,
-	}
 
+		mooseConsentLevelFunc:      moose.MooseNordvpnappSetConsentLevel,
+		mooseSetConsentIntoCtxFunc: moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesConsentLevel,
+		mooseSetTokenRenewDateFunc: moose.NordvpnappSetContextApplicationNordvpnappConfigCurrentStateTokenRenewDateValue,
+		mooseSetTPLiteUserPrefFunc: moose.NordvpnappSetContextApplicationNordvpnappConfigUserPreferencesThreatProtectionLiteEnabledValue,
```
thegravi: Since you introduced function pointer please add tests, otherwise we can directly use moose function.

DoYouEvenCpp: Have you considered extracting those function pointers under an intermediate member that would aggregate all of them?
This way we would group all of them in a single place, keep the number of Subscriber members sane. and group logically member fields of similar type and functionality. Only drawback is we'll have to use this middleware member (like `s.mooseFuncs.xxx`)
WDYT?

devzbysiu: Yeah, we can go this way.

TBH I have an urge to introduce separate interface with **all** moose stuff there and inject the implementation to constructor of subscriber.. but won't go this way now, we would need to have regression testing for moose part which was not planned

devzbysiu: Put it in a separate PR https://github.com/NordSecurity/nordvpn-linux/pull/1340/changes?w=1

devzbysiu: Added tests

events/moose/moose.go
mariusSincovici, line 368: Shouldn't context value be unset when custom DNS is set to something?

devzbysiu: Addressed. Now enabling CustomDNS unsets TP Lite and enabling TP Lite unsets Custom DNS

events/moose/moose.go
```diff
@@ -551,13 +587,35 @@ func (s *Subscriber) NotifyPeerUpdate([]string) error { return nil }
 
 func (s *Subscriber) NotifySelfRemoved(any) error { return nil }
 
-func (s *Subscriber) NotifyThreatProtectionLite(data bool) error {
+func (s *Subscriber) NotifyThreatProtectionLite(isTPLiteEnabled bool) error {
+	if err := s.setTPLite(isTPLiteEnabled); err != nil {
```
keliramu: Line #321 does not swallow error when doing the same thing as here, why error here does not break out with error?

events/moose/moose.go
```diff
@@ -551,13 +587,35 @@ func (s *Subscriber) NotifyPeerUpdate([]string) error { return nil }
...
+		return err
+	}
+
+	// TP Lite is not compatible with custom DNS - if TP Lite is on, Custom DNS should be off
+	if isTPLiteEnabled {
+		disable := events.DataDNS{} // empty custom DNS
```
keliramu, line 642: This line and new variable is redundant, all the same could be done in single line on the following line

events/moose/moose_test.go
```diff
@@ -402,3 +402,343 @@ func TestHandleTokenRenewDateChange(t *testing.T) {
...
 }
+
+func TestNotifyThreatProtectionLite_CallsUserPreferenceSetter(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	tests := []struct {
+		name                      string
+		enabled                   bool
+		mooseUserPrefErrCode      uint32
+		mooseCurrentStateErrCode  uint32
```
keliramu: variable is not used, consider cleanup?

events/moose/moose.go
```diff
@@ -551,13 +587,35 @@ func (s *Subscriber) NotifyPeerUpdate([]string) error { return nil }
...
+	}
+
+	return nil
+}
+
+func (s *Subscriber) setTPLite(isTPLiteEnabled bool) error {
+	// User Preferences field in moose context is used to see what's the setting
+	// user selected - no matter if VPN is actively used or not.
+	if err := s.response(s.mooseSetTPLiteUserPrefFunc(isTPLiteEnabled)); err != nil {
+		log.Println(internal.WarningPrefix, "failed to set TP Lite in User Preferences:", err)
```
keliramu, line 656: Line #321 does not swallow error when doing the same thing as here, why error here does not break out with error?

events/moose/moose.go
devzbysiu: It adds semantic value. Otherwise we'll have `s.setCustomDNS(events.DataDNS{})` - it's not clear what it means without looking on method body (i.e.: is it setting custom DNS to empty? Or to defaults? Or disables? etc.)

events/moose/moose.go
devzbysiu: Addressed. It was done to not break early, because this block and block below provides different info. But yes, we loose the error this way. Now I'm joining errors  and returning at the end

events/moose/moose_test.go
devzbysiu: Good catch, done

events/moose/moose.go
```diff
@@ -340,10 +360,26 @@ func (s *Subscriber) NotifyDefaults(any) error {
...
+		return s.setTPLite(false)
+	}
+
+	return nil
+}
+
+func (s *Subscriber) setCustomDNS(data events.DataDNS) error {
+	dnsIPCount := len(data.Ips)
+	if err := s.response(s.mooseSetCustomDNSMetaFunc(fmt.Sprintf(`{"count":%d}`, dnsIPCount))); err != nil {
+		return err
```
bartoszWojciechO: Can you add error wrapping somewhere along the line here and in the other cases?

bartoszWojciechO (approved):

events/moose/moose.go
devzbysiu: Addressed

PR #1337, opened 2026-02-11
LVPN-9757: add DNS configuration error events

bartoszWojciechO:


daemon/dns/analytics.go
```diff
@@ -4,120 +4,210 @@ import (
...
+		WithGlobalContextPaths(globalPaths...)
+
+	return debuggerEvent
+}
+
+type errorEvent struct {
+	event
+	ErrorType string `json:"error_type"`
+	// Critical should be set to true if the given error prevents DNS configuration
+	Critical bool `json:"cricital"`
```
thegravi:
```suggestion
	Critical bool `json:"critical"`
```

daemon/dns/analytics_test.go
```diff
@@ -71,3 +71,97 @@ func Test_emitDNSConfiguredEvent(t *testing.T) {
...
+				Path:  debuggerEventBaseKey + "." + debuggerEventErrorTypeKey,
+				Value: test.errorType.String(),
+			})
+			assert.Contains(t, event.KeyBasedContextPaths, events.ContextValue{
+				Path:  debuggerEventBaseKey + "." + debuggerEventCriticalKey,
+				Value: test.critical,
+			})
+
+			expectedJson :=
+				fmt.Sprintf("{\"event\":\"dns_configuration_error\",\"namespace\":\"nordvpn-linux\",\"management_service\":\"%s\",\"error_type\":\"%s\",\"cricital\":%s}",
```
thegravi:
```suggestion
				fmt.Sprintf("{\"event\":\"dns_configuration_error\",\"namespace\":\"nordvpn-linux\",\"management_service\":\"%s\",\"error_type\":\"%s\",\"critical\":%s}",
```

daemon/dns/dns.go
```diff
@@ -46,7 +47,13 @@ func (e dnsManagementService) String() string {
 	}
 }
 
-var ErrDNSNotSet = errors.New("DNS unsetter not set")
+var (
+	ErrDNSNotSet = errors.New("DNS unsetter not set")
+	// errDNSSetFailed is returned when all DNS set methods fail
+	errDNSSetFailed = errors.New("all dns set attempts failed")
+	// errDNSSetFailedNoBinaries is returned when no binaries necessary to set DNS were found
+	errDNSSetFailedNoBinaries = errors.New("no binaries to set DNS with this method wre found")
```
mariusSincovici: spelling `wre`

daemon/dns/analytics.go
```diff
@@ -4,120 +4,210 @@ import (
...
+	}
+}
+
 type analytics interface {
-	setManagementService(dnsManagementService)
-	emitResolvConfOverwrittenEvent()
-	emitDNSConfiguredEvent()
+	emitResolvConfOverwrittenEvent(managementService dnsManagementService)
+	emitDNSConfiguredEvent(managementService dnsManagementService)
+	emitDNSConfigurationErrorEvent(managementService dnsManagementService, errorType errorType, critical bool)
```
DoYouEvenCpp: I'd split this method into two separate ones:
- `emitDNSConfigurationErrorEvent`
- `emitCriticalDNSConfigurationErrorEvent`

And get rid completely of isCritical boolean argument.
This will make the code self-documented and less prone to mistakes.
This indeed needs to call `newErrorEvent` under the hood with proper `isCritical` paramater, but I think this should be rather an implementation detail, hidden from the users of this interface.

daemon/dns/analytics_test.go
```diff
@@ -71,3 +71,97 @@ func Test_emitDNSConfiguredEvent(t *testing.T) {
 		})
 	}
 }
+
+func Test_emidDNSConfigurationErrorEvent(t *testing.T) {
```
DoYouEvenCpp: [nit[ a typo:
```suggestion
func Test_emitDNSConfigurationErrorEvent(t *testing.T) {
```

daemon/dns/analytics_test.go
```diff
@@ -71,3 +71,97 @@ func Test_emitDNSConfiguredEvent(t *testing.T) {
...
+		},
+		{
+			name:              "failed to read for systemd-resolved, critical",
+			managementService: systemdResolved,
+			errorType:         resolvConfReadFailedErrorType,
+			critical:          true,
+		},
+	}
+
+	for _, test := range tests {
```
DoYouEvenCpp, line 139: Can you please assign to this a test category? (same also for other tests in this suite, please!)

daemon/dns/dns.go
```diff
@@ -46,7 +47,13 @@ func (e dnsManagementService) String() string {
 	}
 }
 
-var ErrDNSNotSet = errors.New("DNS unsetter not set")
+var (
+	ErrDNSNotSet = errors.New("DNS unsetter not set")
```
DoYouEvenCpp: `ErrDNSNotSet` is a bit of a misnomer here - at least compared to context it's used for. What about `ErrDNSMissingUnsetter`?
Can you also provide documentation for this error?

daemon/dns/dns.go
```diff
@@ -46,7 +47,13 @@ func (e dnsManagementService) String() string {
...
+	// errDNSSetFailed is returned when all DNS set methods fail
+	errDNSSetFailed = errors.New("all dns set attempts failed")
```
DoYouEvenCpp: Is there any other place where those two errors could be actually used? (currently it's just a single use)

bartoszWojciechO: It's used in a single place, but we need this error assigned to a variable so that it can be recognized by the caller(so that caller will be able to send a proper event).

daemon/dns/analytics.go
```diff
@@ -4,120 +4,221 @@ import (
...
+			Path:  debuggerEventBaseKey + "." + debuggerEventManagementServiceKey,
+			Value: e.ManagementService,
+		},
+	}
+}
+
+func (e event) toDebuggerEvent() *events.DebuggerEvent {
+	jsonData, err := json.Marshal(e)
+	if err != nil {
+		log.Println(internal.DebugPrefix, dnsPrefix, "failed to serialize event json for resolv.conf overwrite:", err)
```
DoYouEvenCpp: In such a case, I guess we can safely emit a warning instead of a debug

daemon/dns/analytics.go
```diff
@@ -4,120 +4,221 @@ import (
...
+		},
+	}
+	contextPaths = append(contextPaths, e.event.toContextPaths()...)
+	return contextPaths
+}
+
+func (e errorEvent) toDebuggerEvent() *events.DebuggerEvent {
+	jsonData, err := json.Marshal(e)
+	if err != nil {
+		log.Println(internal.DebugPrefix,
```
DoYouEvenCpp:
```suggestion
		log.Println(internal.WarningPrefix,
```

daemon/dns/analytics.go
```diff
@@ -4,120 +4,221 @@ import (
...
+		log.Println(internal.DebugPrefix, dnsPrefix, "failed to serialize event json for resolv.conf overwrite:", err)
+		jsonData = []byte("{}")
+	}
+
+	debuggerEvent := events.NewDebuggerEvent(string(jsonData)).
+		WithKeyBasedContextPaths(e.toContextPaths()...).
+		WithGlobalContextPaths(globalPaths...)
+
+	return debuggerEvent
+}
```
DoYouEvenCpp, line 66: This is now almost 1:1 duplicated code of `toDebuggerEvent` implemented for `errorEvent`.
Can we maybe unify it, like:
- introduce an intermediary interface 
```
type debuggerEventConverter interface {
	toContextPaths() []events.ContextValue
}
```


- then implement only one version of `toDebuggerEvent`

```
func toDebuggerEvent(e debuggerEventConverter) *events.DebuggerEvent {
	jsonData, err := json.Marshal(e)
	if err != nil {
		log.Println(internal.DebugPrefix, dnsPrefix, "failed to serialize event json:", err)
		jsonData = []byte("{}")
	}

	debuggerEvent := events.NewDebuggerEvent(string(jsonData)).
		WithKeyBasedContextPaths(e.toContextPaths()...).
		WithGlobalContextPaths(globalPaths...)

	return debuggerEvent
}
```
- and finally _slightly_ modify usage of how we emit events:
```
debuggerEvent := toDebuggerEvent(newEvent(resolvConfOverwrittenEventType,
		managementService))
```



Pros:
- we keep only a single implementation of `toDebugger...` method, thus no DRY rule violation
- all the error types events already implement `toContextPaths` so no extra implementation is required, yet we gain type safety here as well
- this will be more extensible, with a new event type is now only a matter of providing an implementation for `toContextPaths`

Cons:
- we don emit a diagnostics log upon failure that contains specific information what very event conversion failed

daemon/dns/analytics.go
```diff
@@ -4,120 +4,221 @@ import (
...
+		managementService).toDebuggerEvent()
 
-	d.managementService = managementService
-}
+	log.Printf("%s%s publishing event: %+v", internal.DebugPrefix, dnsPrefix, debuggerEvent)
 
-func (d *dnsAnalytics) emitResolvConfOverwrittenEvent() {
-	d.mu.Lock()
-	defer d.mu.Unlock()
+	d.debugPublisher.Publish(*debuggerEvent)
```
DoYouEvenCpp: [nit] this is also some repeated code, in the methods responsible for events being emitted.

We can extract this into:
```
func (d *dnsAnalytics) publishEvent(e debuggerEventConverter) {
	debuggerEvent := toDebuggerEvent(e)
	log.Printf("%s%s publishing event: %+v", internal.DebugPrefix, dnsPrefix, debuggerEvent)
	d.debugPublisher.Publish(*debuggerEvent)
}
```

and then use it:
```
func (d *dnsAnalytics) emitResolvConfOverwrittenEvent(managementService dnsManagementService) {
	d.publishEvent(newEvent(resolvConfOverwrittenEventType, managementService))
}
```

daemon/dns/dns.go
```diff
@@ -46,7 +47,13 @@ func (e dnsManagementService) String() string {
 	}
 }
 
-var ErrDNSNotSet = errors.New("DNS unsetter not set")
+var (
+	errDNSMissingUnsetter = errors.New("DNS unsetter not set")
+	// errDNSSetFailed is returned when all DNS set methods fail
+	errDNSSetFailed = errors.New("all dns set attempts failed")
+	// errDNSSetFailedNoBinaries is returned when no binaries necessary to set DNS were found
+	errDNSSetFailedNoBinaries = errors.New("no binaries to set DNS with this method wre found")
```
DoYouEvenCpp: [nit] typo: `wre` (should have been `were` right?)

daemon/dns/dns.go
```diff
@@ -274,16 +293,24 @@ func (d *DNSMethodSetter) Set(iface string, nameservers []string) error {
 		return errors.New("nameservers not provided")
 	}
 
+	binariesAvailable := false
 	for _, method := range d.methods {
 		d.publisher.Publish("set dns for interface [" + iface + "] using: " + method.Name())
 		if err := method.Set(iface, nameservers); err != nil {
+			if !errors.Is(err, exec.ErrNotFound) {
+				binariesAvailable = true
```
DoYouEvenCpp, line 343: Correct me if I'm wrong, but I guess this won't work well for a scenario when first setter fails with something else than  `binary unavailable` and then following method can't bring `binariesAvailable` back to a `false` value, which eventual might shadow actual issue?

bartoszWojciechO: If theres a single binary available on the system that would facilitate a method available on the system, then we wouldn't want to bring the value back to false, even if that method failed for some other reason. We only want _binaries unavailable_ kind of events if no binaries at all were available for a method. We need this event primarily for `NetworkManager` method, to determine if dbus interface needs an implementation(if there are any cases at all where NM is used to manage DNS but `nmcli` is not available on the OS).

DoYouEvenCpp: OK, this approach makes sense.
Thanks!

daemon/dns/analytics.go
bartoszWojciechO: Maybe I'm missing something, but in order to facilitate this approach we'd need an interface for `toContextPaths` a generic method(or a method that takes `any`) for common `toDebuggerEvent`. As this is a very simple constructor/cast method and we don't anticipate having more event types at this point, imo it's a worthy tradeoff to have a slightly less dry code for less complexity. We can revisit this if more event types would need to be added. Do you agree?

DoYouEvenCpp: Sure, let's keep it as is for now.

PR #1344, opened 2026-02-17
LVPN-9955: retry autoconnect when server is not available

bartoszWojciechO:
Added more granular server list errors so that autoconnect job could handle them.
1. DataManager now stores the state of the server data. Caller can check if data was loaded from disk or fetched from the backend api. Server selection returns an error if server data is not ready.
2. If selected server is not available and servers data is outdated, autoconnect will keep retrying to connect. This is done because server might actually be available in an up to date list.
3. If selected server is not available and servers data is up to date, autoconnect will not reattept connection.

### DIFF
--- a/daemon/jobs.go
+++ b/daemon/jobs.go
@@ -273,6 +273,8 @@ type autoconnectServer struct {
 	err error
 }
 
+var errServersUnavailable = errors.New("servers unavailable")
+
 func (autoconnectServer) SetHeader(metadata.MD) error  { return nil }
 func (autoconnectServer) SendHeader(metadata.MD) error { return nil }
 func (autoconnectServer) SetTrailer(metadata.MD)       {}
@@ -284,7 +286,7 @@ func (a *autoconnectServer) Send(data *pb.Payload) error {
 	case internal.CodeFailure:
 		a.err = errors.New("autoconnect failure")
 	case internal.CodeServerUnavailable:
-		a.err = errors.New("servers list not ready")
+		a.err = errServersUnavailable
 	}
 	return nil
 }
@@ -320,10 +322,10 @@ func (r *RPC) StartAutoConnect(timeoutFn network.CalculateRetryDelayForAttempt) 
 			return nil
 		}
 
-		if err := r.doAutoConnect(); err == nil {
-			return nil
-		} else {
+		if err := r.doAutoConnect(); err != nil {
 			log.Println(internal.ErrorPrefix, "autoconnect failed:", err)
+		} else {
+			return nil
 		}
 		tryAfterDuration := timeoutFn(tries)
 		tries++
--- a/daemon/rpc_set_autoconnect.go
+++ b/daemon/rpc_set_autoconnect.go
@@ -58,8 +58,6 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest) 
 
 		serverSelection, err := selectServer(r, &insights, cfg, serverTag, serverGroup)
 		if err != nil {
-			// convert ErrServerDataIsNotReady to ErrServerIsUnavailable. ErrServerDataIsNotReady is used internally and
-			// it's not supposed to be seen by the users.
 			log.Println(internal.ErrorPrefix, "no server found for autoconnect", serverTag, err)
 
 			var errorCode *internal.ErrorWithCode
--- a/daemon/jobs_test.go
+++ b/daemon/jobs_test.go
@@ -90,6 +90,25 @@ func TestStartAutoConnect(t *testing.T) {
 	}
 }
 
+func TestDoAutoconnectHandlesServerAvailabilityIssues(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	rpc := testRPC()
+	rpc.serversAPI = mockFailingServersAPI{}
+	mockConfigManager := newMockConfigManager()
+	updateAutoconnectData(mockConfigManager, config.AutoConnectData{Country: "DE"})
+	rpc.cm = mockConfigManager
+
+	rpc.dm.SetServersData(time.Now(), []core.Server{}, "")
+
+	err := rpc.doAutoConnect()
+	assert.ErrorIs(t, err, errServersUnavailable, "doAutoconnect has ignored server availability errors")
+
+	rpc.dm.SetServersData(time.Now(), serversList(), "")
+	err = rpc.doAutoConnect()
+	assert.Nil(t, err, "unexpected error returned by doAutoconnect")
+}
+
 func TestDoAutoConnect(t *testing.T) {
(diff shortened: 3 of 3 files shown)


daemon/data_models.go
```diff
@@ -104,24 +105,29 @@ type ServersData struct {
...
 	err := encoder.Encode(data)
 	if err != nil {
-		return err
+		return fmt.Errorf("encoding data: %w", err)
 	}
 
 	err = internal.FileWrite(data.filePath, buffer.Bytes(), internal.PermUserRWGroupROthersR)
 	if err != nil {
-		return err
+		return fmt.Errorf("writign data: %w", err)
```
keliramu: Typo `writign` -> `writing`

daemon/data_manager.go
```diff
@@ -146,7 +157,14 @@ func (dm *DataManager) SetServersData(updatedAt time.Time, servers core.Servers,
 		Hash:      hash,
 		filePath:  dm.serversData.filePath,
 	}
-	return dm.serversData.save()
+	log.Println(internal.InfoPrefix, "saving new servers data")
+	err = dm.serversData.save()
+	if err != nil {
+		return fmt.Errorf("saving servers data: %W", err)
```
keliramu: Typo? `W` -> `w`

cmd/daemon/main.go
```diff
@@ -710,11 +710,23 @@ func main() {
...
+		} else {
+			log.Println(internal.InfoPrefix, "data successfully loaded from disk")
 		}
+		close(dataLoadChan)
 	}()
+
+	select {
+	case <-dataLoadChan:
+		log.Println(internal.InfoPrefix, "data loaded in expected time")
+	case <-time.After(10 * time.Second):
```
keliramu: Why 10s? above `dm.LoadData()` should be local disk read and should run in less than one/few seconds usually

keliramu: Above goroutine for dm.LoadData() then makes no sense if we block waiting for its results right after it...

bartoszWojciechO: Leftover from draft stage, will remove in next commit.

cmd/daemon/main.go
```diff
@@ -712,9 +712,12 @@ func main() {
 
 	go func() {
 		if err := dm.LoadData(); err != nil {
-			log.Println(internal.WarningPrefix, err)
+			log.Println(internal.WarningPrefix, "loading data:", err)
```
DoYouEvenCpp: Maybe we could add more information into this diagnostic, like:
```suggestion
			log.Println(internal.WarningPrefix, "loading data from DataManager failed:", err)
```

?

daemon/data_manager.go
```diff
@@ -62,6 +63,8 @@ func (dm *DataManager) LoadData() error {
 	}
 	if err := dm.serversData.load(); err != nil {
 		errs = append(errs, fmt.Errorf("loading servers data: %w", err))
+	} else {
+		dm.serversDataReady = true
```
DoYouEvenCpp: Maybe we could emit here a log(info maybe) that server data is actually ready?

daemon/data_manager.go
```diff
@@ -146,7 +157,14 @@ func (dm *DataManager) SetServersData(updatedAt time.Time, servers core.Servers,
...
 		filePath:  dm.serversData.filePath,
 	}
-	return dm.serversData.save()
+	log.Println(internal.InfoPrefix, "saving new servers data")
+	err = dm.serversData.save()
+	if err != nil {
+		return fmt.Errorf("saving servers data: %w", err)
+	}
+
+	dm.serversDataReady = true
```
DoYouEvenCpp: Shouldn't we set this to false when `save` right above fails?

daemon/rpc_set_autoconnect_test.go
```diff
@@ -284,3 +285,14 @@ func TestAutoconnect_SavesCorrectAutoconnectData(t *testing.T) {
...
+
+func TestAutoconnect_HandleServersDataIsNotReady(t *testing.T) {
+	category.Set(t, category.Unit)
+	mockAuthChecker := mockAutoconnectAuthChecker{}
+	mockConfigManager := newMockConfigManager()
+	dm := DataManager{serversData: ServersData{Servers: serversList()}, serversDataReady: false}
+	r := RPC{cm: mockConfigManager, ac: mockAuthChecker, dm: &dm}
+	request := pb.SetAutoconnectRequest{Enabled: true}
+	_, err := r.SetAutoConnect(context.Background(), &request)
+	assert.Equal(t, err.Error(), internal.ServerUnavailableErrorMessage)
```
DoYouEvenCpp: [nit] using `EqualError` might be more verbose regarding the expected output here.

```suggestion
	assert.EqualError(t, err, internal.ServerUnavailableErrorMessage)
```

daemon/data_manager.go
bartoszWojciechO: This is just a minimal change for the autoconnect the startup fix. I think we should keep the state bookeeping as simple as possible and plan a proper fix/refactor here.

devzbysiu (approved):

daemon/rpc_set_autoconnect.go
```diff
@@ -58,6 +58,8 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 
 		serverSelection, err := selectServer(r, &insights, cfg, serverTag, serverGroup)
 		if err != nil {
+			// convert ErrServerDataIsNotReady to ErrServerIsUnavailable. ErrServerDataIsNotReady is used internally and
+			// it's not supposed to be seen by the users.
```
thegravi: Stale comment?

daemon/jobs.go
```diff
@@ -295,6 +322,8 @@ func (r *RPC) StartAutoConnect(timeoutFn network.CalculateRetryDelayForAttempt)
 
 		if err := r.doAutoConnect(); err == nil {
 			return nil
+		} else {
+			log.Println(internal.ErrorPrefix, "autoconnect failed:", err)
 		}
```
thegravi: More go-idiomatic way would be to swap err check parts. 

```suggestion
  if err := r.doAutoConnect(); err != nil {
      log.Println(internal.ErrorPrefix, "autoconnect failed:", err)
  } else {
      return nil
  }
```

daemon/jobs.go
```diff
@@ -258,6 +283,8 @@ func (a *autoconnectServer) Send(data *pb.Payload) error {
 	switch data.GetType() {
 	case internal.CodeFailure:
 		a.err = errors.New("autoconnect failure")
+	case internal.CodeServerUnavailable:
+		a.err = errors.New("servers list not ready")
```
thegravi: Can you cover with UT this new error?

Image

PR #1345, opened 2026-02-17
LVPN-9904: set analytics debug field when DNS is managed by NetworkManager

DoYouEvenCpp:
Changes under this PR:
- set setManagementService to  `nmcliManaged` when DNS configuration is resolved  to be used with NetworkManager (see changes done to `setUsingAvailable` under `daemon/dns/dns.go`)
- addressed a [comment](https://github.com/NordSecurity/nordvpn-linux/pull/1305#discussion_r2735818672) from https://github.com/NordSecurity/nordvpn-linux/pull/1305 to move file system related operations to a common interface
- adjusted usage of fs mock across UTs

### DIFF
--- a/daemon/dns/dns.go
+++ b/daemon/dns/dns.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os/exec"
 	"strings"
 
 	"github.com/NordSecurity/nordvpn-linux/events"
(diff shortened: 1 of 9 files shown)


daemon/dns/dns.go
```diff
@@ -210,6 +189,7 @@ func (d *DNSServiceSetter) setUsingAvailable(iface string, nameservers []string)
 		log.Println(internal.WarningPrefix, dnsPrefix,
 			"failed to configure DNS using nmcli, attempting with resolv.conf, %w", err)
 	} else {
+		d.analytics.setManagementService(nmcliManaged)
```
bartoszWojciechO: Maybe it would be better to change all the fields in the enum to end with `ManagementService`? `nmcliManaged` seems out of step with other names, on the other hand we use `<field_name><enum_type>` as a convention throughout the app. `nmcliManagementService`, `systemdResolvedManagementService` etc.

bartoszWojciechO (approved):

PR #1347, opened 2026-02-18
LVPN-9636: Improve http response limit handling

thegravi:
- Apply HTTP response size limit (20MB) to compressed data before decompression            
  - Fixes issue where large server list responses (~31MB decompressed, ~1MB compressed) were
  incorrectly rejected

### DIFF
--- a/core/cdn_test.go
+++ b/core/cdn_test.go
@@ -151,8 +151,7 @@ func TestCDNAPI_Request_InvalidGzipData(t *testing.T) {
 	api := NewCDNAPI("test-agent", server.URL, &http.Client{}, response.NoopValidator{})
 	_, err := api.GetRemoteFile("/test")
 
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "gzip")
+	assert.ErrorContains(t, err, "gzip")
 }
 
 func TestCDNAPI_Request_CorruptedGzipData(t *testing.T) {
@@ -192,8 +191,7 @@ func TestCDNAPI_Request_OversizedResponse(t *testing.T) {
 	api := NewCDNAPI("test-agent", server.URL, &http.Client{}, response.NoopValidator{})
 	_, err := api.GetRemoteFile("/test")
 
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "max limit")
+	assert.ErrorContains(t, err, "max limit")
 }
 
 func TestCDNAPI_Request_ValidationFailure(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dfetti (approved):

core/core_test.go
```diff
@@ -258,3 +259,89 @@ func TestMaxBytes(t *testing.T) {
...
 }
+
+func TestSimpleAPI_SizeLimitAppliedToCompressedResponse(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	// Create a large repetitive string that compresses well
+	largeData := make([]byte, 1024*1024) // 1MB of data
+	for i := range largeData {
+		largeData[i] = 'A'
+	}
```
DoYouEvenCpp:
```suggestion
    largeData := bytes.Repeat([]byte{'A'}, 1024*1024) // 1MB of data
```

core/cdn_test.go
```diff
@@ -82,3 +86,237 @@ func setupCdnApi() (*CDNAPI, context.CancelFunc) {
...
+		_, _ = w.Write([]byte(`{"data": "test"}`))
+	}))
+	defer server.Close()
+
+	api := NewCDNAPI("test-agent", server.URL, &http.Client{}, failingValidator{})
+	_, err := api.GetRemoteFile("/test")
+
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "cdn api")
+	assert.Contains(t, err.Error(), "validation failed")
```
DoYouEvenCpp: This can be shortened to:
```suggestion
	assert.ErrorContains(t, err, "cdn api: validation failed")
```

DoYouEvenCpp: If you happen to go for that direction. please consider also altering other similarly constructed assertions in the newly added test cases

core/cdn_test.go
```diff
@@ -82,3 +86,237 @@ func setupCdnApi() (*CDNAPI, context.CancelFunc) {
...
+	_, err := api.GetRemoteFile("/test")
+
+	assert.Error(t, err)
+}
+
+func TestCDNAPI_Request_SizeLimitAppliesToCompressedData(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	// Create data that compresses well - large when decompressed, small when compressed
+	largeData := make([]byte, 1024*1024) // 1MB of repetitive data
```
DoYouEvenCpp: `bytes.Repeat([]byte{'X'}, 1024*1024)` also here?

core/cdn.go
```diff
@@ -66,38 +66,35 @@ func (api *CDNAPI) request(path, method string) (*CDNAPIResponse, error) {
...
+				return nil, err
 			}
-		} else {
-			if err = api.validator.Validate(resp.StatusCode, resp.Header, body); err != nil {
-				return nil, fmt.Errorf("cdn api: %w", err)
+			defer reader.Close()
+			body, err = io.ReadAll(reader)
+			if err != nil {
+				return nil, err
 			}
```
DoYouEvenCpp, line 92: This is now 1:1 exact copy of the bit we got for the `simple_client_api.go` - care to extract this to a method, and simply reuse it in those two places to avoid DRY violation?
(possibly should be a rather trivial extraction, but in case this imposes quite some modification let's drop it)

thegravi: That read+decompress block is identical, but it's surrounded by context differences (error prefixes, headers validation, return type). I'd leave it as it is :)

DoYouEvenCpp: Ack, reasonable :wink:

PR #1348, opened 2026-02-18
LVPN-9970: Fix calculateRulePriority logic

dfetti:
When only the local and main rules exist, `calculateRulePriority` fails. This change fixes that scenario and also handles cases where our rules can be inserted after the "from all lookup main" rule in the priority order.
    
Some edge cases described in the `calculateRulePriority` documentation are still not handled and will be addressed separately.

### DIFF
--- a/daemon/routes/iprule/iprule.go
+++ b/daemon/routes/iprule/iprule.go
@@ -226,7 +226,7 @@ func (r *Router) TableID() uint {
 	return r.tableID
 }
 
-// calculateRulePriority find out what priority id to use for Fwmark rule
+// calculateRulePriority find out what priority id to use for a new rule
 //
 // On some environments already existing IP rules can exist with very high priority
 // (close to 0) and our rules do not fit into ip rules table. E. g.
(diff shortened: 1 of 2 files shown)


daemon/routes/iprule/iprule.go
```diff
@@ -258,25 +260,27 @@ func calculateRulePriority() (uint, error) {
 	// 32767:	from all lookup default
 	// # EXPECTED RESULT: 32766 - 1 = 32765 (check if priority/id is not in use)
 
-	var prioID uint
-
 	rules, err := netlink.RuleList(netlink.FAMILY_V4)
 	if err != nil {
 		return 0, fmt.Errorf("listing ip rules: %w", err)
 	}
+	return findRulePriorityCandidate(rules)
```
mariusSincovici, line 267: Would it be possible to log a message with the calculated priority?

mariusSincovici (approved):

daemon/routes/iprule/iprule.go
```diff
@@ -481,6 +485,19 @@ func isRuleSame(rule netlink.Rule, target netlink.Rule) bool {
 			rule.SuppressPrefixlen == target.SuppressPrefixlen)
 }
 
+// isFromAllLookup checks if the rule is the "from all lookup main"
```
thegravi: [small] Incorrect function name in docstring

thegravi (approved):

PR #1354, opened 2026-02-23
LVPN-9811: Use EnableECH feature in NordWhisper

devzbysiu:
@dfetti update:
The feature setting is fetched from the remote config. If fetching fails, we default it to true.

I tried to keep as much as I could from the libtelio approach to make it easier to add features in the future. There are no tests yet. I just tested it manually.

### DIFF
--- a/cmd/daemon/vpn_factory.go
+++ b/cmd/daemon/vpn_factory.go
@@ -13,17 +13,23 @@ import (
 
 var ErrNordWhisperDisabled = errors.New("NordWhisper technology was disabled in compile time")
 
-func getVpnFactory(eventsDbPath string, fwmark uint32, envIsDev bool,
-	cfg vpn.LibConfigGetter, appVersion string, eventsPublisher *vpn.Events,
+func getVpnFactory(
+	eventsDbPath string,
+	fwmark uint32,
+	envIsDev bool,
+	libtelioCfg vpn.LibConfigGetter,
+	libquenchCfg vpn.NordWhisperConfigGetter,
+	appVersion string,
+	eventsPublisher *vpn.Events,
 ) daemon.FactoryFunc {
-	nordlynxVPN, nordLynxErr := getNordlynxVPN(envIsDev, eventsDbPath, fwmark, cfg, appVersion, eventsPublisher)
+	nordlynxVPN, nordLynxErr := getNordlynxVPN(envIsDev, eventsDbPath, fwmark, libtelioCfg, appVersion, eventsPublisher)
 	if nordLynxErr != nil {
 		// don't exit with `err` here in case the factory will be called with
 		// technology different than `config.Technology_NORDLYNX`
 		log.Println(internal.ErrorPrefix, "getting NordLynx vpn:", nordLynxErr)
 	}
 
-	nordWhisperVPN, nordWhisperErr := getNordWhisperVPN(fwmark, envIsDev, eventsPublisher)
+	nordWhisperVPN, nordWhisperErr := getNordWhisperVPN(fwmark, envIsDev, eventsPublisher, libquenchCfg)
 	if nordWhisperErr != nil {
 		log.Println(internal.ErrorPrefix, "getting NordWhisper vpn:", nordWhisperErr)
 	}
--- a/cmd/daemon/vpnconfig_no_whisper.go
+++ b/cmd/daemon/vpnconfig_no_whisper.go
@@ -1,0 +1,19 @@
+//go:build !quench
+
+package main
+
+import (
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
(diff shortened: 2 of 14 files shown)


cmd/daemon/vpnconfig_no_whisper.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/NordSecurity/nordvpn-linux/daemon/vpn"
+)
+
+type noopNordwhisperConfigGetter struct{}
+
+func (noopNordwhisperConfigGetter) GetConfig() (vpn.NordwhisperFeatureConfig, error) {
+	return vpn.NordwhisperFeatureConfig{EnableECH: true}, nil
+}
+
+func vpnNordwhisperConfigGetterImplementation(_ config.Manager, _ remote.ConfigGetter) noopNordwhisperConfigGetter {
```
keliramu: make the same return type `vpn.NordwhisperConfigGetter` (interface), why use concrete type (struct)?

dfetti: By mistake. I fixed it. When you finish the review, I will push the changes.

daemon/vpn/quench/libquench.go
```diff
@@ -148,22 +148,32 @@ type Quench struct {
...
-func New(fwmark uint32, envIsDev bool, events *vpn.Events) *Quench {
+func New(fwmark uint32, envIsDev bool, events *vpn.Events, cfg vpn.NordwhisperConfigGetter) *Quench {
 	logLevel := quenchBindigns.LogLevelInfo
 	if envIsDev {
 		logLevel = quenchBindigns.LogLevelDebug
 	}
 	logger := Logger{}
 	quenchBindigns.SetLogCallback(logLevel, &logger)
 
+	features, err := cfg.GetConfig()
```
keliramu: This part could be moved into `Start()` method, then we would have RC changes used on every connect - a bit more flexible.

daemon/vpn/quench/config.go
```diff
@@ -14,3 +22,29 @@ type Protocol struct {
...
+	remoteConfigGetter remote.ConfigGetter
+}
+
+func NewNordwhisperConfig(rc remote.ConfigGetter) *NordwhisperConfig {
+	return &NordwhisperConfig{
+		remoteConfigGetter: rc,
+	}
+}
+
+func (qc *NordwhisperConfig) GetConfig() (vpn.NordwhisperFeatureConfig, error) {
```
thegravi: [nit] Missing docstrings for consistency for exported methods. (And other places)

daemon/vpn/libconfig.go
```diff
@@ -4,3 +4,17 @@ package vpn
 type LibConfigGetter interface {
 	GetConfig() (string, error)
 }
+
+type NordwhisperFeatureConfig struct {
```
thegravi: [nit] Naming case inconsistency, Mostly everywhere is `NordWhisper` naming and here `Nordwhisper`.

cmd/daemon/vpnconfig_no_whisper.go
thegravi: [small] `noopNordwhisperConfigGetter` -> `vpn.NordwhisperConfigGetter`

cmd/daemon/vpnconfig_no_whisper.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/NordSecurity/nordvpn-linux/config"
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
+	"github.com/NordSecurity/nordvpn-linux/daemon/vpn"
+)
+
+type noopNordWhisperConfigGetter struct{}
+
+func (noopNordWhisperConfigGetter) GetConfig() (vpn.NordWhisperFeatureConfig, error) {
+	return vpn.NordWhisperFeatureConfig{EnableECH: true}, nil
```
devzbysiu: Maybe `NordWhisperFeatureConfig()`? If we decide to change default value, we would have to change it only in constructor. Now you need to search for struct instantiation

cmd/daemon/vpn_factory.go
```diff
@@ -13,17 +13,23 @@ import (
...
 var ErrNordWhisperDisabled = errors.New("NordWhisper technology was disabled in compile time")
 
-func getVpnFactory(eventsDbPath string, fwmark uint32, envIsDev bool,
-	cfg vpn.LibConfigGetter, appVersion string, eventsPublisher *vpn.Events,
+func getVpnFactory(
+	eventsDbPath string,
+	fwmark uint32,
+	envIsDev bool,
+	libtelioCfg vpn.LibConfigGetter,
+	libquenchCfg vpn.NordWhisperConfigGetter,
```
devzbysiu, line 21: ~~[small] go's idiomatic rules suggest to accept interface in arguments, but return specific types as return values~~

It is an interface

dfetti: NordWhisperConfigGetter is an interface already

PR #1358, opened 2026-02-25
Moose bindings PoC

devzbysiu:
Those are quick and dirty changes to test internal processes.

Not supposed to be merged.


third-party/moose-events
devzbysiu, line 1: This can be removed now, we are not bumping moose version. Same with the `third-party/moose-worker` below

PR #1359, opened 2026-02-25
LVPN-10034: Cover ECH Remote Config option with tests

bbrutall:
- Implemented set of autotests to cover ECH feature

### DIFF
--- a/test/qa/test_remote_config.py
+++ b/test/qa/test_remote_config.py
@@ -4,13 +4,14 @@ import subprocess
 import time
 import uuid
 import os
+from typing import Any
 
 import pytest
 import sh
 
-from lib import daemon
+from lib import daemon, set_technology_and_protocol, STANDARD_TECHNOLOGIES_NO_NORDWHISPER, Defer
 from lib.log_reader import LogReader
-from lib.remote_config_manager import LOCAL_CACHE_DIR
+from lib.remote_config_manager import LOCAL_CACHE_DIR, RemoteConfigManager
 from lib.daemon import (
     Env,
     disable_rc_local_config_usage,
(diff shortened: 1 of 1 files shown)


test/qa/test_remote_config.py
```diff
@@ -832,3 +863,212 @@ def printdir():
...
+    assert not daemon_log_reader.wait_for_messages(
+        messages=RC_USE_LOCAL_CONFIG_MESSAGE,
+        cursor=cursor,
+    )
+
+    sh.nordvpn.connect()
+
+    assert daemon_log_reader.wait_for_messages(
+        messages=RC_USE_ECH_MESSAGE, cursor=cursor
+    ), f"Expected {RC_USE_ECH_MESSAGE} to be absent in the daemon log after reverting."
```
thegravi:
```suggestion
    ), f"Expected {RC_USE_ECH_MESSAGE} to be present in the daemon log after reverting."
```

test/qa/test_remote_config.py
```diff
@@ -33,15 +33,46 @@
...
+    hash_path = f"{LOCAL_CACHE_DIR}/{RC_NORDWHISPER_HASH_FILE}"
+
+    with open(config_path, encoding="utf-8") as config_file:
+        config_data = json.load(config_file)
+
+    for config_entry in config_data.get("configs", []):
+        if config_entry.get("name") == "enable_ech":
+            for setting in config_entry.get("settings", []):
+                setting["value"] = enabled
+                setting["app_version"] = ">=4.4.0"
```
thegravi, line 75: Shall we guard for possible edge cases?
```suggestion
    modified = False
    for config_entry in config_data.get("configs", []):
        if config_entry.get("name") == "enable_ech":
            for setting in config_entry.get("settings", []):
                setting["value"] = enabled
                setting["app_version"] = ">=4.4.0"
                modified = True

    if not modified:
        raise ValueError(
            f"enable_ech config entry not found in {config_path}. "
            "Check if the remote config format has changed."
        )

```

test/qa/test_remote_config.py
```diff
@@ -33,15 +33,46 @@
     f"[Info] feature [libtelio] config loaded from: {LOCAL_CACHE_DIR}",
     f"[Info] feature [nordvpn] config loaded from: {LOCAL_CACHE_DIR}",
 ]
```
thegravi, line 38: Missing `nordwhisper` here?

test/qa/test_remote_config.py
bbrutall: fixed

test/qa/test_remote_config.py
bbrutall: We have test that modifies local config files and verifies for mismatches. Also, added new tests `test_local_config_usage_missing_enable_ech_field`

test/qa/test_remote_config.py
bbrutall: fixed

test/qa/test_remote_config.py
```diff
@@ -832,3 +877,290 @@ def printdir():
...
+      9. Disconnect from VPN.
+      10. Disable 'use local RC config' mode.
+
+    :raises AssertionError: If the expected log message 'RC_USE_LOCAL_CONFIG_MESSAGE' is not found in the daemon log.
+    :raises AssertionError: If the expected missing-parameter warning message is not found in the daemon log.
+    :raises AssertionError: If the expected log message 'RC_USE_ECH_MESSAGE' is not found in the daemon log
+    """
+    rc_config_manager.set_permissions_cache_dir()
+
+    set_ech_for_local_config(enable_ech, entries_to_remove=[("name", "enable_ech")])
```
devzbysiu, line 1133: Does it make sense to pass `enable_ech` when this field config will be removed inside that `set_ech_for_local_config` function?

test/qa/test_remote_config.py
```diff
@@ -832,3 +877,290 @@ def printdir():
...
+):
+    """
+    Test to verify that ECH is disabled via local RC usage and restored after disabling local RC usage.
+
+    Verify that disabling ECH via local RC config takes effect when local config usage is
+    enabled, and that it reverts to normal RC config behavior once local config usage
+    is disabled.
+
+    Test steps:
+      1. Set the local cached RC config to set 'enable_ech' to the parametrized value
```
devzbysiu: Copy-paste? There is no `enable_ech` parameter 😄

test/qa/test_remote_config.py
```diff
@@ -832,3 +877,290 @@ def printdir():
...
+   """
+    set_technology_and_protocol(tech, proto, obfuscated)
+
+    sh.nordvpn.connect()
+
+    assert not daemon_log_reader.wait_for_messages(
+        messages=RC_USE_ECH_MESSAGE, cursor=daemon_log_cursor
+    ), f"Expected {RC_USE_ECH_MESSAGE} to be absent in the daemon log."
+
+    sh.nordvpn.disconnect()
```
devzbysiu, line 914: If the assertion above fails, we'll never call disconnect, right? Or is there some cleanup defined somewhere else? 🤔 
Same in other tests here

test/qa/test_remote_config.py
```diff
@@ -26,22 +27,66 @@
...
+    for config_entry in config_data.get("configs", []):
+        if config_entry.get("name") == "enable_ech":
+            for setting in config_entry.get("settings", []):
+                setting["value"] = enabled
+                setting["app_version"] = ">=4.4.0"
+
+    for config_entry in config_data.get("configs", []):
+        for key, expected_value in entries_to_remove:
+            if config_entry.get(key) == expected_value:
+                config_entry.pop(key, None)
```
DoYouEvenCpp, line 80: [nit] wouldn't be more natural to keep those two for loops under the same context manager (` with open(config_path, encoding="utf-8") as config_file:`)?

test/qa/test_remote_config.py
bbrutall: fixed

test/qa/test_remote_config.py
bbrutall: It doesn’t make much functional sense, but it’s still fine to pass it for testing consistency and to confirm the “missing field” behavior.

test/qa/test_remote_config.py
bbrutall: There is a logout cleanup procedure in `nordvpnd_scope_function` fixture.

test/qa/test_remote_config.py
bbrutall: Yes, it would be more natural to keep them together since both operations modify  the same config_data, but I separated them intentionally to keep the logic clearer (first apply the modifications, then perform optional deletions) and to avoid mixing two different behaviors in one loop.

PR #1360, opened 2026-02-25
LVPN-10062: Update netlink library to 1.3.1

edvyno1:
Main library observed changes:

fwmark in netlink.Rule by default is now uint32=0 instead of int=-1
[reference](https://github.com/vishvananda/netlink/commit/8f96fd8b2f8c1ffc057e32e4a0021c075d943f0c#diff-d550ae620c1f4d7833c6b74d5c1898a6699529bf6f5a54688403426ca0e86ff2L54)

netlink.RouteListFiltered api has changed, main difference for us is that `0.0.0.0/0` IPs were returned as `nil`, now they are properly returned as `0.0.0.0/0`, breaking some of our logic
[reference](https://github.com/vishvananda/netlink/commit/b54f85093f4aed534bf4e2b77fd8e57642b34b8d)

This PR also revives some of our netlink tests that were not running on CI(the tests actually caught the errors and failed during the library upgrade) 

Some of the tests were moved to NotWorking as they were not working and completely unrelated to netlink, they need to be investigated later

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -9,7 +9,7 @@ set -euxo pipefail
 excluded_packages="moose\|telio"
 excluded_packages=$excluded_packages"\|events\/moose"
 excluded_packages=$excluded_packages"\|daemon\/vpn\/quench"
-excluded_categories="root,link,firewall,route,file,integration"
+excluded_categories="root,link,firewall,route,file,integration,notworking"
 
 tags="internal"
 parallel=""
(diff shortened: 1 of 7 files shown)


daemon/routes/iprule/iprule.go
dfetti: We should make sure this logic is not affected. As far as I remember the new update will make  other fields of the rule 0 instead of -1. We should be safe, but it's worth validating it. My changes regarding rule priority algorithm were merged into the release branch: https://github.com/NordSecurity/nordvpn-linux/pull/1348

edvyno1: yep [this](https://github.com/NordSecurity/nordvpn-linux/pull/1348/changes#diff-4be8fb506b6e94496c913c8bcbd2a737f6703ba151c0805d9115ef12249777aaR495) specifically will need to get changed to `rule.Mark == 0`

would it maybe be possible to cherry-pick this commit into main as well so I could rebase and make the change there before merging?

dfetti: Yes, on it

PR #1362, opened 2026-02-26
LVPN-10075: Integrate nftables

mariusSincovici:
Update networker to configure the nftables when there are changes that are affecting the firewall.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -16,7 +16,7 @@ parallel=""
 
 # In case 'full' was specified, do not exclude anything and run
 # everything
-if [ "full" = "full" ]; then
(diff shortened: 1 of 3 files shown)


cmd/daemon/main.go
```diff
@@ -434,11 +433,11 @@ func main() {
 		kernel.NewSysctlSetter(
 			networker.ArpIgnoreParamName, 1,
 		),
+		cfg.AutoConnectData.Allowlist,
 		// TODO: during nft mesh netw integration
```
edvyno1: [reminder] needs to be integrated in this PR

mariusSincovici: Done

PR #1366, opened 2026-03-03
LVPN-9574: Kill GUI application before uninstall

dfetti:
Kill norvpn-gui before removing gui package or main package

### DIFF
--- a/gui/templates/scriptlets/deb/prerm_template
+++ b/gui/templates/scriptlets/deb/prerm_template
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+case "$1" in
+    remove|deconfigure)
+        pkill -TERM -x nordvpn-gui || :
+    ;;
+    upgrade)
+        # Do nothing for upgrade
+    ;;
+    *)
+        echo "prerm called with unknown argument \`$1'" >&2
+        exit 1
+    ;;
+esac
+
+exit 0
--- a/gui/templates/scriptlets/rpm/preun_template
+++ b/gui/templates/scriptlets/rpm/preun_template
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+case "$1" in
+    0)
+        # Remove
+        pkill -TERM -x nordvpn-gui || :
+    ;;
+    1)
+        # Do nothing for upgrade
+    ;;
+    *)
+        echo "preun called with unknown argument \`$1'" >&2
+        exit 1
+    ;;
+esac
+
+exit 0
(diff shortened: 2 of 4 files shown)


contrib/scriptlets/deb/prerm
```diff
@@ -8,6 +8,7 @@ case "$1" in
             init)
                 /etc/init.d/nordvpn stop &>/dev/null || :
                 update-rc.d -f nordvpn remove &>/dev/null || :
+                pkill -TERM -x nordvpn-gui || :
```
edvyno1: are these needed? we will remove the `nordvpn-gui` package when removing `nordvpn` as it's a direct dependency, so the GUI package's scriptlets will be ran regardless, or was there some edge case that also needed these?

gui/templates/scriptlets/rpm/preun_template
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+ENV=$(ps --no-headers -o comm 1)
```
edvyno1: unused variable

devzbysiu: Do we need this?

contrib/scriptlets/deb/prerm
```diff
@@ -17,6 +18,7 @@ case "$1" in
                 systemctl disable nordvpnd.service &>/dev/null || :
                 systemctl disable nordvpnd.socket &>/dev/null || :
                 systemctl daemon-reload &>/dev/null || :
+                pkill -TERM -x nordvpn-gui || true :
```
devzbysiu: Any purpose behind `|| true :`? Usually it's either `|| true` or `|| :` 🤔

contrib/scriptlets/rpm/preun
```diff
@@ -8,13 +8,15 @@ case "$1" in
             init)
                 /etc/init.d/nordvpn stop &>/dev/null || :
                 update-rc.d -f nordvpn remove &>/dev/null || :
+                pkill -TERM -x nordvpn-gui || true :
```
devzbysiu: Do we need to kill GUI here? I think when daemon is uninstalled, GUI is being uninstalled also as GUI has a dependency on daemon

gui/templates/scriptlets/deb/prerm_template
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+case "$1" in
+    remove|deconfigure)
+        pkill -TERM -x nordvpn-gui || :
```
devzbysiu, line 5: Did you check if `pkill` is universally available across distros?

gui/templates/scriptlets/rpm/preun_template
devzbysiu: Duplicate comment

contrib/scriptlets/rpm/preun
devzbysiu: Duplicated comment

contrib/scriptlets/deb/prerm
thegravi: Can we be consistent here, and stick with `|| :`?  It's slightly more "safe" way.

gui/templates/scriptlets/deb/prerm_template
dfetti: Yes, it seems to be available across all the distros

contrib/scriptlets/deb/prerm
dfetti: No need, I removed these.

PR #1369, opened 2026-03-05
LVPN-10116: simple NetworkManager DNS management

bartoszWojciechO:


daemon/dns/dns_resolvenmcli.go
```diff
@@ -9,212 +9,82 @@ import (
...
-	cdmaType     = "cdma"
-	ethernetType = "ethernet"
-	gsmType      = "gsm"
-	wirelessType = "wireless"
-
-	nmCliConKey             = "con"
-	nmCliExecutable         = "nmcli"
-	nmCliIPv4DNSKey         = "ipv4.dns"
-	nmCliIPIgnoreAutoDnsKey = "ipv4.ignore-auto-dns"
+	// execResolvectl defines resolvectl executable
```
devzbysiu: Leftover?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -9,212 +9,82 @@ import (
...
-}
 
-// Name returns name of the DNS management method.
-func (nmcli *NMCli) Name() string {
-	return "nmcli"
-}
+	if out, err := n.runNMCliCommandFunc("general", "reload"); err != nil {
+		log.Println(internal.ErrorPrefix,
+			dnsPrefix,
+			"failed to reaload after adding a config file NetworkManager, command output: %s",
```
devzbysiu: [typo] `reaload`

Also `log.Println` does not support format strings (`output: %s`)

daemon/dns/dns_resolvenmcli.go
```diff
@@ -9,212 +9,82 @@ import (
...
-	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
-	if err != nil {
-		log.Println(internal.WarningPrefix, dnsPrefix, "Failed to get active connections upon UnsetDNS", err)
-		return fmt.Errorf("failed to get active connection upon UnsetDNS: %w", err)
-	}
-	for _, con := range connections {
-		args := []string{nmCliConKey, "modify", con, nmCliIPv4DNSKey, ""}
-		args = append(args, nmCliIPIgnoreAutoDnsKey, "no")
+func (n *NMCli) Set(iface string, nameservers []string) error {
+	configContents := fmt.Sprintf(`[global-dns-domain-*]
```
devzbysiu: Isn't 
```
[global-dns]
enabled=yes
```
needed? 🤔

daemon/dns/dns_resolvenmcli.go
```diff
@@ -9,212 +9,82 @@ import (
...
-func (nmcli *NMCli) Unset(_ string) error {
-	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
-	if err != nil {
-		log.Println(internal.WarningPrefix, dnsPrefix, "Failed to get active connections upon UnsetDNS", err)
-		return fmt.Errorf("failed to get active connection upon UnsetDNS: %w", err)
-	}
-	for _, con := range connections {
-		args := []string{nmCliConKey, "modify", con, nmCliIPv4DNSKey, ""}
-		args = append(args, nmCliIPIgnoreAutoDnsKey, "no")
+func (n *NMCli) Set(iface string, nameservers []string) error {
```
devzbysiu, line 88: Should we have a precondition check if `nameservers` is empty?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -10,211 +11,83 @@ import (
...
-	name          string
-	ipv4DNS       string
-	ignoreAutoDNS string
+type nmCliCommandFunc func(...string) ([]byte, error)
+
+func execNMCliCommand(args ...string) ([]byte, error) {
+	return exec.Command(execNMCli, args...).CombinedOutput()
 }
 
+// Systemd-resolved and resolvectl based DNS handling method
```
DoYouEvenCpp: This comment seems off?

daemon/dns/dns_resolvenmcli.go
```diff
@@ -10,211 +11,83 @@ import (
...
-// Parameters:
-//   - iface: unused
-//
-// Returns an error if the nmcli command fails or if reloading the connection fails.
-func (nmcli *NMCli) Unset(_ string) error {
-	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
-	if err != nil {
-		log.Println(internal.WarningPrefix, dnsPrefix, "Failed to get active connections upon UnsetDNS", err)
-		return fmt.Errorf("failed to get active connection upon UnsetDNS: %w", err)
+func (n *NMCli) Set(iface string, nameservers []string) error {
```
DoYouEvenCpp, line 88: [small] please restore comments back

daemon/dns/dns_resolvenmcli.go
```diff
@@ -10,211 +11,83 @@ import (
...
-	for _, con := range connections {
-		state, err := nmcli.getConnectionState(con)
-		if err != nil {
-			log.Println(internal.WarningPrefix, dnsPrefix, "Failed to get state for connection", con, ":", err)
-			return fmt.Errorf("failed to get connection state: %w", err)
-		}
-		originalStates[con] = state
+func (n *NMCli) removeConfigFile() error {
+	if err := n.filesystemHandle.Remove(networkManagerConfigFilePath); err != nil {
+		return fmt.Errorf("removing DNS config file: %w", err)
```
DoYouEvenCpp: Can't we simply return n.filesystemHandle.Remove(networkManagerConfigFilePath) here?
Thing is, with `Unset` if this fails, then we gonna to wrap the original error twice, cluttering it a bit too much, I'd say.

daemon/dns/dns_resolvenmcli.go
```diff
@@ -1,220 +1,144 @@
...
+	if out, err := n.runNMCliCommandFunc("general", "reload"); err != nil {
+		log.Println(internal.ErrorPrefix,
+			dnsPrefix,
+			"failed to reload after adding a config file NetworkManager, command output:",
+			strings.TrimSpace(string(out)))
+
+		if err := n.removeConfigFile(); err != nil {
+			log.Println(internal.ErrorPrefix, dnsPrefix, "removing config file after a failed reload operation:", err)
+		}
+		return fmt.Errorf("failed to reload NetworkManager config: %w", err)
```
devzbysiu, line 120: Should we maybe join the `err` of `runNMCliCommandFunc` and `removeConfigFile` here?

daemon/dns/dns.go
```diff
@@ -351,7 +353,9 @@ func (d *DNSMethodSetter) Unset(iface string) error {
 	return errors.Join(errs...)
 }
 
-// RestoreResolvConfFile try to restore resolv.conf if target file contains Nordvpn changes
-func RestoreResolvConfFile() {
+// RestoreDNS try to restore resolv.conf if target file contains Nordvpn changes
+func RestoreDNS() {
 	tryToRestoreDNS()
+	// #nosec G104 -- the operation is expected to fail in most cases
+	newNMCli().removeConfigFile()
```
devzbysiu: Maybe we could log an error here? If something weird happens and we can't do cleanup, we won't see the error in logs

daemon/dns/dns_resolvenmcli.go
```diff
@@ -1,220 +1,144 @@
...
-func (nmcli *NMCli) Set(iface string, nameservers []string) error {
-	connections, err := nmcli.getConnectionFromPhysicalInterfaces()
+func (n *NMCli) removeConfigFile() error {
+	return n.filesystemHandle.Remove(networkManagerConfigFilePath)
+}
+
+// higherPriorityFileExists returns true if a config file with higher priority(i.e lexicographically greater) exists
+// within the NetworkManager config directory.
+func (n *NMCli) higherPriorityFileExists() (bool, error) {
+	directoryNames, err := n.getConfigFilesFunc()
```
devzbysiu: Should we be more defensive here and remove our own config from the list if it exists there?

daemon/dns/dns_resolvenmcli_test.go
```diff
@@ -1,263 +1,160 @@
...
 			} else {
-				assert.NoError(t, err)
-			}
-
-			for connName, expectedDNS := range tt.expectedFinalDNS {
-				assert.Equal(t, expectedDNS, currentState.connections[connName],
-					"Connection %s should have DNS restored to %s", connName, expectedDNS)
+				assert.Nil(t, err, "Unexpected error returned by Unset.")
+				_, ok := mockFs.GetFile(networkManagerConfigFilePath)
+				assert.True(t, ok, "File was not removed after Unset.")
```
devzbysiu: Shouldn't this be `assert.False`? 🤔

devzbysiu: Also, `mockFs` is not really removing files so tests pass with `assert.True`. When you update mock to actually remove the file this case will start failing. You can either update the mock to start removing files and test that this side effect in mock happens, or you can track that `Remove` was called with proper argument so we know that the contract is fulfilled

PR #1371, opened 2026-03-06
LVPN-10075: integrate nft [NFT]

edvyno1:


daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+		conn:   &nftables.Conn{},
+		fwmark: fwmark,
+	}
+}
+
+func (n *nft) Configure(config firewall.Config) error {
+	if err := n.configure(config); err != nil {
+		return fmt.Errorf("nft configure: %w", err)
+	}
+	return nil
```
thegravi: [nit] Do we really need this log here? Why not just simply return `n.configure()` and test the result in the caller?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+		KeyByteOrder: binaryutil.NativeEndian,
+		// Constant:     true, // disable for strings https://github.com/google/nftables/issues/177
+	}
+
+	elems := []nftables.SetElement{
+		{Key: ifname("lo")},
+	}
+
+	if len(config.TunnelInterface) > 0 {
+		if len(config.TunnelInterface) > unix.IFNAMSIZ {
```
thegravi: [not] Reuse iface len.

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+func (n *nft) buildLanRangesSet(table *nftables.Table) (*nftables.Set, error) {
+	lanNets := &nftables.Set{
+		Table:    table,
+		Name:     lanPrivateIpsSetName,
+		KeyType:  nftables.TypeIPAddr,
+		Interval: true,
+		Constant: true,
+	}
+
+	elems, err := convertCidrToSetElements([]string{
```
thegravi: [small] Can we have this as a source of truth? The same subnets are hardcoded in the `allowlist_go` (networket) package.

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("parse allowlist IP: %s %w", subnet, err)
+		}
+
+		elements = append(elements,
+			nftables.SetElement{Key: startIp}, nftables.SetElement{Key: endIp, IntervalEnd: true},
+		)
+	}
+	if err := n.conn.AddSet(set, elements); err != nil {
+		return nil, fmt.Errorf("add allowlist set: %v", err)
```
thegravi: [nit] Why not wrap the error (%w)?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+
+		allowedIncomingConnections, err = n.buildAllowedIncomingConnections(table, config.MeshnetInfo.MeshnetMap)
+		if err != nil {
+			return fmt.Errorf("add peers allowed to connect set: %w", err)
+		}
+	}
+
+	if err := n.conn.Flush(); err != nil {
+		return err
+	}
```
thegravi: Why do we have this flush before populating tables?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+			Exprs: buildRules(
+				expr.VerdictAccept,
+				checkPortInSet(tcpPortsSet, unix.IPPROTO_TCP, MATCH_SOURCE),
+				setMetaMark(n.fwmark),
+			),
+		})
+	}
+
+	if udpPortsSet != nil {
+		// udp sport @ports_tcp meta mark set 0xe1f1 accept
```
thegravi:
```suggestion
		// udp sport @ports_udp meta mark set 0xe1f1 accept
```

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+//		type filter hook input priority filter; policy drop;
+//	}
+func (n *nft) addInput(
+	config firewall.Config,
+	table *nftables.Table,
+	excludedInterfaces *nftables.Set,
+	filesharePeers *nftables.Set,
+	allowIncomingConnections *nftables.Set,
+) {
+	dropPolicy := nftables.ChainPolicyAccept
```
thegravi: [small] Why not `chainPolicy`?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+				Table: table,
+				Chain: outputChain,
+				Exprs: buildRules(
+					expr.VerdictDrop,
+					checkIpInSet(lanPrivateRanges, MATCH_DESTINATION),
+					checkPortNumber(53, unix.IPPROTO_UDP, MATCH_DESTINATION),
+				),
+			})
+		}
+	}
```
thegravi, line 422: [small] Can we have a constant here without magic number?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+				Exprs: buildRules(
+					expr.VerdictDrop,
+					checkIpInSet(lanPrivateRanges, MATCH_DESTINATION),
+					checkPortNumber(53, unix.IPPROTO_TCP, MATCH_DESTINATION),
+				),
+			})
+		}
+
+		if !config.Allowlist.Ports.UDP[53] {
+			// ip daddr @lan_ranges tcp dport 53 drop
```
thegravi:
```suggestion
			// ip daddr @lan_ranges udp dport 53 drop
```

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+			Exprs: buildRules(
+				expr.VerdictAccept,
+				checkPortInSet(tcpPortsSet, unix.IPPROTO_TCP, MATCH_DESTINATION),
+				setMetaMark(n.fwmark),
+			),
+		})
+	}
+
+	if udpPortsSet != nil {
+		// udp dport @ports_tcp meta mark set 0xe1f1 accept
```
thegravi:
```suggestion
		// udp dport @ports_udp meta mark set 0xe1f1 accept
```

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+	table *nftables.Table,
+	allowedSubnets *nftables.Set,
+	udpPortsSet *nftables.Set,
+	tcpPortsSet *nftables.Set,
+	lanAllowedPeers *nftables.Set,
+	routingAllowed *nftables.Set,
+	lanRangesIps *nftables.Set,
+	meshMap *firewall.MeshInfo,
+) {
+	dropPolicy := nftables.ChainPolicyAccept
```
thegravi: [nit] `chainPolicy`?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,841 @@
...
+			Chain: forwardChain,
+			Exprs: buildRules(expr.VerdictAccept,
+				checkPortInSet(tcpPortsSet, unix.IPPROTO_TCP, MATCH_DESTINATION),
+				setMetaMark(n.fwmark),
+			),
+		})
+	}
+
+	if udpPortsSet != nil {
+		// udp dport @ports_tcp meta mark set 0xe1f1 accept
```
thegravi:
```suggestion
		// udp dport @ports_udp meta mark set 0xe1f1 accept
```

daemon/firewall/nft/nft_utils.go
```diff
@@ -0,0 +1,449 @@
...
+func ifname(n string) []byte {
+	b := make([]byte, unix.IFNAMSIZ)
+	copy(b, []byte(n))
+	return b
+}
+
+// calculateFirstAndLastV4Prefix returns:
+//   - first: network address (inclusive)
+//   - lastExclusive: (broadcast + 1), i.e. exclusive upper bound
+func calculateFirstAndLastV4Prefix(cidr string) (net.IP, net.IP, error) {
```
thegravi, line 347: [small] Some unit tests here would be useful.

daemon/firewall/firewall.go
```diff
@@ -4,180 +4,88 @@ Package firewall provides firewall service to the caller
...
+
+	return fw.impl.Configure(config)
 }
 
-// Enable restores firewall operations from no-ops.
 func (fw *Firewall) Enable() error {
 	fw.mu.Lock()
 	defer fw.mu.Unlock()
+
+	log.Println(internal.InfoPrefix, logPrefix, "enabling firewall")
```
thegravi, line 51: [small] With `Disable()` we also log previous state. I suggest keep the consistency (add here/ or remove there).

daemon/firewall/firewall.go
```diff
@@ -4,180 +4,88 @@ Package firewall provides firewall service to the caller
...
+
 	if fw.enabled {
-		return NewError(ErrFirewallAlreadyEnabled)
+		return errors.New("firewall already enabled")
```
thegravi: Do we want/need these as package level errors?

networker/allowlist_lan.go
```diff
@@ -15,8 +15,7 @@ func addLANPermissions(allowlist config.Allowlist) config.Allowlist {
 		"192.168.0.0/16",
 		"169.254.0.0/16"}
 
-	var newSubnets []string
-	copy(newSubnets, allowlist.Subnets)
+	newSubnets := append([]string{}, allowlist.Subnets...)
 	for _, network := range localNetworks {
 		if !slices.Contains(newSubnets, network) {
 			newSubnets = append(newSubnets, network)
```
thegravi: Can you add unit test to cover this fix?

edvyno1: this was already merged to 4.5.0 release branch with accompanying python integration test
https://github.com/NordSecurity/nordvpn-linux/commit/8c55fc1afbda0cd9d7b4472f616ab2883739521e

cmd/daemon/main.go
```diff
@@ -206,21 +202,8 @@ func main() {
...
-		iptables.FilterSupportedIPTables([]string{"iptables", "ip6tables"}),
-	)
-	fw := firewall.NewFirewall(
-		&notables.Facade{},
-		iptablesAgent,
-		debugSubject,
-		cfg.Firewall,
-	)
+	nftImpl := nft.New(cfg.FirewallMark)
+	fw := firewall.NewFirewall(nftImpl, cfg.Firewall)
```
keliramu: Get rid of unneeded line 205:
```	// Firewall
	fw := firewall.NewFirewall(
		nft.New(cfg.FirewallMark),
		cfg.Firewall)
```

daemon/firewall/service.go
```diff
@@ -1,31 +1,85 @@
...
-// Agent carries out required firewall changes.
-//
-// Used by implementers.
-type Agent interface {
-	// Add a firewall rule
-	Add(Rule) error
-	// Delete a firewall rule
-	Delete(Rule) error
-	// Flush removes all nordvpn rules
+type FirewallImpl interface {
```
keliramu: [naming] suggesting have `FirewallBackend` - otherwise it's misleading: interface is not implementation

daemon/firewall/nft/nft.go
edvyno1: seems to work without it, will leave still for Marius to have a look when he comes back, maybe some edge case covered that I am not aware of

daemon/firewall/nft/nft.go
edvyno1: applied

daemon/firewall/nft/nft.go
edvyno1: applied, moved constant to internal constants.go

daemon/firewall/nft/nft.go
edvyno1: maybe better to leave comment for what port 53 does, not sure if this makes it more readable if 53 is defined as `defaultDnsPort`, wdyt

daemon/firewall/nft/nft.go
keliramu: Only one `Flush` should be, otherwise it's bad transaction/atomicity.

Single `Flush` is no problem here:
All sets, chains and rules are sent to the kernel in a single netlink batch.
The google/nftables library assigns set IDs locally on AddSet(), so rules can
reference them by ID within the same batch without a prior Flush().
This single Flush() makes the entire table replacement atomic: the kernel
commits the old table deletion and the new table with all its rules simultaneously.

daemon/firewall/nft/nft.go
keliramu: imo, `defaultDnsPort` is more readable/right approach

networker/allowlist_lan.go
thegravi: Yes, I remember. So if this is already covered by e2e test, then good.

daemon/firewall/nft/nft_utils.go
edvyno1: added a unit test

daemon/firewall/firewall.go
edvyno1: got rid of the log in the disable func, as the subsequent log line kind of does the same thing

daemon/firewall/nft/nft.go
DoYouEvenCpp: Second that. I guess we quite often end up with an error being wrapped way to many times, leaving the caller with actually a bit cluttered information.

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,837 @@
...
+		)
+	}
+	if err := n.conn.AddSet(set, elements); err != nil {
+		return nil, fmt.Errorf("add allowlist set: %w", err)
+	}
+
+	return set, nil
+}
+
+func addMainTable(conn *nftables.Conn) *nftables.Table {
```
DoYouEvenCpp: [nit] why not make this method implemented for `*nft` (this way the parameter could be skipped)?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,837 @@
...
+		allowedIncomingConnections, err = n.buildAllowedIncomingConnections(table, config.MeshnetInfo.MeshnetMap)
+		if err != nil {
+			return fmt.Errorf("add peers allowed to connect set: %w", err)
+		}
+	}
+
+	// if err := n.conn.Flush(); err != nil {
+	// 	return err
+	// }
+
```
DoYouEvenCpp: Dead code?

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,837 @@
...
+	if config.MeshnetInfo != nil && routingAllowed != nil {
+		n.addMeshnetNat(table, routingAllowed)
+	}
+
+	return n.conn.Flush()
+}
+
+//	chain prerouting {
+//			type filter hook prerouting priority mangle; policy accept;
+//	}
```
DoYouEvenCpp: [nit] please convert this to a proper comment

daemon/firewall/nft/nft_utils.go
```diff
@@ -0,0 +1,449 @@
...
+
+func buildJumpRules(chainName string, parts ...[]expr.Any) []expr.Any {
+	var n int
+	for _, p := range parts {
+		n += len(p)
+	}
+	out := make([]expr.Any, 0, n+1)
+	for _, p := range parts {
+		out = append(out, p...)
+	}
```
DoYouEvenCpp: [nit] this is literally the same block of code as in `buildRules`. Why not to extract this to a common helper method (say `concatExpressions`, and then reuse?

daemon/firewall/nft/nft_utils_test.go
```diff
@@ -0,0 +1,40 @@
+package nft
```
DoYouEvenCpp, line 1: adding more tests would be awesome... same for `nft.go` 😉

daemon/firewall/nft/nft.go
```diff
@@ -0,0 +1,837 @@
...
+	})
+
+	// ct mark 0xe1f1 accept
+	n.conn.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: inputChain,
+		Exprs: buildRules(expr.VerdictAccept, checkConntrack(n.fwmark)),
+	})
+
+	// meshnet
```
DoYouEvenCpp, line 271: [nit] maybe it might be worth to extract this to a separate method.
I, personally, find methods longer than 100LOC a bit too long 😉

daemon/firewall/firewall.go
```diff
@@ -4,180 +4,88 @@ Package firewall provides firewall service to the caller
...
-		}
-		err = fw.current.Delete(rule)
-		if err != nil {
-			return NewError(fmt.Errorf("deleting %s: %w", name, err))
-		}
-		err = fw.rules.Delete(rule.Name)
-		if err != nil {
-			return NewError(fmt.Errorf("deleting %s from memory: %s", name, err))
-		}
+	if !fw.enabled {
```
DoYouEvenCpp, line 38: Don't we want to at first test against `fw.enabled` and only when actually enabled perform any other actions (including log)?

daemon/firewall/firewall.go
DoYouEvenCpp: This is slightly inconsistent, for `Configure` and `Flush` if firewall is disabled we merely print, while here and for `Disable` an error is returned?
Don't we have to unify the approach here for the whole API of the Firewall?

daemon/firewall/nft/nft.go
mariusSincovici: The flush from here is not needed, I've forgot it while I was debugging a bug from nftables, regarding the interface names.

daemon/firewall/nft/nft_utils_test.go
mariusSincovici: They will be, but not in this PR.

daemon/firewall/firewall.go
mariusSincovici: I'll change to use the same errors as before.
But not sure what it means at package level regarding the errors. 🤔 

----
Regarding the difference between enable/disable and flush. They are used in different situations. 
Enable/disable are a result of a user action and because of this the errors are returned. 
Flush is not, so user cannot do anything to fix this. For example flush is executed for `nordvpn d` when not connected to VPN, as safety to cleanup the firewall in case something remained by mistake in the firewall.

PR #1373, opened 2026-03-06
LVPN-10128: Implement interactive secure login with token

thegravi:
Adds secure interactive token input for `nordvpn login --token` command when no token argument is provided.
> Note: Copy's pending.

[demo-login-token-interactive.webm](https://github.com/user-attachments/assets/4349c4f9-a7c5-4064-aa15-97a6f9e9dfe4)

### DIFF
--- a/test/qa/test_login.py
+++ b/test/qa/test_login.py
@@ -2,6 +2,7 @@ import pytest
 import pexpect
 import sh
 
+
 import lib
 from lib import (
     daemon,
@@ -10,6 +11,7 @@ from lib import (
     settings,
     selenium,
 )
+
 
 pytestmark = pytest.mark.usefixtures("collect_logs")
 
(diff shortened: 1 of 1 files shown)


cli/terminal.go
```diff
@@ -20,6 +20,51 @@ func isStdoutATerminal() bool {
...
+// ErrTokenEmpty is returned when the token is empty or contains only whitespace.
+var ErrTokenEmpty = errors.New("token cannot be empty")
+
+// ErrTokenTooLong is returned when the token exceeds MaxTokenLength.
+var ErrTokenTooLong = fmt.Errorf("token is too long; maximum length is %d characters", MaxTokenLength)
+
+// ErrStdinNotTerminal is returned when stdin is not a terminal but interactive input is required.
+var ErrStdinNotTerminal = errors.New("stdin is not a terminal")
+
+// validateToken trims whitespace and validates the token is non-empty and within MaxTokenLength.
```
DoYouEvenCpp, line 37: It also returns empty string when it exceeds the allowed maximum length.

cli/cli_login_test.go
```diff
@@ -0,0 +1,63 @@
...
+	ctx := cli.NewContext(app, set, &cli.Context{Context: context.Background()})
+
+	err := c.loginWithToken(ctx)
+
+	assert.NoError(t, err)
+	assert.True(t, loginCalled, "LoginWithToken should be called")
+	assert.Equal(t, "abc123def456", capturedToken)
+}
+
+func TestLoginWithToken_NoArgument_NonTTY_ReturnsError(t *testing.T) {
```
DoYouEvenCpp, line 43: Would it be of a reasonable effort to cover with tests other erroneous cases for the cli token login here?

cli/terminal.go
```diff
@@ -20,6 +20,51 @@ func isStdoutATerminal() bool {
 	return term.IsTerminal(int(os.Stdout.Fd()))
 }
 
+// isStdinATerminal returns true if stdin is connected to a terminal.
```
DoYouEvenCpp: [nit] I do appreciate keeping name here on par with `isStdoutATerminal`, but usually we don't put articles into methods names, don't we? 😉

PR #1411, opened 2026-03-31
LVPN-10156: Cleanup connection card

dfetti:
Review theme usage, review spacing and add small improvements.

### DIFF
--- a/gui/lib/theme/connection_card_theme.dart
+++ b/gui/lib/theme/connection_card_theme.dart
@@ -3,12 +3,26 @@ import 'package:theme_tailor_annotation/theme_tailor_annotation.dart';
 
 part 'connection_card_theme.tailor.dart';
 
-final class ConnectionCardLabelTheme {
+@tailorMixinComponent
+final class ConnectionCardLabelTheme
+    extends ThemeExtension<ConnectionCardLabelTheme>
+    with _$ConnectionCardLabelThemeTailorMixin {
+  @override
   final Color disconnectedColor;
+
+  @override
   final Color connectingColor;
+
+  @override
   final Color connectedColor;
+
+  @override
   final Color serverTypeColor;
+
+  @override
   final double spacing;
+
+  @override
   final TextStyle font;
 
   ConnectionCardLabelTheme({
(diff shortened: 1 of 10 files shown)


gui/lib/theme/connection_card_theme.dart
```diff
@@ -24,24 +24,36 @@ final class ConnectionCardLabelTheme {
...
     required this.dipIconHeight,
     required this.borderConnectedColor,
     required this.borderConnectingColor,
     required this.disconnectedBackgroundColor,
     required this.disconnectedIcon,
+    required this.disconnectedPadding,
+  });
+}
+
+final class ConnectionCardButtonTheme {
```
devzbysiu: Going back to this discussion https://github.com/NordSecurity/nordvpn-linux/pull/1395#discussion_r2975170671 

I think those nested classes should have `@tailorMixinComponent`. Was there a reason to skip it?

gui/lib/vpn/connection_card.dart
```diff
@@ -52,26 +43,32 @@ final class ConnectionCard extends StatelessWidget {
...
-            ConnectionCardIcon(status: vpnStatus),
-            SizedBox(width: connectionCardTheme.smallSpacing),
-            Column(
-              crossAxisAlignment: CrossAxisAlignment.start,
+            Row(
               children: [
-                ConnectionCardServerInfo(vpnStatus: vpnStatus),
-                ConnectionCardLabel(vpnStatus: vpnStatus),
+                ConnectionCardIcon(status: vpnStatus),
+                SizedBox(width: connectionCardTheme.smallSpacing),
```
devzbysiu: [small] there is `spacing` field on `Row` which could be used instead of `SizedBox`es

gui/lib/vpn/connection_card_server_info.dart
```diff
@@ -12,11 +12,9 @@ final class ConnectionCardServerInfo extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final statusCardTheme = context.connectionCardTheme;
-    var label = t.ui.connectToVpn;
+    var label = t.ui.fastestServer;
 
-    if (vpnStatus.isDisconnected()) {
-      label = t.ui.fastestServer;
-    } else {
+    if (!vpnStatus.isDisconnected()) {
```
devzbysiu: There is already some nesting in this method. Maybe it would be easier to read with
```dart
if (vpnStatis.isDisconnected()) {
  // return early
}

... rest ...
```

Or you could even extract  producing label to separate function with something like (⚠️  not tested):

```dart
String _resolveLabel() {
  if (vpnStatus.isDisconnected()) return t.ui.fastestServer;

  if (vpnStatus.isConnected()) {
    assert(
      vpnStatus.isMeshnetRouting ||
          (vpnStatus.country != null && vpnStatus.city != null),
      "Status is connected, but we don't know to what we connected",
    );
  }

  if (vpnStatus.isMeshnetRouting) return vpnStatus.hostname ?? vpnStatus.ip ?? "";

  if (vpnStatus.country == null) return t.ui.fastestServer;

  final city = vpnStatus.city != null ? "${vpnStatus.city!}, " : "";
  final virtual = vpnStatus.isVirtualLocation ? " - ${t.ui.virtual}" : "";
  return "$city${vpnStatus.country!.localizedName}$virtual";
}
```

gui/lib/vpn/connection_card_server_info.dart
```diff
@@ -12,39 +12,35 @@ final class ConnectionCardServerInfo extends StatelessWidget {
...
     );
   }
+
+  String _buildServerInfoLabel() {
+    if (vpnStatus.isDisconnected()) {
+      return t.ui.fastestServer;
+    }
+
+    if (vpnStatus.isConnected()) {
+      assert(
```
keliramu, line 32: It's for dev/testing, I assume, and for prod would be good to have this situation logged into log file?

keliramu (approved):

devzbysiu (approved):

PR #1413, opened 2026-03-31
LVPN-10082: Add context menu widget

keliramu:
The purpose of this PR is to introduce context menu widget with easy to use API (just a widget - the use of the widget will be done in a separate PR) - for example, see widgets_showcase.dart.

### DIFF
--- a/gui/lib/theme/context_menu_theme.dart
+++ b/gui/lib/theme/context_menu_theme.dart
@@ -1,0 +1,34 @@
+// ignore_for_file: annotate_overrides
+
+import 'package:flutter/material.dart';
+import 'package:theme_tailor_annotation/theme_tailor_annotation.dart';
+
+part 'context_menu_theme.tailor.dart';
+
+@tailorMixin
+final class ContextMenuTheme extends ThemeExtension<ContextMenuTheme>
+    with _$ContextMenuThemeTailorMixin {
+  final double menuWidth;
+  final BorderRadius menuRadius;
+  final EdgeInsets menuPadding;
+  final Color menuColor;
+  final Color menuBorderColor;
+  final double menuBorderWidth;
+  final double itemHeight;
+  final EdgeInsets itemPadding;
+  final Color itemHoverColor;
+  final TextStyle itemTextStyle;
+
+  ContextMenuTheme({
+    required this.menuWidth,
+    required this.menuRadius,
+    required this.menuPadding,
+    required this.menuColor,
+    required this.menuBorderColor,
+    required this.menuBorderWidth,
+    required this.itemHeight,
+    required this.itemPadding,
+    required this.itemHoverColor,
+    required this.itemTextStyle,
+  });
+}
--- a/gui/lib/theme/context_menu_theme.dart
+++ b/gui/lib/theme/context_menu_theme.dart
@@ -18,6 +18,9 @@ final class ContextMenuTheme extends ThemeExtension<ContextMenuTheme>
   final EdgeInsets itemPadding;
   final Color itemHoverColor;
   final TextStyle itemTextStyle;
+  final Duration animationDuration;
+  final Curve animationCurve;
+  final List<BoxShadow> menuBoxShadow;
 
   ContextMenuTheme({
     required this.menuWidth,
@@ -30,5 +33,8 @@ final class ContextMenuTheme extends ThemeExtension<ContextMenuTheme>
     required this.itemPadding,
     required this.itemHoverColor,
     required this.itemTextStyle,
+    required this.animationDuration,
+    required this.animationCurve,
+    required this.menuBoxShadow,
   });
 }
--- a/gui/lib/widgets/context_menu/context_menu.dart
+++ b/gui/lib/widgets/context_menu/context_menu.dart
@@ -1,0 +1,248 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:nordvpn/theme/aurora_design.dart';
+import 'package:nordvpn/theme/context_menu_theme.dart';
+import 'package:nordvpn/widgets/context_menu/context_menu_item.dart';
+
+export 'package:nordvpn/widgets/context_menu/context_menu_item.dart';
+
+/// A drop-over context menu anchored to an [anchorBuilder] widget.
+///
+/// Opens below the anchor when tapped. Tapping outside the menu closes it
+/// without propagating the tap to elements underneath.
+///
+/// Example:
+/// ```dart
+/// ContextMenu(
+///   items: [
+///     ContextMenuItem(label: 'Option A', onTap: () {}),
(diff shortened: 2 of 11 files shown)


gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,248 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:nordvpn/theme/aurora_design.dart';
```
devzbysiu: Design file should be used only when instantiating themes in `theme.dart` and widgets should get values through theme. 

The reason is because design will be changing and we want to restrict blast radius of such changes. That's why usage of  design values should be terminated on theme so if breaking changes happen, we only need to adjust it in theme, not in all widgets.

dfetti (approved):
Looks good to me, but remove the accessibility things that @devzbysiu mentioned.

gui/lib/widgets/widgets_showcase.dart
```diff
@@ -159,6 +162,124 @@ class _WidgetsShowcaseState extends ConsumerState<WidgetsShowcase> {
...
+                      ),
+                      ContextMenuItem(
+                        label: "Danger zone",
+                        labelColor: context.appTheme.textErrorColor,
+                        onTap: () => logger.i("danger"),
+                      ),
+                    ],
+                    anchorBuilder: (toggleMenu) => ElevatedButton(
+                      onPressed: toggleMenu,
+                      child: const Text("Options"),
```
DoYouEvenCpp: [nit] do we need the other context menus in the showcase?

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,231 @@
...
+    );
+  }
+
+  Widget _buildOverlay(BuildContext context) {
+    final theme = context.contextMenuTheme;
+    final menuWidth = widget.width ?? theme.menuWidth;
+
+    return Stack(
+      children: [
+        // Barrier: absorbs all taps outside the menu panel.
```
DoYouEvenCpp, line 122: This comment seems a bit off?

gui/lib/theme/context_menu_theme.dart
```diff
@@ -0,0 +1,34 @@
...
+  final Color menuColor;
+  final Color menuBorderColor;
+  final double menuBorderWidth;
+  final double itemHeight;
+  final EdgeInsets itemPadding;
+  final Color itemHoverColor;
+  final TextStyle itemTextStyle;
+
+  ContextMenuTheme({
+    required this.menuWidth,
```
DoYouEvenCpp, line 26: Shouldn't we also define menu height?

DoYouEvenCpp (approved):

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,231 @@
+import 'package:flutter/material.dart';
+import 'package:nordvpn/theme/aurora_design.dart';
```
devzbysiu: Design file should be used only when instantiating themes in `theme.dart` and widgets should get values through theme. 

The reason is because design will be changing and we want to restrict blast radius of such changes. That's why usage of  design values should be terminated on theme so if breaking changes happen, we only need to adjust it in theme, not in all widgets.

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,231 @@
...
+    required this.items,
+    required this.width,
+    required this.onItemTapped,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = context.contextMenuTheme;
+    final isDark = Theme.of(context).brightness == Brightness.dark;
+    final shadows = isDark ? AppBoxShadows.darkPopover : AppBoxShadows.lightPopover;
```
devzbysiu: We are handling light / dark theme already in `theme.dart`, if shadows are not adjusted to this mechanism then it should be updated in `AppDesign`.

Here is how `AppDesign` is instantiated in theme: https://github.com/NordSecurity/nordvpn-linux/blob/lvpn-10082-context-menu-widget/gui/lib/theme/theme.dart#L39-L39

gui/lib/widgets/context_menu/context_menu.dart
keliramu: No, it is just what it says - outside click/tap should close the menu, and not impact underlying items.

gui/lib/theme/context_menu_theme.dart
keliramu: Height is determined by number of items, or you have something different in mind?

gui/lib/widgets/context_menu/context_menu.dart
keliramu: Makes sense, yet found such deviation in one more place.

gui/lib/widgets/context_menu/context_menu.dart
keliramu: Addressed as part of fixes to Stanislavs' feedback

gui/lib/widgets/widgets_showcase.dart
keliramu: Left only one

gui/lib/widgets/context_menu/context_menu.dart
keliramu: Addressed

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,233 @@
...
+        duration: theme.animationDuration,
+      );
+      _fadeAnimation = CurvedAnimation(
+        parent: _animationController,
+        curve: theme.animationCurve,
+      );
+      _sizeAnimation = CurvedAnimation(
+        parent: _animationController,
+        curve: theme.animationCurve,
+      );
```
devzbysiu: Those are the same so maybe we could have just one `Animation` object?

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -0,0 +1,233 @@
...
+        vsync: this,
+        duration: theme.animationDuration,
+      );
+      _fadeAnimation = CurvedAnimation(
+        parent: _animationController,
+        curve: theme.animationCurve,
+      );
+      _sizeAnimation = CurvedAnimation(
+        parent: _animationController,
+        curve: theme.animationCurve,
```
devzbysiu: [small] maybe we can have one Animation object as those are the same?

devzbysiu (approved):

gui/lib/widgets/context_menu/context_menu.dart
keliramu: yes, now have the same/one animation object

PR #1417, opened 2026-04-07
LVPN-10168: Fix token renew race condition

thegravi:
Changes:
- Guarding access token renewal process with mutex
- If there is a time de-sync between client and server a force token renewal is performed

### DIFF
--- a/session/access_token_session_store.go
+++ b/session/access_token_session_store.go
@@ -3,6 +3,7 @@ package session
 import (
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
@@ -33,6 +34,7 @@ type AccessTokenSessionStore struct {
 	cfgManager         config.Manager
 	errHandlerRegistry *internal.ErrorHandlingRegistry[error]
 	renewAPICall       AccessTokenRenewalAPICall
+	renewMu            sync.Mutex
 }
 
 // NewAccessTokenSessionStore creates a new AccessTokenSessionStore instance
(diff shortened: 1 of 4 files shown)


bartoszWojciechO (approved):

session/access_token_session_store.go
```diff
@@ -58,17 +60,39 @@ func (s *AccessTokenSessionStore) Renew(opts ...RenewalOption) error {
...
+// Manual tokens cannot be renewed via API.
+func (s *AccessTokenSessionStore) isManualToken() bool {
+	cfg, err := s.getConfig()
+	return err == nil && cfg.ExpiresAt.Equal(ManualAccessTokenExpiryDate)
+}
+
+// doRenew performs the actual token renewal
+func (s *AccessTokenSessionStore) doRenew(options renewalOptions) error {
+	// re-check after acquiring lock - another goroutine may have renewed
+	if !options.forceRenewal {
```
devzbysiu, line 105: Should this check be here? If two goroutines will be called with `forceRenewal=true` and one goroutine goes behind the mutex and does the renew correctly and leaves critical section, then second goroutine will enter critical section and it will skip validation because of `forceRenewal` and will call API again when it's not needed at this point because first goroutine renewed the token 🤔

session/access_token_session_store_race_test.go
```diff
@@ -0,0 +1,232 @@
...
+	}, core.ErrNotFound)
+
+	store := session.NewAccessTokenSessionStore(cfgManager, errorRegistry, api.Renew)
+
+	const numGoroutines = 5
+	var wg sync.WaitGroup
+	renewErrors := make([]error, numGoroutines)
+
+	for i := 0; i < numGoroutines; i++ {
+		wg.Add(1)
```
devzbysiu: You can use `wg.Go(...)` instead of `wg.Add(1) ... wg.Done()` since go 1.25. Same with the test below

session/access_token_session_store_race_test.go
```diff
@@ -0,0 +1,232 @@
...
+		t.Logf("ERROR HANDLER CALLED (would trigger logout): %v", err)
+	}, core.ErrNotFound)
+
+	store := session.NewAccessTokenSessionStore(cfgManager, errorRegistry, api.Renew)
+
+	const numGoroutines = 5
+	var wg sync.WaitGroup
+	renewErrors := make([]error, numGoroutines)
+
+	for i := 0; i < numGoroutines; i++ {
```
devzbysiu: [small] can be done with `range`. Same with the test below

session/access_token_session_store_race_test.go
```diff
@@ -0,0 +1,232 @@
...
+		currentRenewToken: initialRenewToken,
+		requestDelay:      50 * time.Millisecond,
+	}
+}
+
+var ErrRenewTokenNotFound = internal.NewCodedError(101202, "Renew token not found", core.ErrNotFound)
+
+func (api *raceConditionRenewalAPI) Renew(token string, idempotencyKey uuid.UUID) (*session.AccessTokenResponse, error) {
+	api.callCount.Add(1)
+	time.Sleep(api.requestDelay)
```
devzbysiu: What's the purpose of this `Sleep`? If this is for sychronization before critical section then it should be done with channel or something similar, timing is unreliable

devzbysiu (approved):

PR #1420, opened 2026-04-08
LVPN-10113: Trigger pause from GUI

keliramu:
Integrate GUI and Daemon changes for Pause connection feature.


dfetti (approved):

gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -52,6 +60,9 @@ class VpnStatusController extends _$VpnStatusController
 
   Future<void> disconnect() => _doAndShowPopup((vpn) => vpn.disconnect());
 
+  Future<void> pauseConnection(int pauseSeconds) =>
+      _doAndShowPopup((vpn) => vpn.pauseConnection(pauseSeconds));
```
DoYouEvenCpp, line 67: Do we expect any popup to be shown upon calling toast/triggering pause?

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -65,33 +66,28 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
+                onTap: () async =>
+                    await _pauseConnection(ref, pauseConnectionTime1Hour),
               ),
               ContextMenuItem(
                 label: t.ui.pauseFor24Hours,
-                onTap: () async => await ref
-                    .read(vpnStatusControllerProvider.notifier)
-                    .disconnect(), // TODO(LVPN-10113): add proper action
+                onTap: () async =>
+                    await _pauseConnection(ref, pauseConnectionTime24Hours),
```
DoYouEvenCpp, line 105: Shall we register a separate bug to allow toast widget to display > 1hour timeouts correctly?

gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -25,6 +26,13 @@ class VpnStatusController extends _$VpnStatusController
     final status = await ref.read(vpnRepositoryProvider).fetchStatus();
     _registerNotifications();
 
+    if (status.state == ConnectionState.PAUSED) {
+      // show the toast if the gui starts while already paused
```
DoYouEvenCpp, line 30: Scenario:
- user connected
- pause triggered
- user discards the toast
- user restarts GUI
- toast is visible again

Is this something we want to persistently display (the toast, after GUI restart)?

gui/lib/data/providers/vpn_status_controller.dart
keliramu: Yes, it's daemon API call - anything could happen and we should show it.

gui/lib/vpn/connection_card_buttons.dart
keliramu: No, it's been solved.

gui/lib/data/providers/vpn_status_controller.dart
keliramu: It's ok (as per internal discussion).

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -65,33 +66,28 @@ final class ConnectionCardButtons extends ConsumerWidget {
             items: [
               ContextMenuItem(
                 label: t.ui.pauseFor5Min,
-                onTap: () async => await ref
-                    .read(vpnStatusControllerProvider.notifier)
-                    .disconnect(), // TODO(LVPN-10113): add proper action
+                onTap: () async =>
+                    await _pauseConnection(ref, pauseConnectionTime5Min),
```
devzbysiu, line 85: Maybe we could put those labels and time in some collection - like list of tuples for example? Then you could have something like:
```
items: menuItems.map((item) => _buildMenuItem(item)).toList()
```

There are few benefits here:
- less duplication of course
- but also we would have pause time and label defined in one place - so it's less likely that someone will update time without updating label

devzbysiu: Ideal way would be to have the label parameterized (like [this](https://github.com/NordSecurity/nordvpn-linux/blob/lvpn-10113-trigger-pause-from-gui/gui/lib/widgets/toast.dart#L80-L80)) and this same parameter would be used to produce label and time, so changing value would produce proper label and time for the pause, but labels are different depending on the value (`minutes`, `hours`, `hour`), so probably not worth trying to make it this way

gui/lib/theme/theme.dart
```diff
@@ -862,13 +862,13 @@ final class NordVpnTheme {
       backgroundColor: design.semanticColors.bgTertiary,
       spacing: AppSpacing.spacing4,
       borderRadius: AppBorderRadius.md,
-      widgetWidth: 356,
+      widgetWidth: 386,
```
DoYouEvenCpp, line 865: Is this related to moving it slightly upper the position(to not overlap with the footer in GUI' settings)?

keliramu: It is related to showing hours in pause message, when needed

devzbysiu (approved):

gui/lib/vpn/connection_card_buttons.dart
keliramu: This is code optimization request, will try to address it in another PR

PR #1421, opened 2026-04-08
LVPN-10276: Don't set fwmark for DNS requests when VPN is connected

mariusSincovici:
To ensure that the DNS requests are not blocked, the socket is created to use the firewall mark.
When the application is connected to VPN, firewall mark is not set anymore on the socket to ensure that the requests will not bypass the VPN.

### DIFF
--- a/network/connection_helper.go
+++ b/network/connection_helper.go
@@ -11,17 +11,9 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const NO_FW_MARK uint32 = 0
+const noFwMark uint32 = 0
 
-func LookupAddressNoFwmark(addr string, dns string, protocol string) ([]netip.Addr, error) {
-	return LookupAddress(addr, dns, protocol, NO_FW_MARK)
-}
-
-func LookupAddressWithFirewallMark(addr string, dns string, protocol string, fwmark uint32) ([]netip.Addr, error) {
-	return LookupAddress(addr, dns, protocol, fwmark)
-}
-
-func LookupAddress(addr string, dns string, protocol string, fwmark uint32) ([]netip.Addr, error) {
+func lookupAddress(addr string, dns string, protocol string, fwmark uint32) ([]netip.Addr, error) {
 	resolver := net.Resolver{
 		PreferGo: true,
 		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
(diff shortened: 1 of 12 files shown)


network/connection_helper.go
```diff
@@ -11,28 +11,40 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// LookupAddressWithCustomDNS looks up address in a specified DNS server
-func LookupAddressWithCustomDNS(addr string, dns string, protocol string, fwmark uint32) ([]netip.Addr, error) {
+const NO_FW_MARK uint32 = 0
+
+func LookupAddressNoFwmark(addr string, dns string, protocol string) ([]netip.Addr, error) {
```
thegravi: These 3 function are in the same package, so we can un-export them.
Also `LookupAddress` and `LookupAddressWithFirewallMark` are identical, we can tailor mark/nomark with only 2 of these.

network/connection_helper.go
```diff
@@ -11,28 +11,40 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// LookupAddressWithCustomDNS looks up address in a specified DNS server
-func LookupAddressWithCustomDNS(addr string, dns string, protocol string, fwmark uint32) ([]netip.Addr, error) {
+const NO_FW_MARK uint32 = 0
```
thegravi: Either:
- Unexport this
- Or ditch `LookupAddressNoFwmark` and leave single function where this constant can be used for no mark case.

network/resolver.go
```diff
@@ -2,25 +2,49 @@ package network
...
+
+	daemonEvents.Connect.Subscribe(func(dc events.DataConnect) error {
+		resolver.updateVpnStatus(dc.EventStatus == events.StatusSuccess)
+		return nil
+	})
+
+	daemonEvents.Disconnect.Subscribe(func(dd events.DataDisconnect) error {
+		resolver.updateVpnStatus(false)
+		return nil
+	})
```
thegravi, line 45: In some universe  `daemonEvents` can be nil 😃

network/connection_helper.go
mariusSincovici: It can also be only one if it is clear enough :)

network/resolver.go
mariusSincovici: It must not be nil, otherwise the app should crash/not work.

thegravi (approved):

edvyno1 (approved):

PR #1423, opened 2026-04-09
LVPN-10085:  add pause context menu analytics

DoYouEvenCpp:
Add UIEvents being emitted upon pause selection from GUI.

Changes under this PR:
- extended protobufs with new values for `ItemName` and `ItemValue`
- extended accordingly moose' helpers used for the `uievent/middleware`
  - extended unit tests to cover new changes
- introduced new enumeration for GUI holding pause possible lengths, along with a converter from enum to UIEvent's ItemValue


dfetti (approved):

gui/lib/data/models/pause.dart
```diff
@@ -0,0 +1,23 @@
...
+
+enum PauseLength {
+  // values in seconds
+  mins5(5*60),
+  mins15(15*60),
+  mins30(30*60),
+  hour1(60*60),
+  hours24(24*60*60);
+  
+  const PauseLength(this.value);
```
thegravi: [suggestion] What if we would use this type `const PauseLength(this.seconds, this.uiValue)` of ctor? When enum's can be declared: ` mins5(5 * 60, UIEvent_ItemValue.PAUSE_5_MIN)`, and we can eliminate the switch by using const `uiValue` var.

DoYouEvenCpp: Good idea!
Adapted the enumeration.

thegravi (approved):

PR #1424, opened 2026-04-09
LVPN-10110: Add location hint for quick connect

dfetti:
Show location hint for quick connect. The happiest of the paths.

- Adds a gRPC to query the recommended server
- The GUI calls this only once for now when it starts
- The value is used only to display it in disconnected state. When we quick connect, the daemon will do a refetch and will decide on the server.

Tests will be added in the next PR


gui/lib/vpn/connection_card_server_info.dart
```diff
@@ -46,4 +57,16 @@ final class ConnectionCardServerInfo extends StatelessWidget {
     final virtual = vpnStatus.isVirtualLocation ? " - ${t.ui.virtual}" : "";
     return "$city${vpnStatus.country!.localizedName}$virtual";
   }
+
+  String _buildDisconnectedServerInfo(
+    FastestServerLocation? fastestServerLocation,
+  ) {
+    final country = fastestServerLocation?.countryName;
+    final city = fastestServerLocation?.cityName;
+    if (country != null && city != null) {
```
DoYouEvenCpp: Could there be occurrences where GUI is provided with country XOR city?

daemon/rpc_recommended_server.go
```diff
@@ -0,0 +1,47 @@
...
+
+// FastestServer fetches the fastest server location
+func (r *RPC) FastestServer(ctx context.Context, in *pb.Empty) (*pb.FastestServerLocation, error) {
+	var cfg config.Config
+	if err := r.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+	}
+	insights := r.dm.GetInsightsData().Insights
+
+	serverSelection, err := selectServer(r, &insights, cfg, "", "")
```
DoYouEvenCpp, line 21: Question: shall we try to reuse already fetched fastest server here as well? (in case someone's frantically goes between GUI connection cards and settings, for example)

Problem is, this might require some heuristic, which might also be not something quickly implementable. So probably this is a bright future, not something we must consider now. 🤔

daemon/rpc_recommended_server.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/log"
+)
+
+// FastestServer fetches the fastest server location
+func (r *RPC) FastestServer(ctx context.Context, in *pb.Empty) (*pb.FastestServerLocation, error) {
+	var cfg config.Config
+	if err := r.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
```
DoYouEvenCpp, line 16: Shouldn't we return error here instead of proceeding further without/with invalid config?

daemon/rpc_fastest_server.go
```diff
@@ -0,0 +1,47 @@
...
+	if err := r.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+	}
+	insights := r.dm.GetInsightsData().Insights
+
+	serverSelection, err := selectServer(r, &insights, cfg, "", "")
+	if err == nil {
+		country := serverSelection.server.Country()
+		// Cache the recommended server
+		r.lastServerSelection = serverSelection
```
DoYouEvenCpp: Can `lastServerSelection` be simultaneously updated?
Consider scenario:
- users boots their computer
- user made GUI to always run upon OS startup
- user however opens terminal, right after the startup and tries to connect

Won't this cause a (possible) race here?

daemon/rpc_recommended_server.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	insights := r.dm.GetInsightsData().Insights
+
+	serverSelection, err := selectServer(r, &insights, cfg, "", "")
+	if err == nil {
```
bartoszWojciechO, line 22: Can we log this `err` in case it's not `nil`?

daemon/rpc_connect.go
```diff
@@ -26,9 +26,23 @@ func isDedicatedIP(server core.Server) bool {
...
+	return r.lastServerSelection.server != nil && r.lastServerSelection.recommendationUUID != ""
+}
+
 // Connect initiates and handles the VPN connection process
 func (r *RPC) Connect(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer) (retErr error) {
 	r.pauseManager.CancelReconnection()
+
+	// We already fetched and cached the recommended server
+	if isQuickConnect(in) && r.alreadyHaveRecommendation() {
+		return r.connectFromLastSelection(srv, pb.ConnectionSource_MANUAL)
```
bartoszWojciechO: Will the connection fail if the last selected server goes down?

daemon/rpc_recommended_server.go
mariusSincovici, line 21: You cannot use `cfg` if there was an error.

daemon/rpc_fastest_server.go
```diff
@@ -0,0 +1,47 @@
...
+	var cfg config.Config
+	if err := r.cm.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+	}
+	insights := r.dm.GetInsightsData().Insights
+
+	serverSelection, err := selectServer(r, &insights, cfg, "", "")
+	if err == nil {
+		country := serverSelection.server.Country()
+		// Cache the recommended server
```
mariusSincovici: If cache is needed `caching.Cache` can be used, but this will not save for race condition, because there is server pointer inside.

daemon/rpc_fastest_server.go
```diff
@@ -0,0 +1,47 @@
...
+		r.lastServerSelection = serverSelection
+		return &pb.FastestServerLocation{
+			CityName:    country.City.Name,
+			CountryCode: country.Code,
+			CountryName: country.Name,
+		}, nil
+	}
+
+	// We failed to select a server. Check if we already have a cached option
+	if r.alreadyHaveRecommendation() {
```
mariusSincovici: Not knowing when it was cached could result in using something old and invalid.

gui/lib/vpn/connection_card_server_info.dart
```diff
@@ -1,30 +1,41 @@
...
   final VpnStatus vpnStatus;
 
   const ConnectionCardServerInfo({super.key, required this.vpnStatus});
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     final statusCardTheme = context.connectionCardTheme;
 
+    final fastestServerLocation = ref
```
mariusSincovici: Without caching, every time is entered in this screen a new API for recommended servers is made.
Even when user is connected to the VPN.

daemon/rpc_connect.go
```diff
@@ -26,9 +26,23 @@ func isDedicatedIP(server core.Server) bool {
...
+func (r *RPC) alreadyHaveRecommendation() bool {
+	return r.lastServerSelection.server != nil && r.lastServerSelection.recommendationUUID != ""
+}
+
 // Connect initiates and handles the VPN connection process
 func (r *RPC) Connect(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer) (retErr error) {
 	r.pauseManager.CancelReconnection()
+
+	// We already fetched and cached the recommended server
+	if isQuickConnect(in) && r.alreadyHaveRecommendation() {
```
mariusSincovici: Not knowing when the information was cached, it will connect the user to incorrect server location.
Maybe this data should be used as a backup, if recommended servers API fails?

daemon/rpc_connect.go
```diff
@@ -26,9 +26,23 @@ func isDedicatedIP(server core.Server) bool {
 	return index != -1
 }
 
+func isQuickConnect(in *pb.ConnectRequest) bool {
+	return in.GetServerTag() == "" && in.GetServerGroup() == ""
+}
+
+func (r *RPC) alreadyHaveRecommendation() bool {
+	return r.lastServerSelection.server != nil && r.lastServerSelection.recommendationUUID != ""
```
mariusSincovici: This is incorrect. `r.lastServerSelection` contains also the latest connected server, not only the latest recommended server..
e.g.: 
nordvpn c <some_country>
nordvpn c - will swill connect me to the <some_country> not the recommended one.

daemon/rpc_recommended_server.go
mariusSincovici, line 21: It would be good to only store the result if it came from `remote`, otherwise if is returned from local it can be a very bad location.

gui/lib/vpn/connection_card_server_info.dart
dfetti: Not as I recall

daemon/rpc_recommended_server.go
dfetti: There is no such case right now. Right now this gRPC is called only once when GUI starts. Not everytime you switch from settings screen to vpn screen.
Later edit: My bad, I forgot to make keepalive true for the GUI provider and this actually happened, but I will fix it.

daemon/rpc_connect.go
dfetti: Well that's true right now. I should fallback this to the normal flow.

daemon/rpc_recommended_server.go
dfetti: I don't think it's critical. We handle it the same way in connect.

bartoszWojciechO (approved):

gui/lib/data/mocks/daemon/mock_daemon.dart
```diff
@@ -375,4 +375,9 @@ final class MockDaemon extends DaemonServiceBase {
   Future<Payload> pauseConnection(ServiceCall call, PauseRequest request) {
     throw UnimplementedError();
   }
+
+  @override
+  Future<FastestServerLocation> fastestServer(ServiceCall call, Empty request) {
```
DoYouEvenCpp: [minor] this is now going to be called every time on the main screen under a debug run - shall we snap here any dummy implementation to not pollute logs too drastically?

DoYouEvenCpp (approved):

gui/lib/data/mocks/daemon/mock_daemon.dart
dfetti: Yep, I will. I also plan to use it in the tests I'm preparing.

PR #1425, opened 2026-04-10
LVPN-7836: Set tray allows change tray status for any arbitrary os users fix

thegravi:


daemon/rpc_set_tray.go
```diff
@@ -7,15 +7,34 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/NordSecurity/nordvpn-linux/log"
+	"google.golang.org/grpc/peer"
 )
 
 func (r *RPC) SetTray(ctx context.Context, in *pb.SetTrayRequest) (*pb.Payload, error) {
+	peer, ok := peer.FromContext(ctx)
```
mariusSincovici: [small] this seams like a duplicate from rpc_set_notify.go.
It would have been nice to have this code in a single place and use in both.

mariusSincovici (approved):

daemon/rpc_set_tray.go
keliramu: `rpc_settings.go` and `rpc_state.go` also has the same logic, now it is in 4 places the same logic with small variations, maybe it is good time for DRYing it?

keliramu (approved):

PR #1427, opened 2026-04-10
LVPN-10165: update service quality events for pause

bartoszWojciechO:


cmd/daemon/events_linux.go
```diff
@@ -14,41 +14,42 @@ import (
...
+func (*dummyAnalytics) NotifyDebuggerEvent(events.DebuggerEvent) error       { return nil }
+func (*dummyAnalytics) NotifyRequestAPI(events.DataRequestAPI) error         { return nil }
+func (*dummyAnalytics) NotifyUiItemsClick(events.UiItemsAction) error        { return nil }
+func (*dummyAnalytics) NotifyHeartBeat(time.Duration) error                  { return nil }
+func (*dummyAnalytics) NotifyDeviceLocation(core.Insights) error             { return nil }
+func (*dummyAnalytics) NotifyLANDiscovery(bool) error                        { return nil }
+func (*dummyAnalytics) NotifyVirtualLocation(bool) error                     { return nil }
+func (*dummyAnalytics) NotifyPostquantumVpn(bool) error                      { return nil }
+func (*dummyAnalytics) NotifyAppStartTime(int64) error                       { return nil }
+func (*dummyAnalytics) NotifyPauseCancelled(events.DataPauseCancelled) error { return nil }
```
thegravi: Let's keep on moving towards `On..` pattern :)

cmd/daemon/main.go
```diff
@@ -435,11 +435,12 @@ func main() {
 	connectionInfo := state.NewConnectionInfo()
 	statePublisher := state.NewState()
 	internalVpnEvents.Subscribe(connectionInfo)
-	connectionInfo.Subscribe(statePublisher)
+	connectionInfo.SubscribeToInternalStateChanges(statePublisher)
 	daemonEvents.Service.Connect.Subscribe(connectionInfo.ConnectionStatusNotifyConnect)
 	daemonEvents.Service.Disconnect.Subscribe(connectionInfo.ConnectionStatusNotifyDisconnect)
 	daemonEvents.User.Subscribe(statePublisher)
 	configEvents.Subscribe(statePublisher)
+	connectionInfo.SubscribeToPauseCancelled(analytics)
```
DoYouEvenCpp: [ultra nit] can you please move it up a few lines, so all the connectionInfo registrations are grouped together?

daemon/rpc_pause.go
```diff
@@ -35,3 +35,7 @@ func (r *RPC) PauseConnection(ctx context.Context, in *pb.PauseRequest) (*pb.Pay
 
 	return &pb.Payload{Type: internal.CodeSuccess}, nil
 }
+
+func (r *RPC) CancellPause() {
```
DoYouEvenCpp: A typo?

```suggestion
func (r *RPC) CancelPause() {
```

daemon/state/connection_info.go
```diff
@@ -207,12 +225,42 @@ func (c *ConnectionInfo) Pause(pausedAt time.Time, duration time.Duration) {
...
-func (c *ConnectionInfo) Unpause() {
+// Unpause unsets the pause data, sends pause cancelled notification and returns the pause duration.
+func (c *ConnectionInfo) CancelPause(cancelStartTime time.Time) time.Duration {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	var pauseInterval time.Duration
+	if c.pauseData != nil {
+		pauseInterval = c.unpause()
+		duration := time.Since(cancelStartTime)
```
DoYouEvenCpp: Probably this is negligible, but shouldn't we calculate time interval since `cancelStartTime` the moment we compose the event, in `moose.go`?

daemon/state/connection_info.go
```diff
@@ -207,12 +225,42 @@ func (c *ConnectionInfo) Pause(pausedAt time.Time, duration time.Duration) {
 	}
 }
 
-// Unpause unsets the pause data.
-func (c *ConnectionInfo) Unpause() {
+// Unpause unsets the pause data, sends pause cancelled notification and returns the pause duration.
```
DoYouEvenCpp:
```suggestion
// CancelPause unsets the pause data, sends pause cancelled notification and returns the pause duration.
```

daemon/state/connection_info.go
```diff
@@ -207,12 +225,42 @@ func (c *ConnectionInfo) Pause(pausedAt time.Time, duration time.Duration) {
...
+
+// Unpause unsets the pause data and returns the pause duration.
+func (c *ConnectionInfo) Unpause() time.Duration {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return c.unpause()
+}
+
+func (c *ConnectionInfo) unpause() time.Duration {
```
DoYouEvenCpp: Maybe let's introduce slightly different name, to make it more distinct from `Unpause`?
```suggestion
func (c *ConnectionInfo) doUnpause() time.Duration {
```

daemon/state/connection_info.go
```diff
@@ -207,12 +225,42 @@ func (c *ConnectionInfo) Pause(pausedAt time.Time, duration time.Duration) {
 	}
 }
 
-// Unpause unsets the pause data.
-func (c *ConnectionInfo) Unpause() {
+// Unpause unsets the pause data, sends pause cancelled notification and returns the pause duration.
+func (c *ConnectionInfo) CancelPause(cancelStartTime time.Time) time.Duration {
```
DoYouEvenCpp: Could this be tested, with golang tests - some scenario to cover that `CancelPause` actuall changes internal state of the connection info, and triggers new event?

daemon/reconnection_scheduler.go
```diff
@@ -13,10 +13,10 @@ import (
 
 type ReconnectScheduler interface {
 	ScheduleReconnection(duration time.Duration)
-	CancelReconnection()
+	CancelReconnection() time.Duration
```
DoYouEvenCpp, line 16: Do we need to make those public API?
This is all being called in the context of the RPC, so possibly it might be private ones?

daemon/rpc_connect.go
```diff
@@ -147,14 +150,22 @@ func (r *RPC) connectWithStoredServerSelection(ctx context.Context, srv pb.Daemo
...
+		false,
+		pauseDuration)
 }
 
 func (r *RPC) connectWithParameters(ctx context.Context,
 	in *pb.ConnectRequest,
 	srv pb.Daemon_ConnectServer,
 	source pb.ConnectionSource,
 ) (didFail bool, retErr error) {
+	pauseDuration := r.pauseManager.CancelReconnection()
```
DoYouEvenCpp, line 168: So now with scenario:
- user connects
- user trigger pause
- pause finishes
- user disconnects
- user finally connects to any location (`nordvpn connect UK`)

And because we never resets to nil  `reconnectCancelFunc` in https://github.com/NordSecurity/nordvpn-linux/blob/LVPN-10165-pause-events/daemon/reconnection_scheduler.go once pause has been gracefully handled, we'll be calling here the whole logic inside [CancelReconnection](https://github.com/NordSecurity/nordvpn-linux/blob/LVPN-10165-pause-events/daemon/reconnection_scheduler.go#L68) right?

If I'm not wrong, then probably this is redundant in this case (a fix would be to reset the reconnection callback back to nil, once pause is handled and finished).

daemon/state/connection_info.go
bartoszWojciechO: This is done for parity with `Connect`/`Disconnect` events, for those we also calculate the event duration before the notification call. Do you think it should be moved into the moose wrapper?

daemon/reconnection_scheduler.go
bartoszWojciechO: Needs to be exported for common mocking(we store common mocks in a separate package).

daemon/rpc_connect.go
bartoszWojciechO: Done. Fixed with a channel that's closed by the reconnect goroutine rather than clearing the callback for more thread safety.

events/moose/moose.go
```diff
@@ -867,6 +886,34 @@ func (s *Subscriber) NotifyDisconnect(data events.DataDisconnect) error {
...
+			EventTrigger:  moose.NordvpnappEventTriggerUser,
+		},
+		moose.TargetConnectionParams{
+			TargetServerListSource:    serverListOriginToInternalType(data.ServerFromAPI),
+			TargetServerSelectionRule: serverSelectionRuleToInternalType(data.ServerSelectionRule),
+			TargetServerType:          moose.NordvpnappServerTypeNone,
+		},
+		moose.ConnectionOnPauseParams{
+			ConnectionFunnel:     connectionFunnel,
+			VpnConnectionTrigger: moose.NordvpnappVpnConnectionTriggerAfterSnooze,
```
DoYouEvenCpp: This should be `NordvpnappVpnConnectionTriggerNone`

daemon/reconnection_scheduler.go
thegravi, line 39: Can this be called twice? If so, there would be 2 go routines racing for that channel.

daemon/state/connection_info.go
DoYouEvenCpp: Nah, I guess this is all fine now.

daemon/reconnection_scheduler.go
```diff
@@ -64,14 +69,27 @@ func (s *ReconnectSchedulerImpl) ScheduleReconnection(duration time.Duration) {
...
 	defer s.mu.Unlock()
-	if s.reconnectCancelFunc != nil {
+
+	if s.reconnectCancelFunc != nil && s.isReconnectionScheduled() {
 		log.Println(internal.DebugPrefix, "cancelling the reconnection after a pause")
-		s.connectionInfo.Unpause()
+		cancelStartTime := time.Now()
 		s.reconnectCancelFunc()
 		s.reconnectCancelFunc = nil
+		return s.connectionInfo.CancelPause(cancelStartTime)
```
thegravi: Isn't this meaningless, since time delta between `cancelStartTime := time.Now()` and `return s.connectionInfo.CancelPause(cancelStartTime)` is < ms?

daemon/reconnection_scheduler.go
DoYouEvenCpp: Oh, right. Good point!

daemon/reconnection_scheduler.go
bartoszWojciechO: Good point. Changed it so that:

- The goroutine gets a reference to the channel, it won't touch `s.reconnectionScheduledChan` directly
- On subsequent `ScheduleReconnection` call, new channel is generated and stored in `s.reconnectionScheduledChan`
- When the goroutine is cancelled, it calls `close` on the reference to the old channel

DoYouEvenCpp (approved):

dfetti (approved):

DoYouEvenCpp (approved):

thegravi (approved):

PR #1428, opened 2026-04-10
LVPN-10305: update copy for GUI settings footer

DoYouEvenCpp:
This PR updates copy texts in GUI's settings, in its footer.


mariusSincovici (approved):

gui/lib/settings/settings_wrapper_widget.dart
```diff
@@ -150,7 +150,7 @@ Widget _buildVpnStatus(BuildContext context) {
         final vpnStatus = status.value!;
         if (vpnStatus.isDisconnected() || vpnStatus.isPaused()) {
           return Text(
-            t.ui.notConnected,
```
devzbysiu, line 153: Can this be removed now or it's still used somewhere?

devzbysiu (approved):

thegravi (approved):

gui/lib/settings/settings_wrapper_widget.dart
DoYouEvenCpp: Good point! Removed (was actually used in one place, but that's just a disabled test case).

bartoszWojciechO (approved):

PR #1432, opened 2026-04-14
LVPN-10303: Handle paused connection toast message visibility based o…

keliramu:
In case of paused connection, make paused connection toast message not to be shown in blocking screens: loading screen, login screen, error screen, snap missing permissions screen.

### DIFF
--- a/gui/lib/router/metadata.dart
+++ b/gui/lib/router/metadata.dart
@@ -11,6 +11,7 @@ abstract class RouteMetadata with _$RouteMetadata {
   const factory RouteMetadata({
     required AppRoute route,
     required Widget screen,
+    required bool isBlocking,
     String? displayName,
     Function(BuildContext)? onPressed,
   }) = _RouteMetadata;
--- a/gui/lib/router/routes.dart
+++ b/gui/lib/router/routes.dart
@@ -22,6 +22,7 @@ import 'package:nordvpn/settings/vpn_connection_settings.dart';
 import 'package:nordvpn/vpn/vpn.dart';
 import 'package:nordvpn/widgets/widgets_showcase.dart';
 
+final Map<String, RouteMetadata> routeRegistry = {};
 final Map<String, RouteMetadata> routeToNameMap = {};
 
 enum AppRoute {
@@ -46,15 +47,6 @@ enum AppRoute {
   loadingScreen,
   errorScreen,
   consentScreen;
-
-  bool get blocksToast => switch (this) {
-    AppRoute.errorScreen ||
-    AppRoute.login ||
-    AppRoute.loadingScreen ||
-    AppRoute.consentScreen =>
-      true,
-    _ => false,
-  };
 
   @override
   String toString() {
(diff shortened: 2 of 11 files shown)


gui/lib/router/routes.dart
```diff
@@ -47,6 +47,15 @@ enum AppRoute {
   errorScreen,
   consentScreen;
 
+  bool get blocksToast => switch (this) {
```
devzbysiu: This is not used

gui/lib/widgets/toasts_listener.dart
```diff
@@ -11,12 +12,19 @@ final class ToastsListener extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    Duration? timeout = ref.watch(toastsProvider);
+    final Duration? timeout = ref.watch(toastsProvider);
+    final redirectState = ref.watch(redirectStateProvider);
+    final isBlocking =
```
devzbysiu: Would it be possible to go like that?
- have another method on `GoRouterExt`, like `String currentRouteId()`
- then add `required` `isBlocking` field or similar in `RouteMetadata` and set this field in  `configureRoutes()`
- then in `ToastsListener`, something like:
```dart
final isBlocking = routeToNameMap[context.currentRouteId()].isBlocking;
```

The reasoning behind this is that if we'll add new blocking screen, we need to remember that this list exists here and needs to be updated, with field on `RouteMetadata` devs will be forced to fill in this field while adding new route and we automatically have invisible toast on new blocking route for free without the risk of regression

gui/lib/router/routes.dart
keliramu: Removed

gui/lib/widgets/toasts_listener.dart
keliramu: Optimized code, now routes are defined in single place with all metadata.

gui/lib/router/metadata.dart
```diff
@@ -11,6 +11,7 @@ abstract class RouteMetadata with _$RouteMetadata {
   const factory RouteMetadata({
     required AppRoute route,
     required Widget screen,
+    required bool isBlocking,
```
DoYouEvenCpp, line 14: Why not make it an optional c-tor parameter, and default it to `false` - this way, we'd need to explicitly assign it only when we actually want specific route be a blocking. What do you think?

```suggestion
    @Default(false) bool isBlocking,
```

DoYouEvenCpp (approved):

gui/lib/router/metadata.dart
keliramu: IMO, mandatory/explicit option/field would make developer think on what is really needed rather than relying blindly on some default.

devzbysiu: Agree with @keliramu. In this case, if we'll add blocking screen and don't specify that screen is blocking, then we have regression and required param will force devs to specify this so it's safer IMO

DoYouEvenCpp: Yeah, this is really a minor one, so nothing blocking from my side :smile:

dfetti (approved):

devzbysiu (approved):

PR #1434, opened 2026-04-15
LVPN-10224: fix firewall python tests

edvyno1:
This PR replaces iptables based firewall/allowlisting python tests with behavioral tests to be able to test with nft implementation

### DIFF
--- a/test/qa/lib/firewall.py
+++ b/test/qa/lib/firewall.py
@@ -1,459 +1,164 @@
 import os
-import re
+import socket
+import struct
+import time
 
 import sh
 
-from . import Port, Protocol, daemon, logging, dns
+from . import Port, Protocol, ssh
 
 IP_ROUTE_TABLE = 205
 ENDPOINTS = "endpoints"
+SOCK_TIMEOUT = 5
+TCP_DST_PORT = 1234
+UDP_DST_PORT = 1235
+DST_ADDR = "qa-peer"
+# This is used when communicating with the socket server
+# on the qa-peer container, used in port allowlisting behavioral testing
+try:
+    IP = socket.gethostbyname(DST_ADDR)
+except Exception as e: # noqa: BLE001
+    print(e)
+    print("Unable to get qa-peer addr, does it exist?")
 
-# Rules for killswitch
-# mangle
-# -A PREROUTING -i {iface} -m comment --comment nordvpn -m comment --comment drop-IPv4 -j DROP
-# -A POSTROUTING -o {iface} -m comment --comment nordvpn -m comment --comment drop-IPv4 -j DROP
-
-# Rules for firewall
-# mangle
-# -A PREROUTING -i {iface} -m connmark --mark 0xe1f1 -m comment --comment nordvpn -m comment --comment api_allowlist -j ACCEPT
-# -A PREROUTING -i {iface} -m comment --comment nordvpn -m comment --comment drop-IPv4 -j DROP
-# -A POSTROUTING -o {iface} -m mark --mark 0xe1f1 -m comment --comment nordvpn -m comment --comment api_allowlist -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff
-# -A POSTROUTING -o {iface} -m connmark --mark 0xe1f1 -m comment --comment nordvpn -m comment --comment api_allowlist -m comment --comment nordvpn -j ACCEPT
-# -A POSTROUTING -o {iface} -m comment --comment nordvpn -m comment --comment drop-IPv4 -j DROP
-
-# Rules for allowlisted subnet
-# mangle
-# -A PREROUTING -s {subnet_ip} -i {iface} -m comment --comment nordvpn -m comment --comment allowlist_subnets -j ACCEPT
-# -A PREROUTING -i {iface} -m connmark --mark 0xe1f1 -m comment --comment nordvpn -m comment --comment api_allowlist -j ACCEPT
-# -A PREROUTING -i {iface} -m comment --comment nordvpn -m comment --comment drop-IPv4 -j DROP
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: nft
+      triggered-ref: LVPN-10224_fix_firewall_python_tests
```
edvyno1: change back after gitlab merge

test/qa/lib/firewall.py
```diff
@@ -1,459 +1,163 @@
...
-    logging.log("\nCurrent rules:")
-    for rule in current_rules:
-        print(f"{rule}")
-        logging.log(rule)
-
-    print()
+    print(out)
     print(sh.nordvpn.settings())
-    return current_rules == expected_rules
+    return "nordvpn" in out
```
mariusSincovici, line 59: [small] the check is not really needed, because nft will fail if table is not found

test/qa/lib/firewall.py
```diff
@@ -1,459 +1,163 @@
...
+        s.connect((IP, TCP_DST_PORT))
+        s.send(b"ping")
+        print("TCP data sent")
+        data = s.recv(4096)
+        print("data received: ", data)
+        retval = True
+    except TimeoutError:
+        print(f"timeout of {SOCK_TIMEOUT} hit with TCP, source port {src_port}, dst port : {TCP_DST_PORT}",)
+        retval = False
+    # `OSError: [Errno 99] Cannot assign requested address` thrown when unable to connect
```
mariusSincovici: The comment is not correct.
When `errno 99` is returned it means that the `bind` actually failed with success 😆

test/qa/lib/firewall.py
```diff
@@ -1,459 +1,163 @@
...
+    except OSError as e:
+        print(e)
+        retval = False
+    s.close()
```
mariusSincovici: if it fails do we need to close the port, like for UDP? 🤔

test/qa/lib/firewall.py
```diff
@@ -1,459 +1,163 @@
...
-
-    filter_fw_lines = os.popen("sudo iptables -S -t filter").read()
-    filter_fw_list = filter_fw_lines.split("\n")[3:-1]
-    fw_list = mangle_fw_list + filter_fw_list
-
-    dns_full = dns.DNS_NORD + dns.DNS_TPL
-
-    return [rule for rule in fw_list if not any(dns in rule for dns in dns_full)]
-
+def is_ip_routed_via_VPN(subnets: list[str]) -> bool:
```
mariusSincovici, line 67: [optional]  not to hardcode, to use `get_network_interface`

mariusSincovici (approved):

test/qa/lib/firewall.py
edvyno1: will add this to finally block as with UDP

test/qa/lib/firewall.py
edvyno1: I would prefer to keep this func with 1 arg and not have to pass tech every time, but if you think it's better, I can switch it to that as well

test/qa/lib/firewall.py
edvyno1: yep, this should always return true, but just in case nft api somehow changes and table not being there does not error out, this would still return false, so I prefer keeping it this way

test/qa/lib/firewall.py
edvyno1: edited comment, weird how the bind is outside the try block, and I think this specific error happens on the connect line of code, but did not dig into what is under the hood

mariusSincovici: It is a bit strange, because even if you print the socket info it will say is bound to the correct IP and port, but when sending is actually not working because bind fails :smile:

test/qa/test_allowlist_subnet.py
```diff
@@ -50,11 +50,11 @@ def test_connect_allowlist_subnet(tech, proto, obfuscated):
     ip_provider_addresses = socket.gethostbyname_ex(urlparse(lib.API_EXTERNAL_IP).netloc)[2]
     ip_addresses_with_subnet = [ip + CIDR_32 for ip in ip_provider_addresses]
     allowlist.add_subnet_to_allowlist(ip_addresses_with_subnet)
-    assert firewall.is_active(None, ip_addresses_with_subnet)
+    assert not firewall.is_ip_routed_via_VPN([ip_addresses_with_subnet])
```
devzbysiu: Isn't `ip_addresses_with_subnet` already a list?

test/qa/test_firewall.py
```diff
@@ -137,16 +144,19 @@ def test_firewall_03_allowlist_ports_range(tech, proto, obfuscated, ports):
...
             lib.set_firewall("on")
             allowlist.add_ports_to_allowlist([ports])
-            assert not firewall.is_active([ports])
+            assert not firewall.is_active()
+            assert firewall.is_source_port_reachable([ports])
 
             sh.nordvpn.connect()
             assert network.is_connected()
-            assert firewall.is_active([ports])
+            assert firewall.is_active
```
devzbysiu: `is_active()`?

edvyno1: looks like python doesn't care, tests were running and passing, anyway fixed it now

test/qa/test_allowlist_subnet.py
edvyno1: fixed all occurences of this double list wrapping

devzbysiu (approved):

test/qa/test_firewall.py
devzbysiu: It's probably checking if the reference is not None which is always true :man_shrugging:

PR #1436, opened 2026-04-15
LVPN-10250: add transparency to bgGlass color

bartoszWojciechO:


gui/lib/theme/aurora_design.dart
```diff
@@ -189,10 +189,8 @@ final class AppSemanticColorsLight implements SemanticColors {
   @override
   Color get bgOverlay => Color(0x80141415);
 
-  // TODO (dfe) LVPN-10250: Change back to Color(0xB3F7F7F8) after figuring out how to make
-  // transparency work as it should.
   @override
-  Color get bgGlass => Color(0xFFF7F7F8);
+  Color get bgGlass => Color.fromRGBO(0xF7, 0xF7, 0xF8, 0xB3);
```
dfetti, line 193: Does it really work? :o 
You should also change it in the AppSemanticColorsDark

PR #1437, opened 2026-04-15
Fix async callbacks causing state access on a dead provider

keliramu:
Fix async callbacks (_pingDaemon, stream listeners) continued to execute after the ProviderContainer was disposed, causing state access on a dead provider.
Failure log: https://github.com/NordSecurity/nordvpn-linux/actions/runs/24440994843/job/71405741638?pr=1427#step:9:4251


keliramu:
Included in another PR

PR #1447, opened 2026-04-20
LVPN-3630: Log level watcher

devzbysiu:
This PR adds runtime file in `/run/nordvpn/loglevel` (or `/var/snap/nordvpn/common/run/nordvpn/loglevel` for snap) that allows controlling log level in the application.

Example:
```bash
echo "debug" | sudo tee /run/nordvpn/loglevel
```
Valid values are `debug`, `info`, `warn`, `error`, `fatal` and `off`.

Changes:

- readme:
  - added `## Troubleshooting`
  - updated `The application manages:` as it was outdated
  - the rest is formatting

- man page to mention log level config

- `log.SetupLogger` function the configures watcher on runtime file and adjust log level on changes
- `log.DefaultLevel` to set default level to `info`, or `debug` (for dev builds)
- log funcs like `log.Info`, `log.Warn` - there is a separate PR with replacing current `log.PrintXYZ` with those new ones

### DIFF
--- a/filewatch/filewatch.go
+++ b/filewatch/filewatch.go
@@ -21,7 +21,7 @@ func GetFileWatcher(pathsToMonitor ...string) (watcher *fsnotify.Watcher, err er
 
 	for _, file := range pathsToMonitor {
 		if err := watcher.Add(file); err != nil {
-			return nil, fmt.Errorf("adding group file to watcher: %w", err)
+			return nil, fmt.Errorf("adding file to watcher: %w", err)
 		}
 	}
 
--- a/filewatch/filewatch.go
+++ b/filewatch/filewatch.go
@@ -1,0 +1,29 @@
+package filewatch
+
+import (
+	"fmt"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+// GetFileWatcher returns a fsnotify file watcher that is monitoring files provided in pathsToMonitor
+func GetFileWatcher(pathsToMonitor ...string) (watcher *fsnotify.Watcher, err error) {
+	watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("creating new watcher: %w", err)
+	}
+
+	defer func() {
+		if err != nil && watcher != nil {
+			_ = watcher.Close()
+		}
+	}()
+
+	for _, file := range pathsToMonitor {
+		if err := watcher.Add(file); err != nil {
+			return nil, fmt.Errorf("adding group file to watcher: %w", err)
+		}
+	}
+
+	return watcher, nil
+}
--- a/log/log_level_watcher.go
+++ b/log/log_level_watcher.go
@@ -1,0 +1,120 @@
+package log
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+// WatchLevelFile starts a goroutine that watches path for writes and updates
+// the active log level whenever the file content changes. The file should
+// contain one of "debug", "info", or "error" (case-insensitive). The parent
+// directory is watched so that file creation is also detected. The goroutine
+// stops when the returned CancelFunc is called.
+func WatchLevelFile(path string) (CancelFunc, error) {
+	Info("setting log level watcher on path", path)
+	if err := os.MkdirAll(filepath.Dir(path), 0o750); err != nil {
+		return nil, err
+	}
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+
+	applyLevelFile(path)
+
+	if err := watcher.Add(filepath.Dir(path)); err != nil {
+		if closeErr := watcher.Close(); closeErr != nil {
+			Warn("closing log level watcher:", closeErr)
+		}
+		return nil, err
+	}
+
+	go func() {
+		defer func() {
+			if err := watcher.Close(); err != nil {
+				Warn("closing log level watcher:", err)
+			}
+		}()
+		var debounce <-chan time.Time
+
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				if event.Name != path {
+					continue
+				}
+				if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
+					debounce = time.After(50 * time.Millisecond)
+				}
+			case <-debounce:
+				Info("applying log level from file")
+				applyLevelFile(path)
+			case _, ok := <-watcher.Errors:
+				if !ok {
+					return
+				}
+			}
+		}
+	}()
+
+	return func() {
+		if err := watcher.Close(); err != nil {
+			Warn("closing log level watcher:", err)
+		}
+	}, nil
+}
+
+func applyLevelFile(path string) {
+	root, err := os.OpenRoot(filepath.Dir(path))
+	if err != nil {
+		return
+	}
+	defer func() {
+		if err := root.Close(); err != nil {
+			Errorf("failed to close dir root '%s': %v", filepath.Dir(path), err)
+		}
+	}()
+
+	f, err := root.Open(filepath.Base(path))
+	if err != nil {
+		return
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			Errorf("failed to close file '%s': %v", path, err)
+		}
+	}()
+
+	data, err := io.ReadAll(f)
+	if err != nil {
+		return
+	}
+
+	text := strings.TrimSpace(strings.ToLower(string(data)))
+	switch text {
+	case "debug":
+		SetLevel(levelDebug)
+	case "info":
+		SetLevel(levelInfo)
+	case "warn":
+		SetLevel(levelWarn)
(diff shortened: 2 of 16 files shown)


log/logger.go
```diff
@@ -19,6 +95,91 @@ func SetFlags(flag int) {
 	log.SetFlags(flag)
 }
 
+func Debug(v ...any) {
+	if atLevel(levelDebug) {
```
DoYouEvenCpp: [minor]
Maybe by providing helper methods:
```
func logAt(l logLevel, prefix string, v []any) {
    if atLevel(l) {
        output(logMsg(prefix, v))
    }
}

func logAtf(l logLevel, prefix, format string, v []any) {
    if atLevel(l) {
        output(fmt.Sprintf(prefix+" "+format, v...))
    }
}
```

we can get rid of this repeating `if` condition?

To finally  have:

```
func Info(v ...any)                    { logAt(levelInfo, infoPrefix, v) }
func Infof(format string, v ...any)    { logAtf(levelInfo, infoPrefix, format, v) }
func Warn(v ...any)                    { logAt(levelWarn, warningPrefix, v) }
func Warnf(format string, v ...any)    { logAtf(levelWarn, warningPrefix, format, v) }
func Defer(v ...any)                   { logAt(levelInfo, deferPrefix, v) }
func Deferf(format string, v ...any)   { logAtf(levelInfo, deferPrefix, format, v) }
```

What do you think?

log/log_level_watcher.go
```diff
@@ -0,0 +1,120 @@
...
+// contain one of "debug", "info", or "error" (case-insensitive). The parent
+// directory is watched so that file creation is also detected. The goroutine
+// stops when the returned CancelFunc is called.
+func WatchLevelFile(path string) (CancelFunc, error) {
+	Info("setting log level watcher on path", path)
+	if err := os.MkdirAll(filepath.Dir(path), 0o750); err != nil {
+		return nil, err
+	}
+
+	watcher, err := fsnotify.NewWatcher()
```
DoYouEvenCpp: You can use https://github.com/NordSecurity/nordvpn-linux/blob/main/internal/filesystem.go#L623 here. The `applyLevelFile(path)` in between probably won't be a problem?

log/logger.go
```diff
@@ -1,16 +1,92 @@
...
+	if err != nil {
+		Error("failed to start log level watcher:", err)
+		return func() {}
+	}
+	return cancel
+}
+
+func SetLevel(l logLevel) {
+	Info("setting log level to", l)
+	level.Store(uint32(l))
```
DoYouEvenCpp, line 87: Does it make sense to check if new call to `SetLevel` actually changes the level/or check it in the `log_level_watcher`?

log/log_level_watcher.go
devzbysiu: I didn't use it on purpose because of the circular dep, but can take a look if we can re-arrange stuff to avoid it

log/logger.go
devzbysiu: Yeah, we can go like that. Need to adjust the depth for `output` though to report correct source of the log

log/logger.go
devzbysiu: I think it doesn't really matter, but won't hurt either

devzbysiu: Actually it won't give us anything as I still have to read the file and figure out the level, so the only thing we would get is that `Store` is not called and maybe the log "setting log level to .." does not appear, but it's not worth it, so I'll leave it as is

log/logger.go
devzbysiu: Addressed

log/log_level_watcher.go
devzbysiu: Proposal https://github.com/NordSecurity/nordvpn-linux/pull/1454

log/logger.go
DoYouEvenCpp: OK, reasonable!

log/log_level_watcher.go
devzbysiu: Addressed

DoYouEvenCpp (approved):

log/logger.go
```diff
@@ -1,16 +1,92 @@
...
+
+// CancelFunc stops a logger resource such as the level file watcher.
+type CancelFunc func()
+
+var level atomic.Uint32
+
+// DefaultLevel returns LevelDebug for dev environments and LevelInfo otherwise.
+func DefaultLevel(devEnv bool) logLevel {
+	return levelDebug
+}
```
thegravi: Ignores argument.

log/log_level_watcher.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/filewatch"
+	"github.com/fsnotify/fsnotify"
+)
+
+// WatchLevelFile starts a goroutine that watches path for writes and updates
+// the active log level whenever the file content changes. The file should
+// contain one of "debug", "info", or "error" (case-insensitive). The parent
```
thegravi: Description stated more levels: `debug, info, warn, error, fatal`

log/logger.go
```diff
@@ -1,16 +1,92 @@
...
+		Error("failed to start log level watcher:", err)
+		return func() {}
+	}
+	return cancel
+}
+
+func SetLevel(l logLevel) {
+	Info("setting log level to", l)
+	level.Store(uint32(l))
+}
```
thegravi, line 88: In case we have a lower level, the log message would not be printed before the change, and will be silently dropped. Is this by design?

filewatch/filewatch.go
```diff
@@ -0,0 +1,29 @@
...
+
+	defer func() {
+		if err != nil && watcher != nil {
+			_ = watcher.Close()
+		}
+	}()
+
+	for _, file := range pathsToMonitor {
+		if err := watcher.Add(file); err != nil {
+			return nil, fmt.Errorf("adding group file to watcher: %w", err)
```
thegravi: what's `group` here?

log/log_level_watcher.go
```diff
@@ -0,0 +1,114 @@
...
+// directory is watched so that file creation is also detected. The goroutine
+// stops when the returned CancelFunc is called.
+func WatchLevelFile(path string) (CancelFunc, error) {
+	Info("setting log level watcher on path", path)
```
thegravi: Perhaps this should sound like "trying to" because there is no "set/done" later. And e.g. on first error (MkdirAll) nothing is set.

log/logger.go
```diff
@@ -19,6 +95,47 @@ func SetFlags(flag int) {
...
+func Fatalf(format string, v ...any) {
+	logAtf(levelFatal, fatalPrefix, format, v)
+	os.Exit(1)
+}
+
+func logAt(l logLevel, prefix string, v []any) {
+	if level.Load() <= uint32(l) {
+		msg := fmt.Sprintln(v...)
+		// -1 to trim trailing \n
+		output(prefix + " " + msg[:len(msg)-1])
```
thegravi: What if something more robust?
```suggestion
        // trim trailing \n
        msg := strings.TrimRight(fmt.Sprintln(v...), "\n")                                                                                                                                        
        output(prefix + " " + msg) 
```

log/log_level_watcher.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	applyLevelFile(path)
+
+	go func() {
+		defer func() {
+			if err := watcher.Close(); err != nil {
+				Warn("closing log level watcher:", err)
+			}
+		}()
```
thegravi, line 35: Do we really need this and cannot get by with returned cancel func? Nothing buggish, but which is the true owner of canceling? :D

log/logger.go
```diff
@@ -1,16 +1,92 @@
...
+		fallthrough
+	default:
+		return "unknown"
+	}
+}
+
+// CancelFunc stops a logger resource such as the level file watcher.
+type CancelFunc func()
+
+var level atomic.Uint32
```
thegravi, line 64: this gets `levelUnknown` by default and passes any filter until setup is called. Wouldn't that make any issues?

log/logger.go
devzbysiu: Addressed

log/log_level_watcher.go
devzbysiu: Addressed

log/logger.go
devzbysiu: Yes

filewatch/filewatch.go
devzbysiu: The same as it was before code was moved. Removed.

log/log_level_watcher.go
devzbysiu: Addressed

log/logger.go
devzbysiu: Changed

log/log_level_watcher.go
devzbysiu: In services, cancel func may never be called

log/logger.go
devzbysiu: It will print all messages, so same as we have now. In case of an issue with setup logger, we have all logs for debugging

thegravi (approved):

PR #1450, opened 2026-04-21
LVPN-10355: fix context menu spacing and width

keliramu:
LVPN-10355: Fix context menu spacing and width

### DIFF
--- a/gui/lib/theme/theme.dart
+++ b/gui/lib/theme/theme.dart
@@ -825,7 +825,7 @@ final class NordVpnTheme {
 
   ContextMenuTheme _contextMenuThemeExt() {
     return ContextMenuTheme(
-      menuWidth: 260,
+      menuWidth: 268,
       menuRadius: AppBorderRadius.md,
       menuPadding: EdgeInsets.symmetric(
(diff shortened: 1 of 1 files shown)


gui/lib/theme/theme.dart
```diff
@@ -850,7 +850,7 @@ final class NordVpnTheme {
           ? AppBoxShadows.lightPopover
           : AppBoxShadows.darkPopover,
       menuShadowMargin: 6,
-      menuGap: 4,
+      menuGap: 8,
```
dfetti: Can't we use AppSpacing? From Aurora?

dfetti (approved):

gui/lib/theme/theme.dart
DoYouEvenCpp: +1, `spacing2` should fit in here nicely.

DoYouEvenCpp (approved):

gui/lib/theme/theme.dart
keliramu: Thx, addressed

PR #1451, opened 2026-04-21
LVPN-10360: Ignore meshnet changes that are not relevant to firewall

mariusSincovici:
Not to reconfigure the firewall at every meshnet refresh, the meshnet map is compared with the last one that was used to configure the firewall. If the fields relevant for firewall didn't changed then the change is ignored.

### DIFF
--- a/daemon/firewall/service.go
+++ b/daemon/firewall/service.go
@@ -71,22 +71,20 @@ func (m *MeshInfo) IsSimilar(meshInfo *MeshInfo) bool {
 				p1.DoIAllowRouting == p2.DoIAllowRouting &&
 				p1.DoIAllowLocalNetwork == p2.DoIAllowLocalNetwork &&
 				p1.DoIAllowFileshare == p2.DoIAllowFileshare
-
 		},
 	)
 
-	if !arePeersEqual {
-		return false
-	}
-
-	return true
+	return arePeersEqual
 }
 
 func NewMeshInfo(meshnetMap mesh.MachineMap, meshInterface string) *MeshInfo {
-	return &MeshInfo{
+	info := &MeshInfo{
 		MeshnetMap:    meshnetMap,
 		MeshInterface: meshInterface,
 	}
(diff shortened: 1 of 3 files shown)


daemon/firewall/service.go
```diff
@@ -30,16 +33,58 @@ type Config struct {
...
+func (m *MeshInfo) IsSimilar(meshInfo *MeshInfo) bool {
+	if meshInfo == nil {
+		return false
+	}
+
+	if m.MeshInterface != meshInfo.MeshInterface {
+		return false
+	}
+
+	sortMeshnetMap(&meshInfo.MeshnetMap)
```
devzbysiu: This will change the `MeshnetMap` of the arg that is passed in, probably would be safer to pass `MeshInfo` by value to `IsSimilar` as this method is just asking question "is this similar" so may be surprising that it also changes something in passed arg

daemon/firewall/service.go
```diff
@@ -83,6 +128,9 @@ func WithTunnelInterface(tunnelInterface string) Option {
 func WithMeshnetInfo(meshInfo *MeshInfo) Option {
 	return func(c *Config) {
 		c.MeshnetInfo = meshInfo
+		if meshInfo != nil {
+			sortMeshnetMap(&c.MeshnetInfo.MeshnetMap)
```
devzbysiu: Maybe we don't need this sort here, since the ctor is already sorting it?

devzbysiu (approved):

daemon/firewall/service.go
mariusSincovici: Removed all the sorting and the check is on the lists.

daemon/firewall/service.go
mariusSincovici: The problem with the passed param was even if it is sent by value, it will still not make a copy on the []peers.
To prevent issues, I've just remove the sorting completely.

PR #1453, opened 2026-04-23
LVPN-10375: refactor meshnet key into device key

bartoszWojciechO:
Connection to the dedicated servers will be established using keys generated on the client's side instead of NordLynx key received from the core. For simplicity, we will be using the same key as the one generated for meshnet connections.


device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+// found in the provided config.
+func (d *DeviceKeyManagerImpl) getDeviceKey(cfg config.Config) (string, bool) {
+	if cfg.DeviceKey == "" {
+		return d.keyGenerator.Private(), true
+	}
+
+	return cfg.DeviceKey, false
+}
+
+// RegisterMeshnet registers the device key for meshnet if it isn't registered and returns true if it is successfully
```
devzbysiu: Wrong name, should be `CheckAndRegisterMeshnet`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	var cfg config.Config
+	if err := d.configManager.Load(&cfg); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+		return false
+	}
+
+	if isMeshnetRegistrationInfoCorrect(cfg) && cfg.DeviceKey != "" {
```
devzbysiu: Looks like `isMeshnetRegistrationInfoCorrect` already checks for `cfg.DeviceKey != ""`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+	if err != nil {
+		return fmt.Errorf("registering device key: %w", err)
+	}
+
+	if err := d.configManager.SaveWith(keyConfig(newConfig)); err != nil {
+		return err
+	}
+
+	if !isMeshnetRegistrationInfoCorrect(newConfig) {
+		return fmt.Errorf("meshnet registration failure")
```
devzbysiu: Would it be possible to add more details why it failed? Maybe it should be `func checkMeshnetRegistrationInfo(..) error` or something similar 🤔

device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+		// There is a delay in the backend between registering a new key and when that key is recognized, so we need to wait
+		// some time, otherwise connection will fail.
+		const delayAfterNewKey time.Duration = time.Second * 5
+		d.delayFunc(delayAfterNewKey)
+	}
+
+	return cfg, nil
+}
+
+func (d *DeviceKeyManagerImpl) registerDedicateServer(deviceKey string,
```
devzbysiu: [typo] `Dedicated`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+	if errors.Is(err, core.ErrConflict) {
+		// We try to keep the same keys as long as possible, but if relogin with different account happens
+		// then they have to be regenerated. There's no way to check if the current mesh device data
+		// belongs to this account or not, so handling this on registering error is the best approach.
+		deviceKey = d.keyGenerator.Private()
+		cfg = invalidateKeyData(cfg)
+
+		uuid, uuidParseErr := uuid.Parse(resp.UUID)
+		if uuidParseErr != nil {
+			return cfg, fmt.Errorf("parsing UUID: %w", err)
```
devzbysiu: Shouldn't be `uuidParseErr`?

core/simple_client_api.go
```diff
@@ -434,6 +440,54 @@ func (api *SimpleClientAPI) Server(id int64) (*Server, error) {
...
+		return DevicesResponse{}, fmt.Errorf("marshaling the request data: %w", err)
+	}
+	// TODO: replace MockServerBaseURL with api.baseURL once the real API becomes available
+	req, err := request.NewRequestWithBearerToken(http.MethodPatch, api.agent, MockServerBaseURL, fmt.Sprintf(DevicesUpdateURL, deviceUUID.String()), "application/json", "", "gzip, deflate", bytes.NewBuffer(data), token)
+	if err != nil {
+		return DevicesResponse{}, fmt.Errorf("creating nc credentials request: %w", err)
+	}
+	resp, err := api.do(req)
+	if err != nil {
+		return DevicesResponse{}, fmt.Errorf("executing HTTP POST request: %w", err)
```
devzbysiu: `PATCH` instead of `POST`?

core/urls.go
```diff
@@ -174,4 +174,14 @@ const (
 
 	// RepoTypeTest defines non-production (qa, development) repo type
 	RepoTypeTest = "nordvpn-test"
+
+	DedicatedServersURL = "/v1/dedicated-servers"
+
+	DevicesURL = DedicatedServersURL + "/devices"
+
+	// /v1/dedicated-servers/<device-uuid>
+	DevicesUpdateURL = DedicatedServersURL + "/devices/%s"
```
devzbysiu: Maybe `DevicesURL + "/%s"`?

core/urls.go
```diff
@@ -174,4 +174,14 @@ const (
 
 	// RepoTypeTest defines non-production (qa, development) repo type
 	RepoTypeTest = "nordvpn-test"
+
+	DedicatedServersURL = "/v1/dedicated-servers"
+
+	DevicesURL = DedicatedServersURL + "/devices"
+
+	// /v1/dedicated-servers/<device-uuid>
```
devzbysiu: This will be `/v1/dedicated-servers/devices/<device-uuid>`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,310 @@
...
+
+type DedicatedServersKeyManager interface {
+	CheckAndRegisterDedicatedServers() bool
+}
+
+// DeviceKeyManagerImpl does both registration checks and registration, if it's not done.
+type DeviceKeyManagerImpl struct {
+	configManager       config.Manager
+	keyGenerator        KeyGenerator
+	meshnetRegistery    cmesh.Registry
```
devzbysiu: [typo] `Registery`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+	return nil
+}
+
+type registerFunc func(deviceKey string,
+	distroName string,
+	isKeyNew bool,
+	cfg config.Config) (config.Config, error)
+
+// CheckAndRegisterDedicatedServers checks if device key is registered for the dedicated servers and registers it if it
+// isn't. Return true if the key was successfully registered.
```
DoYouEvenCpp: Something odd with the wording here, maybe we can go with a slightly modified comment:

```suggestion
// CheckAndRegisterDedicatedServers checks if device key is registered for the dedicated servers and registers it if not.
// Returns true if the key was successfully registered.
```

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func invalidateKeyData(cfg config.Config) config.Config {
+	cfg.MeshDevice = nil
+	cfg.DeviceUUID = uuid.Nil
+	return cfg
+}
+
+// getDeviceKey returns device key from cfg or generates a new key if it doesn't exist. Returns true if key was not
+// found in the provided config.
```
DoYouEvenCpp: What about stating here a tuple is returned, first element is the key, while the second holds true if a new key got generated, otherwise false?

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return fmt.Errorf("registering device key: %w", err)
+	}
+
+	if err := d.configManager.SaveWith(keyConfig(newConfig)); err != nil {
+		return err
+	}
+
+	if err := isMeshnetRegistrationInfoCorrect(newConfig); err != nil {
+		return fmt.Errorf("meshnet registration info is not correct after registration: %w", err)
```
DoYouEvenCpp: I feel this (the error coming from `isMeshnetRegistrationInfoCorrect`)  might be wrapped within too many layers already.
First layer is `isMeshnetRegistrationInfoCorrect`
Then second layer is here, `ForceRegisterMeshnet`
Then third layer could be `StartMeshnet`
So finally we can end up with `error: reason: more detailed reason` - I doubt its readability might be a bit problematic while reading logs.
What do you think?

devzbysiu (approved):

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+	cmesh "github.com/NordSecurity/nordvpn-linux/core/mesh"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/log"
+	sysinfo "github.com/NordSecurity/nordvpn-linux/sysinfo"
+
+	"github.com/google/uuid"
+)
+
+// KeyGenerator for use in device key generation.
+type KeyGenerator interface {
```
mariusSincovici, line 25: Isn't this duplicated from `type KeyGenerator interface {` mesh.go?

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+func (d *DeviceKeyManagerImpl) registerKey(cfg config.Config,
+	registerFunc registerFunc) (config.Config, error) {
+	deviceKey, newKey := d.getDeviceKey(cfg)
+	if newKey {
+		cfg = invalidateKeyData(cfg)
+	}
+
+	distroName, err := sysinfo.GetHostOSName()
+	if err != nil {
+		return cfg, err
```
DoYouEvenCpp: Isn't  returning `nil, err` a safer option here (also a few lines below) - worst case scenario, someone who still tries to use value with error returned hits an early null dereference, rather than risking app running in an invalid (unexpected) state due to erroneous config returned?

mariusSincovici (approved):

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+		deviceKey = d.keyGenerator.Private()
+		cfg = invalidateKeyData(cfg)
+
+		uuid, uuidParseErr := uuid.Parse(resp.UUID)
+		if uuidParseErr != nil {
+			return cfg, fmt.Errorf("parsing UUID: %w", uuidParseErr)
+		}
+
+		resp, err = d.dedicatedServersAPI.UpdateDevice(uuid, core.UpdateDeviceRequest{
+			PublicKey: deviceKey,
```
keliramu: Should be public key - but you are passing in a private key? because on the line 292 `deviceKey = ...Private()`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+
+type DedicatedServersKeyManager interface {
+	CheckAndRegisterDedicatedServers() bool
+}
+
+// DeviceKeyManagerImpl does both registration checks and registration, if it's not done.
+type DeviceKeyManagerImpl struct {
+	configManager       config.Manager
+	keyGenerator        KeyGenerator
+	meshnetReggistry    cmesh.Registry
```
keliramu: typo `meshnetReggistry` -> `meshnetRegistry`

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+	Private() string
+	// Public expects base64 encoded private key and returns base64 encoded public key
+	Public(string) string
+}
+
+// DelayFunc blocks the app for a duration of time
+type DelayFunc func(duration time.Duration)
+
+type MeshnetDeviceKeyManager interface {
+	// RegisterMeshnet checks if the device key is registered for meshnet and registers it if it isn't.
```
keliramu: Update comment to match method name

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+		return false
+	}
+
+	if err := isMeshnetRegistrationInfoCorrect(cfg); err == nil {
+		return true
+	}
+
+	newConfig, err := d.registerKey(cfg, d.registerMeshnet)
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "failed to register new device key: %s", err)
```
keliramu: `Println` does not consume `%s` - use `Printf` instead

device_key/device_key_manager.go
```diff
@@ -0,0 +1,327 @@
...
+		return false
+	}
+
+	if err := d.configManager.SaveWith(keyConfig(newConfig)); err != nil {
+		log.Println(internal.ErrorPrefix, err)
+		return false
+	}
+
+	if err := isMeshnetRegistrationInfoCorrect(newConfig); err != nil {
+		log.Println(internal.ErrorPrefix, "registration failed: %s", err)
```
keliramu: `Println` does not consume `%s` - use `Printf` instead

keliramu (approved):

PR #1456, opened 2026-04-24
LVPN-10278: Accept tunnel traffic before DNS drop

devzbysiu:
## Context
- for main changes:
  - when container is started first, then VPN is connected, container does not have DNS updated so it uses local resolver I - this causes DNS traffic to be dropped in forward chain, so DNS wasn't working in containers
  - also there was no rule that would accept forwarded response so even correct order didn't work

## Changes

There are three changes in this PR

[Main changes](https://github.com/NordSecurity/nordvpn-linux/pull/1456/changes/a0eeaf19338fdc97424d5f4525767e23810fa0a0):
- tunnel traffic is accepted quicker, before DNS is dropped, so as long as DNS goes through the tunnel, it's ok
- also established traffic is accepted on forward, so responses come back so now container vs VPN doesn't matter

Additional changes:
- [reordered](https://github.com/NordSecurity/nordvpn-linux/pull/1456/changes/f4c1d9da1f63498ad515ee16f30586eb1710f276) output chain to optimize local and VPN traffic
- [removed](https://github.com/NordSecurity/nordvpn-linux/pull/1456/changes/c80ff4c335fc1c3461416936a18cfbb97f5420a6) TCP/UDP allowlist rules from forward chain

### DIFF
--- a/daemon/firewall/nft/nft.go
+++ b/daemon/firewall/nft/nft.go
@@ -361,7 +361,31 @@ func (n *nft) addForwardChain(config firewall.Config, nftCtx *nftContext) {
 			),
 			UserData: userdata.AppendString(nil, userdata.TypeComment, "traffic to mesh peer"),
 		})
+	}
 
+	if len(config.TunnelInterface) > 0 {
+		// oif "nordtun" accept
+		n.conn.AddRule(&nftables.Rule{
+			Table: nftCtx.table,
+			Chain: forwardChain,
+			Exprs: buildRules(
+				&expr.Verdict{Kind: expr.VerdictAccept},
+				checkInterfaceName(config.TunnelInterface, ifNameOutput, expr.CmpOpEq),
+			),
+			UserData: userdata.AppendString(nil, userdata.TypeComment, "internet to allowlist IPs"),
+		})
+
+		// iif "nordtun" ct state established,related accept
+		n.conn.AddRule(&nftables.Rule{
+			Table: nftCtx.table,
+			Chain: forwardChain,
+			Exprs: buildRules(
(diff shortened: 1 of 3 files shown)


mariusSincovici (approved):

mariusSincovici (approved):

daemon/firewall/nft/nft.go
```diff
@@ -361,47 +367,58 @@ func (n *nft) addForwardChain(config firewall.Config, nftCtx *nftContext) {
...
+		// oif "nordtun" accept
 		n.conn.AddRule(&nftables.Rule{
 			Table: nftCtx.table,
 			Chain: forwardChain,
 			Exprs: buildRules(
 				&expr.Verdict{Kind: expr.VerdictAccept},
-				checkIPIsInSet(nftCtx.allowlistSubnets, matchDest),
+				checkInterfaceName(config.TunnelInterface, ifNameOutput, expr.CmpOpEq),
 			),
 			UserData: userdata.AppendString(nil, userdata.TypeComment, "internet to allowlist IPs"),
```
thegravi: Update comment

daemon/firewall/nft/nft_test.go
```diff
@@ -65,3 +68,68 @@ func TestConfigure(t *testing.T) {
 		})
 	}
 }
+
```
thegravi, line 71: There is a change to forwarding, can we have test for these rules:
- `oif <tun> accept`
- `iif <tun> ct state established,related accept`

daemon/firewall/nft/nft.go
devzbysiu: Updated

PR #1459, opened 2026-04-27
LVPN-10375: Perserve body of an API response in case of specific error

bartoszWojciechO:
In case of device registration, device ID is still returned in case of 409(Conflict) errors. This ID needs to be used to update the device's data.

Normally, body of an response was always consumed in case of errors so we need some way to perserve it. Methods used for api calls and error handling(SimpleClientAPI.do and ExtractError) was expanded to take an accepted error code upon which response body should be perserved as an argument.

### DIFF
--- a/core/errors.go
+++ b/core/errors.go
@@ -50,8 +50,10 @@ type apiError struct {
 //
 // if an error was returned, do not try to read a response again.
 func ExtractError(resp *http.Response) error {
-	return extractError(resp, -1)
+	return extractError(resp, NoAcceptedErrorCode)
 }
+
+const NoAcceptedErrorCode = -1
 
 // extractError from the response if it exist.
 //
--- a/core/errors.go
+++ b/core/errors.go
@@ -1,9 +1,11 @@
 package core
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/NordSecurity/nordvpn-linux/internal"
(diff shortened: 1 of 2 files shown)


core/simple_client_api.go
```diff
@@ -95,11 +95,21 @@ func (api *SimpleClientAPI) request(path, method string, data []byte, token stri
...
-// do request regardless of the authentication.
-func (api *SimpleClientAPI) do(req *http.Request) (*http.Response, error) {
+// doRequest request regardless of the authentication.
+//
+// Response's body will be consumed if an error code is returned. Do not try to read the body in this case.
+func (api *SimpleClientAPI) doRequest(req *http.Request) (*http.Response, error) {
+	return api.do(req, -1)
+}
+
+// doRequest request regardless of the authentication.
```
DoYouEvenCpp:
```suggestion
// do request regardless of the authentication.
```

core/simple_client_api.go
```diff
@@ -450,7 +460,7 @@ func (api *SimpleClientAPI) RegisterDevice(token string, deviceRequest DevicesRe
 	if err != nil {
 		return DevicesResponse{}, fmt.Errorf("creating nc credentials request: %w", err)
 	}
-	resp, err := api.do(req)
+	resp, err := api.do(req, 409)
```
DoYouEvenCpp:
```suggestion
	resp, err := api.do(req, http.StatusConflict)
```

DoYouEvenCpp (approved):

core/errors.go
```diff
@@ -48,6 +50,16 @@ type apiError struct {
 //
 // if an error was returned, do not try to read a response again.
 func ExtractError(resp *http.Response) error {
+	return extractError(resp, -1)
```
thegravi: [nit] I would still make a constant for `-1`, just to eliminate magic numbers.

thegravi (approved):

PR #1460, opened 2026-04-27
LVPN-9936: Reduce firewall reconfigure at network change

mariusSincovici:
* At network changes remove some of firewall reconfigure, because it is set to block all.
* Monitor only interfaces that have default route for VPN reconnect
* Make again `nordvpn d` cleanup the firewall, if is executed when not connected to VPN and killswitch is off.
* Change kill switch type from bool to enum to better understand its state from networker

### DIFF
--- a/daemon/vpn/quench/libquench.go
+++ b/daemon/vpn/quench/libquench.go
@@ -176,6 +176,7 @@ func (q *Quench) Start(ctx context.Context, creds vpn.Credentials, server vpn.Se
 
 	opts := quenchBindigns.NewVnicOptions()
 	opts.SetFwmark(q.fwmark)
+	// Set infinite timeout so the application can control the start/stop of the VPN connection
 	opts.SetConnectionTimeout(math.MaxUint64)
 
 	vnic, err := quenchBindigns.VnicFromName(q.vnicName, q.observer, &opts)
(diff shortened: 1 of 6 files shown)


devzbysiu (approved):

daemon/vpn/quench/libquench.go
```diff
@@ -175,6 +176,7 @@ func (q *Quench) Start(ctx context.Context, creds vpn.Credentials, server vpn.Se
 
 	opts := quenchBindigns.NewVnicOptions()
 	opts.SetFwmark(q.fwmark)
+	opts.SetConnectionTimeout(math.MaxUint64)
```
keliramu, line 180: What problem does it solve to set almost infinite timeout?

mariusSincovici: It is actually infinite. Otherwise the library will stop retrying to connected to VPN. 
But the stopping must be controlled by the application and must never stop when there is a reconnect in the background.

mariusSincovici: Also, having this will make NordWhisper behave like NordLynx and OpenVPN.

keliramu: Pls make some comment in code about this

thegravi (approved):

daemon/vpn/quench/libquench.go
mariusSincovici: Added

PR #1461, opened 2026-04-28
LVPN-10344: show connection details button when connected to meshnet peer

keliramu:
LVPN-10344: show connection details button when connected to meshnet peer; 
LVPN-10254: make button secure my connection text not bold; 
Add on-tap ripple effect to buttons secure my connection and cancel (same as pause connection and connection details);

### DIFF
--- a/gui/lib/vpn/connection_card_buttons.dart
+++ b/gui/lib/vpn/connection_card_buttons.dart
@@ -23,6 +23,14 @@ final class ConnectionCardButtons extends ConsumerWidget {
   static const disconnectMenuItemKey = Key("disconnectMenuItem");
   static const disconnectButtonKey = Key("vpnDisconnectButton");
 
+  static const _pauseLengths = [
+    PauseLength.mins5,
+    PauseLength.mins15,
+    PauseLength.mins30,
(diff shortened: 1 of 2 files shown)


dfetti (approved):

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -195,4 +183,39 @@ final class ConnectionCardButtons extends ConsumerWidget {
   Future<void> _pauseConnection(WidgetRef ref, PauseLength pauseLength) async {
     ref.read(vpnStatusControllerProvider.notifier).pauseConnection(pauseLength);
   }
+
+  static String _pauseLabel(PauseLength pause) => switch (pause) {
```
DoYouEvenCpp, line 187: Shouldn't this land in `gui/lib/data/models/pause.dart` ?

DoYouEvenCpp (approved):

gui/lib/vpn/connection_card_buttons.dart
keliramu: These labels are used only here - imo, no point to have them elsewhere.

PR #1464, opened 2026-04-28
LVPN-10430: Add fwmark to mqtt nc connection

keliramu:
Add fwmark to mqtt nc connection

### DIFF
--- a/nc/nc.go
+++ b/nc/nc.go
@@ -10,7 +10,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
@@ -20,6 +22,7 @@ import (
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
 	mqttp "github.com/eclipse/paho.mqtt.golang/packets"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -129,6 +132,7 @@ type Client struct {
 	subjectPeerUpdate events.Publisher[[]string]
 	credsFetcher      CredentialsGetter
 	retryDelayFunc    CalculateRetryDelayForAttempt
+	fwmark            uint32
 
 	startMu          sync.Mutex
 	started          bool
@@ -143,6 +147,7 @@ func NewClient(
 	subjectErr events.Publisher[error],
 	subjectPeerUpdate events.Publisher[[]string],
 	credsFetcher CredentialsGetter,
+	fwmark uint32,
 ) *Client {
 	return &Client{
 		clientBuilder:     clientBuilder,
@@ -151,6 +156,7 @@ func NewClient(
 		subjectPeerUpdate: subjectPeerUpdate,
 		credsFetcher:      credsFetcher,
 		retryDelayFunc:    network.ExponentialBackoff,
+		fwmark:            fwmark,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


nc/nc.go
```diff
@@ -162,6 +173,46 @@ type mqttMessage struct {
...
+	if err != nil {
+		return "", "", fmt.Errorf("resolving %s: %w", host, err)
+	}
+	if len(ips) == 0 {
+		return "", "", fmt.Errorf("no IPs resolved for %s", host)
+	}
+
+	ip := ips[0]
+	var resolvedHost string
+	if ip.Is6() {
```
mariusSincovici: IPv6 will be blocked/disabled in the system when connected to VPN, won't this break the connection?

nc/nc.go
```diff
@@ -162,6 +173,46 @@ type mqttMessage struct {
...
+// (for TLS ServerName verification), and any error that occurred.
+func (c *Client) resolveEndpoint(endpoint string) (string, string, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return "", "", fmt.Errorf("parsing endpoint URL: %w", err)
+	}
+
+	host := u.Hostname()
+	port := u.Port()
+	if port == "" {
```
mariusSincovici: Is this really needed? Because so far we didn't had to do this 🤔

nc/nc.go
```diff
@@ -162,6 +173,46 @@ type mqttMessage struct {
...
+		switch u.Scheme {
+		case "ssl", "tls", "mqtts", "tcps":
+			port = "8883"
+		default:
+			port = "1883"
```
mariusSincovici: Probably this should not be used in production since is usually unencrypted.

nc/nc.go
```diff
@@ -170,12 +221,56 @@ func (c *Client) createClientOptions(
...
 	opts.SetUsername(credentials.Username)
 	opts.SetPassword(credentials.Password)
 	opts.SetClientID(credentials.UserID.String())
 	opts.SetConnectTimeout(timeout)
 
+	// Set TLS config with original hostname for certificate verification when using resolved IP
+	if originalHostname != "" {
+		// #nosec G402 -- minimum TLS version is controlled by the standard library
+		opts.SetTLSConfig(&tls.Config{
+			ServerName: originalHostname,
```
mariusSincovici: isn't `RootCAs: pool,` also needed?

keliramu: no, `nil` means `go` goes with system's root CA pool

nc/nc.go
keliramu: ok

nc/nc.go
keliramu: removed

nc/nc.go
keliramu: made to ignore IPv6 addresses (with log msg)

nc/nc.go
```diff
@@ -170,12 +179,69 @@ func (c *Client) createClientOptions(
...
+				}
+			} else {
+				// Fallback to original endpoint if resolution fails
+				log.Println(logPrefix, "DNS resolution failed, using original endpoint:", resolveErr)
+				opts.AddBroker(credentials.Endpoint)
+			}
+		} else {
+			opts.AddBroker(credentials.Endpoint)
+		}
+	} else {
```
devzbysiu: Is this correct? If we have error when parsing `credentials.Endpoint`, then we use it anyway? 🤔

nc/nc.go
```diff
@@ -170,12 +179,69 @@ func (c *Client) createClientOptions(
...
+		opts.AddBroker(credentials.Endpoint)
+	}
```
devzbysiu, line 218: Small suggestion - I think this whole block could be significantly simplified by moving broker URLs preparation to separate func and using De Morgan's to revert conditions and exit early, something like that (⚠️ not tested) :
```go
func (c *Client) brokerURLs(credentials config.NCData) []string {
	u, err := url.Parse(credentials.Endpoint)
	if err != nil {
		return []string{credentials.Endpoint}  // <-------------   pls double check this as per other comment
	}

	// Resolve hostname to IPs using resolver with fwmark (bypasses killswitch).
	// Add each IP as a separate broker - MQTT library will try them sequentially.
	if c.resolver == nil {
		return []string{credentials.Endpoint}
	}

	ips, resolveErr := c.resolver.Resolve(u.Hostname())
	if resolveErr != nil || len(ips) == 0 {
		return []string{credentials.Endpoint}
	}

	return c.composeURLs(ips, u)
}

func (*Client) composeURLs(ips []netip.Addr, u *url.URL) []string {
	urls := []string{}
	for _, ip := range ips {
		if ip.Is6() {
			log.Println(logPrefix, "got IPv6 address:", ip, " ignore.")
			continue
		}
		brokerURL := fmt.Sprintf("%s://%s", u.Scheme, ip.String())
		if u.Port() != "" {
			brokerURL = fmt.Sprintf("%s:%s", brokerURL, u.Port())
		}
		urls = append(urls, brokerURL)
	}
	return urls
}
```

And it's easier to read with those preconditions:
- can't parse endpoint? - return default
- no resolver? - return default
- can't resolve? - return default
- then the actual meat

keliramu: Simplified, but on the same base, you probably missed one extra case where url parse outcome is used for tls setup as well - so cannot move it away.

mariusSincovici (approved):

dfetti (approved):

PR #1469, opened 2026-04-29
LVPN-10278: Add snapshot tests for fw

devzbysiu:
Context:
- to defend against unexpected ruleset changes, we are doing snapshot tests for our NFT rules

Changes:
- [add](https://github.com/NordSecurity/nordvpn-linux/commit/0346f1be4eb30a38d9b84b5fb952b08522469a04) snapshot tests, including support for updating golden files via `UPDATE_GOLDEN` env var
- [remove](https://github.com/NordSecurity/nordvpn-linux/commit/2753bcf326851bb7a5ac151b064903f3cbf4038f) previous test checking rules as those are covered now by snapshot tests
- [new](https://github.com/NordSecurity/nordvpn-linux/commit/a2c356b37b9f09003b492f462b6a1d6c479a55cb) `Snapshot (golden file) tests` section in `TESTING.md`
- [formatting and linter warns](https://github.com/NordSecurity/nordvpn-linux/commit/2fdec76b728031e97669cf53f44c498e4ef06eb8) of `TESTING.md`


mariusSincovici (approved):

TESTING.md
```diff
@@ -1,63 +1,111 @@
...
+
+Some tests verify the exact output of a subsystem by comparing it against a
+stored golden file rather than writing per-field assertions. The nftables
+firewall ruleset tests in `daemon/firewall/nft/` use this approach: each
+subtest renders the full `nft list table` output and diffs it against a file
+under `daemon/firewall/nft/testdata/<TestFunctionName>/<subtest_name>.nft`.
+
+When the expected output changes (e.g. after modifying firewall rule
+generation), regenerate the golden files by running the full cgo test suite
+with `UPDATE_GOLDEN_FILES=1`:
```
edvyno1, line 71: [optional] maybe have some option to have this on by default, I can see a lot of potential time wasted pushing code and then forgetting to regenerate only to be told by CI later, having it on by default would still have you consciously needing to commit the newly regenerated golden files

edvyno1: it's possible to also just have this variable in `.env` to serve the same purpose of having it on by default, maybe a small mention of this in the .env.sample could be nice

edvyno1 (approved):

TESTING.md
devzbysiu: I want it to be conscious effort to update the files, because you need to know what you are doing when changing the rules. With defaulting to update, it's too easy to have all the job done by automation and just commit without paying attention what changed and why :sweat_smile: 

Also if we go with default to update and the tests were not run locally, then CI will fail anyway and you'll have to run tests locally and commit the changes so we won't save the time this way

devzbysiu: Added a note to `.env.sample` as you mentioned

PR #1471, opened 2026-04-30
LVPN-10376: add device sync flow

bartoszWojciechO:
Device sync flow is triggered upon user login and on app startup. If user has dedicated servers service, daemon will call the device registration endpoint in the API to register the device key.


test/mock/devicekey/device_key_manager.go
mariusSincovici, line 1: filename spelling `devcice_key_manager.go`

daemon/device_sync.go
```diff
@@ -0,0 +1,22 @@
+package daemon
+
+import "fmt"
+
+// SyncDevice registers the device for dedicated servers if dedicated servers service is available to the user.
```
mariusSincovici: [small] Based on the description, this function is registering the device, not syncing it 🤔 . Also it is not working with anything, it is specially made for dedicated servers.

Wouldn't be more appropriate something like `RegisterDedicatedServers` or similar?

cmd/daemon/main.go
```diff
@@ -753,6 +753,9 @@ func main() {
 
 	if ok, _ := authChecker.IsLoggedIn(); ok {
 		go daemon.StartNC("[startup]", notificationClient)
+		if err := rpc.SyncDevice(); err != nil {
```
mariusSincovici: Can this be an issue for meshnet startup, when kill switch is on? 🤔

test/mock/devicekey/device_key_manager.go
devzbysiu, line 1: [typo] `devcice`

auth/auth.go
```diff
@@ -204,6 +208,21 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	services, err := r.fetchServices()
+	if err != nil {
+		return false, fmt.Errorf("fetching services: %w", err)
+	}
+
+	return slices.ContainsFunc(services, func(service core.ServiceData) bool {
+		return service.ID == DedicatedIPServiceID && r.expChecker.IsExpired(service.ExpiresAt)
```
devzbysiu: This should check `DedicatedServersServiceID` right? 😄

devzbysiu: Also shouldn't be `!IsExpired()`?

mariusSincovici: wrong condition?

daemon/rpc_login.go
```diff
@@ -214,6 +218,10 @@ func (r *RPC) LoginOAuth2(ctx context.Context, in *pb.LoginOAuth2Request) (paylo
 	// (dont forget to reset it after login/signup is completed)
 	r.initialLoginType.set(in.GetType())
 
+	if err := r.SyncDevice(); err != nil {
```
devzbysiu: Shouldn't this be called in `LoginOAuth2Callback`?

daemon/rpc_login.go
```diff
@@ -35,6 +35,10 @@ func (r *RPC) LoginWithToken(ctx context.Context, in *pb.LoginWithTokenRequest)
 		}, nil
 	}
 
+	if err := r.SyncDevice(); err != nil {
```
devzbysiu: This should probably be called when we know that user is logged in successfully right? And we know this only after we have success response from `loginWithToken` call below 🤔

auth/auth.go
```diff
@@ -36,11 +37,14 @@ type Checker interface {
 	// GetDedicatedIPServices returns all available server IDs, if server is not selected by the user it will set
 	// ServerID for that service to NoServerSelected
 	GetDedicatedIPServices() ([]DedicatedIPService, error)
+	// HasDedicatedServerService
```
devzbysiu: No value in this doc comment 😄

auth/auth.go
```diff
@@ -204,6 +208,21 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
 	return dipServices, nil
 }
 
+// HasDedicatedServerService
```
devzbysiu: Same here, no value in this comment

daemon/rpc_login.go
```diff
@@ -35,6 +35,10 @@ func (r *RPC) LoginWithToken(ctx context.Context, in *pb.LoginWithTokenRequest)
...
+		log.Println(internal.ErrorPrefix, "failed to sync device for dedicated servers")
```
devzbysiu: Logging the error here would be useful

auth/auth.go
DoYouEvenCpp: Shouldn't this get unit-tested? 😉

daemon/device_sync.go
```diff
@@ -0,0 +1,22 @@
...
+		return fmt.Errorf("checking service status: %w", err)
+	}
+
+	if !hasDedicatedServerService {
+		return nil
+	}
+
+	ok := r.dedicatedServersKeyManager.CheckAndRegisterDedicatedServers()
+	if !ok {
+		return fmt.Errorf("failed to register device")
```
DoYouEvenCpp, line 18: Would it be a big of a hassle to pass errors from `CheckAndRegisterDedicatedServers` into the upper layers? Currently there might be a few ways to fail `CheckAndRegisterDedicatedServers`, yet those got merely logged. Is this something worth a change, what do you think?

DoYouEvenCpp (commented):
Would it be a significant effort to introduce a test, integration ideally, into `rpc_login_test` so we got this new bit covered there as well?

daemon/device_sync.go
bartoszWojciechO: Actually, this is already handled in the next PR(connection flow).

DoYouEvenCpp (approved):

cmd/daemon/main.go
bartoszWojciechO: Even if new device key is generated and it invalidates the mesh data, this key can be then registered on demand for meshnet functions(key is checked/generated before any mesh action is taken).

devzbysiu (approved):

PR #1475, opened 2026-05-06
LVPN-10383: add support of dedicated servers into tray

DoYouEvenCpp:
This PR introduces support for dedicated servers.

Changes under this PR:
- dedicated server is now selectable from tray
- dedicated server is now present always in the tray
- `nordvpn groups` now lists also dedicated servers
- while trying to connect to a dedicated server an OS notification popups are now triggered when user doesn't have either a valid licence or they didn't finish configuration of their dedicated server subscription
- aligned copy text for error messages with the latest designs


daemon/data_manager.go
```diff
@@ -377,6 +381,11 @@ func (dm *DataManager) Groups(
 		}
 	}
 
+	// Dedicated Server is to be always present in Tray
+	if !groupsSet.Contains(dedicatedServersGroupTitle) {
```
bartoszWojciechO: Is the check needed if group is always present?

tray/recents.go
```diff
@@ -46,6 +46,7 @@ var groupTitles = map[config.ServerGroup]string{
 	config.ServerGroup_THE_AMERICAS:                     "The Americas",
 	config.ServerGroup_ASIA_PACIFIC:                     "Asia Pacific",
 	config.ServerGroup_AFRICA_THE_MIDDLE_EAST_AND_INDIA: "Africa The Middle East and India",
+	config.ServerGroup_DEDICATED_SERVERS:                "Dedicated Server",
```
bartoszWojciechO: Should this be `"Dedicated Server"` or `"Dedicated Servers"`? I'm starting to get lost with the naming 😄

bartoszWojciechO (approved):

daemon/data_manager.go
DoYouEvenCpp: In the matter of fact, true - we can skip this check.

tray/recents.go
DoYouEvenCpp: `Dedicated Server` - plural form. :wink:

keliramu (approved):

edvyno1 (approved):

PR #1487, opened 2026-05-12
LVPN-10513: Allow responses from peer without incoming access

devzbysiu:
Details:
- when host denies incoming connections for peer, it should still be able to communicate with peer and receive responses
- this PR modifies `meshnet_input` chain to allow that:
```nft
chain mesh_input {
  ip saddr 100.64.0.0/29 accept comment "meshnet private IP"
  ct state established,related ct original ip saddr <host_mesh_ip> accept comment "responses to my connections only"   <---- new
  tcp dport 49111 ip saddr @fileshare_allowed_peers accept comment "meshnet to fileshare"
  tcp dport 49111 drop comment "meshnet to fileshare"
  ip saddr @allow_incoming_connections accept comment "meshnet to local"
  drop
}
```

### DIFF
--- a/daemon/firewall/nft/nft.go
+++ b/daemon/firewall/nft/nft.go
@@ -165,7 +165,7 @@ func (n *nft) addInputChain(config firewall.Config, nftCtx *nftContext) {
 	// meshnet
 	if config.MeshnetInfo != nil {
 		// Add chain for the meshnet and the jump rule to it
-		meshChain := n.addMeshnetInputChain(nftCtx, config.MeshnetInfo.MeshnetMap.Machine.Address)
+		meshChain := n.addMeshnetInputChain(nftCtx, config.MeshnetInfo.MeshnetMap.Address)
 
 		// iifname "nordlynx" ip saddr 100.64.0.0/10 jump mesh_input
 		n.conn.AddRule(&nftables.Rule{
--- a/daemon/firewall/nft/nft.go
+++ b/daemon/firewall/nft/nft.go
@@ -2,6 +2,7 @@ package nft
 
 import (
 	"fmt"
+	"net/netip"
 
 	"github.com/NordSecurity/nordvpn-linux/config"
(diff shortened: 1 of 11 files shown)


test/helpers/fw_config_builder.go
```diff
@@ -52,6 +56,7 @@ func (b *FirewallConfigBuilder) BlockFileshare() *FirewallConfigBuilder {
 
 func (b *FirewallConfigBuilder) Meshnet(iface string) *FirewallConfigBuilder {
 	b.cfg.MeshnetInfo = &firewall.MeshInfo{MeshInterface: iface}
+	b.cfg.MeshnetInfo.MeshnetMap.Machine.Address = netip.MustParseAddr(DefaultMachineAddr)
```
thegravi: [nit] Will we always need this by default? Maybe something like `WithSelfAddress` would suffice?

thegravi (approved):

test/helpers/fw_config_builder.go
devzbysiu: Yes, makes sense. I changed it to be passed through `Meshnet` func as it's not possible to have mechnet without IP so it's required argument when setting meshnet config

edvyno1 (approved):

daemon/firewall/nft/nft.go
```diff
@@ -519,6 +520,20 @@ func (n *nft) addMeshnetInputChain(nftCtx *nftContext) *nftables.Chain {
 		UserData: userdata.AppendString(nil, userdata.TypeComment, "meshnet private IP"),
 	})
 
+	// ct state established,related ct original saddr <selfMeshIP> accept
+	if selfMeshIP.IsValid() {
```
mariusSincovici: [small] .Is4()?

mariusSincovici (approved):

daemon/firewall/nft/nft.go
devzbysiu: Updated

PR #1490, opened 2026-05-13
LVPN-10100: Deprecating region based specialty servers

thegravi:
Merging a feature branch.


daemon/jobs.go
```diff
@@ -335,16 +335,26 @@ func (r *RPC) doAutoConnect() error {
 
 	server := connectServer{}
 
+	serverTag := cfg.AutoConnectData.ServerTag
 	groupTag := ""
-	if cfg.AutoConnectData.Group != config.ServerGroup_UNDEFINED &&
+	groupParam := cfg.AutoConnectData.Group
+
+	// runtime migration for deprecated regional groups.
```
devzbysiu: Can we make the migration on daemon start? Then we have logic in one place so the auto-connect doesn't have to change and we can also remove migration from `rpc_settings.go`

daemon/jobs.go
mariusSincovici, line 336: Could you please add a check if autoconnect is still enabled and only then to execute the connect?

daemon/jobs.go
thegravi: fixed in https://github.com/NordSecurity/nordvpn-linux/pull/1504

daemon/jobs.go
thegravi: fixed in https://github.com/NordSecurity/nordvpn-linux/pull/1504

edvyno1 (approved):

devzbysiu (approved):

PR #1494, opened 2026-05-13
LVPN-10473: Implement missing autotests covering Firewall cases

bbrutall:
- Implemented tests for "VPN Connected" section


test/qa/test_connect.py
```diff
@@ -46,6 +48,58 @@ def disconnect_base_test():
...
+
+    :param server: IP address of the DNS server to probe.
+    :param timeout: Socket timeout in seconds.
+    :return: True if port 53 responded, False if dropped or unreachable.
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        sock.settimeout(timeout)
+        dns_query = struct.pack(">HHHHHH", 1, 0x0100, 1, 0, 0, 0)
+        sock.sendto(dns_query, (server, 53))
```
devzbysiu: This can raise, probably should be guarded with `with ... as ..:` to avoid leaks. Same in `_port_53_tcp_reachable`

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
         sh.nordvpn(get_alias(), virtual_country)
 
     assert "The specified server is not available at the moment or does not support your connection settings." in ex.value.stdout.decode(), "Should show server unavailable error"
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
+def test_dns_udp_53_to_lan_resolver_dropped(tech, proto, obfuscated):
```
devzbysiu: Do we have Test Cases defined already for this test and others so we can reference this with our `Manual TC: <ticket-id>` comment?

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
...
+                 f"http://127.0.0.1:{local_http_port}"],
+                capture_output=True, text=True, check=False,
+            )
+
+            assert result.stdout.strip() in ("200", "301", "302"), (
+                "Loopback traffic must be accepted when VPN is connected"
+            )
+        finally:
+            http_server.terminate()
+            http_server.wait(timeout=5)
```
devzbysiu: Isn't this also raising exception if the timeout passed? Then we may have dangling server running, maybe we could do `http_server.kill()` if the `wait` fails

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
...
+
+        assert not _port_53_tcp_reachable("127.0.0.53"), (
+            "TCP port 53 to 127.0.0.53 must be blocked by the firewall when VPN is connected"
+        )
+
+        disconnect_base_test()
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.TECHNOLOGIES)
+def test_tcp_established_connection_response_accepted_via_tunnel(tech, proto, obfuscated):
```
devzbysiu: This does not verify that the traffic goes through the tunnel. Was this on purpose or just missed?

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
...
+      1. Connect to VPN.
+      2. Send raw UDP to 127.0.0.53:53.
+      3. Assert the query fails, - port 53 must be blocked by the firewall.
+      4. Disconnect from VPN.
+
+    :raises AssertionError: If UDP port 53 is reachable when VPN is connected.
+    """
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+
+    with lib.Defer(sh.nordvpn.disconnect):
```
thegravi: Shouldn't we try to reach local resolver first before connecting to VPN just to make sure it works, and before "confirming" we cannot indeed reach it?

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
...
+        time.sleep(1)
+
+        subprocess.run(
+            ["curl", "-s", "-o", "/dev/null",
+             "--connect-timeout", "3", "--max-time", "3",
+             f"http://{other_lan_ip}"],
+            check=False,
+        )
+
+        phys_thread.join()
```
thegravi: Should we have a timeout in case tool is stuck?

test/qa/test_connect.py
```diff
@@ -660,3 +714,189 @@ def test_obfuscation_prevents_virtual_location_connection(tech, proto, obfuscate
...
+                ["curl", "-s", "-o", "/dev/null", "-w", "%{http_code}",
+                 "--connect-timeout", "5", "--max-time", "5",
+                 f"http://127.0.0.1:{local_http_port}"],
+                capture_output=True, text=True, check=False,
+            )
+
+            assert result.stdout.strip() in ("200", "301", "302"), (
+                "Loopback traffic must be accepted when VPN is connected"
+            )
+        finally:
```
thegravi: Didn't we forget to `disconnect_base_test()` here?

test/qa/test_connect.py
bbrutall: Makes sense. Done

test/qa/test_connect.py
bbrutall: Updated docstrings.

test/qa/test_connect.py
bbrutall: Fixed.

test/qa/test_connect.py
bbrutall: Improved.

test/qa/test_connect.py
bbrutall: Reused `capture_utils`

test/qa/test_connect.py
bbrutall: Now, it's handled by `lib.Defer(sh.nordvpn.disconnect)`

PR #1495, opened 2026-05-14
LVPN-10378: display dedicated servers service status

bartoszWojciechO:


auth/auth.go
```diff
@@ -26,6 +26,11 @@ type DedicatedIPService struct {
 	ServerIDs []int64
 }
 
+type DedicatedServersService struct {
```
DoYouEvenCpp: This is now akin to `DedicatedIPService` - shall we have a single, unified struct here maybe?

auth/auth.go
```diff
@@ -26,6 +26,11 @@ type DedicatedIPService struct {
...
+	ExpiresAt string
+	Active    bool
```
DoYouEvenCpp, line 31: Can't this be concluded by running `ExpiresAt > now()`, or we need specifically `Active` somewhere?

auth/auth.go
```diff
@@ -37,8 +42,8 @@ type Checker interface {
 	// GetDedicatedIPServices returns all available server IDs, if server is not selected by the user it will set
 	// ServerID for that service to NoServerSelected
 	GetDedicatedIPServices() ([]DedicatedIPService, error)
-	// HasDedicatedServerService returns true if user has dedicated servers service that is not expired.
-	HasDedicatedServerService() (bool, error)
+	// GetDedicatedServersService returns dedicated servers service status and expiration date
```
DoYouEvenCpp: I'm aware this PR is still one commit behind the branch, but just to remind - currently we wanted to go for the singular version (eg. `DedicatedServer`).

auth/auth.go
DoYouEvenCpp: Or we can leverage `core/checkers.go` - `IsExpired` ?

auth/auth.go
```diff
@@ -208,19 +213,33 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
...
+		return DedicatedServersService{Active: false}, nil
 	}
 
-	return slices.ContainsFunc(services, func(service core.ServiceData) bool {
-		return service.Service.ID == DedicatedServersServiceID && !r.expChecker.IsExpired(service.ExpiresAt)
-	}), nil
+	return DedicatedServersService{
+		ExpiresAt: dedicatedServersServiceData.ExpiresAt,
+		Active:    true,
+	}, nil
```
DoYouEvenCpp, line 238: We can shortcut this to:

```suggestion
	dedicatedServersServiceData := services[dedicatedServersServiceIndex]
	return DedicatedServersService{
 	  ExpiresAt: dedicatedServersServiceData.ExpiresAt,
	  Active:    !r.expChecker.IsExpired(dedicatedServersServiceData.ExpiresAt),
	}, nil
```

right?

daemon/rpc_account.go
```diff
@@ -103,6 +103,22 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
 		}
 	}
 
+	dedicatedServerService, err := r.ac.GetDedicatedServersService()
+	if err != nil {
+		log.Println(internal.ErrorPrefix, "gettig dedicated server service:", err)
```
DoYouEvenCpp:
```suggestion
		log.Println(internal.ErrorPrefix, "getting dedicated server service:", err)
```

daemon/rpc_account.go
```diff
@@ -103,6 +103,22 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
...
+		if errors.Is(err, core.ErrUnauthorized) {
+			return &pb.AccountResponse{Type: internal.CodeExpiredAccessToken}, nil
+		} else {
```
DoYouEvenCpp: [nit] this else is redundant here

auth/auth.go
bartoszWojciechO: This is mostly for simplicity, because each caller would have to implement some sort of an mockable interface for `IsExpired` if we wanted to test it. It's simpler to assume that `GetDedicatedServerService` gives us proper service status. Also, the service might not even have an expiration date(because it was not available in the first place), so having a separate field for availability makes it more readable imo.

I just look at it like a more verbose version of `GetDedicatedServerService() (string, bool, error)` where the `bool` represents service availability.

auth/auth.go
bartoszWojciechO: I'm not sure how to unift it wihout overcomplicating. Do you have some suggestions?

DoYouEvenCpp: Yeah, come to think about it - agree. That would require a bit of code to make it unified. Let's drop it.

DoYouEvenCpp (approved):

daemon/rpc_account.go
```diff
@@ -103,6 +103,21 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
 		}
 	}
 
+	dedicatedServerService, err := r.ac.GetDedicatedServerService()
```
thegravi, line 106: Now this and `GetDedicatedIPServices()` calls the same API twice and just uses different underlying data. Can we somehow eliminate this?

daemon/device_sync.go
```diff
@@ -4,12 +4,12 @@ import "fmt"
 
 // RegisterDedicatedServers registers the device for dedicated servers if dedicated servers service is available to the user.
 func (r *RPC) RegisterDedicatedServers() error {
-	hasDedicatedServerService, err := r.ac.HasDedicatedServerService()
+	dedicatedServersService, err := r.ac.GetDedicatedServerService()
```
keliramu: naming: should it be `dedicatedServerService` (singular)

daemon/device_sync_test.go
```diff
@@ -16,29 +17,29 @@ func TestSyncDevice(t *testing.T) {
 
 	tests := []struct {
 		name                             string
-		hasDedicatedServersService       bool
+		dedicatedServersService          auth.DedicatedServerService
```
keliramu: naming: should it be dedicatedServerService (singular)

protobuf/daemon/account.proto
```diff
@@ -21,6 +21,8 @@ message AccountResponse {
   repeated DedidcatedIPService dedicated_ip_services = 7;
   TriState mfa_status = 8;
   string created_on = 9;
+  int64 dedicated_servers_status = 10;
```
keliramu: naming: `servers` -> `server`

auth/auth.go
```diff
@@ -208,19 +213,29 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
 	return dipServices, nil
 }
 
-// HasDedicatedServerService returns true if user has dedicated servers service that is not expired.
-func (r *RenewingChecker) HasDedicatedServerService() (bool, error) {
+// GetDedicatedServerService returns dedicated servers service status and expiration date
```
keliramu: naming: `servers` -> `server`

keliramu (approved):

daemon/rpc_account.go
bartoszWojciechO: Next PR adds caching in case of those DIP/DedicatedServers services https://github.com/NordSecurity/nordvpn-linux/pull/1506. Do you think it's enough or should we have some unified call at the upper layer?

thegravi (approved):

PR #1498, opened 2026-05-15
LVPN-10518: Fix lan discovery reset

devzbysiu:
Context:
- when networker is initialized it gets config values passed in (or instances of classes that had config values pass to those)
- when `nordvpn set defaults` is called some of those values are not reset in networker which makes it use the stale values event though config has changed
- this PR fixes lan-discovery specifically

Changes:
- reset lan discovery value (and allowlist as it holds lan ranges)
- add logs
- add test for found issue

### DIFF
--- a/daemon/rpc_set_defaults.go
+++ b/daemon/rpc_set_defaults.go
@@ -68,7 +68,14 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
 		}, nil
 	}
 	r.netw.SetVPN(v)
-	_ = r.netw.SetARPIgnore(cfg.ARPIgnore.Get())
+
+	if err = r.netw.SetARPIgnore(cfg.ARPIgnore.Get()); err != nil {
+		log.Warn("resetting arp ignore failed:", err)
+	}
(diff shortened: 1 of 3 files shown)


daemon/rpc_set_defaults.go
```diff
@@ -68,7 +68,14 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
 		}, nil
 	}
 	r.netw.SetVPN(v)
-	_ = r.netw.SetARPIgnore(cfg.ARPIgnore.Get())
+
+	if err = r.netw.SetARPIgnore(cfg.ARPIgnore.Get()); err != nil {
+		log.Warn("resetting arp ignore failed:", err)
+	}
+	r.netw.SetLanDiscovery(cfg.LanDiscovery)
+	if err := r.netw.SetAllowlist(cfg.AutoConnectData.Allowlist); err != nil {
```
thegravi: [nit] reuse `err` without declaring new one.

mariusSincovici (approved):

thegravi (approved):

daemon/rpc_set_defaults.go
devzbysiu: Done

PR #1499, opened 2026-05-15
LVPN-5412: Remove config encryption

edvyno1:


config/manager.go
```diff
@@ -216,29 +209,35 @@ func (f *FilesystemConfigManager) load(c *Config, copy *Config) error {
...
 	if err != nil {
 		return err
 	}
-
-	decryptedJSON, err := internal.Decrypt(data, pass)
-	if err != nil {
-		return err
+	var decryptedData []byte
+	// Checking if file header matches the expected decrypted file header
+	if !bytes.Equal(data[:4], decryptedConfigFileHeader) {
```
thegravi: `4` -> `len(decryptedConfigFileHeader)`

I also would suggest adding something like `len(data) < len(header)` to be sure this does not panic

config/manager_test.go
```diff
@@ -266,3 +268,34 @@ func TestLoadWithCopyReturnsIndependentCopies(t *testing.T) {
 	// Verify second copy is unchanged (independent)
 	assert.NotEqual(t, first, second, "Modifications to the first copy should not affect the second copy")
 }
+
+func TestConfigMigratesFromEncryptedToUnencrypted(t *testing.T) {
```
thegravi, line 272: Can we have a test where header is != 4B?

config/manager.go
```diff
@@ -130,22 +132,14 @@ func (f *FilesystemConfigManager) SaveWith(fn SaveFunc) error {
...
-	if err != nil {
-		return err
-	}
-
 	data, err := json.Marshal(c)
 	if err != nil {
 		return err
 	}
+	dataWithHeader := decryptedConfigFileHeader
+	dataWithHeader = append(dataWithHeader, data...)
```
thegravi: With a coincidence by a chance we allocate a new slice when appending data. Because on the first step `dataWithHeader := decryptedConfigFileHeader` we assign it a pointer and it's shared with the second part `dataWithHeader = append(dataWithHeader, data...)`, but since the capacity is 4B we need to allocate a new slice and this make a new slice and does not write directly to `decryptedConfigFileHeader` memory.
I would suggest to preallocate with `make()` the total capacity needed for `dataWithHeader`.

config/manager.go
edvyno1: addressed

config/manager_test.go
edvyno1: added a new test for empty config

config/manager.go
edvyno1: changed now, good spot

thegravi (approved):

config/manager_test.go
```diff
@@ -266,3 +268,49 @@ func TestLoadWithCopyReturnsIndependentCopies(t *testing.T) {
...
+	var cfg Config
+	err = configManager.Load(&cfg)
+	require.NoError(t, err)
+	err = configManager.SaveWith(func(c Config) Config {
+		c.MachineID = uuid.MustParse("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
+		c.LanDiscovery = true
+		return c
+	})
+	require.NoError(t, err)
+	encrypt_removed_config, _ := internal.FileRead("testdata/settings_5.0.0.dat")
```
keliramu: `camelCase` plz - it's go convention:
`encrypt_removed_config` -> `encryptRemovedConfig`
`post_save_config` -> `postSaveConfig`

config/manager.go
keliramu: Plz update fn comment

keliramu (approved):

config/manager_test.go
edvyno1: changed

config/manager.go
edvyno1: updated

PR #1500, opened 2026-05-18
LVPN-10246: DIP events anonymization

thegravi:


events/moose/moose.go
```diff
@@ -246,13 +298,16 @@ func (s *Subscriber) Init(consent config.AnalyticsConsent) error {
 		internal.IsProdEnv(s.buildTarget.Environment),
 		s,
 		s,
-		s.canSendAllEvents.Load(),
```
mariusSincovici, line 249: Isn't a bit of ambiguity with `canSendAllEvents`? 
Is used in function `isEnabled` only, but this only used in `response(respCode uint32)` for when moose is not initialized error.
On the same time there is a `isInitialized` which keeps track if moose was initialized.

mariusSincovici (approved):

edvyno1 (approved):

events/moose/moose.go
thegravi: https://github.com/NordSecurity/nordvpn-linux/pull/1522

PR #1505, opened 2026-05-19
LVPN-10410: Fix hover colors for connection card buttons

dfetti:
It covers LVPN-10416 and LVPN-10410.

### DIFF
--- a/gui/lib/theme/theme.dart
+++ b/gui/lib/theme/theme.dart
@@ -387,79 +387,106 @@ final class NordVpnTheme {
       ),
       buttonTheme: ConnectionCardButtonTheme(
         maxConnectButtonWidth: 330,
-        secureMyConnectionButtonStyle: OutlinedButton.styleFrom(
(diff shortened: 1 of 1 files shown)


gui/lib/theme/theme.dart
```diff
@@ -387,79 +387,126 @@ final class NordVpnTheme {
...
+            ).copyWith(
+              backgroundColor: WidgetStateProperty.resolveWith<Color?>((
+                states,
+              ) {
+                if (states.contains(WidgetState.hovered)) {
+                  return design.semanticColors.bgPrimary;
+                }
+
+                return design.semanticColors.bgTertiary;
+              }),
```
devzbysiu: [small] maybe worth considering if we should extract this to helper function as this is repeated 3 times

devzbysiu (approved):

DoYouEvenCpp (approved):

thegravi (approved):

PR #1506, opened 2026-05-19
LVPN-10379: handle user services updates in NC

bartoszWojciechO:
NC provides us with user service updates. We can fetch user's services upon receiving such notifications instead of fetching the services every time when needed. This will reduce the load on the backend and we will be able to use this for GUI in the future.

The user's services are only cached for Dedicated IP/Servers services. Regular VPN service is fetched based on the service's expiration date.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -84,7 +84,7 @@ type RenewingChecker struct {
 	mu                  sync.Mutex
 	accountUpdateEvents *daemonevents.AccountUpdateEvents
 	sessionStores       []session.SessionStore
-	servicesState       ServicesState
+	servicesState       *ServicesState
 }
 
 // NewRenewingChecker is a default constructor for RenewingChecker.
@@ -94,7 +94,7 @@ func NewRenewingChecker(cm config.Manager,
 	logoutPub events.Publisher[events.DataAuthorization],
 	errPub events.Publisher[error],
 	accountUpdateEvents *daemonevents.AccountUpdateEvents,
-	servicesState ServicesState,
+	servicesState *ServicesState,
 	sessionStores ...session.SessionStore,
 ) *RenewingChecker {
 	return &RenewingChecker{
@@ -193,9 +193,6 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
 }
 
 func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
 	services, err := r.servicesState.fetchServices()
 	if err != nil {
 		return nil, fmt.Errorf("fetching available services: %w", err)
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -68,6 +68,12 @@ func (systemTimeExpirationChecker) IsExpired(expiryTime string) bool {
 	return time.Now().After(expiry)
 }
 
+func hasDedicatedServerService(services core.ServicesResponse, expirationChecker core.ExpirationChecker) bool {
+	return slices.ContainsFunc(services, func(service core.ServiceData) bool {
+		return service.Service.ID == DedicatedServersServiceID && !expirationChecker.IsExpired(service.ExpiresAt)
+	})
+}
+
 // NewTokenExpirationChecker
 func NewTokenExpirationChecker() core.ExpirationChecker {
 	return &systemTimeExpirationChecker{}
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -195,7 +195,7 @@ func (r *RenewingChecker) IsVPNExpired() (bool, error) {
 func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error) {
 	services, err := r.servicesState.fetchServices()
 	if err != nil {
-		return nil, fmt.Errorf("fetching available services: %w", err)
+		return nil, fmt.Errorf("fetching available dedicated IP services: %w", err)
 	}
 
 	dipServices := []DedicatedIPService{}
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -84,6 +84,7 @@ type RenewingChecker struct {
 	mu                  sync.Mutex
 	accountUpdateEvents *daemonevents.AccountUpdateEvents
 	sessionStores       []session.SessionStore
+	servicesState       ServicesState
 }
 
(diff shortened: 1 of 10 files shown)


nc/nc.go
```diff
@@ -420,8 +426,13 @@ func (c *Client) handleMessage(client mqtt.Client, msg mqtt.Message, ctx context
 		)
 	}
 
-	if payload.Message.Data.Event.Type == typeMeshnet {
+	switch payload.Message.Data.Event.Type {
+	case typeUserServiceUpdate:
+		c.subjectServicesUpdate.Publish(true)
```
DoYouEvenCpp: [nit] isn't more idiomatic here to emit `struct{}{}` ? (at least more coherent with similar type of listeners)

auth/services_state.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewServicesState(credentialsAPI core.CredentialsAPI) ServicesState {
+	return ServicesState{
+		credentialsAPI: credentialsAPI,
+	}
+}
+
+// fetchServices fetches user's services from the API and saves them so that they will be returned on subsequent calls.
+func (s *ServicesState) fetchServices() (core.ServicesResponse, error) {
+	if s.servicesFetched {
```
keliramu: what if new services were purchased? or if existing service expired? now it is fetch once and use from cache forever, need some logic to refresh cache after some time. Or is cache updated on NC msg about DS service change? but still, it's not enough, imo.

keliramu: NC msg is sent for every service change Ok, time based cache invalidation/update would be nice to have just in case NC is down or blocked.

cmd/daemon/main.go
```diff
@@ -540,6 +544,7 @@ func main() {
 		daemonEvents.User.Logout,
 		errSubject,
 		accountUpdateEvents,
+		servicesState,
```
keliramu, line 524: Passing by value will make a copy, is that desired result? probably not

auth/services_state.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/log"
+)
+
+// ServicesState is responsible for fetching and storing user's available services.
+type ServicesState struct {
+	credentialsAPI core.CredentialsAPI
+
+	servicesFetched bool
+	services        core.ServicesResponse
```
keliramu: `ServicesState` needs concurrency protection, NC msg handling happens on separate goroutine, I guess.

cmd/daemon/main.go
DoYouEvenCpp: Yeah, agree - now notifications about services updates and cache, they both work on two separate objects.

auth/auth.go
```diff
@@ -188,7 +191,7 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
 	r.mu.Lock()
 	defer r.mu.Unlock()
```
DoYouEvenCpp: This now can be dropped, since cache protects shared resources by itself - right?
The same  goes about the `HasDedicatedServerService`

auth/auth.go
```diff
@@ -213,7 +216,7 @@ func (r *RenewingChecker) HasDedicatedServerService() (bool, error) {
...
 
-	services, err := r.fetchServices()
+	services, err := r.servicesState.fetchServices()
```
DoYouEvenCpp, line 224: User's logout shall clear cached values, right?

auth/services_state.go
```diff
@@ -0,0 +1,34 @@
...
+type ServicesState struct {
+	cache *caching.Cache[core.ServicesResponse]
+}
+
+func NewServicesState(credentialsAPI core.CredentialsAPI) ServicesState {
+	cache := caching.NewCacheWithTTL(time.Hour*5, credentialsAPI.Services)
+	return ServicesState{
+		cache: cache,
+	}
+}
```
DoYouEvenCpp: We should be returning this by pointer, aren't we?
```suggestion
func NewServicesState(credentialsAPI core.CredentialsAPI) *ServicesState {
	cache := caching.NewCacheWithTTL(time.Hour*5, credentialsAPI.Services)
	return &ServicesState{
		cache: cache,
	}
}
```

auth/auth.go
```diff
@@ -210,10 +210,7 @@ func (r *RenewingChecker) GetDedicatedIPServices() ([]DedicatedIPService, error)
 
 // HasDedicatedServerService returns true if user has dedicated servers service that is not expired.
 func (r *RenewingChecker) HasDedicatedServerService() (bool, error) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	services, err := r.fetchServices()
+	services, err := r.servicesState.fetchServices()
 	if err != nil {
 		return false, fmt.Errorf("fetching services: %w", err)
```
DoYouEvenCpp: maybe let's make the error a bit more specific here?

```suggestion
		return false, fmt.Errorf("fetching dedicated server service availability: %w", err)
```

DoYouEvenCpp (approved):

cmd/daemon/main.go
keliramu: Plz fix this `Println` with `%s` as well

auth/auth.go
```diff
@@ -218,7 +224,7 @@ func (r *RenewingChecker) GetDedicatedServerService() (DedicatedServerService, e
 	r.mu.Lock()
```
keliramu: Is lock needed here? It is not used in above `GetDedicatedIPServices` also `fetchServices` is using `Cache` with its internal locking.

auth/services_state.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/internal/caching"
+	"github.com/NordSecurity/nordvpn-linux/log"
+)
+
+// ServicesState is responsible for fetching and storing user's available services.
+type ServicesState struct {
+	cache                    *caching.Cache[core.ServicesResponse]
+	dedicatdServerKeyManager devicekey.DedicatedServersKeyManager
```
keliramu: typo `dedicatdServerKeyManager` -> `dedicatedServerKeyManager`

keliramu (approved):

PR #1518, opened 2026-05-25
LVPN-10473: Implement missing autotests covering Firewall cases

bbrutall:
- Implemented firewall autotests for "VPN Connected" section

### DIFF
--- a/test/qa/lib/capture_utils.py
+++ b/test/qa/lib/capture_utils.py
@@ -1,6 +1,7 @@
 import asyncio
 import threading
 import os
+import time
 
 import pyshark
 
@@ -40,9 +41,13 @@ class BackgroundCapture:
     def start(self):
         self._thread = threading.Thread(target=self._run, daemon=True)
         self._thread.start()
+        # Give tshark time to initialise and start capturing before traffic is generated
+        time.sleep(1)
 
 
     def stop(self):
+        # Give tshark time to flush the last packets before closing the capture
+        time.sleep(1)
         try:
             if self._capture is not None:
                 self._capture.close()
--- a/test/qa/test_firewall.py
+++ b/test/qa/test_firewall.py
@@ -1,11 +1,19 @@
+import contextlib
 import random
 
 import pytest
+import socket
+import subprocess
+import time
+
+import requests
 import sh
 
 import lib
 from lib import IS_NIGHTLY, allowlist, firewall, network
 from lib.dynamic_parametrize import dynamic_parametrize
+from lib import capture_utils
+from lib.firewall import tun_interface_names
 
 pytestmark = pytest.mark.usefixtures("nordvpnd_scope_module", "collect_logs")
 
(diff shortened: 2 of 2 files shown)


test/qa/test_firewall.py
```diff
@@ -287,3 +345,204 @@ def test_firewall_lan_allowlist_work_together(tech, proto, obfuscated):
...
+
+        assert not _port_53_tcp_reachable("192.168.1.1"), (
+            "TCP port 53 to 192.168.1.1 must be blocked by the firewall when VPN is connected"
+        )
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER)
+def test_firewall_tcp_established_connection_response_accepted_via_tunnel(tech, proto, obfuscated):
+    """
+    Verify that response traffic for an established TCP connection is accepted via the tunnel.
```
devzbysiu, line 363: Do we have test case ID for this one?

test/qa/test_firewall.py
```diff
@@ -14,6 +22,56 @@ def setup_module(module):  # noqa: ARG001
...
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
+            sock.settimeout(timeout)
+            sock.sendto(dns_query, (server, 53))
+            sock.recvfrom(512)
+        return True
+    except Exception:  # noqa: BLE001
+        return False
+
+
+def _port_53_tcp_reachable(server: str, timeout: float = 3.0) -> bool:
```
devzbysiu: We already have `is_port_accessible_TCP` function, can it be used instead?

test/qa/test_firewall.py
```diff
@@ -14,6 +22,56 @@ def setup_module(module):  # noqa: ARG001
     firewall.setup_port_sock_server(None)
 
 
+def _port_53_udp_reachable(server: str, timeout: float = 3.0) -> bool:
```
devzbysiu: We already have `is_port_accessible_UDP`, can it be used instead?

test/qa/test_firewall.py
```diff
@@ -287,3 +345,204 @@ def test_firewall_lan_allowlist_work_together(tech, proto, obfuscated):
...
+      2. Start packet capture on all interfaces filtered to 192.168.1.1.
+      3. Attempt HTTP request to 192.168.1.1 — no response expected.
+      4. Assert 192.168.1.1 is routed via VPN.
+      5. Assert no packets to 192.168.1.1 appeared on the physical interface.
+      6. Disconnect from VPN.
+
+    :raises AssertionError: If 192.168.1.1 is not routed via VPN or traffic appears on the physical interface.
+    """
+    other_lan_ip = "192.168.1.1"
+    phys_iface = _get_physical_interface()
```
devzbysiu: This `_get_physical_interface` does not return physical interface, but interface of default route. Then the test is checking that there are no packets on this interface, but if something will change, then the assertion will most likely still pass because we might be checking wrong interface. 

Maybe instead of checking that there are no packets on some interface returned by this function, we could check that all captured packets go though tunnel interface which we already know? Capture is already done on "any" so if all packets go through known tunnel it means no packets go through something else

test/qa/test_firewall.py
```diff
@@ -287,3 +345,204 @@ def test_firewall_lan_allowlist_work_together(tech, proto, obfuscated):
...
+    lib.set_technology_and_protocol(tech, proto, obfuscated)
+
+    with lib.Defer(sh.nordvpn.disconnect):
+        sh.nordvpn.connect()
+
+        cap = capture_utils.BackgroundCapture(
+            "any",
+            bpf_filter=f"host {other_lan_ip}",
+        )
+        cap.start()
```
devzbysiu, line 488: I see that every time we do `cap.start()`, we sleep for some time and then before every `cap.stop()` we also sleep for some time. This sleeping should probably be moved to `BackgroundCapture` then, because now everyone writing tests should remember to insert those sleeps. We could have sleep duration as a param to both `start()` and `stop()` and we can even have some defaults there if needed

test/qa/test_firewall.py
```diff
@@ -287,3 +345,204 @@ def test_firewall_lan_allowlist_work_together(tech, proto, obfuscated):
...
+        time.sleep(1)
+
+        with contextlib.suppress(requests.exceptions.RequestException):
+            requests.get(f"http://{other_lan_ip}", timeout=3, allow_redirects=False)
+
+        time.sleep(1)
+        cap.stop()
+
+        assert firewall.is_ip_routed_via_VPN([other_lan_ip + "/32"]), (
```
devzbysiu: This tests configuration instead of behavior. We most likely don't need this as checking that all captured traffic goes through the tunnel is enough

test/qa/test_firewall.py
```diff
@@ -287,3 +345,204 @@ def test_firewall_lan_allowlist_work_together(tech, proto, obfuscated):
                 sh.nordvpn.connect()
                 assert not firewall.is_ip_routed_via_VPN(["1.1.1.1"]), "Allowlisted subnet is not going through VPN"
                 assert firewall.is_ip_routed_via_VPN(["1.0.0.1"]), "Not whitelisted subnet is going through VPN"
+
+
+@pytest.mark.parametrize(("tech", "proto", "obfuscated"), lib.STANDARD_TECHNOLOGIES_NO_NORDWHISPER)
```
edvyno1: why is nordwhisper not parametrized as part of these tests? it should either be full technologies or just one, as tech does not impact the way firewall is working

test/qa/test_firewall.py
bbrutall: Fixed

test/qa/test_firewall.py
bbrutall: Fixed

test/qa/test_firewall.py
bbrutall: Not yet

test/qa/test_firewall.py
bbrutall: Yes, it makes sense. Updated test logic.

test/qa/test_firewall.py
bbrutall: Moved `sleep` to `BackgroundCapture` methods `start/stop`

test/qa/test_firewall.py
bbrutall: Fixed.

test/qa/test_firewall.py
bbrutall: Replaced with `lib.TECHNOLOGIES`

edvyno1 (approved):

test/qa/lib/capture_utils.py
```diff
@@ -40,9 +41,13 @@ def _run(self):
     def start(self):
         self._thread = threading.Thread(target=self._run, daemon=True)
         self._thread.start()
+        # Give tshark time to initialise and start capturing before traffic is generated
+        time.sleep(1)
 
 
     def stop(self):
+        # Give tshark time to flush the last packets before closing the capture
+        time.sleep(1)
```
mariusSincovici, line 50: If flush is needed, then tshark has `USR2` signal for this.

test/qa/test_firewall.py
devzbysiu: Isn't it this one maybe `LVPN-10476` or it's a different one?

test/qa/lib/capture_utils.py
bbrutall: `USR2` for flushing is a tcpdump-specific feature  and tshark does not support it. Sleeps are needed to avoid a race condition in `pyshark` async packet queue  and not for tshark buffer flushing.

test/qa/test_firewall.py
bbrutall: Yes, this one is relevant.

mariusSincovici (approved):

devzbysiu (approved):

PR #1523, opened 2026-05-26
LVPN-10402: dedicated servers autoconnect

bartoszWojciechO:
Adds handling of dedicated server related errors and a fallback in of expired 

When user's autoconnect target is a dedicated server and their service expires, autoconnect target will be set to the fastest available server.

### DIFF
--- a/cli/messages.go
+++ b/cli/messages.go
@@ -138,13 +138,12 @@ Example: nordvpn set meshnet off
 Supported values for <enabled>: 1, true, enable, on, enabled
 Example: nordvpn set meshnet on`
 
-	MsgSetMeshnetSuccess            = "Meshnet is set to '%s' successfully."
-	MsgMeshnetAlreadyEnabled        = "Meshnet is already enabled."
-	MsgMeshnetAlreadyDisabled       = "Meshnet is already disabled."
-	MsgMeshnetNotEnabled            = "Meshnet is not enabled. Use the \"nordvpn set meshnet on\" command to enable it."
-	MsgMeshnetNordlynxMustBeEnabled = "NordLynx technology must be set to use this feature."
-	MsgMeshnetVersionNotSupported   = "Current application version does not support the Meshnet feature."
-	MsgMeshnetUsage                 = "Meshnet is a way to safely access other devices, no matter where in the world they are. Once set up, Meshnet functions just like a secure local area network (LAN) — it connects devices directly. It also allows securely sending files to other devices. Use the \"nordvpn set meshnet on\" command to enable Meshnet. Learn more: https://meshnet.nordvpn.com/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=meshnet-documentation&nm=app&ns=nordvpn-linux-cli&nc=meshnet-documentation"
+	MsgSetMeshnetSuccess          = "Meshnet is set to '%s' successfully."
+	MsgMeshnetAlreadyEnabled      = "Meshnet is already enabled."
+	MsgMeshnetAlreadyDisabled     = "Meshnet is already disabled."
+	MsgMeshnetNotEnabled          = "Meshnet is not enabled. Use the \"nordvpn set meshnet on\" command to enable it."
+	MsgMeshnetVersionNotSupported = "Current application version does not support the Meshnet feature."
+	MsgMeshnetUsage               = "Meshnet is a way to safely access other devices, no matter where in the world they are. Once set up, Meshnet functions just like a secure local area network (LAN) — it connects devices directly. It also allows securely sending files to other devices. Use the \"nordvpn set meshnet on\" command to enable Meshnet. Learn more: https://meshnet.nordvpn.com/?utm_medium=app&utm_source=nordvpn-linux-cli&utm_campaign=meshnet-documentation&nm=app&ns=nordvpn-linux-cli&nc=meshnet-documentation"
 
 	MsgMeshnetRefreshUsage = "Refreshes the Meshnet in case it was not updated automatically."
 	MsgMeshnetPeerUnknown  = "Peer '%s' is unknown."
@@ -397,10 +396,14 @@ Your browsing activities remain private, regardless of your choice.
 	SetARPIgnoreNothingToSet = "ARP ignore is already set to '%s'."
 	SetARPIgnoreWarning      = "You’ve turned off arp-ignore. This is an advanced privacy setting and should only be off if your network setup requires ARP responses."
 
-	DedicatedServersNoServiceMessage       = "Get a server that's exclusively yours. Enjoy your own IP with resources and speed reserved entirely for you. Use it with port forwarding to host services and access devices remotely. To get a personal dedicated server, continue in the browser: %s"
-	DedicatedServersNoServersAvailable     = "Setup your dedicated server. One last step - pick a server location.\n%s"
-	DedicatedServersServerNotReadyMessage  = "Activating your dedicated server. This may take a few minutes."
-	DedicatedServersNoNordlynxMessage      = "NordLynx technology must be set to use this feature."
-	DedicatedServersCanNotConnectMessage   = "Couldn't connect. Please try again."
-	DedicatedServersConnectionLimitReached = "Too many connections or device limit reached"
+	DedicatedServersNoServiceMessage           = "Get a server that's exclusively yours. Enjoy your own IP with resources and speed reserved entirely for you. Use it with port forwarding to host services and access devices remotely. To get a personal dedicated server, continue in the browser: %s"
+	DedicatedServersNoServersAvailable         = "Setup your dedicated server. One last step - pick a server location.\n%s"
+	DedicatedServersServerNotReadyMessage      = "Activating your dedicated server. This may take a few minutes."
+	DedicatedServersNoNordlynxMessage          = "NordLynx technology must be set to use this feature."
+	DedicatedServersCanNotConnectMessage       = "Couldn't connect. Please try again."
+	DedicatedServersConnectionLimitReached     = "Too many connections or device limit reached"
+	DedicatedServersAutoconnectNordlynxMessage = "Auto-connect to Dedicated Server requires NordLynx protocol."
+	DedicatedServerPQMessage                   = "Post-quantum encryption is not compatible with Dedicated Server."
+
+	MsgNordlynxMustBeEnabled = "NordLynx technology must be set to use this feature."
 )
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -744,7 +744,7 @@ func main() {
 	if ok, _ := authChecker.IsLoggedIn(); ok {
 		go daemon.StartNC("[startup]", notificationClient)
 		if err := rpc.RegisterDedicatedServers(); err != nil {
-			log.Println(internal.ErrorPrefix, "failed to sync device:", err)
+			log.Println(internal.WarningPrefix, "failed to sync device:", err)
 		}
 	}
 	if cfg.Mesh {
--- a/daemon/jobs.go
+++ b/daemon/jobs.go
@@ -319,6 +319,32 @@ func (r *RPC) fallbackTechnology(targetTechnology config.Technology) error {
 	return nil
 }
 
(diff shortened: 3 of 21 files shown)


daemon/jobs.go
```diff
@@ -319,6 +319,31 @@ func (r *RPC) fallbackTechnology(targetTechnology config.Technology) error {
...
 }
 
+// fallbackDedicatedServer changes autoconnection target to a generic(fastest) server if the current target is set to a
+// dedicated server and user doesn't have a dedicated server subscription or feature is disabled in the remote config.
+// New config is saved and returned.
+func (r *RPC) fallbackDedicatedServer(cfg config.Config) config.Config {
+	serviceData, err := r.ac.GetDedicatedServerService()
+	if err != nil {
+		log.Println(internal.ErrorPrefix,
+			"getting dedicated servers service status to determine if fallback is needed: %w", err)
```
keliramu: Remove `%w` from `Println`

daemon/jobs.go
```diff
@@ -319,6 +319,31 @@ func (r *RPC) fallbackTechnology(targetTechnology config.Technology) error {
...
+
+	if err != nil || !serviceData.Active || !r.remoteConfigGetter.IsFeatureEnabled(remote.FeatureDedicatedServer) {
+		cfg.AutoConnectData.Group = config.ServerGroup_UNDEFINED
+		cfg.AutoConnectData.ServerTag = ""
+		if err := r.cm.SaveWith(func(c config.Config) config.Config {
+			c.AutoConnectData = cfg.AutoConnectData
+			return c
+		}); err != nil {
+			log.Println(internal.ErrorPrefix,
+				"failed to save config after a fallback from dedicated server: %w", err)
```
keliramu: Remove `%w` from `Println`

daemon/rpc_connect.go
```diff
@@ -39,6 +39,12 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
 	return ids
 }
 
+// IsDedicatedServer returns true if either serverTag or serverGroup represent the dedicated server group
+func IsDedicatedServer(servetTag string, serverGroup string) bool {
```
keliramu: fix typo `servetTag` -> `serverTag`

daemon/rpc_set_autoconnect.go
```diff
@@ -30,6 +31,16 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 		}, nil
 	}
 
+	if IsDedicatedServer(in.ServerTag, in.ServerGroup) {
```
keliramu: Should it be `if in.GetEnabled() && IsDedicatedServer(in.ServerTag, in.ServerGroup) {` ?
to perform DS validation only when autoconnect is being enabled

keliramu (approved):

cmd/daemon/main.go
```diff
@@ -737,7 +737,7 @@ func main() {
 	if ok, _ := authChecker.IsLoggedIn(); ok {
 		go daemon.StartNC("[startup]", notificationClient)
 		if err := rpc.RegisterDedicatedServers(); err != nil {
-			log.Println(internal.ErrorPrefix, "failed to sync device: %s", err)
+			log.Println(internal.ErrorPrefix, "failed to sync device:", err)
```
thegravi: Would you consider switching to a warning, since the flow here is continued?

daemon/jobs.go
```diff
@@ -319,6 +319,31 @@ func (r *RPC) fallbackTechnology(targetTechnology config.Technology) error {
...
 
+// fallbackDedicatedServer changes autoconnection target to a generic(fastest) server if the current target is set to a
+// dedicated server and user doesn't have a dedicated server subscription or feature is disabled in the remote config.
+// New config is saved and returned.
+func (r *RPC) fallbackDedicatedServer(cfg config.Config) config.Config {
+	serviceData, err := r.ac.GetDedicatedServerService()
+	if err != nil {
+		log.Println(internal.ErrorPrefix,
+			"getting dedicated servers service status to determine if fallback is needed:", err)
+	}
```
thegravi, line 331: What about if the call failure is temporary? Shouldn't we check the exact error before forcing config reset to fastest?
Also, our cold start-up when network is not yet reachable could impact this in the same way.
Maybe let it fails again and again without changing the autoconnect setting under such circumstances?

daemon/rpc_connect.go
```diff
@@ -39,6 +39,12 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
 	return ids
 }
 
+// IsDedicatedServer returns true if either serverTag or serverGroup represent the dedicated server group
+func IsDedicatedServer(serverTag string, serverGroup string) bool {
```
thegravi: Do we need it exposed? It's under the same `deamon` package if I'm not mistaken.

daemon/rpc_connect.go
```diff
@@ -217,6 +222,10 @@ func (r *RPC) connectWithParameters(ctx context.Context,
 		if cfg.Technology != config.Technology_NORDLYNX {
 			return true, srv.Send(&pb.Payload{Type: internal.CodeDedicatedServersNoNordlynx})
 		}
+		// third, if technology is correct, check if post quantum is enabled(pq is not supported for dedicated servers)
+		if cfg.AutoConnectData.PostquantumVpn {
+			return true, srv.Send(&pb.Payload{Type: internal.CodeServerUnavailable})
+		}
```
thegravi, line 232: Don't we need to check "RPC set PQ" call to discard change if auto connect is set to DS? Just like as wit h"RPC set technology".

daemon/rpc_set_autoconnect.go
```diff
@@ -30,6 +31,20 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
...
+		if !r.remoteConfigGetter.IsFeatureEnabled(remote.FeatureDedicatedServer) {
+			return &pb.Payload{Type: internal.CodeFailure}, nil
+		}
+
+		if cfg.Technology != config.Technology_NORDLYNX {
+			return &pb.Payload{Type: internal.CodeDedicatedServersNoNordlynx}, nil
+		}
+
+		if cfg.AutoConnectData.PostquantumVpn {
+			return &pb.Payload{Type: internal.CodeServerUnavailable}, nil
```
thegravi: We have a dedicated message for "no-nordlynx", but when there is an issue with PQ we don't. Is that by design?

daemon/rpc_connect.go
thegravi, line 204: Do we handle the case when we're connected to DS and hit pause, and during that period we set `PQ=true` and after pause ends we resume and cannot connect back.

daemon/rpc_set_autoconnect.go
```diff
@@ -30,6 +31,20 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 		}, nil
 	}
 
+	if in.GetEnabled() && IsDedicatedServer(in.ServerTag, in.ServerGroup) {
+		if !r.remoteConfigGetter.IsFeatureEnabled(remote.FeatureDedicatedServer) {
+			return &pb.Payload{Type: internal.CodeFailure}, nil
```
thegravi: If DS from RC is disabled we return `CodeFailure`, and this gets translated to:
```go
case internal.CodeFailure, internal.CodeEmptyPayloadError:
	return formatError(fmt.Errorf(client.ConnectCantConnect))
```
and `ConnectCantConnect` is -> `"We couldn't connect you to the VPN. Please check your internet connection and try again. If the issue persists, contact our customer support."`

daemon/jobs.go
```diff
@@ -319,6 +319,31 @@ func (r *RPC) fallbackTechnology(targetTechnology config.Technology) error {
 	return nil
 }
 
+// fallbackDedicatedServer changes autoconnection target to a generic(fastest) server if the current target is set to a
+// dedicated server and user doesn't have a dedicated server subscription or feature is disabled in the remote config.
+// New config is saved and returned.
+func (r *RPC) fallbackDedicatedServer(cfg config.Config) config.Config {
```
thegravi, line 325: Can we have some tests for this?

daemon/jobs_test.go
```diff
@@ -193,6 +194,63 @@ func TestDoAutoConnect(t *testing.T) {
...
+	category.Set(t, category.Unit)
+
+	mockConfigManager := newMockConfigManager()
+	updateAutoconnectData(mockConfigManager, config.AutoConnectData{
+		Group:     config.ServerGroup_DEDICATED_SERVER,
+		ServerTag: "Dedicated Server",
+	})
+
+	remoteConfigMock := mock.NewRemoteConfigMock()
+	remoteConfigMock.AddFeatureToggle(remote.FeatureDedicatedServer, true)
```
thegravi: Isn't this dead code?

I see `testRPC()` has defaults on DS.

cli/messages.go
```diff
@@ -397,8 +397,9 @@ Your browsing activities remain private, regardless of your choice.
...
-	DedicatedServersNoServersAvailable    = "Setup your dedicated server. One last step - pick a server location.\n%s"
-	DedicatedServersServerNotReadyMessage = "Activating your dedicated server. This may take a few minutes."
-	DedicatedServersNoNordlynxMessage     = "NordLynx technology must be set to use this feature."
+	DedicatedServersNoServiceMessage           = "Get a server that's exclusively yours. Enjoy your own IP with resources and speed reserved entirely for you. Use it with port forwarding to host services and access devices remotely. To get a personal dedicated server, continue in the browser: %s"
+	DedicatedServersNoServersAvailable         = "Setup your dedicated server. One last step - pick a server location.\n%s"
+	DedicatedServersServerNotReadyMessage      = "Activating your dedicated server. This may take a few minutes."
+	DedicatedServersNoNordlynxMessage          = "NordLynx technology must be set to use this feature."
```
thegravi, line 402: [nit] Duplicates `MsgMeshnetNordlynxMustBeEnabled`.

daemon/rpc_connect.go
```diff
@@ -39,6 +39,12 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
 	return ids
 }
 
+// IsDedicatedServer returns true if either serverTag or serverGroup represent the dedicated server group
```
thegravi:
```suggestion
// IsDedicatedServer returns true if either serverTag or serverGroup represents the dedicated server group
```

daemon/rpc_connect.go
```diff
@@ -39,6 +39,12 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
...
+func IsDedicatedServer(serverTag string, serverGroup string) bool {
+	return groupConvert(serverTag) == config.ServerGroup_DEDICATED_SERVER ||
+		groupConvert(serverGroup) == config.ServerGroup_DEDICATED_SERVER
```
thegravi, line 45: What would happen here (ant in a call site) if we provide:
`serverTag = p2p` and `serverGroup = dedicated_server`? Or would that be handlers earlier?

bartoszWojciechO: It's handled later actually, there's an error returned by server selection in cases where server tag and server group are a group.

daemon/jobs.go
bartoszWojciechO: Yeah, there was some discussion about this already. I've removed the error check later. In case of no-net the fallback will not happen and the autoconnect job will keep looping until it either succeeds or hits the no-service condition.

daemon/rpc_connect.go
bartoszWojciechO: Added check in the post-pause connection flow.

daemon/jobs.go
bartoszWojciechO: There already are tests for this `TestDoAutoConnect_DedicatedServerFallback_ServiceExpired`(also updated to cover the fetch failure case) and `TestDoAutoConnect_DedicatedServerFallback_FeatureDisabledInRemoteConfig`. Do you have some specific case in mind?

thegravi (approved):

PR #1531, opened 2026-05-28
LVPN-10581, LVPN-10512: Fix tray status is not updating

mariusSincovici:
Make `renderChan` buffered channel to not miss rebuilds because of daemon notifications.
To prevent deadlock remove locks from the functions called by `rebuildTray`.

### DIFF
--- a/tray/daemon_state_handler.go
+++ b/tray/daemon_state_handler.go
@@ -7,6 +7,7 @@ import (
 )
 
 func (ti *Instance) onDaemonStateEvent(item *pb.AppState) {
+	log.Debugf("app state received %v\n", item)
 	changed := false
 	switch st := item.GetState().(type) {
 	case *pb.AppState_Error:
--- a/tray/state_listener.go
+++ b/tray/state_listener.go
@@ -46,7 +46,7 @@ func (l *stateListener) Stop() {
 	}
 }
 
-func (l *stateListener) consumeStream(server grpc.ServerStreamingClient[pb.AppState]) {
+func (l *stateListener) consumeStream(ctx context.Context, server grpc.ServerStreamingClient[pb.AppState]) {
 	for {
 		state, err := server.Recv()
 		if err != nil {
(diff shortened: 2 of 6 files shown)


tray/daemon_state_handler.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 func (ti *Instance) onDaemonStateEvent(item *pb.AppState) {
+	log.Debugf("app state received %v\n", item)
```
thegravi, line 10: `item` has full content of `pb.AppState`, let's cut it down to `item.State`.

tray/state_listener.go
```diff
@@ -57,19 +57,20 @@ func (l *stateListener) consumeStream(server grpc.ServerStreamingClient[pb.AppSt
 		select {
 		case l.queue <- state:
 		case <-time.After(time.Second):
-			log.Printf("%s %s App state consumer's queue is full, dropping\n", logTag, internal.WarningPrefix)
+			log.Warnf(logTag, "App state consumer's queue is full, dropping: %v\n", state)
```
thegravi: `logTag` here is treated as log message format.

tray/tray.go
```diff
@@ -260,7 +260,9 @@ func (ti *Instance) Start() {
 
 	ti.state.vpnStatus = pb.ConnectionState_DISCONNECTED
 	ti.state.notificationsStatus = Invalid
-	ti.renderChan = make(chan struct{})
+	// allow more events to be queued in case one goroutine is busy building the UI,
+	// while multiple events are received from the daemon
+	ti.renderChan = make(chan struct{}, 5)
```
thegravi, line 265: 🪄

thegravi (approved):

bartoszWojciechO (approved):

tray/tray.go
mariusSincovici: If you mean from where `5`. Just random value, to be sure we don't miss something on the tray. :smile:

PR #1536, opened 2026-05-29
LVPN-9620: Error Notification Service

thegravi:
**E**rror **N**otification **S**ervice feature branch


test/mock/core/servers.go
```diff
@@ -1,29 +1,23 @@
-package core
+package core_test
 
-import (
-	"net/http"
+// type ServersAPIMock struct {
```
DoYouEvenCpp: I do think this might be off?

DoYouEvenCpp (approved):

edvyno1 (approved):

PR #1541, opened 2026-06-01
LVPN-10602: Fix DNS address validation

mariusSincovici:
Add a single place where the DNS address is verified if is valid or not. 
Add test to check that the settings are kept after daemon reboot.

### DIFF
--- a/daemon/migrations.go
+++ b/daemon/migrations.go
@@ -28,7 +28,7 @@ func MigrateDeprecatedRegionalAutoconnect(cm config.Manager) error {
 	})
 }
 
-// configCleanup - validate/cleanup DNS addresses, allowlist subnets
+// ConfigCleanup - validate/cleanup DNS addresses, allowlist subnets
 func ConfigCleanup(c config.Config) config.Config {
 	// Remove all nameservers with IPv6 addresses
 	var dnsList []string
--- a/daemon/migrations.go
+++ b/daemon/migrations.go
@@ -33,7 +33,7 @@ func ConfigCleanup(c config.Config) config.Config {
 	// Remove all nameservers with IPv6 addresses
 	var dnsList []string
 	for _, addr := range c.AutoConnectData.DNS {
-		if internal.IsDNSAddressValid(addr) {
+		if internal.IsAddressValidAsDNSServer(addr) {
 			dnsList = append(dnsList, addr)
 		} else {
 			log.Warn("remove invalid DNS address from the list", addr)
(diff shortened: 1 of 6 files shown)


internal/validators.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+import "net"
+
+func IsDNSAddressValid(address string) bool {
```
thegravi: [nit] Please add docstring and let's mention it's for ipv4 only

daemon/migrations.go
```diff
@@ -25,3 +27,24 @@ func MigrateDeprecatedRegionalAutoconnect(cm config.Manager) error {
...
+			dnsList = append(dnsList, addr)
+		} else {
+			log.Warn("remove invalid DNS address from the list", addr)
+		}
+	}
+	c.AutoConnectData.DNS = dnsList
+
+	// Remove overlapping, invalid and IPv6 subnets, if any
+	c.AutoConnectData.Allowlist.NormalizeSubnets(func(removed, reason string) {
+		log.Println(internal.WarningPrefix, "On start, allowlist remove subnet:", removed, "; reason:", reason)
```
thegravi: [small] `log.Println(internal.WarningPrefix` -> `log.Warn`

thegravi (approved):

daemon/migrations.go
```diff
@@ -25,3 +27,24 @@ func MigrateDeprecatedRegionalAutoconnect(cm config.Manager) error {
 		return c
 	})
 }
+
+// configCleanup - validate/cleanup DNS addresses, allowlist subnets
+func ConfigCleanup(c config.Config) config.Config {
```
thegravi: [nit] Now it's exported -> emphasis on a comment

mariusSincovici: Yes, because it was moved in another module.

mariusSincovici:
I've saw also that the implementation was not quite correct, since it accepted `::ffff:192.168.1.1`. 
I've changed to use `netip.ParseAddr` instead, which gives more correct results.

internal/validators.go
mariusSincovici: I don't think that adding docstring for the function brings any real value, since it contains a single line of code. Also the entire application is only working with IPv4 only. :thinking: 

I've changed the name of the function to make it more clear what it does: `IsAddressValidAsDNSServer`.

daemon/migrations.go
mariusSincovici: Done

dfetti (approved):

PR #1545, opened 2026-06-02
LVPN-10547: Add validation of login response

devzbysiu:
Changes:

- login response is validated before used (for both browser and token flows)
- in dev builds, allow to configure proxy

### DIFF
--- a/core/authentication.go
+++ b/core/authentication.go
@@ -5,10 +5,12 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"sync"
 
+	"github.com/NordSecurity/nordvpn-linux/daemon/response"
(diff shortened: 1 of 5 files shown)


dfetti (approved):

core/authentication.go
```diff
@@ -114,6 +106,22 @@ func (o *OAuth2) Token(exchangeToken string) (*LoginResponse, error) {
...
+	var tokenResp LoginResponse
+	err = json.Unmarshal(body, &tokenResp)
+	if err != nil {
+		return nil, err
+	}
+
+	return &tokenResp, nil
+}
+
+func (o *OAuth2) doValidatedRequest(req *http.Request, errContext string) ([]byte, error) {
```
mariusSincovici, line 124: Not sure it it makes sense to pass `errContext string` only for the error format 🤔

mariusSincovici (approved):

PR #1547, opened 2026-06-02
LVPN-10603: Add pause menu in tray

dfetti:
This also fixes IsMeshPeer field received from Status RPC which was uninitialized and caused some errors in the GUI.

### DIFF
--- a/daemon/rpc_status.go
+++ b/daemon/rpc_status.go
@@ -61,6 +61,7 @@ func (r *RPC) Status(context.Context, *pb.Empty) (*pb.StatusResponse, error) {
 		Obfuscated:                status.IsObfuscated,
 		PausedAt:                  timestamppb.New(status.PausedAt),
 		PauseRemainingDurationSec: status.PauseRemainingTimeSec,
+		IsMeshPeer:                status.IsMeshnetPeer,
 	}, nil
 }
 
--- a/tray/actions.go
+++ b/tray/actions.go
@@ -258,6 +258,25 @@ func (ti *Instance) disconnect() bool {
 	return true
 }
 
+func (ti *Instance) pause(pauseLength PauseLength) bool {
+	resp, err := ti.client.PauseConnection(context.Background(), &pb.PauseRequest{Seconds: pauseLength.DurationSeconds})
+	if err != nil {
+		ti.notify(NoForce, "Pause error: %s", err)
+		return false
+	}
+
+	switch resp.Type {
+	case internal.CodePauseAttemptWhenConnectedToMeshPeer:
+		log.Printf("%s Pause attempt when connected to meshnet peer", internal.ErrorPrefix)
+		return false
+	case internal.CodeFailure:
+		log.Printf("%s Pause attempt failed", internal.ErrorPrefix)
+		ti.notify(NoForce, "Pause attempt failed")
+		return false
+	}
+	return true
+}
+
 func (ti *Instance) setNotify(flag bool) bool {
 	flagText := getFlagText(flag)
 	resp, err := ti.client.SetNotify(context.Background(), &pb.SetNotifyRequest{
--- a/tray/actions.go
+++ b/tray/actions.go
@@ -259,7 +259,8 @@ func (ti *Instance) disconnect() bool {
 }
 
 func (ti *Instance) pause(pauseLength PauseLength) bool {
-	resp, err := ti.client.PauseConnection(context.Background(), &pb.PauseRequest{Seconds: pauseLength.DurationSeconds})
+	ctx := attachUIEventMetadata(context.Background(), pb.UIEvent_PAUSE, pauseLength.EventValue)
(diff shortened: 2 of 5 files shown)


tray/actions.go
```diff
@@ -242,6 +242,25 @@ func (ti *Instance) disconnect() bool {
...
+func (ti *Instance) pause(pauseLength PauseLength) bool {
+	resp, err := ti.client.PauseConnection(context.Background(), &pb.PauseRequest{Seconds: pauseLength.DurationSeconds})
+	if err != nil {
+		ti.notify(NoForce, "Pause error: %s", err)
+		return false
+	}
+
+	switch resp.Type {
+	case internal.CodePauseAttemptWhenConnectedToMeshPeer:
+		log.Printf("%s Pause attempt when connected to meshnet peer", internal.ErrorPrefix)
```
keliramu: As a user, I would like to see some feedback what has happened when I tried to Pause. Now only log message is written.

keliramu: I understand, this situation should not happen, but if it happens, would be good to notify user what is the problem

daemon/rpc_status.go
```diff
@@ -61,6 +61,7 @@ func (r *RPC) Status(context.Context, *pb.Empty) (*pb.StatusResponse, error) {
 		Obfuscated:                status.IsObfuscated,
 		PausedAt:                  timestamppb.New(status.PausedAt),
 		PauseRemainingDurationSec: status.PauseRemainingTimeSec,
+		IsMeshPeer:                status.IsMeshnetPeer,
```
keliramu, line 64: Naming of this variable could be improved - if I understood it correctly, it should be true if we are connected to mesh peer vs if we are connected to vpn server. I would rename this variable to `IsTargetMeshPeer` or `IsConnectedToMeshPeer`

tray/actions.go
```diff
@@ -242,6 +242,25 @@ func (ti *Instance) disconnect() bool {
...
+		return false
+	case internal.CodeFailure:
+		log.Printf("%s Pause attempt failed", internal.ErrorPrefix)
+		ti.notify(NoForce, "Pause attempt failed")
+		return false
+	}
```
keliramu, line 278: Does not handle `CodeNothingToDo` response, but maybe not a problem

tray/menu.go
```diff
@@ -280,6 +289,62 @@ func buildQuickConnectButton(ti *Instance) {
...
+		DurationSeconds: 24 * 60 * 60,
+	},
+}
+
+func buildPauseMenu(ti *Instance) {
+	if ti == nil {
+		return
+	}
+
+	pauseMenu := systray.AddMenuItem("Pause", "Pause tooltip")
```
keliramu: `Pause tooltip` probably should be replaced with some appropriate text

keliramu (approved):

tray/actions.go
```diff
@@ -242,6 +242,27 @@ func (ti *Instance) disconnect() bool {
...
+	ctx := attachUIEventMetadata(context.Background(), pb.UIEvent_PAUSE, pauseLength.EventValue)
+	resp, err := ti.client.PauseConnection(ctx, &pb.PauseRequest{Seconds: pauseLength.DurationSeconds})
+	if err != nil {
+		ti.notify(NoForce, "Pause error: %s", err)
+		return false
+	}
+
+	switch resp.Type {
+	case internal.CodePauseAttemptWhenConnectedToMeshPeer:
+		log.Printf("%s Pause attempt when connected to meshnet peer", internal.ErrorPrefix)
```
DoYouEvenCpp: log.Error() - please apply to other newly introduced logs within this PR.

tray/menu.go
```diff
@@ -280,6 +290,68 @@ func buildQuickConnectButton(ti *Instance) {
 	go handleQuickConnectClick(ti, item)
 }
 
+type PauseLength struct {
```
DoYouEvenCpp: Does `PauseLength` have to be exported?

tray/actions.go
```diff
@@ -258,6 +258,27 @@ func (ti *Instance) disconnect() bool {
 	return true
 }
 
+func (ti *Instance) pause(pauseLength pauseLength) bool {
+	ctx := attachUIEventMetadata(context.Background(), pb.UIEvent_PAUSE, pauseLength.EventValue)
+	resp, err := ti.client.PauseConnection(ctx, &pb.PauseRequest{Seconds: pauseLength.DurationSeconds})
+	if err != nil {
+		ti.notify(NoForce, "Pause failed. Please try again.")
```
DoYouEvenCpp, line 265: [nit] maybe we could create tray/message.go and keep newly (in future we can move every tray related copy) added texts there?

DoYouEvenCpp (approved):

tray/actions.go
dfetti: I'll do that in the timer PR, which I'm currently working on.

PR #1549, opened 2026-06-03
LVPN-10294: Integrate libtelio ENS feature

thegravi:
Added support for libtelio's Error Notification Service (ENS) in the daemon. VPN connection errors reported by libtelio node events are now extracted, mapped to protocol-agnostic VPN error codes, and published on the internal VPN event bus.

### DIFF
--- a/daemon/vpn/nordlynx/libtelio/libtelio.go
+++ b/daemon/vpn/nordlynx/libtelio/libtelio.go
@@ -71,17 +71,21 @@ func eventCallbackWrap(callbackHandler *telioCallbackHandler) eventCb {
 
 type callbackHandler interface {
 	handleEvent(teliogo.Event) error
-	setMonitoringContext(ctx context.Context)
+	setConnectionMonitoringContext(ctx context.Context)
 }
 
 type telioCallbackHandler struct {
-	mu                sync.Mutex
-	statesChan        chan<- state
-	monitoringContext context.Context
+	mu                          sync.Mutex
+	statesChan                  chan<- state
+	connectionMonitoringContext context.Context
+	vpnErrorsChan               chan<- teliogo.VpnConnectionError
 }
 
-func newTelioCallbackHandler(statesChan chan<- state) *telioCallbackHandler {
-	return &telioCallbackHandler{statesChan: statesChan}
+func newTelioCallbackHandler(
+	statesChan chan<- state,
+	vpnErrorsChan chan<- teliogo.VpnConnectionError,
+) *telioCallbackHandler {
+	return &telioCallbackHandler{statesChan: statesChan, vpnErrorsChan: vpnErrorsChan}
 }
 
 func (t *telioCallbackHandler) handleEvent(e teliogo.Event) error {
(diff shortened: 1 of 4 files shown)


daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -92,6 +96,13 @@ func (t *telioCallbackHandler) handleEvent(e teliogo.Event) error {
 		log.Printf(internal.InfoPrefix+" received event %T: %s\n", e, maskPublicKey(string(eventBytes)))
 	}
 
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	ctx := t.connectionMonitoringContext
```
mariusSincovici, line 101: Any reason to assign it in a local variable? Or it is just not to write so much 😄?

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -845,3 +884,58 @@ func monitorConnection(
...
+
+// startConnectionErrorMonitor starts connection error monitoring in a goroutine and returns only
+// once it is running
+func startConnectionErrorMonitor(
+	ctx context.Context,
+	errorsChan <-chan teliogo.VpnConnectionError,
+	eventsPublisher *vpn.Events,
+) {
+	var wg sync.WaitGroup
+	wg.Add(1)
```
mariusSincovici, line 896: [small] `wg.Go(func() {` is the same as wg.Add + go func. 
wg.wait is still needed.

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -845,3 +884,58 @@ func monitorConnection(
...
+	case teliogo.VpnConnectionErrorConnectionLimitReached:
+		return events.VPNConnectionErrorConnectionLimitReached
+	case teliogo.VpnConnectionErrorServerMaintenance:
+		return events.VPNConnectionErrorServerMaintenance
+	case teliogo.VpnConnectionErrorUnauthenticated:
+		return events.VPNConnectionErrorUnauthenticated
+	case teliogo.VpnConnectionErrorSuperseded:
+		return events.VPNConnectionErrorSuperseded
+	default:
+		return events.VPNConnectionErrorUnknown
```
mariusSincovici, line 940: Wouldn't it make sense to log log.Debug for what we don't handle?

mariusSincovici (approved):

daemon/vpn/nordlynx/libtelio/libtelio.go
thegravi: The latter one 😄

daemon/vpn/nordlynx/libtelio/libtelio.go
thegravi: Why not. But it would only benefit us if we had older lib version, where the library has more errors, and we still haven't support them yet. FYI, if libtelio does not recognize an error incoming from the server, then it's set as `Unknown`.

daemon/vpn/nordlynx/libtelio/libtelio.go
thegravi: In this case, we need to signal about start of go routine, and with `wg.Go()`, we'd signal when the routine has ended. Unfortunately, I cannot apply it here. With negative counter for Add/Done we'd get a panic.

```go
func (wg *WaitGroup) Go(f func()) {
	wg.Add(1)
	go func() {
		defer wg.Done()
		f()
	}()
}
```

dfetti (approved):

PR #1550, opened 2026-06-03
LVPN-10641: Move server selection in a separate package

mariusSincovici:
* A new internal package `serverpicker` was added which contains the server selection from servers.go, split over several files. 
* Use the same filtering rules for all the servers, no matter if they are from API or local
* Update servers API call to include country.id and city.id. Having this will help for the future not to use anymore countries list, but get country/city info from the servers list.
* Removes duplicated mock API classes used for tests. 
* Changed mock packages from core to core_test because there were conflicts with the non-testing core package.


keliramu (approved):

daemon/serverpicker/remote_server_selection.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/NordSecurity/nordvpn-linux/core"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/log"
+	"github.com/google/uuid"
+)
+
+const (
+	recommendationUUIDHeader = "X-Recommendation-Uuid"
+	emptyUUUID               = ""
```
devzbysiu: typo `UUU`

daemon/serverpicker/server_selection.go
```diff
@@ -0,0 +1,193 @@
...
+	if serverGroup == config.ServerGroup_DEDICATED_IP {
+		// DIP servers are selected from the user subscription services
+		return ServerSelection{}, ErrDedicatedIPServer
+	}
+
+	// construct the servers list filters, for matching the current settings
+	localSelFn := selectFilterForLocalServers(input.Tag, serverGroup, obfuscated)
+	filterServersFn := func(s core.Server) bool {
+		return core.IsConnectableWithProtocol(tech, protocol)(s) &&
+			(core.IsObfuscated()(s) == obfuscated) &&
```
devzbysiu: [small] maybe we could extract this as a separate predicate? It's used here, two times in `dip.go` and also in `rpc_server.go`

daemon/serverpicker/tag.go
```diff
@@ -0,0 +1,142 @@
...
+
+func matchInputToServerLocation(server core.Server, tag string) *core.ServerTag {
+	for _, country := range server.Locations {
+		countryName := internal.SnakeCase(country.Name)
+		countryCode := internal.SnakeCase(country.Code)
+
+		if strings.EqualFold(tag, countryName) || strings.EqualFold(tag, countryCode) {
+			if country.ID == 0 {
+				// return only when it contains valid data
+				return nil
```
devzbysiu, line 115: Shouldn't this `continue`? 🤔 Also at line `#122` below

test/mock/core/mock_data.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	for i, server := range servers {
+		servers[i].Keys = generateKeys(server)
+	}
+
+	return servers
+}
+
+// duplicate from jobs_servers.go
```
devzbysiu: Can't we reuse implementation from `job_servers.go`?

devzbysiu (approved):

PR #1553, opened 2026-06-08
LVPN-10449: no net on daemon boot

edvyno1:
This PR addresses these issues:
 * Daemon starting too early, before network is up, causing our autoconnect server picking to fail due to no access to api, falling back to random server
 * HTTP3 transport not having fwmark support causing killswitch users that are not connected to the VPN, cause delays when performing app actions that require interaction with the api, to wait out the HTTP3 rotation timer

They were solved by
 * Adding fwmark support for http3 transport
 * Setting up a second systemd service that acts as a killswitch enabler if the feature is enabled in the user settings. This service starts at the same time as our current daemon would, as early as possible. The normal daemon now runs after systemd target of network.online


cmd/daemon/transports.go
```diff
@@ -92,27 +94,76 @@ func createH1Transport(resolver network.DNSResolver, fwmark uint32) func() http.
...
-	pool, err := x509.SystemCertPool()
-	if err != nil {
-		log.Fatal(err)
+func createH3Transport(resolver network.DNSResolver, fwmark uint32) func() http.RoundTripper {
+	return func() http.RoundTripper {
+		pool, err := x509.SystemCertPool()
+		if err != nil {
+			log.Fatal(err)
+		}
+		// // as of quic-go 0.40.1, GSO handling causes race conditions
```
DoYouEvenCpp:
```suggestion
		// as of quic-go 0.40.1, GSO handling causes race conditions
```

cmd/daemon/transports.go
```diff
@@ -92,27 +94,76 @@ func createH1Transport(resolver network.DNSResolver, fwmark uint32) func() http.
...
+		_ = os.Setenv("QUIC_GO_DISABLE_GSO", "true")
+		udpConn, err := NewMarkedUDPConn(fwmark)
+		if err != nil {
+			log.Error(err)
+		}
+		quicTransport := &quic.Transport{Conn: udpConn}
```
DoYouEvenCpp, line 121: What if `NewMarkedUDPConn` fails to create a new connection and we pass `nil` here?
This is a potential panic, later in the runtime.

Shall we carry on upon failure after calling `NewMarkedUDPConn`?

contrib/systemd/system/nordvpnd.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=NordVPN Daemon
 Requires=nordvpnd.socket
-After=network-pre.target
+Wants=nordvpnd-killswitch.service
+After=nordvpnd-killswitch.service network-online.target
```
DoYouEvenCpp: Don't we need to have `Wants=network-online.target` as per [this](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html?__goaway_challenge=meta-refresh&__goaway_id=661f912f352a3aa969ec844cab538468#network-online.target):
> Units that strictly require a configured network connection should pull in network-online.target (via a Wants= type dependency) and order themselves after it.

?
```suggestion
Wants=network-online.target
After=nordvpnd-killswitch.service network-online.target
```

contrib/systemd/system/nordvpnd-killswitch.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
```
DoYouEvenCpp, line 1: Don't we have to add here `DefaultDependencies=no` ?
Also don't we miss to run it *before* `network-pre.target` so the actual order is: killswitch -> network -> daemon usual service? - to prevent any possible leaks?
This way (combined with the other comment for the `contrib/systemd/system/nordvpnd.service`) we might event remove `Before=nordvpnd.service` below.

What's your view on that?

DoYouEvenCpp: For the `DefaultDependencies` there's a reference: [link](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html?__goaway_challenge=meta-refresh&__goaway_id=20fa2616eaf7e2c1c3eed08e6aa29834&__goaway_referer=https%3A%2F%2Fwww.google.com%2F#Default%20Dependencies)

contrib/systemd/system/nordvpnd.service
DoYouEvenCpp, line 19: Don't we need to have something akin for the killswitch service now (specifically we run it now in the scriplets)? 
This is to make it a bit more bulletproof, eg. upon removal/change forthe `Wants` here, killswitch isn't pulled into the boot process.

contrib/systemd/system/nordvpnd-killswitch.service
edvyno1: imo no need for default dependencies, our daemon used to just have `before=network-pre.target` and it would start before anything would happen so I will be adding that part

contrib/systemd/system/nordvpnd.service
edvyno1: doesn't the original `Wants=nordvpnd-killswitch.service` pull it in already?

cmd/daemon/transports.go
edvyno1: addressed now with log.fatal if there is an err for udpconn creation, will need to address this later properly

cmd/daemon/transports.go
```diff
@@ -102,27 +106,76 @@ func createSimpleH1Transport(environment string) func() http.RoundTripper {
...
+		// #nosec G402 -- minimum tls version is controlled by the standard library
+		return &http3.Transport{
+			QUICConfig: &quic.Config{
+				MaxIdleTimeout: request.TransportTimeout,
+			},
+			TLSClientConfig: &tls.Config{
+				RootCAs: pool,
+			},
+			Dial: func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (*quic.Conn, error) {
```
bartoszWojciechO, line 132: Because we are missing function signature in this case, it would be good to have a comment that generally explains why we need a custom dial function.

DoYouEvenCpp (approved):

bartoszWojciechO (approved):

PR #1554, opened 2026-06-09
LVPN-10676: force device key registration when device is not recognized

bartoszWojciechO:
When device is deleted from dedicated servers web interface, the registration data(primarily device UUID) stored on the device becomes invalidated. In such cases, calls to /v1/dedicated-servers/<server-uuid>/connect return error 910001(device not found).

This PR adds forced key re-registration so that user doesn't have to manually set settings to default to be able to connect.

### DIFF
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -318,14 +318,13 @@ func (r *RPC) connect(
 			log.Println(internal.ErrorPrefix, "failed to fetch the device key for dedicated server connection")
 			return false, internal.ErrUnhandled
 		}
-		creds.NordLynxPrivateKey = dedicatedServersDeviceData.DevicePrivateKey
 
 		dedicatedServerConnectionData, err := getDedicatedServerConnectionData(
 			r.dedicatedServersAPI,
 			serverSelection.server.DedicatedServerUUID,
 			*dedicatedServersDeviceData)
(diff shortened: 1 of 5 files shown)


daemon/rpc_connect.go
```diff
@@ -324,6 +324,18 @@ func (r *RPC) connect(
...
+		if errors.Is(err, core.ErrDedicatedServersDeviceNotFound) {
+			dedicatedServersDeviceData := r.dedicatedServerKeyManager.ForceRegisterDedicatedServers()
+			if dedicatedServersDeviceData == nil {
+				log.Println(internal.ErrorPrefix, "failed to force dedicated server device registration")
+				return true, srv.Send(&pb.Payload{Type: internal.CodeDedicatedServersCanNotConnect})
+			}
+			dedicatedServerConnectionData, err = getDedicatedServerConnectionData(
+				r.dedicatedServersAPI,
+				serverSelection.server.DedicatedServerUUID,
+				*dedicatedServersDeviceData)
```
keliramu, line 335: Is device key unchanged here? maybe need `creds.NordLynxPrivateKey = dedicatedServersDeviceData.DevicePrivateKey` ?

DoYouEvenCpp (approved):

keliramu (approved):

PR #1560, opened 2026-06-11
LVPN-10353: Fix failing tests

ggediminass:
Attempt to address failing tests related mainly to Meshnet.


test/qa/test_meshnet.py
```diff
@@ -95,37 +98,68 @@ def test_mesh_removed_machine_by_other():
...
-        assert f"-A POSTROUTING -s {peer_ip}/32 ! -d 100.64.0.0/10 -m comment --comment nordvpn -j MASQUERADE" not in rules, "MASQUERADE rule should not exist when routing is disabled"
+    tester_hostname = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list()).get_this_device().hostname
+    qapeer_hostname = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list()).get_external_peer().hostname
+
+    # Routing denied test
+    meshnet.set_permissions(qapeer_hostname, False, False, False, False)
+    ssh_client.exec_command("nordvpn mesh peer refresh")
+    with pytest.raises(RuntimeError) as ex:
+        ssh_client.exec_command(f"nordvpn mesh peer connect {tester_hostname}")
+        assert (meshnet.MSG_ROUTING_FAIL % tester_hostname) in ex, "qa-peer should not be able to route traffic"
```
keliramu: In case of exception this line and the next following line is never executed, to check exception condition different approach should be used

test/qa/test_meshnet.py
```diff
@@ -95,37 +98,68 @@ def test_mesh_removed_machine_by_other():
...
-
-    #rules = sh.sudo.iptables("-S", "POSTROUTING", "-t", "nat")
-    rules = os.popen("sudo iptables -S POSTROUTING -t nat").read()
-
-    if routing:
-        assert f"-A POSTROUTING -s {peer_ip}/32 ! -d 100.64.0.0/10 -m comment --comment nordvpn -j MASQUERADE" in rules, "MASQUERADE rule should exist when routing is enabled"
-    else:
-        assert f"-A POSTROUTING -s {peer_ip}/32 ! -d 100.64.0.0/10 -m comment --comment nordvpn -j MASQUERADE" not in rules, "MASQUERADE rule should not exist when routing is disabled"
+    tester_hostname = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list()).get_this_device().hostname
+    qapeer_hostname = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list()).get_external_peer().hostname
```
keliramu: Two lines above make 2 gRPC calls, it could be optimized to single gRPC call:
```
peers = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list())
tester_hostname = peers.get_this_device().hostname
qapeer_hostname = peers.get_external_peer().hostname
```

test/qa/test_meshnet.py
```diff
@@ -95,37 +98,68 @@ def test_mesh_removed_machine_by_other():
...
+    last_qapeer_tsfr = fileshare.get_last_transfer(True, ssh_client)
+    with lib.Defer(lambda: ssh_client.exec_command(f"nordvpn fileshare cancel {last_qapeer_tsfr}")):
+        assert f"File transfer {last_qapeer_tsfr} has started in the background." in transfer_started_msg, "qa-peer should be able to send file to tester"
+
+    # Fileshare denied test
+    meshnet.set_permissions(qapeer_hostname, False, False, False, False)
+    ssh_client.exec_command("nordvpn mesh peer refresh")
+    with pytest.raises(RuntimeError) as ex:
+        ssh_client.exec_command(f"nordvpn fileshare send --background {tester_hostname} /usr/bin/echo")
+        assert "This peer does not allow file transfers from you." in transfer_started_msg, "qa-peer should not be able to send file to tester"
```
keliramu: `with` is not the right approach to use here - if exception is raised then line #162 is not executed, and if no exception then you get `Failed: DID NOT RAISE RuntimeError`

keliramu (approved):

test/qa/test_meshnet_invite.py
```diff
@@ -41,7 +41,8 @@ def test_invite_send_repeated():
     assert "Meshnet invitation for 'test@test.com' already exists." in ex.value.stderr.decode("utf-8"), "Repeated invite should show error message"
 
 
-@pytest.mark.skipif(meshnet.is_meshnet_test_disabled_from_run(), reason="TODO: LVPN-9458")
+@pytest.mark.skip("LVPN-9458")
+#@pytest.mark.skipif(meshnet.is_meshnet_test_disabled_from_run(), reason="TODO: LVPN-9458")
```
bbrutall: Commented decorator can be removed.

test/qa/test_meshnet_routing.py
```diff
@@ -262,7 +262,9 @@ def test_route_to_peer_that_is_disconnected():
 
     ssh_client.exec_command("nordvpn set mesh off")
 
-    time.sleep(140)
+    # Needed for peer to show up as disconnected
+    daemon.stop()
+    daemon.start()
```
bbrutall, line 267: Looks like a workaround, isn't it?

ggediminass: Interested, why'd you call it a workaround here. I wouldn't consider it a workaround here. I'd assume, that Telio for some time holds "connected" status of peer cached, longer than I'd like to see. But when we restart the app daemon, Meshnet finally can see the actual status of peer, as it attempts to contact it.

test/qa/lib/ssh.py
```diff
@@ -67,6 +67,15 @@ def is_running(self):
             else:
                 return True
 
+        def is_connected(self) -> bool:
```
DoYouEvenCpp: Can't we reuse https://github.com/NordSecurity/nordvpn-linux/blob/main/test/qa/lib/daemon.py#L45 instead of snapping new method, akin to daemon's?

test/qa/test_meshnet.py
```diff
@@ -95,37 +98,69 @@ def test_mesh_removed_machine_by_other():
...
-    if routing:
-        assert f"-A POSTROUTING -s {peer_ip}/32 ! -d 100.64.0.0/10 -m comment --comment nordvpn -j MASQUERADE" in rules, "MASQUERADE rule should exist when routing is enabled"
-    else:
-        assert f"-A POSTROUTING -s {peer_ip}/32 ! -d 100.64.0.0/10 -m comment --comment nordvpn -j MASQUERADE" not in rules, "MASQUERADE rule should not exist when routing is disabled"
+    peers = meshnet.PeerList.from_str(sh_no_tty.nordvpn.mesh.peer.list())
+    tester_hostname = peers.get_this_device().hostname
+    qapeer_hostname = peers.get_external_peer().hostname
+
+    # Routing denied test
+    meshnet.set_permissions(qapeer_hostname, False, False, False, False)
```
DoYouEvenCpp: [nit] can we have some predefined values for those False/True (here, and below) that would basically contain info on what those parameters actually mean?

DoYouEvenCpp (approved):

bbrutall (approved):

PR #1564, opened 2026-06-11
LVPN-10606: Pause timer in tray

dfetti:
This PR add the pause timer in tray

### DIFF
--- a/tray/menu.go
+++ b/tray/menu.go
@@ -207,6 +207,11 @@ func buildConnectionSection(ti *Instance) {
 	}
 
 	buildVPNStatusLabel(ti)
+
+	if ti.state.vpnStatus == pb.ConnectionState_PAUSED {
+		buildPauseTimer(ti)
+	}
+
 	if ti.state.vpnStatus == pb.ConnectionState_CONNECTED {
 		buildConnectedToSection(ti)
 		if ti.state.vpnIsMeshPeer {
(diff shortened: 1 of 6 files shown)


tray/menu.go
```diff
@@ -345,6 +350,49 @@ func buildPauseMenu(ti *Instance) {
...
+
+		for ti.pauseTimer.GetTimer() > 0 {
+			select {
+			case _, open := <-timer.ClickedCh:
+				if !open {
+					return
+				}
+			case <-ticker.C:
+				currentValue := ti.pauseTimer.DecrementTimer()
+				timer.SetTitleQuiet(
```
mariusSincovici: This must be updated only when the menu open.

tray/tray.go
```diff
@@ -122,21 +123,23 @@ type notifier interface {
...
+	iconConnected         string
+	iconDisconnected      string
+	state                 trayState
+	quitChan              chan<- norduser.StopRequest
+	stateListener         *stateListener
+	connSensor            *connectionSettingsChangeSensor
+	recentConnections     *recentConnectionsManager
+	checkboxSync          *CheckboxSynchronizer
+	isVisible             atomic.Bool
+	stopVisibilityMonitor chan struct{}
```
DoYouEvenCpp, line 141: Does this get ever initialized? (reading/writing on nil channel is an indefinite block)

tray/menu.go
```diff
@@ -347,6 +352,60 @@ func buildPauseMenu(ti *Instance) {
...
+				if currentValue != 0 {
+					currentValue--
+					ti.state.pauseRemainingSec = currentValue
+				}
+
+				ti.state.mu.Unlock()
+
+				if ti.isVisible.Load() {
+					timer.SetTitleQuiet(
+						buildTimerString(currentValue),
```
DoYouEvenCpp, line 389: So this will update timer only when visible - are we fine with situations when tray got closed, and upon restart the timer value will be (for a really short value) completely off, until it gets recalculated and rendered?

tray/menu.go
```diff
@@ -347,6 +352,60 @@ func buildPauseMenu(ti *Instance) {
...
+			case _, open := <-timer.ClickedCh:
+				if !open {
+					return
+				}
+			case <-ticker.C:
+				ti.state.mu.Lock()
+
+				if currentValue != 0 {
+					currentValue--
+					ti.state.pauseRemainingSec = currentValue
```
mariusSincovici, line 382: What if the `pauseRemainingSec` was changed externally? Wouldn't this replace its value?

tray/tray.go
```diff
@@ -156,6 +159,7 @@ type trayState struct {
 	vpnIsMeshPeer        bool
 	initialSyncCompleted bool
 	connSelector         ConnectionSelector
+	pauseRemainingSec    uint32
```
mariusSincovici: [suggestion] to use int, not to have somewhere -1 and get back to max int32.

mariusSincovici (approved):

DoYouEvenCpp (approved):

tray/menu.go
dfetti: It's a tradeoff to not spam the dbus with updates when the tray is not visible

PR #1565, opened 2026-06-12
Return h3 wrapper with err on failed transport creation

edvyno1:
This PR adds a wrapper that returns an error if h3 transport failed to be created, this can happen in snap if the network plugs are not connected on fresh app install, which would cause the snap to exit with fatal


mariusSincovici (approved):

DoYouEvenCpp (approved):

edvyno1:
merging with red ci due to one consistently failing mesh test(to be replaced soon)

PR #1567, opened 2026-06-12
LVPN-10296: Handle server maintenance ENS event

mariusSincovici:
* On server maintenance event, the application reconnects to the VPN.
* When there a reconnect is made, then internal kill switch is enabled to prevent leaks.

### DIFF
--- a/daemon/ens/ens_monitoring.go
+++ b/daemon/ens/ens_monitoring.go
@@ -1,0 +1,90 @@
+package ens
+
+import (
+	"context"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/log"
+	"github.com/NordSecurity/nordvpn-linux/networker"
+)
+
+const (
+	logPrefix = "[ens]"
+	evChSize  = 2
+)
+
+type ConnectCallback func() error
+
+type Monitor struct {
+	eventsCh    chan events.VPNConnectionErrorEvent
+	ctx         context.Context
+	netw        networker.Networker
+	reconnectFn ConnectCallback
+	rc          remote.ConfigGetter
+}
+
+func NewMonitor(ctx context.Context, netw networker.Networker, rc remote.ConfigGetter, connectCallback ConnectCallback) *Monitor {
+	return &Monitor{
+		eventsCh:    make(chan events.VPNConnectionErrorEvent, evChSize),
+		ctx:         ctx,
+		netw:        netw,
+		rc:          rc,
+		reconnectFn: connectCallback,
+	}
+}
+
+func (m *Monitor) HandleENSNotification(e events.VPNConnectionErrorEvent) error {
+	select {
+	case m.eventsCh <- e:
+	case <-m.ctx.Done():
+		log.Debug(logPrefix, "ignore event because context is done", e)
+	case <-time.After(10 * time.Millisecond):
+		log.Warn(logPrefix, "channel is full dropping ENS event", e)
+	}
+	return nil
+}
+
+func (m *Monitor) Start() {
+	if m.reconnectFn == nil {
+		log.Fatal(logPrefix, "connect callback is nil")
+	}
+	go m.run()
+}
+
+func (m *Monitor) run() {
+	log.Info(logPrefix, "start ENS monitoring")
+
+	for {
+		select {
+		case e, ok := <-m.eventsCh:
+			if !ok {
+				continue
+			}
+
+			if !m.rc.IsFeatureEnabled(remote.FeatureENS) {
+				continue
+			}
+
+			log.Debug(logPrefix, "event received", e)
+			if e.Code != events.VPNConnectionErrorServerMaintenance {
+				log.Debug(logPrefix, "ignoring", e)
+				continue
+			}
+
+			if !m.netw.IsVPNActive() {
+				log.Debug(logPrefix, "ignoring because VPN is not connected", e)
+				continue
+			}
+			// TODO: check if current server == event.server and check r.RequestedConnParams
+			if err := m.reconnectFn(); err != nil {
+				log.Error(logPrefix, "failed to reconnect", err)
+			}
+
+		case <-m.ctx.Done():
+			log.Info(logPrefix, "stop ENS monitoring")
+			return
+		}
+	}
+}
--- a/daemon/ens/ens_monitoring.go
+++ b/daemon/ens/ens_monitoring.go
@@ -26,6 +26,9 @@ type Monitor struct {
 }
 
 func NewMonitor(ctx context.Context, netw networker.Networker, rc remote.ConfigGetter, connectCallback ConnectCallback) *Monitor {
+	if connectCallback == nil {
+		log.Fatal(logPrefix, "connect callback is nil")
+	}
 	return &Monitor{
 		eventsCh:    make(chan events.VPNConnectionErrorEvent, evChSize),
 		ctx:         ctx,
@@ -47,9 +50,6 @@ func (m *Monitor) HandleENSNotification(e events.VPNConnectionErrorEvent) error 
 }
 
 func (m *Monitor) Start() {
-	if m.reconnectFn == nil {
-		log.Fatal(logPrefix, "connect callback is nil")
-	}
 	go m.run()
 }
 
@@ -60,7 +60,8 @@ func (m *Monitor) run() {
 		select {
 		case e, ok := <-m.eventsCh:
 			if !ok {
-				continue
+				log.Warn(logPrefix, "events channel closed, stopping ENS monitoring")
+				return
 			}
 
 			if !m.rc.IsFeatureEnabled(remote.FeatureENS) {
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -31,10 +31,35 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
 
 // Connect initiates and handles the VPN connection process
 func (r *RPC) Connect(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer) (retErr error) {
-	return r.connectFromRequest(in, srv, pb.ConnectionSource_MANUAL)
+	return r.executeConnect(srv, func(ctx context.Context) (bool, error) {
+		return r.connectWithParameters(ctx, in, srv, pb.ConnectionSource_MANUAL, "")
+	})
 }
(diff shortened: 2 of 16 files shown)


daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,88 @@
...
+func (m *Monitor) Start() {
+	if m.connectFn == nil {
+		log.Fatal(logPrefix, "connect callback is nil")
+	}
+	go m.run()
+}
+
+func (m *Monitor) run() error {
+	log.Info(logPrefix, "start ENS monitoring")
+	// go func() {
```
mariusSincovici: Delete this before merging

daemon/rpc_connect.go
```diff
@@ -30,10 +29,36 @@ func determineServerGroupIDs(server *core.Server) []config.ServerGroup {
...
 // Connect initiates and handles the VPN connection process
 func (r *RPC) Connect(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer) (retErr error) {
-	return r.connectFromRequest(in, srv, pb.ConnectionSource_MANUAL)
+	return r.executeConnect(srv, func(ctx context.Context) (bool, error) {
+		return r.connectWithParameters(ctx, in, srv, pb.ConnectionSource_MANUAL, "")
+	})
 }
 
-func (r *RPC) connectFromRequest(in *pb.ConnectRequest, srv pb.Daemon_ConnectServer, source pb.ConnectionSource) error {
+func (r *RPC) ConnectFromLastSelection(srv pb.Daemon_ConnectServer,
```
mariusSincovici, line 39: I've changed the function names to start with upper case, to know which functions are already synchronized and which not.

daemon/serverpicker/serverpicker_test.go
```diff
@@ -684,7 +684,7 @@ func TestPickServer(t *testing.T) {
 					Latitude:  test.latitude,
 				},
 				cfg,
-				NewSearchParams(test.tag, test.groupFlag),
+				NewSearchParams(test.tag, test.groupFlag, ""),
```
keliramu: Would be good to have tests for `excludedServer` case

mariusSincovici: I've added a test to exclude the only server that would be found otherwise. In that case it will return an error.

daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,90 @@
...
+		connectFn: connectCallback,
+	}
+}
+
+func (m *Monitor) HandleENSNotification(e events.VPNConnectionErrorEvent) error {
+	select {
+	case m.eventsCh <- e:
+	case <-m.ctx.Done():
+		log.Debug(logPrefix, "ignore event because context is done", e)
+		close(m.eventsCh)
```
devzbysiu: Is this needed? `ctx` is used only by this monitor and `run` is already handling `ctx.Done()`. I'm not sure what will happen when channel is closed because of `ctx.Done()` and at the same time `ctx.Done()` is received in `select` in `run` below, but if the failed receive will be picked, then this 
```go
case e, ok := <-m.eventsCh:
    if !ok {
	    continue
	}
```
will spin 😅

Maybe it should be 
```go
 if !ok {
	return
}
```

Probably not a big deal because `ctx` will be closed when daemon is being stopped

daemon/rpc_connect.go
```diff
@@ -59,23 +84,34 @@ func (r *RPC) connectFromRequest(in *pb.ConnectRequest, srv pb.Daemon_ConnectSer
...
 	}
 
-	// set connection status to "Disconnected"
-	if didFail || err != nil {
-		r.events.Service.Disconnect.Publish(events.DataDisconnect{})
+	var serverTag string
+	if reqParams.ServerName != "" {
+		// specific server name was selected. Use server country + city instead
+		c := currentServer.Country()
+		serverTag = c.Code + internal.SnakeCase(c.City.Name)
```
devzbysiu: `+ " " +` in between? 🤔

devzbysiu: Also below

daemon/ens/ens_monitoring.go
mariusSincovici: To be sure I'll remove the close part, to prevent possible crash.

> ctx is used only by this monitor 

It is the same context to cancel the HTTP calls.

daemon/rpc_connect.go
mariusSincovici: Changed and added a test to verify that this works.

devzbysiu (approved):

daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,90 @@
...
+	case <-m.ctx.Done():
+		log.Debug(logPrefix, "ignore event because context is done", e)
+	case <-time.After(10 * time.Millisecond):
+		log.Warn(logPrefix, "channel is full dropping ENS event", e)
+	}
+	return nil
+}
+
+func (m *Monitor) Start() {
+	if m.reconnectFn == nil {
```
DoYouEvenCpp: Maybe move this assert up to c-tor, the `NewMonitor` ?

daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,90 @@
...
+			if e.Code != events.VPNConnectionErrorServerMaintenance {
+				log.Debug(logPrefix, "ignoring", e)
+				continue
+			}
+
+			if !m.netw.IsVPNActive() {
+				log.Debug(logPrefix, "ignoring because VPN is not connected", e)
+				continue
+			}
+			// TODO: check if current server == event.server and check r.RequestedConnParams
```
DoYouEvenCpp, line 81: Any particular task to track this one?

daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,90 @@
...
+	return &Monitor{
+		eventsCh:    make(chan events.VPNConnectionErrorEvent, evChSize),
+		ctx:         ctx,
+		netw:        netw,
+		rc:          rc,
+		reconnectFn: connectCallback,
+	}
+}
+
+func (m *Monitor) HandleENSNotification(e events.VPNConnectionErrorEvent) error {
```
DoYouEvenCpp, line 41: So, this returns only `nil` - any plans to actually leverage  `error` on the return type here?

daemon/rpc_connect.go
```diff
@@ -60,23 +85,39 @@ func (r *RPC) connectFromRequest(in *pb.ConnectRequest, srv pb.Daemon_ConnectSer
...
-		return srv.Send(&pb.Payload{Type: internal.CodeNothingToDo})
+func (r *RPC) reconnectOnServerMaintenance(
+	ctx context.Context,
+	srv pb.Daemon_ConnectServer,
+) (bool, error) {
+	reqParams := r.RequestedConnParams.Get()
+	currentServer := r.lastServerSelection.Server
+	log.Debug("[ens]", reqParams, currentServer)
+
+	group := ""
```
DoYouEvenCpp: [nit] maybe let's align this with how `serverTag` and `hostname` are declared below (eg. to leverage a variable here?)

daemon/rpc_connect.go
```diff
@@ -236,11 +280,10 @@ func (r *RPC) connectWithParameters(ctx context.Context,
 
 	inputServerTag := internal.RemoveNonAlphanumeric(in.GetServerTag())
 
-	log.Println(internal.DebugPrefix, "picking servers for", cfg.Technology, "technology", "input",
-		in.GetServerTag(), in.GetServerGroup())
-
-	serverSelection, err := selectServer(r, &insights, cfg, inputServerTag, in.GetServerGroup())
+	log.Debug("picking servers for", cfg.Technology, "technology", "input",
+		in.GetServerTag(), in.GetServerGroup(), excludedServer)
```
DoYouEvenCpp: The current statement in the log might be a bit vague, shall we:
```suggestion
	log.Debug("picking servers for", cfg.Technology, "technology", "input",
		in.GetServerTag(), in.GetServerGroup(), ", server excluded from the lookup: ", excludedServer)
```

?

daemon/serverpicker/serverpicker_test.go
```diff
@@ -593,6 +593,7 @@ func TestPickServer(t *testing.T) {
 		tag                  string
 		groupFlag            string
 		onlyPhysicServers    bool
+		excludedServer       string
 		expectedServerName   string
 		expectedRemoteServer bool
```
DoYouEvenCpp, line 594: This isn't used in any test - shall we drop it?

daemon/serverpicker/serverpicker_test.go
```diff
@@ -593,6 +593,7 @@ func TestPickServer(t *testing.T) {
 		tag                  string
 		groupFlag            string
```
DoYouEvenCpp: Not used across all tests, we can remove it and pass to `PickServer` an empty string instead?

daemon/serverpicker/serverpicker_test.go
DoYouEvenCpp: Unused across tests

DoYouEvenCpp: Unused, always defaulted to `Protocol_UNKNOWN_PROTOCOL` across tests.

daemon/ens/ens_monitoring.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (m *Monitor) run() {
+	log.Info(logPrefix, "start ENS monitoring")
+
+	for {
+		select {
+		case e, ok := <-m.eventsCh:
+			if !ok {
+				continue
```
DoYouEvenCpp: Isn't more robust in regards to open/closed principle to `return` here instead of `continue`?
In a scenario where <due to any change in the future> the channel is already closed, yet the context isn't cancelled we fall here into quite a nice loop.

WDYT?

daemon/ens/ens_monitoring.go
thegravi: We'll see if there is any need some time later, but since this is a generic signature for rpc call, we leave it as it is.

daemon/ens/ens_monitoring.go
thegravi: Agree. Also, if channel is closed an we'd send something to it we'd get a panic before hitting this code.

DoYouEvenCpp (approved):

daemon/ens/ens_monitoring.go
thegravi: No yet, will make one.

thegravi (approved):

PR #1569, opened 2026-06-15
LVPN-10390: Add nordvpn troubleshoot command

mariusSincovici:


cli/cli_troubleshoot.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+
+	"github.com/fatih/color"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/term"
+)
+
+const (
+	TroubleshootUsageText = "Collects diagnostic logs and system information for troubleshooting"
```
thegravi: What about copy for all the sentences?

cli/cli_troubleshoot.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func (c *cmd) Troubleshoot(ctx *cli.Context) error {
+	stream, err := c.client.CollectDiagnostics(context.Background(), &pb.Empty{})
+	if err != nil {
+		return formatError(err)
+	}
+
+	isTTY := term.IsTerminal(int(os.Stdout.Fd()))
```
thegravi: `isStdoutTerminal()`

cli/messages.go
```diff
@@ -393,8 +393,8 @@ Your browsing activities remain private, regardless of your choice.
 	MsgRemoveNarrowConfirmPrompt = "The range you are trying to add includes previously allowed subnets. To prevent conflicts, we'll delete the redundant ones. Do you want to remove overlapping subnets?"
 
 	SetARPIgnoreUsageText    = "Controls whether your device ignores or responds to ARP requests while the VPN is active.\nBy default, arp-ignore is on. This means ARP requests are ignored to prevent VPN exposure."
-	SetARPIgnoreSuccess      = "ARP ignore set to '%s' successfully."
-	SetARPIgnoreNothingToSet = "ARP ignore is already set to '%s'."
+	SetARPIgnoreSuccess      = "ARP ignore set to ‘%s’ successfully."
+	SetARPIgnoreNothingToSet = "ARP ignore is already set to ‘%s’."
```
thegravi: Why not change all quotes, but only these ones?

cli/messages.go
```diff
@@ -405,4 +405,8 @@ Your browsing activities remain private, regardless of your choice.
 	DedicatedServersAutoconnectNordlynxMessage = "Auto-connect to Dedicated Server requires NordLynx protocol."
 	DedicatedServerPQMessage                   = "Post-quantum encryption is not compatible with Dedicated Server."
 	DedicatedServersNoNordlynxMessage          = "Connection to Dedicated Server requires NordLynx technology."
+	// Troubleshoot
+	MsgTroubleshootSuccess    = "Diagnostics collected successfully.\nFile saved to: %s"
+	MsgTroubleshootFailure    = "Failed to collect diagnostics. Please try again or contact support."
+	MsgTroubleshootDisclaimer = "WARNING: this archive contains sensitive information about your system and configuration. Share it only with NordVPN customer support, and only over a secure channel."
```
thegravi: Copy approved?

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+	"google.golang.org/grpc/peer"
+	"google.golang.org/protobuf/encoding/prototext"
+)
+
+const (
+	// maxDaemonLogSize is the maximum size of daemon logs to collect (500 MB).
+	maxDaemonLogSize = 500 * 1024 * 1024
+
+	// maxZipFileSize caps the resulting diagnostics archive (40 MB).
+	maxZipFileSize = 40 * 1024 * 1024
```
thegravi, line 33: Shall we move these under internal constants package?

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+
+	// User-facing messages sent via pb.DiagnosticsProgress.Error. Centralised
+	// here so support can grep them, and so we never accidentally diverge two
+	// copies of the same wording.
+	zipTooLargeMsg       = "Diagnostics file exceeds 40 MB limit. Please contact support for assistance."
+	failedCreateZipMsg   = "Failed to create zip file: %v"
+	failedChownZipMsg    = "Failed to change file ownership: %v"
+	failedCollectMsg     = "Failed to collect diagnostics: %v"
+	failedCloseZipMsg    = "Failed to close zip file: %v"
+	noDaemonLogSourceMsg = "unable to extract daemon logs automatically: the daemon was not started via systemd or snap. Please contact customer support to send the logs manually"
```
thegravi: Copy?

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+	zipFile, err := createDiagnosticsZip(caller.outputDir)
+	if err != nil {
+		log.Error(logPrefix, "failed to create diagnostics zip:", err)
+		return srv.Send(&pb.DiagnosticsProgress{Error: fmt.Sprintf(failedCreateZipMsg, err)})
+	}
+	zipPath := zipFile.Name()
+
+	if snapconf.IsUnderSnap() {
+		if err := os.Chmod(zipPath, internal.PermUserRWGroupROthersR); err != nil {
+			log.Error(logPrefix, "failed to change file permissions:", err)
```
thegravi: If we don't return then I think warning-level is more appropriate here.

magefiles/mage.go
```diff
@@ -678,6 +680,8 @@ func (Test) QADockerFast(ctx context.Context, testGroup, testPattern string) err
 
 // Codeql Runs CodeQL analysis using Docker container
 func (Test) Codeql(ctx context.Context, language string) error {
+	return nil
```
thegravi: Leftovers?

protobuf/daemon/common.proto
```diff
@@ -56,3 +56,9 @@ enum ClientID {
   GUI = 2;
   TRAY = 3;
 }
+
+message DiagnosticsProgress {
+  string step = 2;
+  string file_path = 4;
+  string error = 5;
```
thegravi: Why not 1,2,3?

cli/cli_troubleshoot.go
devzbysiu: There is a ticket in progress on that

cli/messages.go
devzbysiu: No idea why this was changed really. But it shouldn't be changed, I'm bringing it back

cli/messages.go
devzbysiu: Ticket for copy is in progress

daemon/rpc_troubleshoot.go
devzbysiu: Ticket for copy is in progress

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+	}
+
+	// createDiagnosticsZip uses os.CreateTemp under the hood, so we always
+	// get a fresh file with no chance of collision — no existence check, no
+	// TOCTOU race, no overwrite of an old report, even if two collections
+	// land in the same second.
+	zipFile, err := createDiagnosticsZip(caller.outputDir)
+	if err != nil {
+		log.Error(logPrefix, "failed to create diagnostics zip:", err)
+		return srv.Send(&pb.DiagnosticsProgress{Error: fmt.Sprintf(failedCreateZipMsg, err)})
```
bartoszWojciechO: Shouldn't this follow the same convention as most of the other code? Usually we either return an error code from `nordvpn-linux/internal/codes.go` or a code defined in a `proto` file.

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+	gid       uint32
+	outputDir string
+}
+
+// resolveDiagnosticsCaller extracts the caller's UID/GID from the gRPC
+// context, looks up their user record, and picks the directory where the
+// diagnostics zip will land. The actual filename is generated atomically by
+// os.CreateTemp at write time.
+func resolveDiagnosticsCaller(ctx context.Context) (*diagnosticsCaller, error) {
+	p, ok := peer.FromContext(ctx)
```
bartoszWojciechO: We also have to extract caller data in the meshnet server https://github.com/NordSecurity/nordvpn-linux/blob/main/meshnet/server.go#L198 and in the grpc interceptors for `norduser` startup https://github.com/NordSecurity/nordvpn-linux/blob/main/norduser/grpc_interceptor.go#L29 . Maybe it's worth some common methods?

daemon/rpc_troubleshoot.go
```diff
@@ -0,0 +1,888 @@
...
+	// land in the same second.
+	zipFile, err := createDiagnosticsZip(caller.outputDir)
+	if err != nil {
+		log.Error(logPrefix, "failed to create diagnostics zip:", err)
+		return srv.Send(&pb.DiagnosticsProgress{Error: fmt.Sprintf(failedCreateZipMsg, err)})
+	}
+	zipPath := zipFile.Name()
+
+	if snapconf.IsUnderSnap() {
+		if err := os.Chmod(zipPath, internal.PermUserRWGroupROthersR); err != nil {
```
bartoszWojciechO, line 72: Do we have to handle this differently because snap doesn't allow us to change owner? Maybe this approach would be sufficient for non-snap as well, so we wouldn't have to have two different paths?

cli/cli_troubleshoot.go
devzbysiu: Addressed

cli/messages.go
devzbysiu: Addressed

daemon/rpc_troubleshoot.go
devzbysiu: They are not used anywhere else and for now should not be

daemon/rpc_troubleshoot.go
devzbysiu: Addressed

magefiles/mage.go
devzbysiu: Most likely, removed

protobuf/daemon/common.proto
devzbysiu: Don't know. No reason IMO, changed.

daemon/rpc_troubleshoot.go
devzbysiu: Good point. Changed to return error codes from proto. Messages is a CLI responsibility now

daemon/rpc_troubleshoot.go
devzbysiu: Agree. And it turns out we already have helper to do this, but was not applied everywhere. Anyway, this was bigger change, so I moved it to a separate PR [here](https://github.com/NordSecurity/nordvpn-linux/pull/1620)

daemon/rpc_troubleshoot.go
devzbysiu: Discussed

bartoszWojciechO (approved):

ci/build_snap.sh
```diff
@@ -3,8 +3,8 @@ set -euxo pipefail
 
 # configure git safe-directories when running in containerized environment
 if [[ -n ${DOCKER_ENV+x} ]]; then
-  git config --global --add safe.directory "${WORKDIR}"
-  git config --global --add safe.directory "${WORKDIR}/parts/nordvpn/build"
+  git config --global --add safe.directory "${WORKDIR}" || true
+  git config --global --add safe.directory "${WORKDIR}/parts/nordvpn/build" || true
```
edvyno1: leftovers?

ci/env.sh
```diff
@@ -11,7 +11,7 @@
 set -euxo pipefail
 
 NAME=nordvpn
-HASH=$(git rev-parse --short HEAD)
+HASH=$(git rev-parse --short HEAD || echo "badf00")
```
edvyno1: leftovers?

magefiles/mage.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	registryPrefix         = "ghcr.io/nordsecurity/nordvpn-linux/"
-	imageBuilder           = registryPrefix + "builder:1.4.4"
+	imageBuilder           = registryPrefix + "builder:1.4.5"
```
edvyno1: builder was bumped to 1.4.6 with golang upgrade, it's likely that a fresh rebase and dockerfile rebuild/repush is needed as 1.4.7 to combine both changes

snap/snapcraft.yaml
```diff
@@ -113,7 +116,7 @@ parts:
     override-build: |
       # When it is executed in git submodule and contents of `.` are copied to
       # `./parts/nordvpn/build`, .git file points to the invalid host repository
-      git config --global --add safe.directory "*"
+      git config --global --add safe.directory "*" || true
```
edvyno1: leftovers?

ci/build_snap.sh
devzbysiu: Most likely. Removed

ci/env.sh
devzbysiu: Most likely. Removed

magefiles/mage.go
devzbysiu: Good catch. Bumped

snap/snapcraft.yaml
devzbysiu: Most likely. Removed

edvyno1 (approved):

thegravi (approved):

PR #1570, opened 2026-06-15
LVPN-10141: add connection details context menu analytics

keliramu:
LVPN-10141: add connection details context menu analytics:
- click / reconnect;
- click / change vpn settings;
- click / help;


gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -63,6 +63,10 @@ class VpnStatusController extends _$VpnStatusController
   Future<void> reconnect(ConnectionParameters args) =>
       _doAndShowPopup((vpn) => vpn.reconnect(args));
 
+  Future<void> changeSettings() =>
+      _doAndShowPopup((vpn) => vpn.changeSettings());
```
devzbysiu: I think analytics should be transparent for users, so IMO we should skip `_doAndShowPopup()`. Especially in this case where user just navigates inside GUI, we should never show any popup here. Same for `getHelp()` below

daemon/rpc_uievent.go
```diff
@@ -0,0 +1,16 @@
...
+	"context"
+
+	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
+	"github.com/NordSecurity/nordvpn-linux/internal"
+)
+
+// SendUIEvent is a lightweight RPC endpoint for standalone UI analytics events.
+// The actual event extraction and publishing is handled by the gRPC middleware
+// (uievent.Middleware), which intercepts all incoming requests and publishes
+// any UI event metadata to the Moose analytics pipeline.
```
devzbysiu: Wouldn't be better to pass ui event as an arg when calling this endpoint and handling it explicitly here instead of relying on interceptor?

gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -63,6 +63,10 @@ class VpnStatusController extends _$VpnStatusController
   Future<void> reconnect(ConnectionParameters args) =>
       _doAndShowPopup((vpn) => vpn.reconnect(args));
 
+  Future<void> changeSettings() =>
```
devzbysiu: Those two methods have nothing to do with VPN status. I suggest to move it to separate class purely for analytics. 

Then maybe we could introduce methods for other analytics related work we are doing now and getting rid of this interceptor on daemon side and just pass UI Event explicitly in the request and process it in this new `SendUIEvent` endpoint (related to this comment https://github.com/NordSecurity/nordvpn-linux/pull/1570/changes#r3419180205).

The interceptor mechanism was introduced because analytics were related to daemon action that we had to call and we wanted to avoid introducing analytics specific endpoint, but now those two methods are purely UI thing and we have to have the endpoint, so interceptor mechanism is not a best fit anymore. This way we would have consistent way of dealing with analytics - single place for calling endpoint and explicit event handling in the endpoint

keliramu: Reworked - eliminated gRPC interceptors, implemented single gRPC API `ReportUIEvent`.
Here, for this comment, only removed `_doAndShowPopup` but left access to gRPC API through controller.

daemon/rpc_uievent.go
keliramu: Reworked - eliminated gRPC interceptors, implemented single gRPC API `ReportUIEvent`

gui/lib/data/providers/vpn_status_controller.dart
keliramu: Removed `_doAndShowPopup` but left access to gRPC API through controller.

gui/lib/data/providers/vpn_status_controller.dart
keliramu: UIEvent reporting moved to separate repository, for actions which do not have corresponding gRPC calls.

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -184,6 +186,16 @@ final class ConnectionCardButtons extends ConsumerWidget {
     ref.read(vpnStatusControllerProvider.notifier).pauseConnection(pauseLength);
   }
 
+  Future<void> _changeSettings(BuildContext context, WidgetRef ref) async {
```
devzbysiu: This can be `void` and no `async`, same with `_getHelp()`

daemon/rpc_uievent.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/uievent"
+)
+
+// ReportUIEvent receives a UIEvent from any client and publishes the
+// corresponding analytics action to the Moose pipeline.
+func (r *RPC) ReportUIEvent(
+	_ context.Context, in *pb.UIEvent,
+) (*pb.Payload, error) {
+	if in != nil {
```
devzbysiu: `in` can be non-nil, but still may have default, unset values. We had `IsValid` in `uievent/context.go` that verified the fields before publishing. I think it would be good to have it to not spam moose with invalid events

devzbysiu (approved):

dfetti (approved):

daemon/rpc_uievent.go
keliramu: added same checks

gui/lib/vpn/connection_card_buttons.dart
keliramu: Fixed

thegravi (approved):

PR #1573, opened 2026-06-16
LVPN-10420: fix pause menu shadow clipping on the left side

bartoszWojciechO:
SizeTransition automatically clips the size of it's children. Because of that, shadow of it's children(pause menu and context menu) is not visible on the left side - because it's clipped.

To fix that, the relationship was changed so that SizeTransition is a child of a SizedBox that contains the shadow.

### DIFF
--- a/gui/lib/widgets/context_menu/context_menu.dart
+++ b/gui/lib/widgets/context_menu/context_menu.dart
@@ -150,15 +150,26 @@ class _ContextMenuState extends State<ContextMenu>
             alignment: AlignmentDirectional.topStart,
             child: FadeTransition(
               opacity: _animation,
-              child: SizeTransition(
-                sizeFactor: _animation,
-                axisAlignment: -1,
-                child: Padding(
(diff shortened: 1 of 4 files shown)


gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -150,15 +150,26 @@ class _ContextMenuState extends State<ContextMenu>
             alignment: AlignmentDirectional.topStart,
             child: FadeTransition(
               opacity: _animation,
-              child: SizeTransition(
-                sizeFactor: _animation,
-                axisAlignment: -1,
-                child: Padding(
-                  padding: EdgeInsets.only(bottom: theme.menuShadowMargin),
```
devzbysiu, line 157: This can be removed if it's not used anymore `menuShadowMargin`

devzbysiu (approved):

DoYouEvenCpp (approved):

PR #1577, opened 2026-06-17
LVPN-10714: handle resolvconf management service

bartoszWojciechO:
In some cases(especially in non-standard configurations) when resolvconf is managing DNS, app cannot detect this based on resolv.conf comment/link target and falls back to priority list detection. Other management service will be choosen if it's available. This results in DNS resolution outage.

resolvconf detection was added in this PR to prevent such cases.


daemon/dns/dns.go
DoYouEvenCpp: [nit] now with `var err error` above, no need to create a new variable anymore.
```suggestion
	if err = d.setUsingAvailable(iface, nameservers); err != nil {
```

DoYouEvenCpp, line 303: [nit] this comment now requires an update:
```suggestion
//     1b. If network manager is detected, it tries to set DNS using nmcli handler
//     1c. <if resolvconf is detected...>
//  2. If the above fails, infer which method to use based on /etc/resolv.conf link target
```

DoYouEvenCpp, line 307: [nit] similarly below:
```suggestion
//     2b. If network manager sym-link is detected, it tries to set DNS using nmcli handler
//     2c. If resolveconf...
//  3. If the above fails, it attempts to use best available method(see setUsingAvailable)
```

daemon/dns/dns.go
```diff
@@ -293,39 +309,36 @@ func (d *DNSServiceSetter) Set(iface string, nameservers []string) error {
...
-	switch d.currentManagementService {
-	case systemdResolvedManagementService:
-		log.Info(dnsPrefix, "setting DNS using systemd-resolved")
-		err := d.set(d.systemdResolvedSetter, iface, nameservers)
-		if err == nil {
-			return nil
+	var err error
+	if d.currentManagementService != unknownManagementService &&
+		d.currentManagementService != unmanagedManagementService {
+		//nolint:exhaustive
```
mariusSincovici, line 317: [small] is `//nolint:exhaustive` needed since there is a default?

mariusSincovici (approved):

DoYouEvenCpp (approved):

daemon/dns/dns.go
bartoszWojciechO: Unfortunately linter will complain even if there's a default case.

PR #1580, opened 2026-06-17
Sync with main

mariusSincovici:


daemon/rpc_connect.go
mariusSincovici, line 1: This file had conflicts

daemon/servers.go
mariusSincovici, line 1: Also here

daemon/dedicated_server_state.go
mariusSincovici, line 1: This contains the logic for DS from servers.go

.github/workflows/ci-gitlab.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       project-id: ${{ secrets.PROJECT_ID }}
     with:
       cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-      triggered-ref: v1.6.0
+      triggered-ref: v1.7.2
```
DoYouEvenCpp: 1.7.3 is the latest

mariusSincovici: Yes, you're right. I think there were new changes made after I've started the merging. If I'm not mistaken also some of the logger changes are not .
Would it be ok to keep the 1.7.2, and at some point sync again in a different PR, not to have conflicts on the file later?

mariusSincovici: Updated it to 1.7.3, because of some changes needed.

devzbysiu (approved):

daemon/serverpicker/dedicated_servers.go
```diff
@@ -0,0 +1,168 @@
...
+		// Invariant: dedicated servers are always of Nordlynx, hence Wireguard tech.
+		// This is needed to correctly display dedicated servers in the recent connections list.
+		Technologies: core.Technologies{core.Technology{ID: core.WireguardTech}},
+	}
+
+	result := ServerSelection{
+		DedicatedServerStatus: dedicatedServer.Status,
+		Server:                server,
+	}
+	return result, nil
```
thegravi, line 92: Comparison with the original one, there is a regression on returning status: https://github.com/NordSecurity/nordvpn-linux/blob/748baa6c5c198cff0929557d11e5fd61172c38d4/daemon/servers.go#L666

thegravi (approved):

daemon/serverpicker/dedicated_servers.go
mariusSincovici: It should not matter, because  according to the go code conventions, if a function returns an error the other function result must not be used.

PR #1585, opened 2026-06-19
LVPN-10737: fix /user/services request spam

bartoszWojciechO:
Because of faulty caching mechanism in the AccountInfo RPC handler, data is never cached.

Fresh data is fetched each time. When fresh data is fetched and user's VPN service is expired, a notification is emitted. This notification triggers AccountInfo request in tray, resulting in a permanent loop.

### DIFF
--- a/daemon/rpc_account.go
+++ b/daemon/rpc_account.go
@@ -57,7 +57,6 @@ func dipServicesToProtobuf(dipServices []auth.DedicatedIPService) []*pb.Dedidcat
 func (r *RPC) setDedicatedIPServerData(accountInfo *pb.AccountResponse) (*pb.AccountResponse, error) {
 	dipServices, err := r.ac.GetDedicatedIPServices()
 	if err != nil {
-		log.Error("getting dedicated ip services:", err)
 		if errors.Is(err, core.ErrUnauthorized) {
 			return &pb.AccountResponse{Type: internal.CodeExpiredAccessToken},
 				fmt.Errorf("getting dedicated ip services: %w", err)
--- a/daemon/rpc_account.go
+++ b/daemon/rpc_account.go
@@ -13,6 +13,7 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/events"
 	"github.com/NordSecurity/nordvpn-linux/internal"
 	"github.com/NordSecurity/nordvpn-linux/log"
+	"google.golang.org/protobuf/proto"
 )
 
 func findLatestDIPExpirationData(dipServices []auth.DedicatedIPService) (string, error) {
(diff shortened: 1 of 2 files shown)


daemon/rpc_account.go
```diff
@@ -113,27 +115,29 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
...
 	} else {
 		accountInfo.Type = internal.CodeSuccess
 	}
 
-	accountInfo = r.setDedicatedIPServerData(accountInfo)
-	if accountInfo.Type != internal.CodeSuccess {
+	accountInfo, err = r.setDedicatedIPServerData(accountInfo)
+	if err != nil {
+		log.Error("getting dedicated servers/ip service data:", err)
 		return accountInfo, nil
```
mariusSincovici, line 144: Same here?

daemon/rpc_account.go
```diff
@@ -113,27 +115,29 @@ func (r *RPC) AccountInfo(ctx context.Context, req *pb.AccountRequest) (*pb.Acco
 	if accountInfo, ok := r.dm.GetAccountData(req.Full); ok {
 		// because Dedicated IP and Dedicated Servers service data is using it's own caching that is updated with NC
 		// events, we can always try to fetch it to keep the returned information more accurate
-		dedicatedIPServerAccountInfo := r.setDedicatedIPServerData(accountInfo)
-		if dedicatedIPServerAccountInfo.Type == internal.CodeSuccess {
-			return dedicatedIPServerAccountInfo, nil
+		dedicatedIPServerAccountInfo, err := r.setDedicatedIPServerData(accountInfo)
+		if err != nil {
+			log.Error("getting dedicated servers/ip service data:", err)
+			return accountInfo, nil
```
mariusSincovici, line 124: Any reason why is error not returned in this case? Otherwise result might be used by the calling function?

bartoszWojciechO: In case of error `setDedicatedIPServerData` will still set the response `Type` to an appropriate code. This is compliant with how callers used this function before because it worked the same way, also works similarly a few lines bellow in case of config errors.

mariusSincovici: Ok, then the next return is incorrect "return dedicatedIPServerAccountInfo, nil".

daemon/rpc_account.go
```diff
@@ -54,15 +54,16 @@ func dipServicesToProtobuf(dipServices []auth.DedicatedIPService) []*pb.Dedidcat
 }
 
 // setDedicatedIPServerData sets Dedicated IP and Dedicated Server related fields in the accountInfo
-func (r *RPC) setDedicatedIPServerData(accountInfo *pb.AccountResponse) *pb.AccountResponse {
+func (r *RPC) setDedicatedIPServerData(accountInfo *pb.AccountResponse) (*pb.AccountResponse, error) {
 	dipServices, err := r.ac.GetDedicatedIPServices()
 	if err != nil {
 		log.Error("getting dedicated ip services:", err)
```
keliramu: Is there any reason not wrapping error with `fmt.Errorf(...)` and writing to log here?

keliramu (approved):

daemon/rpc_account.go
thegravi: duplicated

thegravi (approved):

daemon/rpc_account_test.go
```diff
@@ -261,3 +267,42 @@ func TestAccountInfo_ContainsServiceData(t *testing.T) {
...
 }
+
+func TestAccountInfo_CacheIsUpdatedIfDedicatedServersIPServiceIsNotAvailable(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	tests := []struct {
+		name                   string
+		dedicatedIPServices    []auth.DedicatedIPService
+		dedicatedServerService auth.DedicatedServerService
+	}{}
```
DoYouEvenCpp: I guess, we actually want to have some actual tests here? 😄

DoYouEvenCpp (changes requested):

daemon/rpc_account.go
bartoszWojciechO: Can you elaborate?

The idea is that we have two caches now:

- AccountInfo cache - updated in this RPC
- DedicatedServers/IP cache - updated by NC events

When handling this RPC, we take AccountInfo cache and try to update it with DedicatedServers/IP cache(because it's more up to date). If fetching the DS/DIP cache fails for some reason we just return whatever we found in AccountInfo cache(`return accountInfo, nil`), otherwise we return the updated data(`return dedicatedIPServerAccountInfo, nil`).

mariusSincovici: Right. sorry. Missed the closing `}`.

DoYouEvenCpp (approved):

PR #1587, opened 2026-06-22
LVPN-10582: Handle duplicated ServerMaintenance event

devzbysiu:
Context:
- `ServerMaintenance` event reconnects user to a different server
- if the event is duplicated, we should not react to it

Changes:
- ENS monitoring reconnects the server only if the event is targeting currently connected server, comparison is based on server public key passed when starting monitoring, then reconnection starts monitoring for this new server again
- tests
- code cleanup

### DIFF
--- a/daemon/ens/ens_monitoring.go
+++ b/daemon/ens/ens_monitoring.go
@@ -1,0 +1,90 @@
+package ens
+
+import (
+	"context"
+	"time"
+
+	"github.com/NordSecurity/nordvpn-linux/config/remote"
+	"github.com/NordSecurity/nordvpn-linux/events"
+	"github.com/NordSecurity/nordvpn-linux/log"
+	"github.com/NordSecurity/nordvpn-linux/networker"
+)
+
+const (
+	logPrefix = "[ens]"
+	evChSize  = 2
+)
+
+type ConnectCallback func() error
+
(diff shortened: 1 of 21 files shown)


daemon/ens/ens_monitoring.go
```diff
@@ -77,8 +77,15 @@ func (m *Monitor) run() {
 				log.Debug(logPrefix, "ignoring because VPN is not connected", e)
 				continue
 			}
-			// TODO: check if current server == event.server and check r.RequestedConnParams
-			if err := m.reconnectFn(); err != nil {
+
+			currServer, _ := m.netw.GetConnectionParameters()
+			eventIsForDifferentServer := currServer.NordLynxPublicKey != e.ServerPublicKey
```
DoYouEvenCpp, line 82: [nit] any benefit from having this check elevated to a separate variable?

DoYouEvenCpp (approved):

daemon/ens/ens_monitoring.go
devzbysiu: Purely to add semantic value to it

devzbysiu:
Closing this one for https://github.com/NordSecurity/nordvpn-linux/pull/1593 because of the conflicts and base branch changes that affect the solution

PR #1589, opened 2026-06-22
LVPN-10715: make sure no stale fileshare process is left

keliramu:
Fileshare process was left stale after turn mesh off (and more situations).\nThe main culprit was not closing all pipes on cleanup after redirected stdout and stderr to lumberjack.\nImplemented additional measures to make sure fileshare process is not left stale.

### DIFF
--- a/cmd/fileshare/redirect_to_lumberjack.go
+++ b/cmd/fileshare/redirect_to_lumberjack.go
@@ -44,8 +44,18 @@ func redirectStdOutputToLumberjack(lj *lumberjack.Logger) (cleanup func() error,
 	}()
 
 	cleanup = func() error {
-		_ = w.Close() // EOF for the goroutine
-		<-done        // wait for drain
+		// Dup2 created independent copies of the pipe write end on fd 1
+		// (stdout) and fd 2 (stderr). w.Close() only closes the original
+		// fd from os.Pipe. Redirect fd 1 and fd 2 to /dev/null so ALL
+		// pipe write ends are closed and io.Copy receives EOF.
+		devNull, err := unix.Open(os.DevNull, unix.O_WRONLY, 0)
+		if err == nil {
+			_ = unix.Dup2(devNull, 1)
+			_ = unix.Dup2(devNull, 2)
+			_ = unix.Close(devNull)
+		}
+		_ = w.Close()
+		<-done
 		return lj.Close()
 	}
 	if internal.IsProdEnv(Environment) {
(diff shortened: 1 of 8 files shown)


cmd/fileshare/redirect_to_lumberjack.go
```diff
@@ -44,8 +44,18 @@ func redirectStdOutputToLumberjack(lj *lumberjack.Logger) (cleanup func() error,
...
 	cleanup = func() error {
-		_ = w.Close() // EOF for the goroutine
-		<-done        // wait for drain
+		// Dup2 created independent copies of the pipe write end on fd 1
+		// (stdout) and fd 2 (stderr). w.Close() only closes the original
+		// fd from os.Pipe. Redirect fd 1 and fd 2 to /dev/null so ALL
+		// pipe write ends are closed and io.Copy receives EOF.
+		devNull, err := unix.Open(os.DevNull, unix.O_WRONLY, 0)
+		if err == nil {
+			_ = unix.Dup2(devNull, 1)
```
bartoszWojciechO, line 53: Maybe we can use `os.Stdout.Fd()` instead of a constant `1`? Same for the line bellow.

internal/filesystem.go
```diff
@@ -619,6 +621,88 @@ func isProcessRunning(executablePath string, readdir readdirFunc, readfile readf
...
+			return
+		}
+		log.Warnf("failed to send SIGTERM to %s (pid %d): %s", binaryName, pid, err)
+		return
+	}
+	log.Infof("sent SIGTERM to stale %s (pid %d)", binaryName, pid)
+
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if err := syscall.Kill(pid, 0); err != nil {
```
bartoszWojciechO, line 690: Is this call made to check if the process is dead? Shouldn't it also check if error is `syscall.ESRCH` and log something if it isn't?

bartoszWojciechO (approved):

thegravi (approved):

internal/filesystem.go
keliramu: it is just a probe - check if process is still alive, it returns err if process not found - this is ok

cmd/fileshare/redirect_to_lumberjack.go
keliramu: constants 1 and 2 in unix/linux are well known, if you dont mind, I will keep them here

PR #1591, opened 2026-06-24
LVPN-10730: Add bindings import change on major bump

devzbysiu:
Context:
- when our CI pipeline is used to build with libtelio version override and that version is a major version change, compilation will fail because bindings version in import paths is hardcoded
- this PR modifies those import paths on the fly when it detects that the version override is a major version change

Changes:
- `set_bindings_version.sh` additionally checks if there is a major version change and then `sed`s import paths to include this new major version
- this is based on a convention that libtelio `vX` has bindings available under `github.com/NordSecurity/libtelio-go/vX`

### DIFF
--- a/ci/set_bindings_version.sh
+++ b/ci/set_bindings_version.sh
@@ -30,7 +30,7 @@ new_module="${repo_path}/${major_version}"
 if [[ -n "${current_major}" && -n "${major_version}" && "${current_major}" != "${major_version}" ]]; then
   # NOTE: GONOSUMDB allows ignoring module checksum check made by go. This is
   # not needed for regular builds, because we commit changed go.sum which is
-  # trusted by go tooling, but when we are changing module on the fly, go
+  # trusted by go tooling. But when we are changing module on the fly, go
   # needs to recompute checksum database and we don't allow to do this in
   # docker, so this temporary override here allows module change on the fly
   # without compilation errors.
--- a/ci/set_bindings_version.sh
+++ b/ci/set_bindings_version.sh
@@ -2,8 +2,8 @@
 set -euxo pipefail
 
 declare -A LIB_NAME_TO_PACKAGE=(
-    [libtelio]=github.com/NordSecurity/libtelio-go
-    [libdrop]=github.com/NordSecurity/libdrop-go
+  [libtelio]=github.com/NordSecurity/libtelio-go
+  [libdrop]=github.com/NordSecurity/libdrop-go
 )
 
 declare -A LIB_NAME_TO_VERSION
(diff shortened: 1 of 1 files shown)


ci/set_bindings_version.sh
```diff
@@ -7,27 +7,37 @@ declare -A LIB_NAME_TO_PACKAGE=(
...
-if [[ -n "${bindings_version}" ]]; then
-  full_package_path=$repo_path/$major_version@$bindings_version
-fi
-
-go get "$full_package_path"
\ No newline at end of file
+[[ -n "${LIBTELIO_BINDINGS_VERSION:-}" ]] && LIB_NAME_TO_BINDINGS_VERSION[libtelio]=${LIBTELIO_BINDINGS_VERSION}
+[[ -n "${LIBDROP_BINDINGS_VERSION:-}" ]] && LIB_NAME_TO_BINDINGS_VERSION[libdrop]=${LIBDROP_BINDINGS_VERSION}
+
+for lib_name in "$@"; do
```
edvyno1: [minor] if we are handling all the variables here in one go then maybe we can reduce the two calls into one call with both libraries then in [here](https://github.com/NordSecurity/nordvpn-linux/blob/9aac3336eeca017268ddcbd6196da0d31d7fdbb4/ci/compile.sh#L62)

devzbysiu: Oof, yeah. That's actually a leftover from a different approach. I think I'll undo this loop and have it separate calls to be consistent because we already have `add_private_bindings.sh` and `remove_private_bindings.sh` that accept one thing at a time. And in current approach I don't need to pass all things at once anyway. Good catch :smile:

devzbysiu: Changed

DoYouEvenCpp (approved):

edvyno1 (approved):

ci/set_bindings_version.sh
```diff
@@ -2,32 +2,38 @@
...
+new_module="${repo_path}/${major_version}"
+
+if [[ -n "${current_major}" && -n "${major_version}" && "${current_major}" != "${major_version}" ]]; then
+  # NOTE: GONOSUMDB allows ignoring module checksum check made by go. This is
+  # not needed for regular builds, because we commit changed go.sum which is
+  # trusted by go tooling, but when we are changing module on the fly, go
+  # needs to recompute checksum database and we don't allow to do this in
+  # docker, so this temporary override here allows module change on the fly
+  # without compilation errors.
+  GONOSUMDB="${repo_path}/*" go get "${new_module}@${new_version}"
```
thegravi, line 37: Now `go get` is only called when `current_major != major_version.` This would not work if there is an update from `6.2.3` to `6.3.0`.

devzbysiu: It shouldn't, because bindings are still `v6` for those versions. This is only for major version change

PR #1592, opened 2026-06-24
LVPN-10181: set hard dependency between GUI and Daemon

DoYouEvenCpp:
This PR introduces a hard dependency between GUI and Daemon, so from now on GUI package strongly depends on Daemon, except for dev builds.
The change is implemented for the sake of DEB and RPM packages (Snap is already provided as a bundled one).

This PR also contains minor changes:
- removed `GetDaemonApiVersionRequest` and `GetDaemonApiVersionResponse` proto definitions
- for the sake of GUI detection of broken connection to Daemon switched from `DaemonApiVersionRequest` to `Ping`


protobuf/daemon/common.proto
keliramu: This enum is also not needed, imo

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -8,6 +8,7 @@ import 'package:nordvpn/grpc/error_handling_interceptor.dart';
 import 'package:nordvpn/logger.dart';
 import 'package:nordvpn/pb/daemon/common.pb.dart';
 import 'package:nordvpn/pb/daemon/service.pbgrpc.dart';
+import 'package:nordvpn/pb/daemon/ping.pb.dart';
```
keliramu: Unused import, `ping` is from `service.pbgrpc.dart` which is already present

gui/lib/data/providers/grpc_connection_controller.dart
```diff
@@ -112,43 +113,25 @@ class GrpcConnectionController extends _$GrpcConnectionController {
...
-      final response = await _client.getDaemonApiVersion(
-        GetDaemonApiVersionRequest(),
+      final _ = await _client.ping(
+        Empty(),
         options: CallOptions(timeout: _pingCallTimeout),
       );
-      _checkApiCompatibility(response.apiVersion);
+
+      // notify dependant providers only when there is actual change in the state
+      if (state is AsyncData<bool> && state.value == false) {
```
keliramu: imo, should be `if (!(state is AsyncData<bool> && state.value == true))` - set to true if not yet true, `state` type could be `AsyncError` after some comm failure, but later when ping succeeds, state would be not reset to `AsyncData(true)` with current condition.

keliramu (approved):

devzbysiu (approved):

PR #1593, opened 2026-06-24
LVPN-10582: Handle duplicated maintenance event

devzbysiu:
Context:\n\n- ServerMaintenance event reconnects user to a different server\n- if the event is duplicated, we should not react to it\n\nChanges:\n\n- ENS monitoring reconnects the server only if the event is targeting currently connected server, comparison is based on server public key\n- tests\n- code cleanup

### DIFF
--- a/daemon/vpn/nordlynx/libtelio/libtelio.go
+++ b/daemon/vpn/nordlynx/libtelio/libtelio.go
@@ -869,7 +869,8 @@ func (l *Libtelio) InjectVPNConnectionError(code int32, publicKey string) error 
 func isConnected(ctx context.Context,
 	stateCh <-chan state,
 	connParams connParameters,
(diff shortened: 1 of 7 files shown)


daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -952,41 +954,31 @@ func monitorConnectionState(
 }
 
 // startConnectionErrorMonitor starts connection error monitoring in a goroutine and returns only
```
DoYouEvenCpp: [nit] sorry, this `once` feel really off there 😄 
```suggestion
// startConnectionErrorMonitor starts connection error monitoring in a goroutine and returns
```

DoYouEvenCpp (approved):

thegravi (approved):

PR #1600, opened 2026-06-26
LVPN-10739: Introduce internet connection speed autotest

bbrutall:
- Added 'test_speed_thresholds' and 'test_speed_degradation' autotests

### DIFF
--- a/test/qa/test_speedtest.py
+++ b/test/qa/test_speedtest.py
@@ -1,0 +1,226 @@
+import json
+import os
+from datetime import datetime
+from statistics import median
+
+import pytest
+import sh
+
+import lib
+from lib import network
+
+pytestmark = pytest.mark.usefixtures("nordvpnd_scope_function")
+
+MIN_DOWNLOAD_MBPS = 70
+MIN_UPLOAD_MBPS = 50
+MAX_PING_MS = 50
+MIN_SPEED_RATIO = 0.25
+
+DOWNLOAD_BYTES = 5000000
+UPLOAD_BYTES = 5000000
+CLOUDFLARE_DOWN_URL = f"https://speed.cloudflare.com/__down?bytes={DOWNLOAD_BYTES}"
+CLOUDFLARE_UP_URL = "https://speed.cloudflare.com/__up"
+PING_HOST = "1.1.1.1"
+PING_COUNT = 5
+UPLOAD_FILE = "/tmp/speedtest_upload.bin"
+
+SPEED_TEST_RUNS = 3
+
+ARTIFACTS_DIR = os.path.join(os.environ.get("WORKDIR", "."), "dist", "test_artifacts")
+
+
+@pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


test/qa/test_speedtest.py
```diff
@@ -0,0 +1,226 @@
...
+pytestmark = pytest.mark.usefixtures("nordvpnd_scope_function")
+
+MIN_DOWNLOAD_MBPS = 10.0
+MIN_UPLOAD_MBPS = 5.0
+MAX_PING_MS = 100.0
+MIN_SPEED_RATIO = 0.5
+
+DOWNLOAD_BYTES = 5000000
+UPLOAD_BYTES = 5000000
+CLOUDFLARE_DOWN_URL = f"https://speed.cloudflare.com/__down?bytes={DOWNLOAD_BYTES}"
```
devzbysiu, line 21: I see that this is  undocumented API, did you consider using something official or maybe some python lib for testing speed?

test/qa/test_speedtest.py
```diff
@@ -0,0 +1,226 @@
...
+def _measure_download() -> float:
+    """Download test file from Cloudflare, return median speed in Mbps."""
+    results = []
+    for _ in range(SPEED_TEST_RUNS):
+        output = sh.curl(
+            "-o", "/dev/null",
+            "-s", "-w", "%{speed_download}",
+            CLOUDFLARE_DOWN_URL,
+            _timeout=60,
+            _ok_code=[0, 56],
```
devzbysiu, line 76: Why `56` exit code is considered ok? Isn't it failure when downloading data?

test/qa/test_speedtest.py
```diff
@@ -0,0 +1,226 @@
...
+    results = []
+    for _ in range(SPEED_TEST_RUNS):
+        output = sh.curl(
+            "-X", "POST",
+            "-o", "/dev/null",
+            "-s", "-w", "%{speed_upload}",
+            "--data-binary", f"@{UPLOAD_FILE}",
+            CLOUDFLARE_UP_URL,
+            _timeout=60,
+            _ok_code=[0, 56],
```
devzbysiu, line 102: Also is `56` code appropriate for upload check? Looks like this `56` means `Failure in receiving network data.`

test/qa/test_speedtest.py
bbrutall: Cloudflare occasionally resets the connection after the data transfer completes. Curl still reports the measured `speed_download` value in stdout before the reset happens. Without `_ok_code=[0, 56]`, sh raises an exception and we lose a valid measurement. We also run few iterations and take the median, so even if one run gets a partial result, it gets filtered out.

test/qa/test_speedtest.py
bbrutall: Same as above.

devzbysiu (approved):

dfetti (approved):

PR #1608, opened 2026-06-30
LVPN-10748: Add mDNS as part of LAN Discovery

devzbysiu:
Context:

- some printers use mDNS-based protocols to verify that found printer is still there
- this causes issues when users are connected to VPN as by default we were not allowing mDNS range
- this PR allows mDNS when LAN Discovery is enabled

Changes:

- mDNS address is added when user enables LAN Discovery
- tests
- cleanup


networker/lan_discovery.go
```diff
@@ -7,11 +7,12 @@ import (
...
-// addLANPermissions creates a new Allowlist. Subnets map is copied and updated with LANs, Port maps
-// remain unchanged.
-func addLANPermissions(allowlist config.Allowlist) config.Allowlist {
+// addLANDiscoverySubnets creates a new Allowlist. Subnets map is copied and
+// updated with LANs, Port maps remain unchanged. mDNS subnet is inserted for
+// service-discovery.
+func addLANDiscoverySubnets(allowlist config.Allowlist) config.Allowlist {
 	newSubnets := append([]string{}, allowlist.Subnets...)
-	for _, network := range internal.LocalNetworks {
+	for _, network := range append(internal.LocalNetworks, internal.MDNSSubnet) {
```
thegravi: Appending to a global list, even though a hidden reallocation happens here (cap == 4) and a new slice is allocated without affecting this var.

networker/lan_discovery.go
```diff
@@ -7,11 +7,12 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/internal"
 )
 
-// addLANPermissions creates a new Allowlist. Subnets map is copied and updated with LANs, Port maps
-// remain unchanged.
-func addLANPermissions(allowlist config.Allowlist) config.Allowlist {
+// addLANDiscoverySubnets creates a new Allowlist. Subnets map is copied and
```
thegravi: [small] subnets are not map anymore, but a string slice

DoYouEvenCpp (approved):

networker/lan_discovery.go
devzbysiu: Addressed

networker/lan_discovery.go
devzbysiu: Addressed

thegravi (approved):

PR #1610, opened 2026-07-01
LVPN-10175: Wrong login error with invalid token

edvyno1:
This PR maps a bad request response due to invalid authorization header from api to an unauthorized error, so we show correct user facing error.


devzbysiu (approved):

core/errors.go
```diff
@@ -128,6 +131,18 @@ func extractError(resp *http.Response, acceptedCode int) error {
...
 }
 
+func extractErrorForLogin(info apiError) error {
+	const (
+		invalidAuthorizationHeader = 100106
+	)
+
+	switch info.Errors.Code {
+	case invalidAuthorizationHeader:
+		return ErrUnauthorized
```
thegravi: We receive `400` response and yet here it states `Unauthorized` `401`, it's confusing. I'd add a new error called `ErrInvalidAuthHeader = errors.New("invalid authorization header")`, and then replace the return.


```suggestion
		return ErrInvalidAuthHeader
```

In `rpc_login.go:101`, I'd add something like this:
```go
if errors.Is(err, core.ErrMalformedAuthHeader) {
	return &pb.LoginResponse{
		Type: internal.CodeTokenInvalid,
	}, nil
}
```

core/errors.go
```diff
@@ -128,6 +133,18 @@ func extractError(resp *http.Response, acceptedCode int) error {
 	}
 }
 
+func extractErrorForLogin(info apiError) error {
```
keliramu, line 160: Would be nice to have unit tests with real/mocked responses code-error mapping validation?

devzbysiu (approved):

core/errors.go
edvyno1: we don't have any unit tests for mesh and dedicated servers either, do you think it's worth adding the tests for all of them?

core/errors.go
edvyno1: new error was created, logs are clearer now

core/errors.go
keliramu: yes, it's a debt, but small table-driven test would lock in the intended behavior cheaply and guard against future regressions

edvyno1: test added

thegravi (approved):

PR #1611, opened 2026-07-01
LVPN-10582: Improve target server identification

devzbysiu:
Context:

- public key is shared per region and can't be used to identify duplicated event
- server endpoint is used instead

Changes:

- use server endpoint instead of public key
- adjust and cleanup tests


daemon/ens/ens_monitoring.go
```diff
@@ -95,13 +95,13 @@ func (m *Monitor) run() {
 			}
 
 			currServer, _ := m.netw.GetConnectionParameters()
-			eventIsForDifferentServer := currServer.NordLynxPublicKey != e.ServerPublicKey
+			eventIsForDifferentServer := currServer.Endpoint != e.ServerEndpoint
```
keliramu: Endpoint is `ip:port` - comparing two as strings could be fragile, and consequences could be that we miss ens events. Maybe need parse `ip:port` with e.g. `netip.AddrPort` and then compare

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -131,10 +131,14 @@ func (t *telioCallbackHandler) handleEvent(e teliogo.Event) error {
 		}
 
 		if evt.Body.VpnConnectionError != nil {
+			if evt.Body.Endpoint == nil {
+				log.Error("dropping ENS connection error: libtelio event missing endpoint", *evt.Body.VpnConnectionError)
+				break
```
keliramu, line 139: Is this really intentional and justified to break out here based on absence of endpoint? some events might be missing endpoint, as i understood. Or we do not care about those events? code comment would be nice here about justification

keliramu: okok, `endpoint` now is server id - if it missing, nothing can be done.

devzbysiu: Added comment explaining it anyway

daemon/ens/ens_monitoring.go
devzbysiu: Addressed

keliramu (approved):

DoYouEvenCpp (approved):

daemon/vpn/vpn.go
```diff
@@ -48,3 +49,12 @@ type ServerData struct {
...
 	DedicatedServerPort int64
 }
+
+func (s ServerData) EndpointEqual(other string) bool {
+	ap, errA := netip.ParseAddrPort(s.Endpoint)
+	bp, errB := netip.ParseAddrPort(other)
+	if errA != nil || errB != nil {
+		return false
+	}
+	return ap == bp
```
thegravi, line 59: [nit] I'd add `Unmap()` here in case we ever use ipv6 for checks or other things.

thegravi (approved):

PR #1612, opened 2026-07-01
LVPN-10079: add a11y support and shadow effects for toast widgets

DoYouEvenCpp:
Changes under this PR:
- toasts widgets are now a11y capable
- when a toast message is displayed, focus is immediately moved onto it (the close button)
  - enter or space buttons (when focused the close button) discards toast widget
  - tab moves focus outside of the toast message and focuses previously focused widget
  - esc discards focus from the toast message and focuses previously focused widget
  - if TTS is enabled, announcement is immediately triggered upon toast message is shown
- UTs to cover changes
- extracted copies specifically for a11y purposes

This PR also resolves LVPN-10418:
- toast widget is now provided with shadow effects as per designs.

PR #1614, opened 2026-07-02
LVPN-10081: Accessibility in connection card

dfetti:
This adds keyboard navigation and TTS.

### DIFF
--- a/gui/lib/i18n/en/a11y.i18n.json
+++ b/gui/lib/i18n/en/a11y.i18n.json
@@ -1,4 +1,10 @@
 {                                                                                                                                                                                                                                                                                                                                                                                                                                 
   "VPNResumesIn": "VPN connection resumes in $minutes minutes $seconds seconds",
-  "VPNResumesInWithHours": "VPN connection resumes in $hours hours $minutes minutes $seconds seconds"
+  "VPNResumesInWithHours": "VPN connection resumes in $hours hours $minutes minutes $seconds seconds",
+  "VPNPanelDisconnected": "VPN Panel. Preferred location $location. Not secured.",
+  "VPNPanelConnecting": "VPN Panel. Connecting to $location.",
+  "VPNPanelConnected": "VPN Panel. Connected to $location.",
+  "VPNPanelLoading": "VPN Panel. Loading.",
+  "moreOptions": "More options",
+  "listWithItems": "List with $number items"
 }
--- a/gui/lib/theme/theme.dart
+++ b/gui/lib/theme/theme.dart
@@ -922,7 +922,8 @@ final class NordVpnTheme {
     required Color hoverColor,
   }) {
     return WidgetStateProperty.resolveWith<Color?>((states) {
-      if (states.contains(WidgetState.hovered)) {
+      if (states.contains(WidgetState.hovered) ||
+          states.contains(WidgetState.focused)) {
         return hoverColor;
       }
 
--- a/gui/lib/vpn/connection_card_buttons.dart
+++ b/gui/lib/vpn/connection_card_buttons.dart
@@ -5,11 +5,13 @@ import 'package:nordvpn/data/models/connect_arguments.dart';
 import 'package:nordvpn/data/models/pause.dart';
 import 'package:nordvpn/data/models/server_info.dart';
 import 'package:nordvpn/data/models/vpn_status.dart';
+import 'package:nordvpn/data/providers/recommended_server_provider.dart';
 import 'package:nordvpn/data/providers/vpn_settings_controller.dart';
 import 'package:nordvpn/data/providers/vpn_status_controller.dart';
 import 'package:nordvpn/data/repository/uievent_repository.dart';
 import 'package:nordvpn/i18n/strings.g.dart';
 import 'package:nordvpn/internal/scaler_responsive_box.dart';
+import 'package:nordvpn/pb/daemon/servers.pb.dart';
 import 'package:nordvpn/router/routes.dart';
 import 'package:nordvpn/internal/urls.dart';
 import 'package:nordvpn/theme/app_theme.dart';
@@ -17,7 +19,7 @@ import 'package:nordvpn/theme/connection_card_theme.dart';
 import 'package:nordvpn/widgets/context_menu/context_menu.dart';
 import 'package:nordvpn/widgets/dynamic_theme_image.dart';
 
-final class ConnectionCardButtons extends ConsumerWidget {
+final class ConnectionCardButtons extends ConsumerStatefulWidget {
   static const secureMyConnectionButtonKey = Key("vpnSecureMyConnectionButton");
   static const cancelButtonKey = Key("vpnCancelButton");
   static const pauseConnectionButtonKey = Key("pauseConnectionButton");
@@ -35,23 +37,45 @@ final class ConnectionCardButtons extends ConsumerWidget {
   final VpnStatus vpnStatus;
 
   const ConnectionCardButtons({super.key, required this.vpnStatus});
+  @override
+  ConsumerState<ConnectionCardButtons> createState() =>
+      _ConnectionCardButtonsState();
+}
+
+final class _ConnectionCardButtonsState
+    extends ConsumerState<ConnectionCardButtons> {
+  late final FocusNode _buttonFocusNode;
 
   @override
-  Widget build(BuildContext context, WidgetRef ref) {
+  void initState() {
+    super.initState();
+    _buttonFocusNode = FocusNode();
+  }
+
+  @override
+  void dispose() {
+    _buttonFocusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
     final appTheme = context.appTheme;
     final buttonTheme = context.connectionCardTheme.buttonTheme;
-
+    final recommendedServerLocation = ref
+        .watch(recommendedServerProvider)
+        .when(
+          data: (fastest) => fastest,
+          error: (_, _) => null,
+          loading: () => null,
+        );
     return ScalerResponsiveBox(
       maxWidth: buttonTheme.maxConnectButtonWidth,
       child: IntrinsicHeight(
-        child: Row(
-          spacing: appTheme.horizontalSpaceSmall,
-          children: _buildButtons(
-            context,
-            ref,
-            appTheme,
-            buttonTheme,
-            vpnStatus,
+        child: FocusTraversalGroup(
+          child: Row(
+            spacing: appTheme.horizontalSpaceSmall,
+            children: _buildButtons(context, recommendedServerLocation),
           ),
         ),
       ),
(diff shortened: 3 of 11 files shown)


gui/lib/vpn/connection_card_label.dart
```diff
@@ -29,29 +33,30 @@ final class ConnectionCardLabel extends StatelessWidget {
...
-      style: labelTheme.font.copyWith(color: _labelColor(labelTheme)),
+    final connectionStatus = ConnectionCardLabel.connectionStatusText(
+      vpnStatus,
+    );
+    return Focus(
+      focusNode: statusFocusNode,
+      skipTraversal: true,
+      child: Semantics(
+        label: connectionStatus,
+        textField: true,
```
devzbysiu: I think this is for input fields, not for static texts 🤔

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -211,19 +235,26 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
+              label: t.ui.changeVPNsettings,
+              onTap: () => _changeSettings(context, ref),
+            ),
+            ContextMenuItem(label: t.ui.getHelp, onTap: () => _getHelp(ref)),
+          ],
+          anchorBuilder: (toggleMenu) => OutlinedButton(
+            style: buttonTheme.connectionDetailsButtonStyle,
+            onPressed: toggleMenu,
+            child: Semantics(
+              label: "More",
```
devzbysiu: I think this needs translations

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -76,37 +78,49 @@ final class ConnectionCardButtons extends ConsumerWidget {
               onPressed: () async => await ref
                   .read(vpnStatusControllerProvider.notifier)
                   .disconnect(),
-              child: Text(t.ui.disconnect),
+              child: Semantics(
```
devzbysiu, line 103: Isn't this redundant? It excludes semantics from the text, so its label is ignored, but then sets label to the same value 🤔

devzbysiu: Same with other buttons, `OutlinedButton` already has a role `button` from flutter and it will interpret the text inside properly, so I think this is redundant in this form

gui/lib/widgets/context_menu/context_menu.dart
```diff
@@ -165,10 +190,32 @@ class _ContextMenuState extends State<ContextMenu>
...
+                          actions: <Type, Action<Intent>>{
+                            DismissIntent: CallbackAction<DismissIntent>(
+                              onInvoke: (intent) {
+                                _close();
+                                return null;
+                              },
+                            ),
+                          },
+                          child: FocusTraversalGroup(
+                            policy: OrderedTraversalPolicy(),
```
devzbysiu: I think this is the default policy set 🤔 , not sure if this does anything here

devzbysiu: Yeah, `FocusTraversalGroup` has this:

```dart
FocusTraversalGroup({
    super.key,
    FocusTraversalPolicy? policy,
    this.descendantsAreFocusable = true,
    this.descendantsAreTraversable = true,
    this.onFocusNodeCreated,
    required this.child,
  }) : policy = policy ?? ReadingOrderTraversalPolicy();
```
And docs on `OrderedTraversalPolicy` say:
```dart
/// Constructs a traversal policy that orders widgets for keyboard traversal
/// based on an explicit order.
///
/// If [secondary] is null, it will default to [ReadingOrderTraversalPolicy].
OrderedTraversalPolicy({this.secondary, super.requestFocusCallback});
```

gui/lib/i18n/en/ui.i18n.json
```diff
@@ -281,5 +281,9 @@
   "reconnect": "Reconnect",
   "changeVPNsettings": "Change VPN settings",
   "getHelp": "Get help",
-  "pauseConnection": "Pause connection"
+  "pauseConnection": "Pause connection",
+  "vpnPanel": "VPN Panel",
```
devzbysiu: Don't know if this was discussed, but there is also this PR which introduces `a11y` namespace for semantics-related copy https://github.com/NordSecurity/nordvpn-linux/pull/1612/changes#diff-0f257800b01af69f7946aea2e754f6c3b34c6a59df1055b2bb4ac91d7033e5a1R1-R4

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -32,26 +42,58 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (mounted && !_primaryButtonFocus.hasFocus) {
+          _primaryButtonFocus.requestFocus();
+        }
+      });
+    }
+  }
+
+  int _stateOf(VpnStatus s) => s.isConnected()
+      ? 2
```
devzbysiu: You can get the values of the status with `s.status.value`, but I think this is not needed. You can compare `status` enums directly with `old.vpnStatus.status != widget.vpnStatus.status`

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -207,25 +276,67 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
-        anchorBuilder: (toggleMenu) => OutlinedButton(
-          style: buttonTheme.connectionDetailsButtonStyle,
-          onPressed: toggleMenu,
-          child: DynamicThemeImage("connection_details.svg"),
         ),
       ),
     );
   }
+
+  String _buildSemanticsText(VpnStatus vpnStatus) {
```
DoYouEvenCpp: Don't we miss here `vpnStatus.isPaused()` branch, currently paused will get announced as `loading`, no?

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -79,41 +121,52 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
+                  onTap: () async => await ref
+                      .read(vpnStatusControllerProvider.notifier)
+                      .disconnect(),
+                ),
+              ],
+              anchorBuilder: (toggleMenu) => OutlinedButton(
+                style: buttonTheme.pauseConnectionButtonStyle,
+                focusNode: _primaryButtonFocus,
+                onPressed: toggleMenu,
+                child: Semantics(
```
devzbysiu, line 141: Is `Semantics` needed here? `OutlinedButton` already has a `button` role and the text inside is the same as `label` set on `Semantics`

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -32,26 +42,58 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
-          children: _buildButtons(
-            context,
-            ref,
-            appTheme,
-            buttonTheme,
-            vpnStatus,
+        child: FocusTraversalGroup(
+          child: Row(
+            spacing: appTheme.horizontalSpaceSmall,
+            children: _buildButtons(
```
devzbysiu: [small] you can reduce number of passed args here to just 2. Theme arguments can be taken from `BuildContext` and `vpnStatus` is accessible through instance field `widget`

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -207,25 +276,67 @@ final class ConnectionCardButtons extends ConsumerWidget {
   Widget _buildConnectionDetailsButton(
```
devzbysiu, line 261: Similarly for this one - `ConnectionCardButtonTheme` can be taken from `BuildContext` and `VpnStatus` is available through `widget`

gui/lib/theme/theme.dart
```diff
@@ -926,6 +926,10 @@ final class NordVpnTheme {
         return hoverColor;
       }
 
+      if (states.contains(WidgetState.focused)) {
+        return hoverColor;
+      }
+
```
DoYouEvenCpp:
```suggestion
 if (states.contains(WidgetState.hovered) || states.contains(WidgetState.focused)) {
        return hoverColor;
 }
```

gui/lib/i18n/en/a11y.i18n.json
```diff
@@ -1,4 +1,9 @@
 {                                                                                                                                                                                                                                                                                                                                                                                                                                 
   "VPNResumesIn": "VPN connection resumes in $minutes minutes $seconds seconds",
-  "VPNResumesInWithHours": "VPN connection resumes in $hours hours $minutes minutes $seconds seconds"
-}
\ No newline at end of file
+  "VPNResumesInWithHours": "VPN connection resumes in $hours hours $minutes minutes $seconds seconds",
+  "VPNPanel": "VPN Panel",
+  "more": "More options",
```
DoYouEvenCpp: [nit] maybe rename it to `moreOptions` ?

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -60,85 +89,116 @@ final class ConnectionCardButtons extends ConsumerWidget {
 
   List<Widget> _buildButtons(
     BuildContext context,
-    WidgetRef ref,
     AppTheme appTheme,
```
DoYouEvenCpp: `appTheme` is not used at all (it's reused from context below)

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -205,27 +274,93 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
-          child: DynamicThemeImage("connection_details.svg"),
         ),
       ),
     );
   }
+
+  String _buildSemanticsText({
+    RecommendedServerLocation? recommendedServerLocation,
+  }) {
+    // VPN Panel. Preferred location: Fastest Server/[City] [Country]. Not secured. Secure my connection push button.
```
DoYouEvenCpp: So, a colon is missing for preferred locations - either add it below, or fix the comment please

gui/lib/vpn/connection_card_buttons.dart
```diff
@@ -205,27 +274,93 @@ final class ConnectionCardButtons extends ConsumerWidget {
...
+    // VPN Panel. Connecting to Fastest Server/[City], [Country]. Cancel push button.
+    // VPN Panel. Connected to [City], [Country]. Pause menu push button.
+
+    var vpnPanel = "${t.a11y.VPNPanel}. ";
+    if (widget.vpnStatus.isDisconnected() || widget.vpnStatus.isPaused()) {
+      return "$vpnPanel ${t.a11y.preferredLocation} ${_buildRecommendedCityAndCountryText(recommendedServerLocation)}. ${t.ui.notSecured}";
+    }
+
+    if (widget.vpnStatus.isConnecting()) {
+      return "$vpnPanel ${t.ui.connecting} to ${_buildCityAndCountryText()}.";
```
DoYouEvenCpp: [nit] those `to` are not translated here at all - maybe we should provide here a high-level translation string for those cases?

DoYouEvenCpp (approved):

keliramu (approved):

devzbysiu (approved):

gui/lib/vpn/connection_card_buttons.dart
dfetti: I made full sentence transaltion with a parameter for location.

PR #1615, opened 2026-07-02
LVPN-10372: fix problem protocol change during pause reconnect server not found

keliramu:
Before fix it was failing to reconnect, when connected to DoubleVPN, pause, then change technology to NordWhisper, confirm reconnect.
Another case without pause: connected to specialty server (DoubleVPN), change technology to NordWhisper, before fix app gui was showing generic error dialog and rendering disconnected state, while daemon in fact was connected (previous connection is not canceled). After fix: try to reconnect but fail to select specialty server for new technology (NordWhisper), then fallback to quickconnect with new selected technology.


daemon/rpc_connect.go
```diff
@@ -77,14 +77,26 @@ func (r *RPC) executeConnect(srv pb.Daemon_ConnectServer, fn func(context.Contex
...
+		r.reconcileStatusAfterFailedConnect()
 	}
 
 	return err
 }
 
+// reconcileStatusAfterFailedConnect fixes up the reported connection status after a connection
+// attempt failed. If the attempt aborted before the existing tunnel was touched, the VPN is still
+// up (netw.IsVPNActive()==true) and we restore the previous status instead of falsely reporting a
+// disconnect. Only when there is genuinely no active tunnel do we publish a Disconnect.
```
DoYouEvenCpp:
```suggestion
// disconnect. Only when there is genuinely no active tunnel we do publish a Disconnect.
```

keliramu: Thx, fixed

gui/lib/data/providers/vpn_status_controller.dart
```diff
@@ -63,6 +63,21 @@ class VpnStatusController extends _$VpnStatusController
...
+  /// Reconnect using [args]; if no suitable server exists for the currently
+  /// selected technology (e.g. Double VPN has no NordWhisper servers), fall
+  /// back to a quick connect so the newly chosen technology is still applied.
+  Future<void> reconnectOrQuickConnect(ConnectionParameters args) async {
+    final status = await _run((vpn) => vpn.reconnect(args));
+    if (status == DaemonStatusCode.serverUnavailable) {
+      logger.i(
+        "no suitable server for reconnect params, falling back to quick connect",
+      );
+      await connect(null); // quick connect for the current (new) technology
```
mariusSincovici, line 75: If connect returns error, then the error is ignored.

gui/lib/widgets/popups_listener.dart
```diff
@@ -64,12 +64,25 @@ final class _PopupsListenerState extends ConsumerState<PopupsListener> {
...
       context: ctx,
       barrierDismissible: barrierDismissible,
       builder: (_) => buildPopup(metadata),
     );
 
+    // The decision action runs only after the dialog is closed, using this
+    // (long-lived) widget's ref. The dialog has its own ref scoped to the
+    // dialog route, which is disposed on close, so running async actions with
+    // it would throw "Cannot use ref after the widget was disposed".
+    if (mounted && metadata is DecisionPopupMetadata) {
```
mariusSincovici: Why is this needed only for this specific yes/no dialogs? Can't it happen also for other types the same "Cannot use ref after the widget was disposed"?

gui/lib/widgets/popups/decision_popup.dart
```diff
@@ -23,30 +23,29 @@ final class DecisionPopup extends Popup {
...
       ],
     );
   }
 
-  Widget _noButton(WidgetRef ref, BuildContext context) {
+  // The yes/no actions are intentionally not invoked here. Closing the dialog
+  // disposes this widget's `ref`, so an async action that uses it after an
+  // `await` would crash with "Cannot use ref after the widget was disposed".
+  // Instead we pop with the user's choice and let PopupsListener run the
+  // action with its long-lived ref once the dialog is gone.
```
devzbysiu: You can take a look on `RichNotificationPopup` how this was handled, we simply execute action first and close the popup when action is executed. 

If the action takes long time, we should add some kind of loading indicator in dialogs, otherwise users will see nothing happening for the duration of executed action and then suddenly dialog will close.

The option you implemented is also fine as long as we do it for all popups - without having to add exceptional `if` in popups listener. It should be the same mechanism for all popups. But it has some drawbacks: with your approach, we are loosing an option to display error in popup and allowing user to click the action again and we are loosing an option to show progress of the action if it takes too long. From user perspective, dialog will be closed immediately with your approach, but action is happening somewhere and user doesn't see it (it's the same now, but now we can add this loading indicator to popups if needed and with changed approach of running action in popup listener we can't do that)

gui/lib/data/providers/vpn_status_controller.dart
keliramu: no, down in the call sequence there is `ref.read(popupsProvider.notifier).show(status);`

gui/lib/widgets/popups_listener.dart
keliramu: Now logic is common for all dialog types, no new/extra progress indicator, just refactored to solve problem with not valid wedget ref being used, and possibility for resulting error (if any) to be shown regardless where user navigates to during longer action execution

gui/lib/widgets/popups/decision_popup.dart
keliramu: As mentioned in above comment, no new/extra progress indicator added, just relying on existing progress indicators.

keliramu:
Part of problem is solved by other PR, other part of problem is not reproduces anymore.

PR #1616, opened 2026-07-03
LVPN-10783: Improve creds management

devzbysiu:
Changes:

- ask for token instead of using it form FS
- `.env.sample` cleanup

### DIFF
--- a/magefiles/scripts.go
+++ b/magefiles/scripts.go
@@ -4,11 +4,19 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"go/build"
+	"maps"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/NordSecurity/nordvpn-linux/log"
+)
+
+const (
+	ciJobTokenEnvVar    = "CI_JOB_TOKEN"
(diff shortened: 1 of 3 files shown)


bartoszWojciechO (approved):

magefiles/scripts.go
```diff
@@ -4,13 +4,21 @@ package main
...
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/NordSecurity/nordvpn-linux/log"
 )
 
+const (
+	ciJobTokenEnvVar    = "CI_JOB_TOKEN"
+	ciJobTokenPassEntry = "nordvpn-linux/ci_job_token"
```
DoYouEvenCpp, line 19: Wouldn't all capitalized CI_JOB_TOKEN (even under `nordvpn-linux` key) be a bit more verbose - I guess we might be kinda used to this form of capitalisation. 😄 
(nothing critical, just an idea to process?)

DoYouEvenCpp (approved):
LGTM, just a single petty nit.

magefiles/scripts.go
devzbysiu: I used a common naming convention for naming keys in secret storage

PR #1620, opened 2026-07-09
LVPN-10390: Move ucred extracton to one place

devzbysiu:
Context:

- ucred extraction from context is repeated in few places
- we have a helper, but it was not used in all places it should
- this PR moves extraction to single place

Changes:

- remove `rpc_helper.go` and `rpc_helper_test.go` -  `_test.go` had only helper func which should be in `test/helpers.go`, helper moved to `internal/socket.go` where socket related operations are already grouped
- replace use of `getCallerCred` with new helper
- replace use of helper test with new helper from `test/helpers/`
- replace of inline extraction of ucreds with new helper
- improved tests
- code cleanup


bartoszWojciechO (approved):

keliramu (approved):

devzbysiu:
Move just the one commit from this one to the base branch to avoid conflicts, so the changes are already here https://github.com/NordSecurity/nordvpn-linux/pull/1569. Closing this one

PR #1622, opened 2026-07-09
Clear GUI logs on package purge.

Wikpi:
## Issue

I noticed that on Debian systems, purging the `nordvpn-gui` package leaves behind GUI log dumps in the users' home directories, which should not be the case.

## Solution

Updated Debian  `postrm_template` script to distinguish and separate `remove` and `purge` actions. 

During a `purge` action, the script now ensures that all GUI-side log dumps are cleared from users' home directories  without affecting the main `nordvpn` package data. The existing `remove` action behavior remains unchanged.

The same behavior is noticeable for RPM packages, but the `postun_template` script was left unchanged, since RPM does not provide any equivalent purge action (dumps are left in users' directories).

Additionally, updated the `build_package.sh` script to call `envsubst` with an explicit variable list (or an empty list where appropriate). This ensures that only intended variables are substituted, while preserving  shell variables used in the generated template scripts.

### DIFF
--- a/gui/templates/scriptlets/deb/postrm_template
+++ b/gui/templates/scriptlets/deb/postrm_template
@@ -1,9 +1,24 @@
 #!/usr/bin/env bash
 
 case "$1" in
-    purge|remove)
-    ${UNINSTALL_SCRIPT}
-    
+    purge)
+        usrInfo=$(
+        while IFS=: read -r user pass uid gid gecos home shell; do
+            if [ "$uid" -ge 1000 ] && [ "$uid" -le 2000 ]; then
+                echo "$home"
+            fi
+        done < /etc/passwd
+        )
+        (
+            IFS=$'\n'
+            for usrLine in $usrInfo; do
+                # for each user remove gui logs
+                rm -f $usrLine/.cache/nordvpn/nordvpn-gui.log
+            done
+        )
+    ;&  # fallthrough
+    remove)
+        ${UNINSTALL_SCRIPT}
     ;;
     disappear|upgrade|failed-upgrade|abort-install|abort-upgrade)
     ;;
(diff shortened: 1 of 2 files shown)


nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/nordvpn-linux?pullRequest=1622) <br/>All committers have signed the CLA.

keliramu (approved):

gui/templates/scriptlets/deb/postrm_template
```diff
@@ -1,9 +1,24 @@
 #!/usr/bin/env bash
 
 case "$1" in
-    purge|remove)
-    ${UNINSTALL_SCRIPT}
-    
+    purge)
+        usrInfo=$(
+        while IFS=: read -r user pass uid gid gecos home shell; do
+            if [ "$uid" -ge 1000 ] && [ "$uid" -le 2000 ]; then
```
keliramu, line 7: `envsubst` is being used in final script generation from the template, it removes variables and final script is broken

keliramu: Check `gui/scripts/build_package.sh` where it's being processed

Wikpi: Good catch! 

`envsubst` by default was substituting all referenced environment variables. This meant that the shell variables in the template scripts were being replaced with empty strings, which resulted in a broken final execution.
 
Fixed it by passing an explicit variable list to the `envsubst` calls in the `build_package.sh` script.

keliramu (approved):

devzbysiu (approved):

DoYouEvenCpp (approved):

PR #1625, opened 2026-07-10
LVPN-10766: Add button to invoke GUI

thegravi:
Adding "open-gui" button in the system tray:
- When system lacks gui binary, this button redirects to the download website
- When system has gui binary, this button opens the gui app itself


tray/actions.go
```diff
@@ -110,6 +111,50 @@ func tryDbus(uri string) error {
...
+	}
+	return internal.IsCommandAvailable(guiBinaryName)
+}
+
+func (ti *Instance) openGui() {
+	var err error
+	if snapconf.IsUnderSnap() {
+		err = tryDbus(guiLaunchURI)
+	} else {
+		err = launchGuiBinary()
```
devzbysiu: Maybe we could use `tryDbus` (probably better name for it would make sense now that it's used in more places) in both cases - for snap and for deb/rpm

tray/actions.go
```diff
@@ -110,6 +111,50 @@ func tryDbus(uri string) error {
...
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	go func() { _ = cmd.Wait() }() // reap the child process
+	return nil
+}
+
+func (ti *Instance) openDownloadPage() {
+	// #nosec G204 -- fixed URL, no user input
+	if err := exec.Command("xdg-open", guiDownloadURL).Run(); err != nil {
```
devzbysiu: `xdg-open` is more error prone than launching via D-Bus, I think we should do `tryDbus` first and then fallback to `xdg-open` if D-Bus fails. 

There is `openURI` func, we can extend it to do `xdg-open` fallback and use it

tray/menu.go
```diff
@@ -605,6 +603,24 @@ func buildSettingsSubitems(ti *Instance, menu *systray.MenuItem) {
 	go handleTrayOption(ti, trayCheckbox)
 }
 
+func buildGuiSection(ti *Instance) {
```
devzbysiu: [small] should be `buildGUISection` to match go's convention. Also in other places in this PR

tray/gui_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/NordSecurity/nordvpn-linux/snapconf"
+	"github.com/NordSecurity/nordvpn-linux/test/category"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsGuiAvailableUnderSnap(t *testing.T) {
+	category.Set(t, category.Unit)
+
+	t.Setenv(snapconf.EnvSnapName, "nordvpn")
```
devzbysiu: [small] I suggest to add `assert.False(t, isGuiAvailable())` before this line to confirm that this line changes the situation

tray/menu.go
```diff
@@ -605,6 +603,24 @@ func buildSettingsSubitems(ti *Instance, menu *systray.MenuItem) {
...
+	if ti == nil {
+		return
+	}
+	if !ti.state.daemonAvailable {
```
devzbysiu: Maybe would be better to display the button even when daemon is down?

tray/actions.go
```diff
@@ -110,6 +111,50 @@ func tryDbus(uri string) error {
...
+	if snapconf.IsUnderSnap() {
+		return true
+	}
+	return internal.IsCommandAvailable(guiBinaryName)
+}
+
+func (ti *Instance) openGui() {
+	var err error
+	if snapconf.IsUnderSnap() {
+		err = tryDbus(guiLaunchURI)
```
devzbysiu: Did you consider using snap's `io.snapcraft.PrivilegedDesktopLauncher.OpenDesktopEntry` D-Bus service?

tray/actions.go
```diff
@@ -82,32 +89,155 @@ func (ti *Instance) login() {
...
+
+// watchGUIInstallation redraws (async) the tray when the native GUI binary appears in the system
+// or disappears.
+func (ti *Instance) watchGUIInstallation(ctx context.Context) {
+	if snapconf.IsUnderSnap() {
+		// GUI is always bundled with the package
+		return
+	}
+
+	const guiInstallDir = "/usr/bin"
```
devzbysiu: `/usr/bin/nordvpn-gui` is a symlink to `/opt/nordvpn-gui/nordvpn-gui`. Maybe we can watch this one as it will produce less events

devzbysiu (approved):

tray/actions.go
```diff
@@ -82,32 +89,155 @@ func (ti *Instance) login() {
...
+// watchGUIInstallation redraws (async) the tray when the native GUI binary appears in the system
+// or disappears.
+func (ti *Instance) watchGUIInstallation(ctx context.Context) {
+	if snapconf.IsUnderSnap() {
+		// GUI is always bundled with the package
+		return
+	}
+
+	const guiInstallDir = "/opt/nordvpn-gui"
+	watcher, err := filewatch.GetFileWatcher(guiInstallDir)
```
DoYouEvenCpp: Consider scenario, when there's only daemon installation - how we gonna monitor path that doesn't exist?
e.g. https://github.com/fsnotify/fsnotify/blob/main/fsnotify.go#L253
> Paths that do not yet exist on the filesystem cannot be
> // watched.

DoYouEvenCpp: Maybe we can switch here and monitor `/opt` and the moment ~subdirectory of for~ when `nordvpn-gui` appears under the `opt` we add another watcher on gui binary level (`opt/nordvpn-gui`)

tray/actions.go
```diff
@@ -82,32 +89,155 @@ func (ti *Instance) login() {
...
+}
+
+const (
+	guiBinaryName  = "nordvpn-gui"
+	guiLaunchURI   = "nordvpn-gui://open"
+	guiDownloadURL = "https://nordvpn.com/download/linux/"
+)
+
+// isGUIAvailable check whether the system already has the application
+func isGUIAvailable() bool {
```
DoYouEvenCpp, line 176: [nit] the moment we (technically speaking) detect GUI gets installed might not be the same moment when `PATH` might get populated with binary full path - shall we depend here on the fully resolved path instead?

DoYouEvenCpp (changes requested):

tray/actions.go
thegravi: True, after changing last path I did not test the whole thing. Most robust and also easiest mechanism to monitor would be to revert back to monitoring `/usr/bin` dir.

DoYouEvenCpp (approved):

PR #1627, opened 2026-07-15
LVPN-10857: Fix meshnet tests

devzbysiu:
Context:

- on nightly, `test_connect_meshnet` is failing with nickname conflict
- on daily `test_incoming_connections` fails peer not reachable error

Changes:

-  for nickname conflict check, exclude non-routable addresses as those do not conflict with meshnet, so in case where IP lookup happens to return the host with the same name as peer nickname, it will not conflict if IP is non-routable
- for peer not reachable added check waiting for peers to report that they are connected and only then proceed with tests

### DIFF
--- a/meshnet/server.go
+++ b/meshnet/server.go
@@ -2028,7 +2028,7 @@ func MakePeerMaps(peers *pb.PeerList) (map[string]*pb.Peer, map[string]*pb.Peer)
 	return peerPubkeyToPeer, peerNameToPeer
 }
 
-// hasRoutableAddress returns true when at least one IP is globally routable.
+// hasRoutableAddress returns true when at least one IP is globally routable, false otherwise.
 // Link-local (fe80::/10, 169.254.0.0/16) and loopback addresses are excluded:
 // they may be returned by host-local resolvers (e.g. myhostname NSS module or
 // Docker DNS forwarding to the host) and do not conflict with meshnet routing.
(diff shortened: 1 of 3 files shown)


meshnet/server.go
```diff
@@ -2030,6 +2027,23 @@ func MakePeerMaps(peers *pb.PeerList) (map[string]*pb.Peer, map[string]*pb.Peer)
 	}
 	return peerPubkeyToPeer, peerNameToPeer
 }
+
+// hasRoutableAddress returns true when at least one IP is globally routable.
```
DoYouEvenCpp:
```suggestion
// hasRoutableAddress returns true when at least one IP is globally routable, false otherwise.
```

meshnet/server.go
```diff
@@ -2030,6 +2027,23 @@ func MakePeerMaps(peers *pb.PeerList) (map[string]*pb.Peer, map[string]*pb.Peer)
...
+func hasRoutableAddress(ips []net.IP, err error) bool {
+	if err != nil {
+		return false
+	}
+	for _, ip := range ips {
+		if !ip.IsLinkLocalUnicast() && !ip.IsLoopback() {
+			return true
+		}
+	}
+	return false
```
DoYouEvenCpp: [nit] we can shorten the same logic here with slice's `ContainsFunc`
```suggestion
	return slices.ContainsFunc(ips, func(ip net.IP) bool {
		return !ip.IsLinkLocalUnicast() && !ip.IsLoopback()
	})
```

DoYouEvenCpp (approved):
LGMT, only some nits.

keliramu (approved):

PR #1629, opened 2026-07-16
LVPN-10858: make cli case insensitive

keliramu:
Make nordvpn invocations behave identically regardless of the letter case used for command and subcommand names, so that all of the following work the same:
- nordvpn set mesh on
- nordvpn SET MESH on
- nordvpn Set Mesh On

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -114,7 +114,7 @@ func main() {
 	}
 
 	// Canonicalize command/subcommand name case so commands can be invoked
-	// case-insensitively (e.g. "SET MESH on" == "set meshnet on").
+	// case-insensitively (e.g. "SET MESH on" == "set mesh on").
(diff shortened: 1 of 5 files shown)


cmd/cli/main.go
```diff
@@ -113,6 +113,10 @@ func main() {
 		args = append(args, clearFormatting(arg))
 	}
 
+	// Canonicalize command/subcommand name case so commands can be invoked
+	// case-insensitively (e.g. "SET MESH on" == "set meshnet on").
```
devzbysiu: typo, should be `"SET MESH on" == "set mesh on"`

devzbysiu (approved):

cmd/cli/main.go
keliramu: Addressed

DoYouEvenCpp (approved):

PR #1637, opened 2026-07-22
LVPN-10608: Refactor fileshare process monitoring test

ggediminass:
Instead of checking rules, to understand if fileshare port is open or closed, try to access port.

### DIFF
--- a/test/qa/lib/fileshare.py
+++ b/test/qa/lib/fileshare.py
@@ -474,9 +474,9 @@ def is_process_running() -> bool:
     return result.returncode == 0
 
 
-def restart_mesh() -> None:
-    sh.nordvpn.set.meshnet.off()
+def restart_mesh(ssh_client: ssh.Ssh) -> None:
+    exec_command = CommandExecutor(ssh_client) if ssh_client else CommandExecutor()
+    exec_command("nordvpn set meshnet off")
     time.sleep(2)
-    sh.nordvpn.set.meshnet.on()
+    exec_command("nordvpn set meshnet on")
     time.sleep(5)
-
--- a/test/qa/lib/fileshare.py
+++ b/test/qa/lib/fileshare.py
@@ -10,9 +10,8 @@ from threading import Thread
 import pytest
 import sh
 import socket
-import os
 
-from . import FILE_HASH_UTILITY, CommandExecutor, logging, ssh
+from . import firewall, FILE_HASH_UTILITY, CommandExecutor, logging, ssh
 
 class FileSize:
     """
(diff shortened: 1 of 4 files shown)


test/qa/lib/fileshare.py
```diff
@@ -437,25 +436,43 @@ def bind_port() -> socket.socket | None:
...
+    Returns True if connection succeeds, False otherwise.
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.settimeout(timeout)
+        result = sock.connect_ex((peer_address, port))
+        sock.close()
+
+        # connect_ex returns 0 on success, error code otherwise
+        return result == 0
```
thegravi: [small] Would you prefer to scope lifetime inside `with`?
```suggestion
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
            sock.settimeout(timeout)
            # connect_ex returns 0 on success, error code otherwise
            return sock.connect_ex((peer_address, port)) == 0
```

thegravi (approved):

test/qa/lib/fileshare.py
```diff
@@ -437,25 +436,43 @@ def bind_port() -> socket.socket | None:
     return None
 
 
-def check_fileshare_port_status(is_allowed: bool, peer_address: str) -> bool:
+def is_port_accessible(peer_address: str, port: int = 49111, timeout: float = 2.0) -> bool:
```
devzbysiu: We already have `is_port_accessible_TCP` and `is_port_accessible_UDP`, maybe it can be reused/combined?

test/qa/lib/fileshare.py
```diff
@@ -437,25 +436,43 @@ def bind_port() -> socket.socket | None:
...
+    """
+    Attempt to connect to the port to verify it's accessible.
+
+    Returns True if connection succeeds, False otherwise.
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.settimeout(timeout)
+        result = sock.connect_ex((peer_address, port))
+        sock.close()
```
devzbysiu: This should be in `finally` block or using `with` as suggested [here](https://github.com/NordSecurity/nordvpn-linux/pull/1637/changes#r3643023158)

test/qa/lib/firewall.py
```diff
@@ -105,18 +105,18 @@ def is_source_port_reachable(ports: list[Port]) -> bool:
...
     print("process both, tcp")
-    tcp_retval = is_port_accessible_TCP(int(port.value))
+    tcp_retval = is_port_accessible_TCP(src_port=int(port.value), dst_ip=IP, dst_port=TCP_DST_PORT)
     print("process both, udp")
     udp_retval = is_port_accessible_UDP(int(port.value))
     return tcp_retval and udp_retval
 
 
-def is_port_accessible_TCP(src_port : int) -> bool :
+def is_port_accessible_TCP(src_port : int, dst_ip : str, dst_port : int, check_connection_only: bool = False) -> bool :
```
devzbysiu: [small] redundant spaces before `:` in type annotations. It was actually like that in original code also, but it's against python style-guide (weird that linter didn't complain 🤔)

test/qa/test_fileshare.py
devzbysiu, line 1522: This needs `ssh_client` now right?

test/qa/lib/fileshare.py
```diff
@@ -437,25 +436,25 @@ def bind_port() -> socket.socket | None:
...
-def check_fileshare_port_status(is_allowed: bool, peer_address: str) -> bool:
+def check_fileshare_port_status(is_allowed: bool, peer_address: str, port: int = 49111) -> bool:
+    """
+    Poll port accessibility status over 3 attempts with 1-second delays.
+
+    Returns True if actual accessibility matches expected state, False otherwise.
+    """
     for _ in range(3):
-        if is_port_allowed(peer_address) == is_allowed:
+        if firewall.is_port_accessible_TCP(53, peer_address, port, True) == is_allowed:
```
devzbysiu: Can we use `0` as source port instead of `53` so OS picks the port for us? `53` is DNS specific so this may be confusing here

devzbysiu (approved):

test/qa/test_fileshare.py
ggediminass: Currently we skip this test, due to bug in app. I have no way to check if this test will work properly, if I add my changes here too. Let's leave it like it is for now

thegravi (approved):

PR #1640, opened 2026-07-27
LVPN-7951-: Change firewall disable popup message when KS=1

mariusSincovici:
Update error messages for popup displayed when kill switch is enabled and user tries to disable the firewall.
Disable kill switch when firewall is disabled.


thegravi (approved):
[out of scope] How hard would it be to cover vice-versa case, where we'd try to enable KS when FW is off? We could tailor existing message to be applicable for this case.

mariusSincovici:
> [out of scope] How hard would it be to cover vice-versa case, where we'd try to enable KS when FW is off? We could tailor existing message to be applicable for this case.

I've changed the application and now when firewall is disabled, then kill switch is also disabled.

keliramu (approved):

PR #1641, opened 2026-07-28
LVPN-10792: sanitize DIP/DS logs

bartoszWojciechO:


tray/daemon_state_handler.go
devzbysiu: This will log values we are trying to sanitize

daemon/rpc_connect.go
```diff
@@ -351,6 +351,9 @@ func (r *RPC) connect(
 	pauseDuration time.Duration,
 	vpnConnReason events.VPNConnectionReason,
 ) (didFail bool, retErr error) {
+	// disable sanitization in case it was previously enabled
+	r.logSanitizer.DisableSanitization()
```
devzbysiu, line 355: Maybe we can disable sanitization only if something was already set? Otherwise we'll be broadcasting this event every time on connect

devzbysiu (approved):

edvyno1 (approved):
add some python test for this logic

PR #1645, opened 2026-07-28
LVPN-10785: add pause status to the status command

bartoszWojciechO:


cli/cli_status.go
```diff
@@ -24,6 +24,19 @@ func (c *cmd) Status(ctx *cli.Context) error {
...
 
+func formatDuration(secs uint32) string {
+	h := secs / 3600
+	m := (secs % 3600) / 60
+	s := secs % 60
+
+	if h > 0 {
+		return fmt.Sprintf("%02d:%02d:%02d", h, m, s)
+	} else if m > 0 {
+		return fmt.Sprintf("%02d:%02d", m, s)
```
thegravi, line 35: [nit] Wouldn't it confuse a user to see e.g. pause for 24h, after some time user checks and sees 02:18. Is this 2h or 2min? Just wondering why not always use hh:mm:ss format?

thegravi (approved):

cli/cli_status.go
edvyno1: imo good as is, 5 out of the 6 options will not show hours anyway so for most of the use cases it would just be leading zeroes making the output more cluttery

edvyno1 (approved):

PR #1651, opened 2026-07-29
LVPN-10786: add pause cli analytics

bartoszWojciechO:


DoYouEvenCpp (approved):

cli/cli_pause.go
```diff
@@ -21,6 +21,23 @@ func pauseArgToDuration(arg string) (uint32, error) {
 	return 0, fmt.Errorf("unrecognized duration")
 }
 
+func pauseArgToItemValue(arg string) pb.UIEvent_ItemValue {
```
devzbysiu, line 24: Did you consider unifying pause-related info in one place? Now, we have it repeated here, in `menu.go` for tray and in `pause.dart`, maybe there is a way to have all that info in one place in proto? 

If not, then maybe we could at least unify `go` side to have CLI and tray stuff in one place

bartoszWojciechO: What would we want to unify in this case? The common factor is the amount of seconds per pause interval and the analytics `ItemValue` per pause interval. But in case of tray/GUI this information is tied to button/label names and in case of cli it's tied to the argument names, specific to their respective domains(labels/buttons are `"Pause for 5 minutes"` and args are `5m`).

bartoszWojciechO: TBH this RPC could be simplified to simply accept an enum of the accepted interval. There is no reason to allow client to specify a number of seconds. But I don't think we should make such change in this PR.

thegravi (approved):

cli/cli_pause.go
devzbysiu: What I mean is that when dev has to add new pause value, he has to do changes in:

- `uievent.proto`
- `tray/menu.go` - new label + new entry in `pauseLengths`
- `cli/cli_pause.go` - new entry in `argToPauseDurationSeconds` + new case in `pauseArgoToItemValue` switch
- `uievent/moose_strings.go` - new case in `itemValueToString` switch

So 6 changes scattered all around the code. This should be minimized as much as possible. Not necessary in this PR if it's too much changes now.
Client-specific stuff for sure can't be unfied, so we may not have single collection with all stuff, but would be good to have as few places to change as possible

bartoszWojciechO: Added a separate task to do this(LVPN-10911) as I don't want to blow up the scope of this PR too much before the release.

devzbysiu (approved):

PR #1653, opened 2026-07-29
LVPN-7955: Fix old account info after switching accounts

edvyno1:
PR fixes old account information showing up after logging into a different account

- Add invalidation of account data on logout event
- Get rid of GUI refresh hack


dfetti (approved):

daemon/data_manager.go
```diff
@@ -419,3 +420,8 @@ func (dm *DataManager) GetAccountData(requestFreshFetch bool) (*pb.AccountRespon
 func (dm *DataManager) InvalidateAccountData() {
 	dm.accountData.unset()
 }
+
+func (dm *DataManager) NotifyLogout(event.DataAuthorization) error {
+	dm.InvalidateAccountData()
```
thegravi, line 425: Maybe putting `Clear()` inside this `InvalidateAccountData()` call would be more safe?

thegravi (approved):

daemon/data_manager.go
edvyno1: I wanted to stay consistent with how we do it for [services cache](https://github.com/NordSecurity/nordvpn-linux/blob/050ca48dd2b885e15d5e9f957b966a95add3b695/auth/services_state.go#L52), but if you think we should do a full cache clear for account cache I could also do that, your call

DoYouEvenCpp: Yeah, probably `Clear` is safer - `InvalidateAccountData` technically speaking still allows user to fetch cached data even if invalid (e.g. `GetEvenIfStale`) so I guess this is a right call here.
Btw, maybe we also can consider updating as well following: https://github.com/NordSecurity/nordvpn-linux/blob/050ca48dd2b885e15d5e9f957b966a95add3b695/auth/services_state.go#L52 ? (but probably not for 5.3.0)

DoYouEvenCpp (approved):

daemon/data_manager.go
edvyno1: addressed the account data, changed the underlying call from invalidate cache to clear cache
kept services as is for now

edit: need to fix tests, ignore this for a bit

edvyno1: fixed

PR #1663, opened 2026-08-03
LVPN-10917: Append " " only if country is set

devzbysiu:
Changes:

- fix server tag calculation when ENS Maintenance even requires reconnect
- trim input arguments when creating search params

### DIFF
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -129,14 +129,17 @@ func (r *RPC) reconnectOnServerMaintenance(
 }
 
 func locationTag(code, city string) string {
-	tag := internal.SnakeCase(code)
-	if city != "" {
-		if tag != "" {
-			tag += " "
-		}
-		tag += internal.SnakeCase(city)
+	codeTag := internal.SnakeCase(code)
+	cityTag := internal.SnakeCase(city)
+
+	if codeTag == "" {
+		return cityTag
 	}
-	return tag
+	if cityTag == "" {
+		return codeTag
+	}
+
+	return codeTag + " " + cityTag
 }
 
 // determineServerSelectionRule determines the server selection rule based on the provided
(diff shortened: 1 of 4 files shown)


daemon/rpc_connect.go
```diff
@@ -137,6 +128,17 @@ func (r *RPC) reconnectOnServerMaintenance(
...
 
+func locationTag(code, city string) string {
+	tag := internal.SnakeCase(code)
+	if city != "" {
+		if tag != "" {
+			tag += " "
+		}
+		tag += internal.SnakeCase(city)
+	}
+	return tag
```
thegravi: [nit] I think this would be more readable instead of branching ifs.
```suggestion
    codeTag := internal.SnakeCase(code)
    cityTag := internal.SnakeCase(city)
    
    if codeTag == "" {
            return cityTag
    }
    if cityTag == "" {
            return codeTag
    }
    
    return codeTag + " " + cityTag
```

thegravi (approved):

dfetti (approved):

mariusSincovici (approved):

daemon/rpc_connect_test.go
```diff
@@ -1591,3 +1591,57 @@ func TestReconnectOnServerMaintenance_CountryOnly(t *testing.T) {
...
+			expected: "it rome",
+		},
+		{
+			name:     "code only",
+			code:     "IT",
+			city:     "",
+			expected: "it",
+		},
+		{
+			name:     "city only, no leading space",
```
DoYouEvenCpp, line 1617: [nit] so we can also have a complementary test for a case with city name starting with a leading space.

DoYouEvenCpp (approved):

daemon/rpc_connect.go
devzbysiu: Updated

daemon/rpc_connect_test.go
devzbysiu: Updated

PR #1664, opened 2026-08-03
LVPN-10915: Fix flaky ens test

devzbysiu:
Changes:

- instead of using external context for cancellation, introduce `Stop()`
- it waits until goroutine exits so there is no race between asynchronous stopping of the monitor and sending event immediately after sending signal to stop the monitor

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -645,12 +645,12 @@ func main() {
 	)
 
 	ensMonitor := ens.NewMonitor(
-		httpGlobalCtx,
 		netw,
 		rcConfig,
 		rpc.ReconnectOnServerMaintenanceEvent,
 		daemonEvents.Debugger.DebuggerEvents,
 	)
+	defer ensMonitor.Stop()
 	internalVpnEvents.ConnectionError.Subscribe(ensMonitor.HandleENSNotification)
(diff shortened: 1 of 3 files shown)


dfetti (approved):

cmd/daemon/main.go
```diff
@@ -645,12 +645,12 @@ func main() {
 	)
 
 	ensMonitor := ens.NewMonitor(
-		httpGlobalCtx,
 		netw,
 		rcConfig,
 		rpc.ReconnectOnServerMaintenanceEvent,
 		daemonEvents.Debugger.DebuggerEvents,
 	)
+	defer ensMonitor.Stop()
```
DoYouEvenCpp: Technically speaking, now there might be a tight window where a graceful shutdown is happening, some events are queued and the ENS monitor is not yet stopped - can this cause `ReconnectOnServerMaintenanceEvent` to be called upon app termination?
Maybe simply instead of deferring the `Stop` let's call it manually, where the shutdown is being handled?

DoYouEvenCpp (approved):

cmd/daemon/main.go
devzbysiu: Changed `Stop()` to be called manually instead of in defer

PR #1667, opened 2026-08-04
LVPN-10933: fix tray login

bartoszWojciechO:
Login from tray will fail if desktop related environment variables(primarily DESKTOP) are not available upon norduser startup(xdg-open exec call will fail).

In such cases, environment will be read via systemctl --user show-environment and set manually.

### DIFF
--- a/tray/actions.go
+++ b/tray/actions.go
@@ -108,7 +108,7 @@ func (ti *Instance) login() {
 				return
 			}
 
-			log.Warn("Desktop related environment variables not set, attemtpting to load them manually")
+			log.Warn("Desktop related environment variables not set, attempting to load them manually")
 			environment, err := getDesktopEnvironment()
 			if err != nil {
 				log.Error("Failed to read desktop environment manually:", err)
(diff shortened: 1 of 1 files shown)


tray/actions.go
```diff
@@ -79,11 +96,33 @@ func (ti *Instance) login() {
...
 				// we want to force a notification here, otherwise there will be no reaction to user action
 				ti.notify(Force, "Continue log in in the browser: %s", url)
+				return
+			}
+
+			if !strings.Contains(string(out), "Error: no DISPLAY environment variable specified") {
+				return
+			}
+
+			log.Warn("Desktop related environment variables not set, attemtpting to load them manually")
```
devzbysiu: [typo] `attemtpting`

devzbysiu (approved):

tray/actions.go
```diff
@@ -79,11 +96,33 @@ func (ti *Instance) login() {
...
+			environment, err := getDesktopEnvironment()
+			if err != nil {
+				log.Error("Failed to read desktop environment manually:", err)
```
DoYouEvenCpp, line 114: [nit] in case of an error this would become:
`Failed to read desktop environment manually: getting desktop environment: <ERROR>`
maybe let's simply return the error, without wrapping it around the extra string, directly from `getDesktopEnvironment` ?

DoYouEvenCpp (approved):

PR #1678, opened 2026-08-06
LVPN-10911: convert pause time on the daemons side

bartoszWojciechO:


protobuf/daemon/pause.proto
```diff
@@ -4,6 +4,14 @@ package pb;
...
+  PAUSE_5_MIN = 0;
+  PAUSE_15_MIN = 1;
+  PAUSE_30_MIN = 2;
+  PAUSE_1_HOUR = 3;
+  PAUSE_24_HOURS = 4;
+}
+
 message PauseRequest {
-  uint32 seconds = 1;
+  PauseInverval interval = 1;
```
devzbysiu: I think you have to also update `test_grpc.py` after this change

cli/cli_pause_test.go
```diff
@@ -67,7 +67,7 @@ func TestPauseArgumentsParsing(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			result, err := pauseArgToDuration(test.pauseDuration)
+			result, err := pauseArgToInterval(test.pauseDuration)
```
devzbysiu: This also requires update of the `expectedPauseDurationResult` values

protobuf/daemon/pause.proto
```diff
@@ -4,6 +4,14 @@ package pb;
 
 option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
 
+enum PauseInverval {
```
devzbysiu: [typo] `Inverval`

protobuf/daemon/pause.proto
devzbysiu: Btw. isn't this breaking API change?

cli/cli_pause.go
```diff
@@ -8,58 +8,45 @@ import (
 
 	"github.com/NordSecurity/nordvpn-linux/daemon/pb"
 	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/pause"
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
 )
 
-var argToPauseDurationSeconds = map[string]uint32{"5m": 300, "15m": 900, "30m": 1800, "1h": 3600, "24h": 86400}
+var argToPauseDurationSeconds = map[string]pb.PauseInverval{
```
devzbysiu: It's not longer map of arg to seconds

protobuf/daemon/pause.proto
```diff
@@ -4,6 +4,14 @@ package pb;
...
+  PAUSE_5_MIN = 0;
```
devzbysiu: It would probably be safer to have 0 value as unspecified to avoid the default, unset interval being `PAUSE_5_MIN`

cli/cli_pause_test.go
```diff
@@ -25,13 +25,13 @@ func TestPauseArgumentsParsing(t *testing.T) {
...
 			expectedPauseDurationResult: 900,
-			expectedUIEventResult:       pb.UIEvent_PAUSE_5_MIN,
+			expectedUIEventResult:       pb.UIEvent_PAUSE_15_MIN,
 		},
 		{
 			name:                        "success 30m",
 			pauseDuration:               "30m",
 			expectedPauseDurationResult: 1800,
-			expectedUIEventResult:       pb.UIEvent_PAUSE_5_MIN,
+			expectedUIEventResult:       pb.UIEvent_PAUSE_30_MIN,
```
devzbysiu: Those two fixes still don't help, because there is no assertion that verifies this field. That's why the test didn't fail in the first place 😄

devzbysiu: Those two fixes still don't help, because there is no assertion that verifies this field. That's why the test didn't fail in the first place 😄

devzbysiu (approved):

tray/menu.go
```diff
@@ -301,37 +302,23 @@ type pauseLength struct {
...
-		Name:            labelPause24H,
-		Tooltip:         labelPause24H,
-		DurationSeconds: 24 * 60 * 60,
-		EventValue:      pb.UIEvent_PAUSE_24_HOURS,
-	},
+	newPauseLength(labelPause5Min, 5*60),
+	newPauseLength(labelPause15Min, 15*60),
+	newPauseLength(labelPause30Min, 30*60),
+	newPauseLength(labelPause1H, 60*60),
+	newPauseLength(labelPause24H, 24*60*60),
```
thegravi: [small] Would it be possible to reuse actual values from uievent package?

thegravi (approved):

PR #1679, opened 2026-08-06
LVPN-10878: Handle connections limit

mariusSincovici:
Handle connection limit reached ENS event while connecting to the VPN.

### DIFF
--- a/daemon/ens/ens_monitoring.go
+++ b/daemon/ens/ens_monitoring.go
@@ -133,7 +133,7 @@ func (m *Monitor) serverMaintenanceEventProcessing(e events.VPNConnectionErrorEv
 	}
 }
 
-func (m *Monitor) connectionLimitReachedEventProcessing(e events.VPNConnectionErrorEvent) {
+func (m *Monitor) connectionLimitReachedEventProcessing(events.VPNConnectionErrorEvent) {
 	if !m.netw.CancelConnecting(ErrConnectionLimitReached) {
 		log.ENS.Info("connection limit reach ignored")
 	}
--- a/daemon/ens/ens_monitoring.go
+++ b/daemon/ens/ens_monitoring.go
@@ -2,6 +2,7 @@ package ens
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
@@ -12,6 +13,8 @@ import (
 )
 
 const evChSize = 2
+
+var ErrConnectionLimitReached = errors.New("connection limit reached")
 
 type ConnectCallback func(serverEndpoint string) error
 
@@ -87,25 +90,14 @@ func (m *Monitor) run() {
 				m.debuggerEvents.Publish(*newVPNConnectionErrorEvent(e.Code).ToDebuggerEvent())
 			}
 
-			if e.Code != events.VPNConnectionErrorServerMaintenance {
+			switch e.Code {
+			case events.VPNConnectionErrorServerMaintenance:
+				m.serverMaintenanceEventProcessing(e)
+			case events.VPNConnectionErrorConnectionLimitReached:
+				m.connectionLimitReachedEventProcessing(e)
+
+			default:
 				log.ENS.Debug("ignoring", e)
-				continue
-			}
-
-			if !m.netw.IsVPNActive() {
-				log.ENS.Debug("ignoring because VPN is not connected", e)
-				continue
-			}
-
-			currServer, _ := m.netw.GetConnectionParameters()
-			eventIsForDifferentServer := !currServer.EndpointEqual(e.ServerEndpoint)
-			if eventIsForDifferentServer {
-				log.ENS.Debug("ignoring ENS event for non-current server", e)
-				continue
-			}
-
-			if err := m.reconnectFn(e.ServerEndpoint); err != nil {
-				log.ENS.Error("failed to reconnect", err)
 			}
 
 		case <-m.ctx.Done():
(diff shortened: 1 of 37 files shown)


networker/networker.go
```diff
@@ -184,42 +187,46 @@ func NewCombined(
...
-) error {
+) (retErr error) {
 	netw.mu.Lock()
 	defer netw.mu.Unlock()
 
-	netw.allowlist = allowlist
-	netw.enableLocalTraffic = enableLocalTraffic
-	if netw.isConnectedToVPN() {
-		tempKSEnabled := false
-		if netw.KillSwitchState != enabledByUser {
```
mariusSincovici, line 207: Removed for now, we'll be implemented on a separate task.

internal/wait.go
```diff
@@ -12,3 +15,42 @@ func GetSignalChan() <-chan os.Signal {
...
+	go func() {
+		wg.Done()
+		fn()
+	}()
+
+	wg.Wait()
+}
+
+// TrySendTimeout - tries to send value on channel.
+// Blocks for the specified duration. If channel is block for the entire duration, value is discarded.
```
DoYouEvenCpp: [nit]
```suggestion
// Blocks for the specified duration. If channel is blocked for the entire duration, value is discarded.
```

networker/networker.go
```diff
@@ -184,42 +188,47 @@ func NewCombined(
...
 		fwConfig:           firewall.Config{Allowlist: allowlist},
 		ipForwardSetter:    ipForwardSetter,
+		cancelConnectingCh: make(chan error),
 	}
 }
 
 // Start VPN connection after preparing the network.
 func (netw *Combined) Start(
-	ctx context.Context,
+	parent context.Context,
```
DoYouEvenCpp: [nit]
```suggestion
	parentContext context.Context,
```

networker/networker.go
```diff
@@ -184,42 +188,47 @@ func NewCombined(
...
+	internal.RunAsync(func() {
+		select {
+		case <-ctx.Done():
+			return
+		case err, ok := <-netw.cancelConnectingCh:
+			if !ok {
+				return
 			}
+			log.Netw.Info("canceling connection", err)
+			cancelFn(err)
```
DoYouEvenCpp, line 222: Don't we also have to guard a non-failed Start path with extra `defer cancelFn(nil)` (around lines 208/210)?

DoYouEvenCpp (approved):

networker/networker.go
mariusSincovici: Changed to `parentCtx` to keep it shorter 😆 .

internal/wait.go
mariusSincovici: Done

internal/wait.go
```diff
@@ -12,3 +15,42 @@ func GetSignalChan() <-chan os.Signal {
 	signal.Notify(signals, os.Interrupt, linux.SIGTERM, linux.SIGHUP, linux.SIGUSR1)
 	return signals
 }
+
+// RunAsync executes the function on a different goroutine and waits until goroutine starts
+func RunAsync(fn func()) {
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
```
devzbysiu, line 24: [small] you can go with `wg.Go(func() { ... })`

daemon/ens/ens_monitoring_test.go
```diff
@@ -165,3 +208,58 @@ func TestENSMonitoringEventHandling(t *testing.T) {
...
+		return nil
+	}, &subs.Subject[events.DebuggerEvent]{})
+	monitor.Start()
+
+	assert.NilError(t, monitor.HandleENSNotification(events.VPNConnectionErrorEvent{
+		Code:           events.VPNConnectionErrorConnectionLimitReached,
+		ServerEndpoint: "",
+	}))
+
+	err := netw.Start(
```
devzbysiu, line 264: Shouldn't this be called before `monitor.HandleENSNotification()`? 🤔, monitor cancels connection which sends through `cancelConnectingCh` in networker, but `Start()` sets up listener on this channel later

devzbysiu (approved):

daemon/ens/ens_monitoring_test.go
```diff
@@ -165,3 +208,58 @@ func TestENSMonitoringEventHandling(t *testing.T) {
 		})
 	}
 }
+
+func TestCombined_ENSConnectionsLimitReached(t *testing.T) {
```
devzbysiu, line 213: Missing category?

internal/wait.go
mariusSincovici: It cannot be user because wg.Go sets done after function terminates:
```
        wg.Add(1)
	go func() {
		defer func() {
			wg.Done()
		}()
		f()
	}()
```

daemon/ens/ens_monitoring_test.go
mariusSincovici: Yes, I'll review this, otherwise it can be problematic

daemon/ens/ens_monitoring.go
```diff
@@ -122,3 +114,34 @@ func (m *Monitor) Stop() {
 	}
 	m.wg.Wait()
 }
+
+func (m *Monitor) serverMaintenanceEventProcessing(e events.VPNConnectionErrorEvent) {
+	if e.Code != events.VPNConnectionErrorServerMaintenance {
```
bartoszWojciechO: Isn't it safe to assume that this function will always be called with `e.Code == events.VPNConnectionErrorServerMaintenance`. Otherwise maybe it would be good to have some sort of log here? Same for `connectionLimitReachedEventProcessing`.

networker/networker.go
```diff
@@ -200,26 +205,30 @@ func (netw *Combined) Start(
...
-			if err := netw.unsetKillSwitch(); err != nil {
-				log.Error("failed to disable temporary KS", err)
+	ctx, cancelFn := context.WithCancelCause(parentCtx)
+	finishCh := make(chan any)
+	defer close(finishCh)
+
+	// check if connection is canceled from outside, e.g. ENS events
+	internal.RunAsync(func() {
+		select {
+		case <-ctx.Done():
```
bartoszWojciechO, line 215: Don't we want to also terminate this goroutine based on the outer context(the `parentContext`)? For example when connection times out, etc.

mariusSincovici: The goroutine will always terminate, because of `case <-finishCh:`. But it will also terminate at timeout on `ctx.Done()`, because `ctx` is a child of the context with the timeout. And parents cancels also child context.

bartoszWojciechO (approved):

daemon/ens/ens_monitoring.go
mariusSincovici: Yes, I've removed it.

PR #1680, opened 2026-08-06
LVPN-10809: Add VPN connection settings announcements

dfetti:
Make the VPN connection settings screen navigable using the keyboard and add announcements for its items.

### DIFF
--- a/gui/lib/settings/vpn_connection_settings.dart
+++ b/gui/lib/settings/vpn_connection_settings.dart
@@ -1,4 +1,6 @@
+import 'dart:async';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:nordvpn/data/models/app_settings.dart';
 import 'package:nordvpn/data/models/vpn_protocol.dart';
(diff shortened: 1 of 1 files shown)


gui/lib/settings/vpn_connection_settings.dart
```diff
@@ -169,11 +211,14 @@ final class VpnConnectionSettings extends ConsumerWidget {
               mainAxisSize: MainAxisSize.min,
               children: [
                 for (final item in protocols)
-                  RadioButton(
+                  _buildAccessibleRadio(
+                    context,
+                    groupLabel: t.ui.vpnProtocol,
+                    label: item.label,
```
DoYouEvenCpp, line 200: For OpenVPN labels,  for a11y announcements, we can consider switching here for
- `OpenVPN TCP`
- `OpenVPN UDP`
(to not read out parenthesis)

DoYouEvenCpp (approved):

gui/lib/settings/vpn_connection_settings.dart
```diff
@@ -83,37 +85,72 @@ final class VpnConnectionSettings extends ConsumerWidget {
...
+    void open() {
+      if (enabled) context.navigateToRoute(AppRoute.settingsAutoconnect);
+    }
+
+    return MergeSemantics(
+      child: Semantics(
+        button: true,
+        enabled: enabled,
+        onTap: enabled ? open : null,
+        child: FocusableActionDetector(
```
DoYouEvenCpp: Can we extract repeating `FocusableActionDetector` into a helper method, that probably can get its child and the action to call for the intent?

keliramu (approved):

PR #1683, opened 2026-08-07
LVPN-10936: Fix Maintenance reconnect to country after auto-connect

devzbysiu:
Context:

- when setting auto-connect, we are filling `AutoConnectData.Country` field with country code (`AutoConnectData` does not have country code field, should be addressed in LVPN-10972)
- then when auto-connect happens, we are re-storing `RequestedConnParams` from that stored config and we expect `ServerParameters.CountryCode` field to be set, but it was not being set, because auto-connect data does not have country code
- later we are reconnecting on ENS Maintenance based on restored `RequestedConnParams.CountryCode` which was not set so we did a fallback to Quick Connect
- this PR restores `ServerParameters.CountryCode` based on `AutoConnectData.Country` field

Changes:

- set `ServerParameters.CountryCode` during auto-connect
- tests


daemon/jobs.go
```diff
@@ -421,8 +421,13 @@ func (r *RPC) doAutoConnect() error {
 			pb.ConnectionSource_AUTO,
 			serverpicker.ServerParameters{
 				Country: cfg.AutoConnectData.Country,
-				City:    cfg.AutoConnectData.City,
-				Group:   cfg.AutoConnectData.Group,
+				// NOTE: This is not a mistake. At the time of writing, `AutoConnectData`
+				// does not store country code, but we need to rely on it so we are setting
+				// `AutoConnectData.Country` field to country code, see
+				// https://github.com/NordSecurity/nordvpn-linux/blob/5.3.0/daemon/rpc_set_autoconnect.go#L95.
+				CountryCode: cfg.AutoConnectData.Country,
```
keliramu: Why add things but not use them? Maybe add them in the PR where it will be really used?

devzbysiu: Why do you think it's not used? This is the change that addresses Server Maintenance reconnect issue

dfetti (approved):

daemon/jobs.go
```diff
@@ -421,8 +421,13 @@ func (r *RPC) doAutoConnect() error {
 			pb.ConnectionSource_AUTO,
 			serverpicker.ServerParameters{
 				Country: cfg.AutoConnectData.Country,
-				City:    cfg.AutoConnectData.City,
-				Group:   cfg.AutoConnectData.Group,
+				// NOTE: This is not a mistake. At the time of writing, `AutoConnectData`
```
mariusSincovici: What if instead rpc_set_autoconnect is fixed, to store the country code?
And in rpc connect to use country code or country, whichever is not empty? 🤔

mariusSincovici (approved):

PR #1686, opened 2026-08-11
LVPN-10812: fix announcements for general settings page

DoYouEvenCpp:
Changes under this PR:
- `RadioButton` and `OnOffSwitch` widgets are now focusable and thus capable of handling a11y semantics
- some changes to `SettingsWrapperWidget.buildListItem` required to correctly handle semantics (specifically between leading/trailing widgets)
- `CustomExpansionTile` now takes `expandButtonSemanticLabel` so the expand button is announced with a meaningful labal, rather than a generic `push button`
- `GeneralSettings` with fixed announcements, and OK(ish) keyboard navigation (however, keyboard navi wasn't in the scope of this PR, per se)

### DIFF
--- a/gui/lib/constants.dart
+++ b/gui/lib/constants.dart
@@ -17,6 +17,8 @@ final daemonDateFormat = DateFormat("yyyy-MM-dd h:m:s");
 const animationDuration = Duration(milliseconds: 200);
 const loginTimeoutDuration = Duration(seconds: 10);
 
+const focusRingWidth = 2.0;
+
 const maxInt32 = 0xffffffff;
 const maxInt16 = 0xffff;
 const maxCustomDnsServers = 3;
--- a/gui/lib/settings/general_settings.dart
+++ b/gui/lib/settings/general_settings.dart
@@ -69,6 +69,7 @@ class GeneralSettings extends ConsumerWidget {
             return SettingsWrapperWidget.buildListItem(
               context,
               title: t.ui.showNotifications,
+              mergeSemantics: false,
               trailing: OnOffSwitch(
                 semanticLabel: t.ui.showNotifications,
                 value: settings.notifications,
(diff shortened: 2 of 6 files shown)


gui/lib/constants.dart
```diff
@@ -17,6 +17,8 @@ final daemonDateFormat = DateFormat("yyyy-MM-dd h:m:s");
 const animationDuration = Duration(milliseconds: 200);
 const loginTimeoutDuration = Duration(seconds: 10);
 
+const focusRingWidth = 2.0;
```
DoYouEvenCpp, line 20: I'll take a look if this isn't defined already somewhere in the theme

gui/lib/settings/general_settings.dart
```diff
@@ -63,12 +63,15 @@ class GeneralSettings extends ConsumerWidget {
               context,
               title: t.ui.appearance,
               trailing: _buildAppearanceTrailing(context, ref, mode),
+              mergeSemantics: false,
             );
           case _GeneralSettingsItems.notifications:
             return SettingsWrapperWidget.buildListItem(
               context,
               title: t.ui.showNotifications,
+              mergeSemantics: false,
```
dfetti, line 72: This will be in conflict with what I have done in the vpn connection settings and in the consent screen. Take a look at the at the [AccesibleItem](https://github.com/NordSecurity/nordvpn-linux/blob/LVPN-10795-consent-screen-accessibility/gui/lib/widgets/accessible_item.dart) and the way it's used [here](https://github.com/NordSecurity/nordvpn-linux/blob/LVPN-10795-consent-screen-accessibility/gui/lib/settings/vpn_connection_settings.dart#L118) or you can find multiple examples on that branch.

dfetti:
Addressed by [this](https://github.com/NordSecurity/nordvpn-linux/pull/1704)

PR #1688, opened 2026-08-12
LVPN-10881: Display popup for connection limit reached

mariusSincovici:
When connection limit reached event is received in the GUI a popup will be displayed to the user.

### DIFF
--- a/gui/integration_test/e2e_tests/ens_test.dart
+++ b/gui/integration_test/e2e_tests/ens_test.dart
@@ -1,0 +1,28 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:nordvpn/data/repository/daemon_status_codes.dart';
+import 'package:nordvpn/i18n/strings.g.dart';
+
+import '../../test/utils/test_helpers.dart';
+
+void runEnsTests() async {
+  group("test ENS", () {
+    testWidgets("connection limit reached popup", (tester) async {
+      final app = await tester.setupIntegrationTests();
+
+      final mainScreen = await app.goToVpnScreen();
+      // set error code returned by the daemon to be
+      app.vpnStatus.connectingErrorStatusCode =
+          DaemonStatusCode.connectionLimitReached;
+
+      await mainScreen.quickConnect();
+
+      final txtFinder = find.text(t.ui.connectionLimitReachedTitle);
+      await tester.pumpUntilFound(txtFinder);
+      expect(txtFinder.evaluate().isNotEmpty, isTrue);
+      expect(
+        find.text(t.ui.connectionLimitReachedTitle).evaluate().isNotEmpty,
+        isTrue,
+      );
+    });
+  });
+}
--- a/gui/integration_test/e2e_tests/ens_test.dart
+++ b/gui/integration_test/e2e_tests/ens_test.dart
@@ -1,6 +1,5 @@
 import 'package:flutter_test/flutter_test.dart';
 import 'package:nordvpn/data/repository/daemon_status_codes.dart';
-import 'package:nordvpn/i18n/strings.g.dart';
 
 import '../../test/utils/test_helpers.dart';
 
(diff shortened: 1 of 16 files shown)


gui/lib/data/mocks/daemon/grpc_server.dart
```diff
@@ -22,17 +22,14 @@ final class GrpcServer {
...
-
   bool get isRunning => _daemon != null;
 
   MockDaemon get daemon => _daemon!;
   MockAccountInfo get account => daemon.account;
   MockApplicationSettings get appSettings => daemon.appSettings;
   MockServersList get serversList => daemon.serversList;
   MockVpnStatus get vpnStatus => daemon.vpnStatus;
-  MockSnapErrorInterceptor get snapInterceptor => _snapInterceptor;
+  MockSnapErrorInterceptor get snapInterceptor => daemon.snapInterceptor;
```
mariusSincovici, line 32: Moved the interceptor to daemon, because here it will never be recreated.

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -0,0 +1,28 @@
...
+      final app = await tester.setupIntegrationTests();
+
+      final mainScreen = await app.goToVpnScreen();
+      // set error code returned by the daemon to be
+      app.vpnStatus.connectingErrorStatusCode =
+          DaemonStatusCode.connectionLimitReached;
+
+      await mainScreen.quickConnect();
+
+      final txtFinder = find.text(t.ui.connectionLimitReachedTitle);
```
devzbysiu: Would it make sense to have a `Key` set there and find stuff with `find.byKey()`?

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -0,0 +1,28 @@
...
+      await tester.pumpUntilFound(txtFinder);
+      expect(txtFinder.evaluate().isNotEmpty, isTrue);
+      expect(
+        find.text(t.ui.connectionLimitReachedTitle).evaluate().isNotEmpty,
```
devzbysiu: Isn't this the same assertion as one line above?

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -0,0 +1,28 @@
...
+      final mainScreen = await app.goToVpnScreen();
+      // set error code returned by the daemon to be
+      app.vpnStatus.connectingErrorStatusCode =
+          DaemonStatusCode.connectionLimitReached;
+
+      await mainScreen.quickConnect();
+
+      final txtFinder = find.text(t.ui.connectionLimitReachedTitle);
+      await tester.pumpUntilFound(txtFinder);
+      expect(txtFinder.evaluate().isNotEmpty, isTrue);
```
devzbysiu: Maybe we could have a higher level helper that finds what we need, similar like we have in  `vpn_screen.dart` test?

```dart
expect(mainScreen.isSubscriptionPopupVisible(), isFalse);
```

devzbysiu (approved):

gui/integration_test/e2e_tests/ens_test.dart
mariusSincovici: Done. Added key for all the popups

gui/integration_test/e2e_tests/ens_test.dart
mariusSincovici: Yes, I've fixed to check also the body message.
Because it contains URL it is a bit more code to check it.

gui/integration_test/e2e_tests/ens_test.dart
mariusSincovici: Done

edvyno1 (approved):

postnlllogistics-boop (commented):
Explore

PR #1694, opened 2026-08-13
LVPN-10981: Add simple http transport recreation on connect

edvyno1:
This PR adds simple http transport recreation logic on connect event, so it's remade immediately after connect is successful and does not have to wait for errors to recreate itself

### DIFF
--- a/request/http_retransport.go
+++ b/request/http_retransport.go
@@ -19,7 +19,7 @@ type HTTPReTransport struct {
 	protoMinor      int
 	proto           string
 	shouldRetryFunc ShouldRetryFunc
-	NeedRecreate    bool
+	needRecreate    bool
 	// counter provides a mechanism to check whether the inner RoundTripper was re-created
 	// in the background during this round trip. Integer overflow is not important here as only
 	// inequality operator is used.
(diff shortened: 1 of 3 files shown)


dfetti (approved):

request/http_retransport.go
```diff
@@ -65,6 +65,9 @@ func (m *HTTPReTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 
 // NotifyConnect initiates re-creating the inner round tripper when called.
 func (m *HTTPReTransport) NotifyConnect(events.DataConnect) error {
+	if !m.NeedRecreate {
```
devzbysiu: [small] this `NeedsRecreate` could be unexported. It's not accesses outside anyway, but no need to have it public. Also this could be put behind mutext maybe? Similar way as `counter` is read. It's RW mutex, so reads should not be problematic anyway and we'll have one less footgun 😄

devzbysiu (approved):

request/http_retransport.go
edvyno1: applied changes

PR #1695, opened 2026-08-13
Bump riverpod version

dfetti:


keliramu (approved):

gui/lib/data/providers/login_status_provider.dart
```diff
@@ -50,9 +50,12 @@ final class LoginStatus extends _$LoginStatus implements AccountObserver {
 
   Future<void> _refreshState() async {
     try {
-      state = AsyncData(await _fetchLoginState());
+      final loginState = await _fetchLoginState();
+      if (!ref.mounted) return;
```
devzbysiu: We could have guard first and then `final loginState = await _fetchLoginState();` to not do that if it won't be used anyway. Or even have a guard before `try` block to not repeat it in `catch`

gui/lib/widgets/interactive_list_view.dart
```diff
@@ -43,19 +43,29 @@ class _InteractiveListViewState extends State<InteractiveListView> {
...
   @override
   void dispose() {
+    _activeController.removeListener(_onQueryChanged);
     _searchController.dispose();
     _searchFieldNodeFocus.dispose();
     super.dispose();
   }
 
+  void _onQueryChanged() {
+    if (mounted) setState(() {});
```
devzbysiu: This will rebuild the widget on every change of the query, cursor move, click etc. Previously we did that only in certain conditions

devzbysiu (approved):

PR #1696, opened 2026-08-13
Allow to set connection limit reached event before connecting

mariusSincovici:
Changes are only for dev builds:
* Moved ENS part from libtelio.
* Added an atomic bool to keep track if connection limit reach event should be returned at the next connect.
* Fix tests to be less likely to be flaky.

### DIFF
--- a/daemon/vpn/nordlynx/libtelio/dev_ens.go
+++ b/daemon/vpn/nordlynx/libtelio/dev_ens.go
@@ -1,0 +1,49 @@
+package libtelio
+
+import (
+	"errors"
+	"sync/atomic"
+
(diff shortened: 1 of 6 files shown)


devzbysiu (approved):

daemon/vpn/nordlynx/libtelio/dev_ens.go
```diff
@@ -0,0 +1,49 @@
+package libtelio
```
bartoszWojciechO, line 1: Would it be possible to add some comments to this file, explaining what's it for, etc?

bartoszWojciechO (approved):

PR #1697, opened 2026-08-13
LVPN-10814 announcements for popups

DoYouEvenCpp:
With this PR, popups are now announced in a following manner:
- popups' content is announced with a `Popup` information (hence added `popupWithContent` into a11y strings)
- popups' buttons are now separately announced (such as cancel, accept, `X` close button)
- popups are announced now upon being shown

### DIFF
--- a/gui/lib/widgets/popups/decision_popup.dart
+++ b/gui/lib/widgets/popups/decision_popup.dart
@@ -19,7 +19,10 @@ final class DecisionPopup extends Popup {
       crossAxisAlignment: CrossAxisAlignment.start,
       spacing: theme.verticalElementSpacing,
       children: [
-        Text(message(ref), style: theme.textSecondary),
+        Semantics(
+          container: true,
+          child: Text(message(ref), style: theme.textSecondary),
+        ),
         Row(
           spacing: theme.gapBetweenElements,
           children: [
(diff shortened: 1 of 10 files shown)


gui/lib/widgets/popups/decision_popup.dart
```diff
@@ -19,7 +19,14 @@ final class DecisionPopup extends Popup {
       crossAxisAlignment: CrossAxisAlignment.start,
       spacing: theme.verticalElementSpacing,
       children: [
-        Text(message(ref), style: theme.textSecondary),
+        Semantics(
```
devzbysiu, line 22: I think this is not needed. `Text` already has semantics and this will just add empty container that does not contribute anything to a11y. Same in other popups types. Was there a reason to add this?

devzbysiu (approved):

dfetti (approved):

gui/lib/widgets/popups/decision_popup.dart
keliramu: This is needed to have buttons announced without prepending dialog's title and message.
Dialog's title and message is being announced once when dialog gets open, then when user navigates through buttons only button label is announced.

PR #1698, opened 2026-08-13
LVPN-10874: Fix flaky `test_killswitch_autoconnect_to_fastest_debug` test

ggediminass:
Sometimes this test fails due to fact, that it expects country it connects to, to be in the same country as our device. This is incorrect approach, since API may return a neighboring country for Quick Connect recommendations, and in such case test fails. Due to this reason, from now on in this test we will compare the country we've been connected to, with a hard-coded list of countries surrounding Germany.

### DIFF
--- a/test/qa/test_killswitch.py
+++ b/test/qa/test_killswitch.py
@@ -285,4 +285,6 @@ def test_killswitch_autoconnect_to_fastest_debug():
     duration = time.monotonic() - start_time
     assert duration < 7, "Autoconnect with killswitch on should connect in less than 7 seconds"
     status_data = daemon.get_status_data()
-    assert device_country in status_data["country"], "Fastest server should be in the same country"
+    # list of closest countries that surround Germany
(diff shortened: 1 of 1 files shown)


test/qa/test_killswitch.py
```diff
@@ -275,8 +275,7 @@ def test_killswitch_mode_daemon_debug():
     assert network.is_available(), "Network should be available"
 
 
-def test_killswitch_autoconnect_to_fastest_debug():
-    device_country = network.get_external_device_country()
```
edvyno1, line 279: maybe keep this and assert that it is actually Germany, just in case this changes at some point so we would know with 100% failing test

test/qa/test_killswitch.py
edvyno1: remove `_debug` from this test as well

edvyno1 (approved):

mariusSincovici (approved):

PR #1702, opened 2026-08-17
LVPN-10882: Tray notifications

devzbysiu:
Context:

- displaying URLs in OS notifications is not the best UX, especially for long URLs or in DE that don't render URLs as clickable
- fileshare notifications support defining actions displayed as button
- tray notifications code was a duplicate of fileshare notifications, but without button support
- this PR introduces common code for displaying notifications, incluing specifying actions as button

Changes:

- introduce `alert` package that manages notifications and their actions
- migrate tray to `alert`
- migrate fileshare to `alert`

### DIFF
--- a/alert/dbus_notifier.go
+++ b/alert/dbus_notifier.go
@@ -1,0 +1,211 @@
+// Package alert delivers API for sending OS notification.
+package alert
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/NordSecurity/nordvpn-linux/internal"
+	"github.com/NordSecurity/nordvpn-linux/log"
+	"github.com/esiqveland/notify"
+	"github.com/godbus/dbus/v5"
+)
+
+type AlertID = uint32
+
+type Urgency int
+
+const (
+	// UrgencyNormal indicates that the alert should respect the user's settings.
+	UrgencyNormal Urgency = iota
+	// UrgencyCritical indicates that the alert should be shown regardless of the user's settings.
+	UrgencyCritical
+)
+
+func (u Urgency) String() string {
+	switch u {
+	case UrgencyNormal:
+		return "normal"
+	case UrgencyCritical:
+		return "critical"
+	default:
+		return "unknown"
+	}
+}
+
+type Notifier interface {
+	Alert(body string) *AlertBuilder
+	Mute()
+	Unmute()
+	Close() error
+}
+
+type Alert struct {
+	Summary string
+	Body    string
+	Actions []Action
+	Urgency Urgency
+}
+
+func (a Alert) String() string {
+	return fmt.Sprintf("[%s] %q (%s)", a.Summary, a.Body, a.Urgency)
+}
+
+type Action struct {
+	Key      string
+	Label    string
+	Callback func()
+}
+
+type DbusNotifier struct {
+	mu        sync.Mutex
+	isActive  bool
+	transient bool
+	notifier  notify.Notifier
+	actions   map[AlertID]map[string]func()
+}
+
+type Option func(*DbusNotifier)
+
+func WithTransient() Option {
+	return func(n *DbusNotifier) {
+		n.transient = true
+	}
+}
+
+func NewDbusNotifier(opts ...Option) (*DbusNotifier, error) {
+	dbusConn, err := dbus.SessionBusPrivate()
+	if err != nil {
+		return nil, fmt.Errorf("creating D-Bus connection: %w", err)
+	}
+
+	var success bool
+	defer func() {
+		if !success {
+			_ = dbusConn.Close()
+		}
+	}()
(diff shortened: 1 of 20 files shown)


alert/dbus_notifier.go
```diff
@@ -0,0 +1,211 @@
...
+type Option func(*DbusNotifier)
+
+func WithTransient() Option {
+	return func(n *DbusNotifier) {
+		n.transient = true
+	}
+}
+
+func NewDbusNotifier(opts ...Option) (*DbusNotifier, error) {
+	dbusConn, err := dbus.SessionBusPrivate()
```
mariusSincovici, line 77: This is not new code, but I'm wandering why do we need a new private session to the dbus and not use the shared one? 🤔

tray/tray.go
```diff
@@ -298,13 +307,13 @@ func (ti *Instance) Start() {
 
 func (ti *Instance) OnExit() {
 	ti.stateListener.Stop()
+	_ = ti.n.Close()
```
mariusSincovici, line 317: If tray is closed when disabled, the notifications are not working anymore.

test/mock/notify/notify.go
```diff
@@ -0,0 +1,31 @@
+package notify
```
mariusSincovici: [small] if you name the package `mocknotify` or similar then IDE would automatically find the package and not conflict with the real version.

tray/notifications.go
```diff
@@ -1,126 +1,102 @@
...
-		if err != nil {
-			if err := dbusConn.Close(); err != nil {
-				log.Error("Failed to close dbus connection:", err)
-			}
-		}
-	}()
-
-	if err = dbusConn.Auth(nil); err != nil {
-		return nil, err
+func (ti *Instance) connectionResultAlert(out *pb.Payload) *alert.AlertBuilder {
```
mariusSincovici, line 31: [small] I don't have a suggestion for name, but since it only contains the errors, and connected/connecting are left out, shouldn't the name somehow reflect? 🤔

tray/notifications.go
```diff
@@ -1,126 +1,102 @@
...
+		return ti.n.Alert(internal.DoubleGroupErrorMessage)
+	case internal.CodeVPNRunning:
+		return ti.n.Alert(client.ConnectConnected)
+	case internal.CodeNothingToDo:
+		return ti.n.Alert(client.ConnectConnecting)
+	case internal.CodeUFWDisabled:
+		return ti.n.Alert(client.UFWDisabledMessage)
+	case internal.CodeDedicatedServersRenewError:
+		link := ti.trustedPassURLOrDefault(client.DedicatedServersUpselURL, client.DedicatedServersUpselURLLogin)
+		return ti.n.Alert(fmt.Sprintf(cli.DedicatedServersNoServiceMessage, link)).Urgent()
```
mariusSincovici, line 65: 🤔 Why so many urgent?

tray/notifications.go
```diff
@@ -1,126 +1,102 @@
...
-		log.Systray.Infof("Notification suppressed: %s (status: %d, force: %v)",
-			fmt.Sprintf(text, a...),
-			notificationsStatus,
-			ntype == Force)
-	}
+// gatedNotifier suppresses every alert, even Urgent ones,
+// until the tray has completed its sync with the daemon.
+type gatedNotifier struct {
+	alert.Notifier
+	state *trayState
```
mariusSincovici: [small] Passing the pointer here can be a bit risky, because one can forget to lock before accessing members.
To prevent possible data race, and since only `initialSyncCompleted` is being used, to have a `func()bool` that takes the lock outside and just returns the state?

alert/dbus_notifier.go
devzbysiu: We probably could go with shared, but then we need to be careful in every place this shared connection is used, to not close it too early, because something else can use it, so I think this is safer.
Also, I think that `notify` lib registers signal handlers related to actions on the connection, so we would have to be careful here as well

tray/tray.go
devzbysiu: Will be addressed separately

tray/notifications.go
devzbysiu: Will clarify with original authors of the notifications

tray/notifications.go
devzbysiu: Done

test/mock/notify/notify.go
devzbysiu: Done

tray/notifications.go
devzbysiu: Discussed

mariusSincovici (approved):

bartoszWojciechO (approved):

PR #1706, opened 2026-08-17
LVPN-10994: Test if peer is unreachable, after removing it

ggediminass:
Refactor old test, which relies on checking `iptables`.

Goal is to have a test among `core_meshnet` tests, which would verify, that peer cannot be reached anymore, if we remove it from our Meshnet.

### DIFF
--- a/test/qa/test_fileshare.py
+++ b/test/qa/test_fileshare.py
@@ -56,12 +56,7 @@ def setup_module(module):  # noqa: ARG001
     sh.nordvpn.mesh.peer.refresh()
     ssh_client.exec_command("nordvpn mesh peer refresh")
 
-    while True:
-        local_peer_list = sh_no_tty.nordvpn.mesh.peer.list()
-        remote_peer_list = ssh_client.exec_command("nordvpn mesh peer list")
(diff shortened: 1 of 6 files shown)


test/qa/test_fileshare.py
```diff
@@ -56,12 +55,7 @@ def setup_module(module):  # noqa: ARG001
     sh.nordvpn.mesh.peer.refresh()
     ssh_client.exec_command("nordvpn mesh peer refresh")
 
-    while True:
-        local_peer_list = sh_no_tty.nordvpn.mesh.peer.list()
-        remote_peer_list = ssh_client.exec_command("nordvpn mesh peer list")
-        if all("Status: connected" in peer_list for peer_list in (local_peer_list, remote_peer_list)):
-            break
-        time.sleep(1)
+    meshnet.are_peers_connected(ssh_client)
```
mariusSincovici: [small] Small because it is not part of this PR, but `are_peers_connected` is a bit misleading. The function doesn't return `true|false`, but instead waits for the peers to be connected.

mariusSincovici (approved):

dfetti (approved):

PR #1709, opened 2026-08-18
LVPN-10979: Introduce retry mechanism to handle flake test scenarios

bbrutall:
- implemented "retry_on_exc" decorator that can be applied also for for a particular call in a particular place.
- unified some common functions which have retry mechanism to use "retry_on_exc" solution

### DIFF
--- a/test/qa/lib/__init__.py
+++ b/test/qa/lib/__init__.py
@@ -4,6 +4,7 @@ import random
 import os
 from collections.abc import Callable
 from enum import Enum
+from functools import wraps
 
 import sh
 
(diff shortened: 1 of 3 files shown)


mariusSincovici (approved):

test/qa/lib/__init__.py
```diff
@@ -404,3 +405,36 @@ def get_random_virtual_country() -> str:
...
+    """
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            for attempt in range(1, attempts + 1):
+                try:
+                    return func(*args, **kwargs)
+                except Exception as err:
+                    if attempt == attempts:
+                        print(
```
devzbysiu, line 428: Shouldn't this use our `logging.log`? Also in other places in this PR

devzbysiu (approved):

test/qa/lib/__init__.py
bbrutall: `print` is already established convention throughout this module, and decorator follows the same pattern for consistency.

PR #1712, opened 2026-08-19
LVPN-10889: Add OVPN DCO support

thegravi:
### Turns on OpenVPN kernel data channel offload (DCO).
We were passing `--disable-dco` always, because servers used `comp-lzo no`. Template stopped pinning a cipher in #1687.
Now we only pass this flag when offload actually can't work — obfuscated connection lives in userspace

### Firewall is now set up before the VPN starts.
The `meta mark -> ct mark set` rule needs to be there during the handshake — that's the one and only time a marked packet comes out of userspace. After that DCO (handed socket to kernel module) never sends one. Miss it and the output of offloaded traffic is gone.

### New 'suggest'
Adding additional suggestion to install (out-of-tree) kernel module: `openvpn-dco-dkms`

### DIFF
--- a/ci/nfpm/template.yaml
+++ b/ci/nfpm/template.yaml
@@ -152,10 +152,14 @@ contents:
     # like rpm-ostree.
     file_info:
       mode: 0755
-    
+
 # All fields above marked as `overridable` can be overridden for a given package format in this section.
 overrides:
   deb:
+    # Package provides the out-of-tree kernel module the bundled OpenVPN 2.6.x needs for
+    # data channel offload.
+    suggests:
+      - openvpn-dco-dkms
     # Deb specific scripts.
     scripts:
       preinstall: ${WORKDIR}/contrib/scriptlets/deb/preinst
@@ -173,6 +177,10 @@ overrides:
       - glibc >= 2.35
       - (libsqlite3-0 or sqlite-libs)
       - (xsltproc or libxslt)
+    # Package provides the out-of-tree kernel module the bundled OpenVPN 2.6.x needs for
+    # data channel offload.
+    suggests:
+      - (kmod-ovpn-dco or ovpn-dco-kmp-default)
     recommends:
       - iptables
       - nordvpn-gui
--- a/daemon/job_templates.go
+++ b/daemon/job_templates.go
@@ -9,40 +9,44 @@ import (
 	"github.com/NordSecurity/nordvpn-linux/log"
 )
 
+// ovpnTemplates maps every OpenVPN config template variant to the file it is cached in.
+var ovpnTemplates = map[core.OvpnTemplateVariant]string{
+	core.OvpnTemplateStandard:   internal.OvpnTemplatePath,
+	core.OvpnTemplateObfuscated: internal.OvpnObfsTemplatePath,
+}
+
 func JobTemplates(cdn core.CDN) func() {
 	return func() {
-		getTemplate := func(isObfuscated bool) {
+		getTemplate := func(variant core.OvpnTemplateVariant, cachePath string) {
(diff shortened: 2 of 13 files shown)


ci/nfpm/template.yaml
```diff
@@ -152,10 +152,14 @@ contents:
...
     file_info:
       mode: 0755
-    
+
 # All fields above marked as `overridable` can be overridden for a given package format in this section.
 overrides:
   deb:
+    # openvpn-dco-dkms provides the out-of-tree kernel module that the bundled OpenVPN 2.6.x
+    # uses for data channel offload
+    suggests:
```
edvyno1, line 161: swap to `recommends` over `suggests`, recommends is usually set up on systems to be installed automatically, where as suggests would need manual user interaction to specifically install it, less likely to be done

thegravi: Yes, but that could pull in a lot of deps. Need to test with QA. Thats why I left it under suggest.

daemon/vpn/openvpn/config.go
```diff
@@ -51,20 +53,53 @@ type ovpnConfigData struct {
...
+
+// dcoKernelSupported reports whether the ovpn-dco kernel module is available.
+func dcoKernelSupported() bool {
+	if _, err := netlink.GenlFamilyGet(ovpnDCONetlinkFamily); err != nil {
+		log.Info("OpenVPN kernel data channel offload unavailable:", err)
+		return false
+	}
+
+	return true
+}
```
edvyno1: no need for us to do this manually I think, openvpn should disable it during init if kernel does not support it

daemon/job_templates.go
```diff
@@ -9,40 +9,44 @@ import (
...
+// ovpnTemplates maps every OpenVPN config template variant to the file it is cached in.
+var ovpnTemplates = map[core.OvpnTemplateVariant]string{
+	core.OvpnTemplateStandard:   internal.OvpnTemplatePath,
+	core.OvpnTemplateObfuscated: internal.OvpnObfsTemplatePath,
+}
+
 func JobTemplates(cdn core.CDN) func() {
 	return func() {
-		getTemplate := func(isObfuscated bool) {
+		getTemplate := func(variant core.OvpnTemplateVariant, cachePath string) {
```
mariusSincovici: [small] What about to move this in a separate function so it can be tested at some point if it is working?

daemon/vpn/openvpn/config.go
```diff
@@ -51,20 +53,53 @@ type ovpnConfigData struct {
 	Identifier string
 }
 
+// ovpnDCONetlinkFamily is the netlink family this compiled OpenVPN uses.
+// Track ci/openvpn/env.sh OPENVPN_VERSION: 2.6.x wants "ovpn-dco-v2", 2.7.x wants "ovpn".
+const ovpnDCONetlinkFamily = "ovpn-dco-v2"
+
+// dcoKernelSupported reports whether the ovpn-dco kernel module is available.
+func dcoKernelSupported() bool {
+	if _, err := netlink.GenlFamilyGet(ovpnDCONetlinkFamily); err != nil {
```
mariusSincovici: I don't think this is needed. It seams to already be checked by the openvpn, also in v 2.6 https://github.com/OpenVPN/openvpn/blob/c9b790f5b9e8ebca5da38c22f479c31bb8d33686/src/openvpn/dco_linux.c#L86

mariusSincovici (approved):

edvyno1 (approved):

daemon/job_templates.go
thegravi: Done

PR #1715, opened 2026-08-20
LVPN-10998: introduce a way to configure telio log level

bartoszWojciechO:
Log level is read from /run/nordvpn/teliologlevel(in parity with regular loglevel file) and provided as an argument to SetGlobalLogger. Default log level is info.

Because telio logger can be configured only once per telio instance, the loglevel file is checked only upon telio init. Subsequent log level changes require a daemon restart.


edvyno1 (approved):

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -294,10 +308,47 @@ func handleTelioConfig(eventPath string, prod bool, vpnLibCfg vpn.LibConfigGette
...
+		return teliogo.TelioLogLevelError
+	case "warn":
+		return teliogo.TelioLogLevelWarning
+	case "info":
+		return teliogo.TelioLogLevelInfo
+	case "debug":
+		return teliogo.TelioLogLevelDebug
+	case "trace":
+		return teliogo.TelioLogLevelTrace
+	default:
```
thegravi, line 345: Wasn't there `off` and `fatal` also? Or we just keep those here "unsupported"?

README.md
```diff
@@ -101,6 +101,26 @@ echo "debug" | sudo tee /run/nordvpn/loglevel
...
+```sh
+echo "debug" | sudo tee /run/nordvpn/teliologlevel
+```
+
+Valid values are `debug`, `info`, `warn`, `error` and `trace`.
+
+libtelio log level is capped by the global log level. E.g. lib level logs will be printed only if global level
+logs are equal or higher. Global `debug` log level matches both `debug` and `trace` for the lib log level.
+
+Unlike the global log level, the changes require a deamon restart to take effect.
```
thegravi:
```suggestion
Unlike the global log level, the changes require a daemon restart to take effect.
```

daemon/vpn/nordlynx/libtelio/libtelio.go
thegravi: Shouldn't this one also be masked? It would dump `PublicKey`

daemon/vpn/nordlynx/libtelio/libtelio.go
```diff
@@ -294,10 +308,47 @@ func handleTelioConfig(eventPath string, prod bool, vpnLibCfg vpn.LibConfigGette
...
+	root, err := os.OpenRoot(filepath.Dir(internal.TelioLogLevelFile))
+	if err != nil {
+		log.Error("failed to open runtime dir:", err)
+		return teliogo.TelioLogLevelInfo
+	}
+	defer root.Close()
+
+	raw, err := root.ReadFile(filepath.Base(internal.TelioLogLevelFile))
+	if err != nil {
+		log.Error("failed to read log level file:", err)
```
thegravi: Since this will be created only by user (if I'm correct), then with each call to this function, we'd get an error for file not existing, until a user makes one.
```suggestion
	    if !errors.Is(err, fs.ErrNotExist) {
            log.Warn("failed to read telio log level file:", err)
        }

```

daemon/vpn/nordlynx/libtelio/libtelio.go
bartoszWojciechO: Unsupported because telio doesn't have those log levels.

thegravi (approved):

PR #1720, opened 2026-08-21
LVPN-10883: UI show analytics for session limit reached popup

thegravi:
Add UI analytics events for the “Session Limit Reached” popup to track when it is displayed.


gui/test/widgets/popups_listener_test.dart
```diff
@@ -0,0 +1,229 @@
...
+      await dismissViaBarrier(tester);
+
+      expect(
+        tester.findPopupWithId(DaemonStatusCode.connectionLimitReached),
+        findsOneWidget,
+      );
+      expect(uiEvents.shownCount, 1);
+    });
+
+    testWidgets('tapping the help link fires one click event and launches', (
```
devzbysiu: This should not be here - it's not related to `PopupsListener`, but to specific popup and `RichTextMarkdownLinks` used in it.
We already have integration test verifying connection limit reached popup, so maybe that's a good place for such specific `Open help guide` test.

gui/test/widgets/popups_listener_test.dart
devzbysiu, line 1: `PopupsListener` is not a regular widget and you can see this in the setup code - there is more setup code than test code. 
I think integration test is a better place for this. Also, we already have code there for mocking grpc, that also suggests that this should be integration test

gui/lib/widgets/rich_text_markdown_links.dart
```diff
@@ -9,8 +9,16 @@ import 'package:url_launcher/url_launcher.dart';
 class RichTextMarkdownLinks extends StatefulWidget {
   final String text;
   final TextStyle? style;
+  // Invoke on clicked link before it is launched.
+  // The URL is launched regardless of this callback.
+  final VoidCallback? onLinkTap;
```
devzbysiu: Maybe this should accept an index of the URL that the callback is associated with. Or array of callbacks each per URL. The reason for this is that with current design 1) all URL clicks would do the same (not a big deal for current state) and 2) (and this is more important) there is no connection between the text and the link callback so it's easy to miss - I can change the copy to have two URLs and we won't notice that second URL calls the same callback.

If we add index/array/something else and call the callback only for specific URL, then we have also an option to detect that number of URLs vs number of callbacks is off. This is still only-runtime check, but it's better than nothing. Having compile time check would be preferable, but it requires more changes

gui/lib/widgets/popups_listener.dart
```diff
@@ -52,13 +53,19 @@ final class _PopupsListenerState extends ConsumerState<PopupsListener> {
...
+    }
+
     final ctx = goRouterKey.currentContext;
     if (ctx == null) {
       logger.e("Can't display popup. Context is null.");
       return;
     }
 
     _visiblePopup = metadata.id;
+    _reportPopupShown(metadata);
```
devzbysiu: Maybe we could have a callback on the popup, something like `onShown` that allows to specify what should happen after it's built. This way we would have the action closer to popup definition, which makes it easier to find and makes it more natural considering that only one popups has it now, and it's also directly connected to actual popup rendering instead of indirectly via `PopupListener`

daemon/events/events.go
```diff
@@ -65,6 +66,7 @@ func NewEvents(
 	disconnect events.PublishSubcriber[events.DataDisconnect],
 	accountCheck events.PublishSubcriber[any],
 	uiItemsClick events.PublishSubcriber[events.UiItemsAction],
+	uiItemsShow events.PublishSubcriber[events.UiItemsAction],
```
devzbysiu, line 69: [small] maybe rename would be useful here to have something like `ShowUIITems` and `ClickUIITems` or something similar, distinct enough to avoid typos

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -19,6 +35,51 @@ void runEnsTests() async {
...
+      final mainScreen = await app.goToVpnScreen();
+      app.vpnStatus.connectingErrorStatusCode =
+          DaemonStatusCode.connectionLimitReached;
+
+      await mainScreen.quickConnect();
+      await tester.pumpUntilFound(
+        tester.findPopupWithId(DaemonStatusCode.connectionLimitReached),
+      );
+
+      await tester.tapOnText(find.textRange.ofSubstring("Open help guide"));
```
devzbysiu: This finder should be `finders.dart` or even in `vpn_screen_handle.dart` (although, popup can be also triggered from auto-connect settings page, so maybe `finders.dart` is more precise). 

Also it would be good to find it by ID, so the code is agnostic of the copy and UX changes as long as opening help guide is a functionality and has the same ID

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -19,6 +35,51 @@ void runEnsTests() async {
...
+      await tester.pumpAndSettle();
```
devzbysiu: We have `appCtl.waitForUiUpdats()` which handles timeout on refresh. 
It may seem redundant to have such methods, but the reasoning behind this kind of API was to have flutter-agnostic API with high-level methods that make it easy to write tests without having flutter knowledge

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -19,6 +35,51 @@ void runEnsTests() async {
...
+      ]);
+      expect(
+        urlLauncher.launchedUrls.single,
+        startsWith("https://support.nordvpn.com/"),
+      );
     });
   });
 }
+
+Future<void> _waitForPopupEvents(AppCtl app, List<UIEvent> expected) async {
```
devzbysiu: [small] maybe this could be part of `AppCtl` API?

gui/integration_test/e2e_tests/ens_test.dart
```diff
@@ -19,6 +35,51 @@ void runEnsTests() async {
...
+
+      // displaying the popup reports one show ui event
+      await _waitForPopupEvents(app, [_sessionLimitShown]);
+    });
+
+    testWidgets("help guide link reports click and launches the URL", (
+      tester,
+    ) async {
+      final urlLauncher = MockUrlLauncher();
+      UrlLauncherPlatform.instance = urlLauncher;
```
devzbysiu: Maybe this could also be part of `AppCtl` 🤔, and then something like 
```dart
expect(appCtl.launchedUrls().single, startsWith("https://support.nordvpn.com/"));
```

devzbysiu (approved):

edvyno1 (approved):

devzbysiu (approved):

PR #1723, opened 2026-08-21
LVPN-10661: fix wrong status after reconnect failure

keliramu:
LVPN-10661: fix wrong status after reconnect failure - in GUI have Nordlynx VPN connect to Double VPN then go to settings and try to switch protocol to Nordwhisper, dialog will ask to reconnect but proper server cannot be found, then stay connected to the same location and show GUI/Tray proper status.

### DIFF
--- a/daemon/rpc_connect.go
+++ b/daemon/rpc_connect.go
@@ -77,12 +77,24 @@ func (r *RPC) executeConnect(srv pb.Daemon_ConnectServer, fn func(context.Contex
 		return srv.Send(&pb.Payload{Type: internal.CodeNothingToDo})
 	}
 
-	// set connection status to "Disconnected"
+	// reconcile connection status after a failed attempt
 	if didFail || err != nil {
-		r.events.Service.Disconnect.Publish(events.DataDisconnect{})
+		r.reconcileStatusAfterFailedConnect()
 	}
 
 	return err
+}
+
+// reconcileStatusAfterFailedConnect fixes up the reported connection status after a connection
+// attempt failed. If the attempt aborted before the existing tunnel was touched, the VPN is still
+// up (netw.IsVPNActive()==true) and we restore the previous status instead of falsely reporting a
+// disconnect. Only when there is genuinely no active tunnel do we publish a Disconnect.
+func (r *RPC) reconcileStatusAfterFailedConnect() {
+	if r.netw.IsVPNActive() {
+		r.connectionInfo.RestorePreviousStatus()
+		return
+	}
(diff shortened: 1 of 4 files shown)


daemon/rpc_connect.go
```diff
@@ -77,14 +77,26 @@ func (r *RPC) executeConnect(srv pb.Daemon_ConnectServer, fn func(context.Contex
 		return srv.Send(&pb.Payload{Type: internal.CodeNothingToDo})
 	}
 
-	// set connection status to "Disconnected"
+	// reconcile connection status after a failed attempt
 	if didFail || err != nil {
-		r.events.Service.Disconnect.Publish(events.DataDisconnect{})
+		r.reconcileStatusAfterFailedConnect()
```
mariusSincovici: At this point there is no protection against parallel connects. Can't this break the status of another connect? 🤔 
e.g. 
go1: The connect failed, restore is executed
go2: in the same time `didFail, err = fn(ctx)` is executed.

To prevent this, wouldn't it work to have the whole restore logic inside the `if` which is executed sync?
```
if !r.connectContext.TryExecuteWith(func(ctx context.Context) {
		didFail, err = fn(ctx)
	}) {
```

daemon/rpc_connect.go
```diff
@@ -77,14 +77,26 @@ func (r *RPC) executeConnect(srv pb.Daemon_ConnectServer, fn func(context.Contex
...
 
 	return err
 }
 
+// reconcileStatusAfterFailedConnect fixes up the reported connection status after a connection
+// attempt failed. If the attempt aborted before the existing tunnel was touched, the VPN is still
+// up (netw.IsVPNActive()==true) and we restore the previous status instead of falsely reporting a
+// disconnect. Only when there is genuinely no active tunnel do we publish a Disconnect.
+func (r *RPC) reconcileStatusAfterFailedConnect() {
+	if r.netw.IsVPNActive() {
```
mariusSincovici, line 91: What if the VPN gets disconnects between `r.netw.IsVPNActive()` and `r.connectionInfo.RestorePreviousStatus()`.
Wouldn't this break the VPN status, because it would show connected instead of disconnected?

daemon/rpc_connect.go
keliramu: Good catch, thx, addressed

daemon/rpc_connect.go
keliramu: yes, it's quite complex to solve, let's do it in LVPN-11027

mariusSincovici (approved):

dfetti (approved):

PR #1725, opened 2026-08-21
LVPN-11019: Add analytics for out of tree DCO usage

thegravi:
Added analytics to track out-of-tree OpenVPN DCO usage through a debugger event.

### DIFF
--- a/daemon/vpn/openvpn/analytics.go
+++ b/daemon/vpn/openvpn/analytics.go
@@ -2,6 +2,7 @@ package openvpn
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


bartoszWojciechO (approved):

mariusSincovici (approved):

daemon/vpn/openvpn/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+func (a *DCOAnalytics) NotifyConnect(e events.DataConnect) error {
+	if e.EventStatus == events.StatusSuccess && e.Technology == config.Technology_OPENVPN {
+		a.publisher.Publish(*newDCOStatusEvent().ToDebuggerEvent())
+	}
+	return nil
+}
+
+// ToDebuggerEvent converts the event to a DebuggerEvent for publishing.
+func (e *dcoStatusEvent) ToDebuggerEvent() *events.DebuggerEvent {
+	// cannot fail: the struct holds only strings and bools
```
devzbysiu: I would still go with error handling as it doesn't cost us anything and when someone will expand the struct, no one will remember to double check if error handling should be updated now or not yet

daemon/vpn/openvpn/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+
+	event := &dcoStatusEvent{
+		Namespace: ovpnNamespace,
+		Subscope:  ovpnSubscope,
+		Event:     dcoStatusEventName,
+		DCOActive: dcoLinkKinds[linkKind],
+		LinkKind:  linkKind,
+	}
+
+	if internal.FileExists(dcoModuleSysfsDir) {
```
devzbysiu, line 82: Is the goal of `ModuleAvailable` to track both modules - in and out of tree? Because `/sys/module/ovpn_dco_v2` is not available on systems with in-tree module, so this will report that DCO is active, but module is not available, which is not precise

thegravi: With current setup we will be able to identify the version users are using for ovpn connection. Whether it's out-of-tree dco (usable), or in-tree dco (unusable), etc.

devzbysiu (approved):

PR #1727, opened 2026-08-24
LVPN-9987: Keep killswitch on during daemon restart

edvyno1:
This PR adds restart kill signal as SIGUSR1, which keeps killswitch on after stopping the daemon, for systemd and initd systems

### DIFF
--- a/contrib/systemd/system/nordvpnd.service
+++ b/contrib/systemd/system/nordvpnd.service
@@ -10,6 +10,7 @@ NonBlocking=true
 KillMode=process
 Restart=always
 RestartSec=5
+RestartKillSignal=SIGUSR1
 # centos7 RuntimeDirectory ignored
 RuntimeDirectory=nordvpn
 RuntimeDirectoryMode=0750
--- a/contrib/initd/nordvpn
+++ b/contrib/initd/nordvpn
@@ -60,7 +60,7 @@ stop() {
   return "$RETVAL"
 }
 
-upgrade-stop() {
+usr1-stop() {
   # Return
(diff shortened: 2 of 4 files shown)


devzbysiu (approved):

contrib/systemd/system/nordvpnd.service
```diff
@@ -10,6 +10,7 @@ NonBlocking=true
 KillMode=process
 Restart=always
 RestartSec=5
+RestartKillSignal=SIGUSR1
```
keliramu, line 13: just a note: RestartKillSignal= needs systemd ≥ 244, but does not break logic on older, also snap is different story.

keliramu (approved):

contrib/systemd/system/nordvpnd.service
edvyno1: yes mentioned in ticket, this is known and accepted

thegravi (approved):
Any e2e test possible to verify?

thegravi:
Also, I see we register these signals not at the beginning of `main`, shall we move it at to top?
```go
signals := internal.GetSignalChan() // main.go 801
```

**Update**: After @edvyno1 tests, moving signal handling to main's top could save us roughly 500ms during init, but calling two restart (systemctl) commands too quickly still results in a timeout and signal(s) stay unhandled.
Anything that takes too long still hits systemd’s 30s timeout and results in SIGKILL. Keeping the state as it is, which also prevents handling of rapid signal generation.

edvyno1:
e2e test added

PR #1729, opened 2026-08-24
LVPN-10961: Add review_snap.sh script

devzbysiu:
Changes:

- new `review_snap.sh` script that calls snap's `review-tools`
- magefiles updates to call this script so in case of issue with snap, it fails locally

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,8 +11,7 @@ Please follow the instructions in the following step for setting up the
 development environment. (Note: This process was tested on a virtual machine
 with Ubuntu 22.04):
 
-1. Install [Go 1.22](https://go.dev/doc/install). Note that newer versions will
-   not work if either `telio` or `drop` are included.
+1. Install [Go 1.26.3](https://go.dev/doc/install).
 1. Install [Mage](https://github.com/magefile/mage#installation). Even though
    currently building using bash scripts directly works well, mage is the
    recommended way to go.
--- a/BUILD.md
+++ b/BUILD.md
@@ -21,7 +21,9 @@ with Ubuntu 22.04):
 ### Docker builds
 
 Mage targets for building with docker are named `build:xxxDocker`. The only
-extra dependency for docker builds is [Docker](https://docs.docker.com/engine/install/ubuntu/).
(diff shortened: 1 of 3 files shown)


ci/review_snap.sh
```diff
@@ -0,0 +1,22 @@
...
+
+mapfile -t FILES < <(find "${WORKDIR}/dist/app/snap/" -type f -name "*_${ARCH}.snap")
+
+if [[ ${#FILES[@]} -ne 1 ]]; then
+  echo "ERROR: expected exactly 1 snap file, found ${#FILES[@]}: ${FILES[*]}" >&2
+  exit 1
+fi
+
+echo "review snap package: ${FILES[0]}"
+review-tools.snap-review "${FILES[0]}"
```
keliramu, line 20: `ci/review_snap.sh: line 20: review-tools.snap-review: command not found`
need to update docker

devzbysiu: This is not running in container as it requires `snapd`. This runs on host after the package was build inside container

keliramu: But `mage build:snapDocker` should work, now it fails

devzbysiu: Did you install `review-tools`? 

I updated `BUILD.md` with info that `review-tools` are needed (two commits - one is just formatting, second is actual change)

BUILD.md
```diff
@@ -1,78 +1,106 @@
...
-1. Install [Mage](https://github.com/magefile/mage#installation). Even though currently buiding
-   using bash scripts directly works well, mage is the recommended way to go.
+
+## How to setup the development environment
+
+Please follow the instructions in the following step for setting up the
+development environment. (Note: This process was tested on a virtual machine
+with Ubuntu 22.04):
+
+1. Install [Go 1.22](https://go.dev/doc/install). Note that newer versions will
```
edvyno1: `Go 1.22` and the note are outdated looks like

edvyno1 (approved):

keliramu (approved):

BUILD.md
devzbysiu: Updated

PR #1737, opened 2026-08-26
LVPN-11020: Diagnostics improvements

devzbysiu:
Changes:

- `openVerifiedUserFile()` opens only files relative to user home dir root, which is opened based on UID, this function does not follow symlinks and double checks if the file UID matches the expected one
- `addUserFileToZip()` uses func above to add file to zip
- `addUserDirectoryToZip()` walks FS rooted at verified home user dir, adds only non-symlink files relative to the root 
- the funcs above are used for accessing cli and norduserd logs
- tests
- cleanup

### DIFF
--- a/daemon/rpc_diagnostics.go
+++ b/daemon/rpc_diagnostics.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"io"
@@ -64,11 +65,16 @@ func (r *RPC) CollectDiagnostics(
 		})
 	}
 
-	// createDiagnosticsZip uses os.CreateTemp under the hood, so we always
-	// get a fresh file with no chance of collision — no existence check, no
-	// TOCTOU race, no overwrite of an old report, even if two collections
-	// land in the same second.
-	zipFile, err := createDiagnosticsZip(caller.outputDir)
+	outputRoot, err := os.OpenRoot(caller.outputDir)
+	if err != nil {
+		log.Diagnostics.Error("failed to open root output dir:", err)
+		return srv.Send(&pb.DiagnosticsProgress{
+			ErrorCode: pb.DiagnosticsErrorCode_DIAGNOSTICS_ERROR_CODE_FAILED_TO_CREATE_ZIP,
+		})
+	}
+	defer outputRoot.Close()
+
+	zipFile, err := createDiagnosticsFile(outputRoot)
 	if err != nil {
 		log.Diagnostics.Error("failed to create diagnostics zip:", err)
 		return srv.Send(&pb.DiagnosticsProgress{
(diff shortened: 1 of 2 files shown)


daemon/rpc_diagnostics.go
```diff
@@ -737,6 +751,90 @@ func addFileToZip(zipWriter *zip.Writer, filePath, zipPath string) error {
...
+		_ = file.Close()
+		return nil, err
+	}
+
+	if !info.Mode().IsRegular() {
+		_ = file.Close()
+		return nil, fmt.Errorf("%s: not a regular file", relPath)
+	}
+
+	stat, ok := info.Sys().(*syscall.Stat_t)
```
mariusSincovici: [small] It might be similar to this:
```
var st unix.Stat_t
	if err := unix.Fstat(int(file.Fd()), &st); err != nil {
		return err
	}
```

daemon/rpc_diagnostics.go
```diff
@@ -737,6 +751,90 @@ func addFileToZip(zipWriter *zip.Writer, filePath, zipPath string) error {
...
+			return nil
+		}
+
+		if d.Type()&fs.ModeSymlink != 0 {
+			return nil
+		}
+
+		relPath, err := filepath.Rel(relDir, path)
+		if err != nil {
+			return err
```
mariusSincovici: [small] Returning error fully stops the whole `WalkDir`. Maybe it is not needed for all cases to return the error? 🤔

mariusSincovici (approved):

daemon/rpc_diagnostics.go
```diff
@@ -737,6 +751,90 @@ func addFileToZip(zipWriter *zip.Writer, filePath, zipPath string) error {
...
+		}
+
+		zipPath := filepath.Join(zipPrefix, relPath)
+
+		file, err := openVerifiedUserFile(homeRoot, path, expectedUID)
+		if err != nil {
+			return err
```
keliramu: Maybe would be better/more robust not to fail-out entirely on some bad file, but just log the problem, and continue with other files

daemon/rpc_diagnostics.go
```diff
@@ -737,6 +751,90 @@ func addFileToZip(zipWriter *zip.Writer, filePath, zipPath string) error {
 	return err
 }
 
+func openVerifiedUserFile(homeRoot *os.Root, relPath string, expectedUID uint32) (*os.File, error) {
+	file, err := homeRoot.OpenFile(relPath, os.O_RDONLY|syscall.O_NOFOLLOW, 0)
```
keliramu: Additional flag `syscall.O_NONBLOCK` would be beneficial here (for pipe as a file case when open blocks until first write - it's not usual case, but would be complete protection)

daemon/rpc_diagnostics.go
devzbysiu: Done

daemon/rpc_diagnostics.go
devzbysiu: Done, added also logs and tests for those cases

daemon/rpc_diagnostics.go
devzbysiu: Done, added also logs and tests for those cases

daemon/rpc_diagnostics.go
devzbysiu: Done, added also test for this case

keliramu (approved):

PR #1748, opened 2026-08-31
LVPN-10980: add rule to block non vpn source packets

edvyno1:
Addresses an issue with some protocols where old living connections established before the VPN tunnel is up can send packets to the server using wrong source address.

### DIFF
--- a/daemon/firewall/service.go
+++ b/daemon/firewall/service.go
@@ -42,7 +42,8 @@ func NewConfig(opts ...Option) Config {
 func (c *Config) HasSimilarMeshInfo(cfg *Config) bool {
 	return c.MeshnetInfo != nil &&
 		cfg.MeshnetInfo != nil &&
-		c.MeshnetInfo.IsSimilar(cfg.MeshnetInfo)
+		c.MeshnetInfo.IsSimilar(cfg.MeshnetInfo) &&
+		c.TunnelIP == cfg.TunnelIP
 }
 
 type MeshInfo struct {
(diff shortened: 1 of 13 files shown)


daemon/firewall/service.go
```diff
@@ -124,6 +126,12 @@ func WithTunnelInterface(tunnelInterface string) Option {
 	}
 }
 
+func WithTunnelIP(tunnelIP netip.Addr) Option {
```
mariusSincovici, line 131: Since the IP and name need to be set in the same time, what about to change the `WithTunnelInterface` and accept 2 params?

mariusSincovici (approved):

daemon/firewall/service.go
edvyno1: addressed

networker/networker.go
```diff
@@ -1040,11 +1032,21 @@ func (netw *Combined) setMesh(
...
+	// If nordlynx was used as vpnet, the interface will change IP, we refresh it here
+	var tunnelIP, ok = netip.Addr{}, false
+	if netw.isVpnSet {
+		tunnelIP, ok = netw.vpnet.Tun().IP()
+	}
 	// configure firewall
 	newCfg := netw.fwConfig.CopyWith(
 		firewall.WithMeshnetInfo(firewall.NewMeshInfo(netw.cfg, netw.mesh.Tun().Interface().Name)),
 	)
+	if ok {
```
mariusSincovici: [small] Wouldn't work?
```
if netw.isVpnSet {
    if tunnelIP, ok := netw.vpnet.Tun().IP(); ok {
           newCfg = newCfg.CopyWith(
			firewall.WithTunnelIP(tunnelIP),
		)
}
}
```

mariusSincovici (approved):

dfetti (approved):

networker/networker.go
edvyno1: applied changes

PR #1750, opened 2026-09-01
LVPN-11033: Remove virtual location from CLI

dfetti:
This should remove everything related to virtual location from the CLI, except the set command.
This should also remove the virtual_server_enabled metric from the moose.

### DIFF
--- a/cli/cli_groups.go
+++ b/cli/cli_groups.go
@@ -24,17 +24,15 @@ func (c *cmd) Groups(ctx *cli.Context) error {
 		return formatError(fmt.Errorf(MsgListIsEmpty, "server groups"))
 	}
 
-	footer := footerForServerGroupsList(resp.Servers)
 	groupList, err := columns(
 		resp.Servers,
 		serverNameLen,
 		formatServerName,
-		footer,
 	)
 
 	if err != nil {
 		log.Error(err)
-		countries, _ := formatTable(resp.Servers, serverNameLen, formatServerName, 1, footer)
+		countries, _ := formatTable(resp.Servers, serverNameLen, formatServerName, 1)
 		fmt.Println(countries)
 	} else {
 		fmt.Println(groupList)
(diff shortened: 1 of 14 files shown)


cli/messages.go
```diff
@@ -364,7 +364,7 @@ Provide a [transfer_id] argument to list files in the specified transfer.`
 	MsgSetVirtualLocationDescription = "Enables or disables access to virtual locations."
 	MsgFooterVirtualLocationNote     = "* Virtual location servers"
```
thegravi: This `MsgFooterVirtualLocationNote` also needs to go

cli/cli_groups.go
```diff
@@ -24,17 +24,16 @@ func (c *cmd) Groups(ctx *cli.Context) error {
 		return formatError(fmt.Errorf(MsgListIsEmpty, "server groups"))
 	}
 
-	footer := footerForServerGroupsList(resp.Servers)
 	groupList, err := columns(
```
thegravi, line 27: Can we drop this `footer` from `columns()` and `formatTable` functions now, since everywhere I checked `""` is passed?

cli/cli_status.go
```diff
@@ -60,9 +60,6 @@ func Status(resp *pb.StatusResponse) string {
 
 	if resp.Name != "" {
 		serverName := resp.Name
-		if resp.VirtualLocation {
-			serverName += " - Virtual"
-		}
 		b.WriteString(fmt.Sprintf("Server: %s\n", serverName))
```
thegravi:
```suggestion
        b.WriteString(fmt.Sprintf("Server: %s\n", resp.Name))
```

thegravi:
Some of E2E tests are failing because they were based on help texts.

DoYouEvenCpp (approved):

thegravi (approved):

PR #1756, opened 2026-09-02
LVPN-11036: Remove virtual label from GUI

dfetti:
This removes the virtual location from GUI. Will require some changes after the protobuf cleanup but they will come in a later PR.


gui/lib/data/mocks/daemon/mock_application_settings.dart
```diff
@@ -84,7 +83,7 @@ final class MockApplicationSettings extends CancelableDelayed {
       fwmark: fwmark ?? val.fwmark,
       technology: technology ?? val.technology,
       protocol: protocol ?? val.protocol,
-      virtualLocation: virtualLocation ?? val.virtualLocation,
+      virtualLocation: true,
```
DoYouEvenCpp, line 86: Do we need to keep this param?

gui/lib/data/mocks/daemon/mock_daemon.dart
```diff
@@ -306,7 +306,7 @@ final class MockDaemon extends DaemonServiceBase {
     ServiceCall call,
     SetGenericRequest request,
   ) {
-    return appSettings.setSettings(virtualLocation: request.enabled);
+    throw UnimplementedError();
```
DoYouEvenCpp, line 309: Can't we get rid of it completely?

DoYouEvenCpp (approved):

gui/lib/data/mocks/daemon/mock_daemon.dart
dfetti: All this will be cleaned after the virtual removal from protobuf

keliramu (approved):

PR #1760, opened 2026-09-03
LVPN-11078: Remove deprecated aliases from the "allowlist" CLI command

thegravi:
- Removing the deprecated `whitelist` alias from the `allowlist` CLI command. The command no longer has any aliases.
- Adding migration of existing user `allowlist` settings on the daemon’s first startup.


config/config.go
```diff
@@ -86,8 +86,11 @@ type AutoConnectData struct {
 	ThreatProtectionLite bool      `json:"cybersec,omitempty"`
 	Obfuscate            bool      `json:"obfuscate,omitempty"`
 	DNS                  DNS       `json:"dns,omitempty"`
-	Allowlist            Allowlist `json:"whitelist,omitempty"`
-	PostquantumVpn       bool      `json:"postquantum_vpn"`
+	Allowlist            Allowlist `json:"allowlist,omitempty"`
+	// LegacyAllowlist is the allowlist saved by versions before 6.0.0, under the
+	// "whitelist" key. Exists for the migration purposes only, and can be removed in the future.
+	LegacyAllowlist *Allowlist `json:"whitelist,omitempty"`
```
mariusSincovici, line 92: Would it be possible to not have this here, because this will continue to be saved into the file until the member is deleted?

mariusSincovici (approved):

dfetti (approved):

PR #1762, opened 2026-09-04
LVPN-10896: NW reported as obfuscated

thegravi:
When protocol is set to NordWhisper simple and auto connections are presented as `Obfuscated` in the GUI and no more responds to `Obfuscate=on/off` settings

### DIFF
--- a/gui/test/data/providers/vpn_status_controller_test.dart
+++ b/gui/test/data/providers/vpn_status_controller_test.dart
@@ -1,0 +1,118 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:nordvpn/data/models/vpn_protocol.dart';
+import 'package:nordvpn/data/providers/app_state_provider.dart';
+import 'package:nordvpn/data/providers/vpn_status_controller.dart';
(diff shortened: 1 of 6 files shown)


gui/test/data/providers/vpn_status_controller_test.dart
```diff
@@ -0,0 +1,118 @@
...
+        vpnRepositoryProvider.overrideWithValue(_FakeVpnRepository(nordLynx)),
+        appStateProvider.overrideWithValue(_FakeAppStateChange()),
+      ],
+    );
+    addTearDown(container.dispose);
+    await container.read(vpnStatusControllerProvider.future);
+    return container;
+  }
+
+  test("a status change refreshes the protocol", () async {
```
mariusSincovici: Aren't this tests already covered by the E2E tests? Because if status change is not working the GUI is not working, right?

mariusSincovici (approved):

DoYouEvenCpp (approved):

gui/test/data/providers/vpn_status_controller_test.dart
thegravi: Yes, removed.

PR #1766, opened 2026-09-07
LVPN-10255: dont destroy token with normal logout

edvyno1:
- Removed `persist-token` flag, added `revoke-token` flag
- Changed default logout behavior, it now should be as follows: 
  - short term lived tokens are always destroyed(tokens created via regular login oauth flow) 
  - long term lived tokens are kept by default, destroyed only on logout with `nordvpn logout --revoke-token` (CLI exclusive),
- Unified logout behavior in `nordvpn logout` and `nordvpn set defaults --logout`
- No longer using the `Logout` api endpoint for logouts with manual token as it did not work nor was intended to work with them
- Added CLI messages to inform about successful revokation of token when using the `revoke-token` flag, as well as possibility to revoke token when doing normal logout


tray/actions.go
```diff
@@ -213,7 +213,7 @@ func (ti *Instance) watchGUIInstallation(ctx context.Context) {
 	}
 }
 
-func (ti *Instance) logout(persistToken bool) bool {
+func (ti *Instance) logout(revokeToken bool) bool {
```
mariusSincovici: [small] this param is always false, maybe it should not be passed as param anymore so it is not set by accident to `true`.

daemon/access/access_logout.go
```diff
@@ -106,8 +110,11 @@ func Logout(input LogoutInput) (logoutResult LogoutResult) {
 	if !input.NcClient.Revoke() {
 		log.Warn("error revoking NC token")
 	}
-
-	if !input.PersistToken {
+	// LVPN-11093 This needs to be validated properly once API endpoint exists
+	isLoggedInWithManualToken := tokenData.TokenExpiry == session.ManualAccessTokenExpiryDateString
+	var isRevokeSuccessful = false
+	// on logout we destroy token if we did not log in using `nordvpn login --token` or if revoke token flag was used
+	if !isLoggedInWithManualToken || input.RevokeToken {
```
mariusSincovici: isLoggedInWithManualToken = 1
input.RevokeToken = 1

daemon/access/access_logout.go
```diff
@@ -146,8 +154,12 @@ func Logout(input LogoutInput) (logoutResult LogoutResult) {
...
 	input.DebugPublisherFunc("user logged out")
 
-	if !input.PersistToken && tokenData.RenewToken == "" {
-		return LogoutResult{Status: internal.CodeTokenInvalidated, Err: nil}
+	if !input.RevokeToken && tokenData.RenewToken == "" {
+		return LogoutResult{Status: internal.CodeTokenStillValid, Err: nil}
+	}
+	// Only show special message if user logged in with token and logged out with revoke token
+	if input.RevokeToken && isRevokeSuccessful && isLoggedInWithManualToken {
+		return LogoutResult{Status: internal.CodeRevokedAccessToken, Err: nil}
```
mariusSincovici: Isn't `CodeRevokedAccessToken` meant for error, but in this case is a success?

daemon/rpc_set_defaults.go
```diff
@@ -34,12 +35,38 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
...
+			UserLogoutEventPublisherFunc: r.events.User.Logout.Publish,
+			DebugPublisherFunc:           r.publisher.Publish,
+			DisconnectFunc:               r.DoDisconnect,
+			DeviceKeyInvalidator:         r.dedicatedServerKeyManager,
+		})
+
+		if result.Err != nil {
+			switch result.Err {
+			case internal.ErrNotLoggedIn:
+				log.Info("trying to log out with set defaults, user already logged out")
```
mariusSincovici, line 53: Shouldn't there be a return in here? Otherwise `switch result.Status {` uses invalid value for `Status`.
Or the `switch` to be on else?

internal/codes.go
```diff
@@ -14,11 +14,11 @@ const (
...
-	CodeNothingToDo      int64 = 2000
-	CodeVPNRunning       int64 = 2002
-	CodeVPNNotRunning    int64 = 2003
-	CodeUFWDisabled      int64 = 2004
-	CodeTokenInvalidated int64 = 2005
+	CodeNothingToDo     int64 = 2000
+	CodeVPNRunning      int64 = 2002
+	CodeVPNNotRunning   int64 = 2003
+	CodeUFWDisabled     int64 = 2004
+	CodeTokenStillValid int64 = 2005
```
mariusSincovici: [small] Shouldn't this be in the 1000 category, because it is success with something extra? 🤔

gui/lib/data/repository/account_repository.dart
```diff
@@ -44,7 +44,7 @@ final class AccountRepository {
       formReference: UIEvent_FormReference.GUI,
       itemName: UIEvent_ItemName.LOGOUT,
     );
-    final result = await _client.logout(LogoutRequest(persistToken: false));
+    final result = await _client.logout(LogoutRequest(revokeToken: false));
```
mariusSincovici, line 47: [small for now] For manual tokens this returns 2005, which generates a popup, but since the entire app moves to a different state also the popup is dropped.
It might be good if the result is dropped. In case the code is moved to 1000 category, then it is automatically ignored.

tray/actions.go
edvyno1: removed arg, passing false to logout request

daemon/access/access_logout.go
edvyno1: introduced new code `CodeTokenRevokedSuccessfully` as a 1000 group code

daemon/rpc_set_defaults.go
edvyno1: addressed by manually setting status as success when hitting this branch

internal/codes.go
edvyno1: moved to 1000 category

gui/lib/data/repository/account_repository.dart
edvyno1: should be returning a 1000 code now with the new code changes

mariusSincovici (approved):

daemon/rpc_set_defaults.go
```diff
@@ -34,12 +35,39 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
...
+			NcClient:                     r.ncClient,
+			ConfigManager:                r.cm,
+			UserLogoutEventPublisherFunc: r.events.User.Logout.Publish,
+			DebugPublisherFunc:           r.publisher.Publish,
+			DisconnectFunc:               r.DoDisconnect,
+			DeviceKeyInvalidator:         r.dedicatedServerKeyManager,
+		})
+
+		if result.Err != nil {
+			switch result.Err {
```
thegravi: [small] We can switch on `nil` also. Would be less nesting

daemon/rpc_set_defaults.go
```diff
@@ -34,12 +35,39 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
...
-		if !r.ncClient.Revoke() {
-			log.Warn("error revoking notification center token")
+		switch result.Status {
+		case internal.CodeSuccess, internal.CodeTokenStillValid:
+			log.Info("set defaults logout successful")
+		default:
+			log.Error("logout returned non success return code", result.Status)
+			return &pb.Payload{
+				Type: result.Status,
+			}, nil
```
thegravi, line 70: Given this happens, then we do not reset anything else, right? Before we did no such thing, and _mainly_ proceed to another step with just logging an error. Basically now we cannot reset things if we're offline.

daemon/access/access_logout.go
```diff
@@ -119,15 +126,16 @@ func Logout(input LogoutInput) (logoutResult LogoutResult) {
...
+		isRevokeSuccessful = true
+	}
+	// Logout endpoint does not work with manual token created via UCP, always returns a 404 after deletion
+	if !isLoggedInWithManualToken {
 		if err := input.CredentialsAPI.Logout(); err != nil {
 			log.Error("logging out:", err)
 			switch {
 			// This means that token is invalid anyway
 			case errors.Is(err, core.ErrUnauthorized):
 			case errors.Is(err, core.ErrBadRequest):
```
thegravi, line 139: This is not a supported return code from API

daemon/access/access_logout.go
thegravi, line 122: When we receive these, we still set `isRevokeSuccessful = true` which is not correct.

cli/cli_logout.go
```diff
@@ -24,10 +23,10 @@ func (c *cmd) Logout(ctx *cli.Context) error {
 		ItemType:      pb.UIEvent_CLICK,
 	})
 
-	persistToken := ctx.IsSet(flagPersistToken)
+	revokeToken := ctx.IsSet(flagRevokeToken)
```
thegravi: What happens if you use it like this: `--revoke-token=false`? :)

daemon/rpc_set_defaults.go
```diff
@@ -34,12 +35,39 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
 	}
 
 	if !in.NoLogout {
-		if err := r.ncClient.Stop(); err != nil {
-			log.Warn("error stoping notification center client:", err)
+		result := access.Logout(access.LogoutInput{
```
thegravi, line 38: Previously NC was stopped/revoked despite "already logged out case" now it's under that switch, so in case logout is performed elsewhere so BE sees it, here we would not stop NC.

daemon/rpc_set_defaults.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
thegravi, line 4: Can you add UT for this new case of inverted flag for logout using set-defaults?

protobuf/daemon/logout.proto
```diff
@@ -5,5 +5,5 @@ package pb;
 option go_package = "github.com/NordSecurity/nordvpn-linux/daemon/pb";
 
 message LogoutRequest {
-  bool persist_token = 1;
+    bool revoke_token = 1;
```
thegravi: Is it safe to reuse this? E.g. case where daemon is updated, but GUI has older version. Why not going with 2nd slot?

daemon/rpc_set_defaults.go
```diff
@@ -34,12 +35,39 @@ func (r *RPC) SetDefaults(ctx context.Context, in *pb.SetDefaultsRequest) (*pb.P
 	}
```
thegravi, line 35: [nit] This (UnSetMesh) is "fatal" when called 2nd time from access.Logout(). Probably just another "strange" warning in logs.

daemon/rpc_set_defaults.go
thegravi: Also now we have 2 disconnects and 2 disconnect-event published when using logout on defaults. Does that make things messy?

daemon/rpc_set_defaults.go
edvyno1: this only runs if we run `nordvpn set defaults --logout`, doing normal `nordvpn set defaults` will not hit this code and will still reset. 
the errors returned in this mostly have things to do with failing to clear local config, api failures are handled in the status switch and will not return an err

daemon/access/access_logout.go
edvyno1: 400 will be returned with potential malformed token, might as well keep it

daemon/rpc_set_defaults.go
edvyno1: changed

daemon/access/access_logout.go
edvyno1: addressed this by moving `isRevokeSuccessful` to be set to true only in else branch

cli/cli_logout.go
edvyno1: addressed this and other few occurences of `IsSet` being just checked instead of taking in the value of the `BoolFlag`

protobuf/daemon/logout.proto
edvyno1: swapped to 2nd slot, regenerated the protobuf

daemon/rpc_set_defaults.go
edvyno1: do you think we should still stop NC then, regardless of `--logout` flag in set defaults?

looks like second disconnect event doesn't fire since we were disconnected with first, second has nothing to do and does an early return with no event

daemon/rpc_set_defaults.go
edvyno1: added new UT

daemon/rpc_set_defaults.go
thegravi: I mean for the `--logout` case, but if token is invalidated on BE and you hit "already logged out" branch, so nothing pass it will not be processed.

edvyno1: re added nc stopping

thegravi (approved):

dfetti (approved):

PR #1780, opened 2026-09-11
LVPN-10937: Implement GUI obfuscated group listing

thegravi:
Showing Obfuscated group in the GUI all the time. It's disabled/inactive only when technology is not `NordWhisper`.


daemon/rpc_servers.go
mariusSincovici, line 15: [small] Why is this not returning NordWhisper? 🤔 
Small because it is not used by the GUI.

daemon/rpc_servers.go
```diff
@@ -51,10 +51,31 @@ func groupFilter(groups core.Groups) []config.ServerGroup {
 		}
 	}
 
-	return desiredGroups
+	return withObfuscatedGroup(desiredGroups, technology)
```
mariusSincovici, line 54: Shouldn't this be based on the server supported technologies?

daemon/rpc_servers.go
thegravi: It's used by GUI when receiving a rpc, but actual value is not used. I can add it, but there would be no consumer.

daemon/rpc_servers.go
thegravi: This is a custom case for `NordWhisper`, because nearly every server supports it, and we need to include `Obfuscated groups` as "alias". This one only goes out to GUI client.

gui/test/vpn/servers_list_card_test.dart
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void main() {
+  setUpAll(() async {
+    SharedPreferencesAsyncPlatform.instance = FakeSharedPreferencesAsync();
+    await initServiceLocator();
+  });
+
+  group("obfuscated servers tile", () {
+    Future<Finder> openSpecialtyServers(
```
devzbysiu: It's hard to read what's the test and what's the helper function with this kind of nesting

gui/test/vpn/servers_list_card_test.dart
```diff
@@ -0,0 +1,134 @@
...
+
+    testWidgets("is offered under NordWhisper", (tester) async {
+      final tile = await openSpecialtyServers(
+        tester,
+        technology: Technology.NORDWHISPER,
+      );
+
+      expect(tester.widget<CustomListTile>(tile).enabled, isTrue);
+      expect(
+        find.descendant(of: tile, matching: find.text(t.ui.obfuscated)),
```
devzbysiu: Keys should be used if possible, because it's less brittle with keys

gui/test/vpn/servers_list_card_test.dart
devzbysiu: I think you can reuse existing integration tests to assert on what you are checking here, because we already have the API there to set up everything. Here you have a lot of setup code which is almost more lines than the tests itself

gui/test/vpn/servers_list_card_test.dart
devzbysiu: It's hard to read what's the test and what's the helper function with this kind of nesting. Is there any value in this?

gui/test/vpn/servers_list_card_test.dart
```diff
@@ -0,0 +1,134 @@
...
+  group("obfuscated servers tile", () {
+    Future<Finder> openSpecialtyServers(
+      WidgetTester tester, {
+      required Technology technology,
+    }) async {
+      final serversList = (await tester.runAsync(
+        () => _serversList(technology: technology),
+      ))!;
+
+      await tester.pumpWidget(
```
devzbysiu: We already have `setupWidgetTest` helper for widget tests

gui/test/vpn/servers_list_card_test.dart
devzbysiu: Keys should be used here and in other cases if possible, because it's less brittle with keys

devzbysiu (approved):

daemon/data_manager.go
```diff
@@ -370,8 +380,8 @@ func (dm *DataManager) Groups(
 				continue
 			}
 
-			// the current obfuscated group is not viable any more. Only the OVPN XOR servers carry
-			// it and they should not be connectable
+			// the obfuscated group is never taken from the server tags. Only the OVPN XOR servers
+			// carry that tag, and those are not connectable anymore.
 			if group.ID == config.ServerGroup_OBFUSCATED {
```
mariusSincovici: I think this is getting a bit confusing.
In recent connections it is used the `ServerGroup_OBFUSCATED`. Here is ignored, but later added and we have lots of comments to differentiate when is real obfuscated and when not.
What if a new group is included, e.g. `ServerGroup_NW` or `ServerGroup_OBFUSCATED_NW`..

daemon/data_manager.go
```diff
@@ -362,6 +368,10 @@ func (dm *DataManager) Groups(
 		}
 
 		for _, group := range server.Groups {
+			if group.ID == config.ServerGroup_STANDARD_VPN_SERVERS {
```
mariusSincovici: Why is this needed, to link the NordWhisper group with standard group? 🤔 
If NordWhisper tech is set, then the server is already connectable, `core.IsConnectableVia`, and obfuscated needs to be displayed.

gui/test/utils/test_helpers.dart
```diff
@@ -128,12 +134,45 @@ extension Helper on WidgetTester {
     return AppCtl(tester: this, urlLauncher: urlLauncher);
   }
 
+  // The servers list the mocked daemon reports for [technology], grouped the
+  // same way the app groups it.
+  Future<ServersList> mockedServersList({
```
mariusSincovici, line 139: Is this needed only to get the servers list?

mariusSincovici (approved):

gui/test/vpn/servers_list_card_test.dart
```diff
@@ -0,0 +1,117 @@
...
+  _connectRequests.add(args);
+}
+
+// Builds the servers list card, opens the specialty servers tab and returns the
+// obfuscated group tile (if listed there).
+Future<Finder> _obfuscatedTile(
+  WidgetTester tester, {
+  required Technology technology,
+}) async {
+  _connectRequests.clear();
```
mariusSincovici: This maybe should go in teardown or setup?

PR #1783, opened 2026-09-14
OSS build CI failure fix

keliramu:
Deduplicate the rustup target list which was causing parallel download of the same thing and clash/failure on renaming local files.


keliramu:
solved by PR #1778

PR #1793, opened 2026-09-15
LVPN-11077: Remove p2p from daemon

dfetti:
This also includes:
- rejection of p2p connect and autoconnect request
- Events cleanup
- Man cleanup


daemon/rpc_set_autoconnect.go
```diff
@@ -32,6 +33,11 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 		}, nil
 	}
 
+	gr := strings.ToLower(strings.TrimSpace(in.ServerGroup))
+	if gr == "p2p" {
+		return &pb.Payload{Type: internal.CodeP2PDeprecated}, nil
+	}
```
thegravi: Would it be possible to handle such that we would not need to explicitly check for `p2p` here?

protobuf/daemon/config/group.proto
```diff
@@ -13,9 +13,8 @@ enum ServerGroup {
 	DEDICATED_IP = 9;
 	STANDARD_VPN_SERVERS = 11;
 	NETFLIX_USA = 13;
-	P2P = 15;
 	OBFUSCATED = 17;
 	DEDICATED_SERVER = 99;
-	reserved 19, 21, 23, 25;
+	reserved 15, 19, 21, 23, 25;
 	reserved "EUROPE", "THE_AMERICAS", "ASIA_PACIFIC", "AFRICA_THE_MIDDLE_EAST_AND_INDIA";
```
thegravi: Missed "P2P'

protobuf/daemon/uievent.proto
```diff
@@ -56,7 +56,7 @@ message UIEvent {
     OBFUSCATED             = 5;
     ONION_OVER_VPN         = 6;
     DOUBLE_VPN             = 7;
-    P2P                    = 8;
+    reserved 8;
```
thegravi, line 59: I'd recommend also adding `reserved "P2P"` to prevent future usage of this naming.

contrib/manual/mantemplate
```diff
@@ -173,16 +173,16 @@ $ \fBnordvpn connect AU\fR
...
+$ \fBnordvpn connect Onion_Over_VPN\fR
 $ \fBnordvpn connect Dedicated_IP\fR
 .fi
 .RE
 .PP
 \fBExample \&6. Connect to a server in a specific group and country\fR
 .RS 4
 .nf
-$ \fBnordvpn connect --group P2P Germany\fR
+$ \fBnordvpn connect --group Double_VPN Germany\fR
```
mariusSincovici: There is no `Germany` in this group.

daemon/rpc_set_autoconnect.go
dfetti: Best I could do was to extract the logic in a IsP2PGroup function and reuse it in rpc_connect.

daemon/rpc_connect.go
```diff
@@ -275,6 +275,11 @@ func (r *RPC) connectWithParameters(ctx context.Context,
 	if err := r.cm.Load(&cfg); err != nil {
 		log.Error(err)
 	}
+
+	if serverpicker.IsP2PGroup(in.ServerTag, in.ServerGroup) {
```
devzbysiu: I think we can do that even before loading config. Now we load config when it may not be needed

daemon/rpc_connect.go
```diff
@@ -275,6 +275,11 @@ func (r *RPC) connectWithParameters(ctx context.Context,
...
+		return true, srv.Send(&pb.Payload{Type: internal.CodeP2PDeprecated})
```
devzbysiu: Should `didFail` be returned as `true` here? 🤔 This causes `r.reconcileStatusAfterFailedConnect()` to be called later, but there is nothing to reconcile, we didn't set connection state at this point

daemon/rpc_connect_test.go
devzbysiu, line 1: We should have test that validates `p2p` case

daemon/rpc_set_autoconnect.go
```diff
@@ -32,6 +32,10 @@ func (r *RPC) SetAutoConnect(ctx context.Context, in *pb.SetAutoconnectRequest)
 		}, nil
 	}
 
+	if serverpicker.IsP2PGroup(in.ServerTag, in.ServerGroup) {
+		return &pb.Payload{Type: internal.CodeP2PDeprecated}, nil
+	}
```
devzbysiu: This will cause such situation:
```bash
$ nordvpn set autoconnect on pl
Auto-connect has been successfully set to 'enabled'.

$ nordvpn set autoconnect off p2p
P2P group is no longer available. All servers now support P2P.
```

dfetti: Fixed this and added some tests for the case

daemon/rpc_connect.go
dfetti: You are right. I fixed it.

daemon/rpc_connect_test.go
dfetti: Added some tests in rpc_connect

daemon/rpc_connect.go
mariusSincovici: [small] If the cfg is not needed here, wouldn't it make sense to move this before config load?

mariusSincovici (approved):

daemon/jobs.go
```diff
@@ -384,6 +384,21 @@ func (r *RPC) doAutoConnect() error {
 		return errAutoConnectDisabled
 	}
 
+	// Migration of deprecated groups failed at init
+	if config.IsDeprecatedP2PGroup(cfg.AutoConnectData.Group) || serverpicker.IsP2PGroup(cfg.AutoConnectData.ServerTag, "") {
+		// Try to migrate again
+		err := MigrateDeprecatedGroupsAutoconnect(r.cm)
+		if err != nil {
+			log.Warn("failed to migrate deprecated groups in doAutoConnect")
```
devzbysiu, line 392: [small] maybe let's add `err` to log msg

devzbysiu (approved):

PR #1795, opened 2026-09-16
LVPN-11122: Fix Snap test failures in local VM

ggediminass:
- Mount directory using rsync, instead of nfs
- Copy daemon.log from VM to Host machine after tests are finished


magefiles/vagrant.go
```diff
@@ -188,6 +188,81 @@ func buildVagrantEnv() (VagrantEnv, error) {
...
+	}
+
+	files := strings.Split(strings.TrimSpace(string(output)), "\n")
+	// Skip the first line (Vagrant's "Using box:..." message)
+	if len(files) > 0 {
+		files = files[1:]
+	}
+	if len(files) == 0 || (len(files) == 1 && files[0] == "") {
+		return fmt.Errorf("no files found in %s", remotePath)
+	}
```
DoYouEvenCpp, line 215: I'd reverse the order here:
- at first verify if the len is 0, or is 1 and is empty
- and then you can simply `files = files[1:]` without the `if len(files) > 0`

magefiles/vagrant.go
```diff
@@ -188,6 +188,81 @@ func buildVagrantEnv() (VagrantEnv, error) {
...
+			fmt.Printf("Warning: failed to create directory %s: %v\n", localDir, err)
+			continue
+		}
+
+		// Write to local file
+		if err := os.WriteFile(localFile, fileContent, 0644); err != nil {
+			fmt.Printf("Warning: failed to write %s: %v\n", localFile, err)
+			continue
+		}
+
```
DoYouEvenCpp, line 259: You can reuse here `WriteFile` from `internal/filesystem.go` - it creates parent directory (if missing), and writes into a file.

magefiles/vagrant.go
```diff
@@ -188,6 +188,81 @@ func buildVagrantEnv() (VagrantEnv, error) {
...
+		// Calculate the local path, preserving directory structure
+		relPath := strings.TrimPrefix(remoteFile, remotePath)
+		relPath = strings.TrimPrefix(relPath, "/")
+		localFile := path.Join(localPath, relPath)
+
+		// Read file from VM
+		catCmd := exec.Command("vagrant", "ssh", vagrantEnv.BoxName, "-c", fmt.Sprintf("cat %s", remoteFile))
```
DoYouEvenCpp, line 232: Just to escape any possible quotes in the filename I'd say let's use %q here
```suggestion
		catCmd := exec.Command("vagrant", "ssh", vagrantEnv.BoxName, "-c", fmt.Sprintf("cat %q", remoteFile))
```

magefiles/vagrant.go
```diff
@@ -188,6 +188,81 @@ func buildVagrantEnv() (VagrantEnv, error) {
...
+		// Write to local file
+		if err := os.WriteFile(localFile, fileContent, 0644); err != nil {
+			fmt.Printf("Warning: failed to write %s: %v\n", localFile, err)
+			continue
+		}
+
+		fmt.Printf("  Copied %s -> %s (%d bytes)\n", remoteFile, localFile, len(fileContent))
+	}
+
+	return nil
```
DoYouEvenCpp, line 263: In case of any file fails, then we emit here merely a warning, and the final result is success. Shall we register the errors anyway and propagate them? For that we could leverage `errors.Join`

magefiles/vagrant.go
```diff
@@ -188,6 +188,81 @@ func buildVagrantEnv() (VagrantEnv, error) {
...
+func copyDirFromVM(vagrantEnv VagrantEnv, remotePath, localPath string) error {
+	// List all files in the remote directory
+	listCmd := exec.Command("vagrant", "ssh", vagrantEnv.BoxName, "-c", fmt.Sprintf("find %s -type f", remotePath))
+	listCmd.Dir = vagrantEnv.VagrantFileDir
+	listCmd.Env = append(os.Environ(),
+		fmt.Sprintf("WORKDIR=%s", vagrantEnv.Cwd),
+		fmt.Sprintf("SNAP_TEST_BOX_URL=%s", vagrantEnv.CustomBoxUrl),
+		fmt.Sprintf("SNAP_TEST_BOX_USER=%s", vagrantEnv.CustomBoxUser.getValue()),
+		fmt.Sprintf("SNAP_TEST_BOX_PASS=%s", vagrantEnv.CustomBoxPass.getValue()),
+	)
```
DoYouEvenCpp, line 201: So this env pre configuration now duplicates what's already in  `runVagrantCmd` - shall we extract this into small helper method and reuse?

bartoszWojciechO (approved):

DoYouEvenCpp (approved):

PR #1798, opened 2026-09-17
LVPN-11039: Migrate deprecated P2P group from recent connections

dfetti:
Migrates deprecated P2P group from recent connections.
There are 2 guards against this
1. In the MigrateDeprecatedP2PGroup where I remove the P2P group from every entry and resolve the duplicates
2. In the rpc_get_recent_connections where I either skip or fix the deprecated group depending on case

This also includes unittests for the migration function.


daemon/recents/recent_connections.go
```diff
@@ -167,6 +167,83 @@ func (r *RecentConnectionsStore) Clean() error {
 	return nil
 }
 
+func (r *RecentConnectionsStore) MigrateDeprecatedP2PGroup() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if err := r.checkExistence(); err != nil {
+		// If the file didn't exist, checkExistence will create a clean one
+		return nil
```
DoYouEvenCpp, line 176: Something's off here?

If `checkExistence` fails to run, we swallow the bug and return early with a no-op.

But I guess what we want here is to:
- check if file exists (if so, we carry-on)
- check if file doesn't exist:
  - if we failed to create one, we return an error
  - if we created one, then we return nil
Was that your intention? If so, then probably the we have to modify the signature a bit (to return `(bool, error)` maybe?)

cmd/daemon/main.go
```diff
@@ -615,6 +615,17 @@ func main() {
...
 
 	sharedContext := sharedctx.New()
+	recentConnections := recents.NewRecentConnectionsStore(
+		internal.RecentVPNConnectionsFilename,
+		&internal.StdFilesystemHandle{},
+		func() {
+			dataUpdateEvents.RecentsUpdate.Publish(events.DataRecentsChanged{})
+		},
+	)
+	if err = recentConnections.MigrateDeprecatedP2PGroup(); err != nil {
```
DoYouEvenCpp, line 625: [nit] Can we group all the `.Migrate*` calls all together? (probably you'd have to reorder the calls a bit)

cmd/daemon/main.go
```diff
@@ -615,6 +615,17 @@ func main() {
...
+		log.Error("failed to migrate deprecated P2P group from recent connections: %w", err)
```
DoYouEvenCpp:
```suggestion
		log.Error("failed to migrate deprecated P2P group from recent connections:", err)
```

daemon/recents/recent_connections.go
```diff
@@ -167,6 +167,83 @@ func (r *RecentConnectionsStore) Clean() error {
...
+
+	connections, err := r.load()
+	if err != nil {
+		log.Recents.Warnf("migrating deprecated P2P group from recent VPN connection: %s", err)
+		if saveErr := r.save([]Model{}); saveErr != nil {
+			return errors.Join(
+				fmt.Errorf("migrating deprecated P2P group from recent vpn connection: %w", err),
+				fmt.Errorf("recreating recent connections file: %w", saveErr))
+		}
+		return nil
```
DoYouEvenCpp, line 187: Again, do we really want to disregard the load-related errors here (if `save` happen to be all fine)?

daemon/recents/recent_connections.go
```diff
@@ -167,6 +167,83 @@ func (r *RecentConnectionsStore) Clean() error {
...
+
+	changed := false
+
+	// Remove the recent connection entries that have only group
+	// Clean entries that have group and country or specific server
+	firstStage := make([]Model, 0, len(connections))
+	for i := range connections {
+		if !config.IsDeprecatedP2PGroup(connections[i].Group) {
+			firstStage = append(firstStage, connections[i])
+			continue
```
DoYouEvenCpp, line 198: Assuming none of the recents is P2P - do we have to run the second stage then?

daemon/recents/recent_connections_test.go
```diff
@@ -1294,3 +1294,133 @@ func TestRecentConnectionsStore_EventPublisher_ConcurrentOperations(t *testing.T
 	assert.Greater(t, finalCount, 0, "Should have published events for successful operations")
 	assert.LessOrEqual(t, finalCount, operations, "Should not publish more events than operations")
 }
+
+func TestRecentConnectionsStore_Migrate_PreservesNonDeprecated(t *testing.T) {
```
DoYouEvenCpp, line 1298: Why not table driven tests here (and for the others)?

daemon/recents/recent_connections_test.go
```diff
@@ -1294,3 +1294,133 @@ func TestRecentConnectionsStore_EventPublisher_ConcurrentOperations(t *testing.T
...
+	category.Set(t, category.Unit)
+	fs := fs.NewSystemFileHandleMock(t)
+	store := NewRecentConnectionsStore("/test/path", &fs, nil)
+
+	france := Model{Country: "France", ConnectionType: config.ServerSelectionRule_COUNTRY}
```
DoYouEvenCpp, line 1303: Shall we assign `ServerGroup_UNDEFINED` for the expected group?

daemon/recents/recent_connections.go
```diff
@@ -167,6 +167,83 @@ func (r *RecentConnectionsStore) Clean() error {
...
+		if connections[i].ConnectionType == config.ServerSelectionRule_GROUP {
+			changed = true
+			continue
+		}
+
+		if connections[i].ConnectionType == config.ServerSelectionRule_COUNTRY_WITH_GROUP {
+			connections[i].ConnectionType = config.ServerSelectionRule_COUNTRY
+		}
+		if connections[i].ConnectionType == config.ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP {
+			connections[i].ConnectionType = config.ServerSelectionRule_SPECIFIC_SERVER
```
keliramu, line 210: How is this migration considered suitable, as `specificServer` is not set `ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP` (looking at `daemon/recent_connections_rpc_helper.go: buildRecentConnectionModel(...)` - that suggests me, that after migration `specificServer` is not set, so, migration is not correct?

keliramu: upd: looks like mentioned `buildRecentConnectionModel` has a bug - for `ServerSelectionRule_SPECIFIC_SERVER_WITH_GROUP` case need to set `specificServer` then migration logic would be correct

keliramu (approved):

daemon/recents/recent_connections.go
```diff
@@ -167,6 +167,93 @@ func (r *RecentConnectionsStore) Clean() error {
...
+		secondStage = firstStage
+	}
+
+	if !changed {
+		return nil
+	}
+
+	if err := r.save(secondStage); err != nil {
+		return fmt.Errorf("migrating deprecated P2P group from recent connections: %w", err)
+	}
```
DoYouEvenCpp: What do you say to change this algo in this way: if nothing changed then return. otherwise always dedup?
this way we can reduce the logic there, and get rid of some variables.

```suggestion
	if !changed {
              return nil
      }

      deduplicated := make([]Model, 0, len(firstStage))
      for _, c := range firstStage {
              if !slices.ContainsFunc(deduplicated, c.Equals) {
                      deduplicated = append(deduplicated, c)
              }
      }

      if err := r.save(deduplicated); err != nil {
              return fmt.Errorf("migrating deprecated P2P group from recent connections: %w", err)
      }
```

daemon/rpc_get_recent_connections.go
```diff
@@ -42,6 +42,13 @@ func (r *RPC) GetRecentConnections(
 			continue
 		}
 
+		if config.IsDeprecatedP2PGroup(v.Group) {
```
DoYouEvenCpp: Can you add here a comment explaining why we keep this safe-guard here?

DoYouEvenCpp (approved):

PR #1799, opened 2026-09-18
LVPN-10942: in tray make pause timer update every minute

keliramu:
LVPN-10942: in tray make pause timer update every minute: this is caused by issue on Linux Mint timer not being updated, now we do not care if tray menu open or not, we update tray menu item every minute (before it was every second).


edvyno1 (approved):

dfetti (approved):

tray/menu.go
```diff
@@ -338,7 +338,7 @@ func buildPauseMenu(ti *Instance) {
 }
 
 func buildPauseTimer(ti *Instance) {
```
thegravi: Each time the tray is re-rendered, this minute ticker is restarted. This can leave the displayed value stale compared to other daemon clients. Try re-render tray each 55s and you would get non-changing value until pause elapses.

PR #1802, opened 2026-09-22
LVPN-11010: remove autoconnect to a specific server

bartoszWojciechO:
Removes autoconnect to a specific server(i.e by <country_code><number>, such as pl206). In cases where autoconnect was already configured, the daemon will try to find the server and set autoconnect to it's country and city. If it's not possible, it will set autoconnect to the fastest server.


cli/cli_set_autoconnect.go
```diff
@@ -108,6 +108,8 @@ func (c *cmd) SetAutoConnect(ctx *cli.Context) error {
 		return errors.New(internal.ServerUnavailableErrorMessage)
 	case internal.CodeDedicatedServersServerNotSetUp:
 		return errors.New(c.injectLinkIntoMessage(client.DedicatedServersSetupURL, client.DedicatedServersSetupURLLogin, DedicatedServersNoServersAvailable))
+	case internal.CodeAutoconnectToSpecificServer:
+		return formatError(argsParseError(ctx))
```
edvyno1, line 112: maybe we should return some specific version that `autoconnect to specific server is not available`

cli/cli_set_autoconnect.go
edvyno1, line 32: remove this line

PR #1803, opened 2026-09-22
LVPN-11160: Disable DNSSEC for NordDNS servers

mariusSincovici:
When application is only using NordDNS servers, then the DNSSEC is disabled because systemd-resolve fails sometimes to detect correctly if is supported or not.


daemon/dns/dns_resolvectl.go
```diff
@@ -41,6 +41,12 @@ func setDNSWithResolvectl(iface string, addresses []string) error {
 	if out, err := exec.Command(execResolvectl, "domain", iface, "~.").CombinedOutput(); err != nil {
 		log.Warn("dns domain routing with resolvectl:", strings.TrimSpace(string(out)), "err:", err)
 	}
+	if areOnlyNordDNSServers(addresses) {
+		// #nosec G204 -- input is properly validated
+		if out, err := exec.Command(execResolvectl, "dnssec", iface, "off").CombinedOutput(); err != nil {
```
keliramu, line 46: How about obtaining current state of `dnssec` for later to get back to what user had in his setup? Or is this executed only against tunnel interface and not relevant to other interfaces?

daemon/dns/nameservers.go
```diff
@@ -100,3 +101,19 @@ func shuffleNameservers(nameservers []string) []string {
...
+func isNordDNSServer(address string) bool {
+	return internal.Contains(defaultServers, address) || internal.Contains(defaultRTPServers, address)
+}
+
+func areOnlyNordDNSServers(addresses []string) bool {
+	if len(addresses) == 0 {
+		return false
+	}
+	areOnlyNordDNS := true
+	for _, address := range addresses {
```
bartoszWojciechO, line 113: Maybe it would be simpler to have:
```suggestion
	for _, address := range addresses {
	    if !isNordDNSServer(address) {
	        return false
	    }
	}
	return true
```

bartoszWojciechO (approved):

daemon/dns/dns_resolvectl.go
mariusSincovici: For tunnel only

keliramu (approved):

daemon/dns/nameservers.go
mariusSincovici: Done

